PR #111, opened 2025-03-21
feat(api-connection): tanstack query hooks for connecting react to fastapi

### LINKED ISSUE #96: (api-connection): use tanstack query hooks for fastapi
### Description

Currently we generate a vanilla client for the website to use. It'd be great if we could generate hooks!

### Motivation

Users don't have to manage state themselves, and more consistent with trpc

### Suggested Implementation

Generate queryOptions, infiniteQueryOptions etc methods.

### Contribute to feature

cogwirrel:
### Reason for this change

* No need for users to manage state manually
* More consistent with tRPC

### Description of changes

This change implements the open-api#ts-hooks generator, which provides factory methods for options to provide to tanstack query hooks, in a similar manner to tRPC.

This saves the need for users to manage state themselves, and provides parity with tRPC.

By default, PUT/PATCH/POST/DELETE http requests are considered "mutations" and all others "queries". This behaviour can be overridden with `x-mutation` and `x-query` vendor extensions however.

Similarly to tRPC, if an operation accepts a `cursor` parameter, we consider this a paginated operation and generate an `infiniteQueryOptions` method alongside the `queryOptions` method. If an operation is paginated but has a different pagination parameter name, this can be specified using the `x-cursor` vendor extension.

We support streaming operations in a similar manner to tRPC, where a streaming query hook will continually have its data updated as chunks are received. Since a single cache entry is shared by all pages for an infinite query hook, we cannot stream multiple pages at once for a "paginated streaming" operation as this risks conflicts/race-conditions. We therefore iterate through the whole stream for each page in this case.
For mutation operations which return a stream, we don't do anything special and just return the iterator for the user to use if required.

* unit tests
* walked through the dungeon adventure tutorial

### Issue # (if applicable)

Fixes #96

### Checklist

----

### DIFF
--- a/docs/src/content/docs/get_started/tutorials/dungeon-game/1.mdx
+++ b/docs/src/content/docs/get_started/tutorials/dungeon-game/1.mdx
@@ -529,9 +529,14 @@ Below is a list of all files which have been generated/updated by the `api-conne
     - src/
       - hooks/
         - useSigV4.tsx used by StoryApi to sign requests
-        - useStoryApi.tsx hooks to call the StoryApi
+        - useStoryApiClient.tsx hook to construct a StoryApi client
+        - useStoryApi.tsx hook for interacting with the StoryApi using TanStack Query
+      - components/
+        - QueryClientProvider.tsx TanStack Query client provider
+        - StoryApiProvider.tsx Provider for the StoryApi TanStack Query hook
+      - main.tsx Instrument the QueryClientProvider and StoryApiProvider
     - .gitignore ignore generated client files
-    - project.json updated to add targets for generating openapi client
+    - project.json updated to add targets for generating openapi hooks
     - ...
   - story_api/
     - scripts/
@@ -541,7 +546,7 @@ Below is a list of all files which have been generated/updated by the `api-conne
 </FileTree>
 
 ```tsx {1,12-15}
-// packages/game-ui/src/hooks/useStoryApi.tsx
+// packages/game-ui/src/hooks/useStoryApiClient.tsx
 import { StoryApi } from '../generated/story-api/client.gen';
 import { useSigV4 } from './useSigV4';
 import { useRuntimeConfig } from './useRuntimeConfig';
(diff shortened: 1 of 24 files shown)


docs/src/content/docs/get_started/tutorials/dungeon-game/1.mdx
```diff
@@ -562,10 +567,39 @@ export const useStoryApi = (): StoryApi => {
...
+};
+
+export default StoryApiProvider;
+```
+
+The above provider component uses the `useStoryApiClient` hook, and instantiates the `StoryApiOptionsProxy`, which is used to build options for TanStack Query hooks. We will use the corresponding hook `useStoryApi` to provide us with this options proxy so that we can let TanStack Query manage the state for our streaming API.
 
 <Aside type="caution">
-The `client.gen.ts` file that is generated should never be modified manually as this will be re-generated any time a change to the backend API is detected.
+The `src/generated/story-api/*.gen.ts` files should never be modified manually as they will be re-generated any time a change to the backend API is detected.
```
agdimech, line 602: Technically only when a build/generate task is invoked*

agdimech: Might be worth adding an info/warning to state they will only be updated once running a build/regen

docs/src/content/docs/get_started/tutorials/dungeon-game/4.mdx
```diff
@@ -612,37 +615,79 @@ function RouteComponent() {
     gameApi.actions.query.queryOptions({ playerName, limit: 100 }),
   );
 
+  // Query for managing the streaming state from the story api
+  const streamingContentQuery = useQuery(
```
agdimech, line 619: I feel like this is more verbose than the vanilla client?

docs/src/content/docs/get_started/tutorials/dungeon-game/1.mdx
cogwirrel: Good shout!

docs/src/content/docs/get_started/tutorials/dungeon-game/4.mdx
cogwirrel: Yeah agreed - we generate the useQuery hook options in a way that's consistent with tRPC but it's definitely more intuitive and easier to manage when you just use the vanilla client :)

I'll keep the codegen as is but revert the tutorial changes :)

Will also use a context/provider so the vanilla client can be reused

agdimech (approved):

PR #117, opened 2025-03-24
feat(fastapi): generate an echo api by default for consistency with trpc

### LINKED ISSUE #92: (fastapi): generate an echo api like trpc instead of read_root
### Description

Generate echo, like we do for trpc

### Motivation

Less confusing

### Suggested Implementation

_No response_

### Contribute to feature

cogwirrel:
### Reason for this change

For tRPC we generate an echo procedure as an example to follow. For FastAPI we were generating a read_root operation which can be confusing for users.

### Description of changes

We update to generate an echo route for consistency with tRPC. Additionally, we define a model for the response returned by the unhandled exception middleware for better error handling in the generated code when using the api-connection.

Deployed and verified the echo operation works!

### Issue # (if applicable)

Fixes #92

### Checklist

----

### DIFF
--- a/docs/src/content/docs/guides/api-connection/react-fastapi.mdx
+++ b/docs/src/content/docs/guides/api-connection/react-fastapi.mdx
@@ -880,6 +880,10 @@ function ItemsAndUsers() {
 ```
 </Drawer>
 
+:::tip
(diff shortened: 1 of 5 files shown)


agdimech (approved):
Nice!

PR #274, opened 2025-08-26
feat(py#strands-agent): add a generator for building agents with strands

cogwirrel:
### Reason for this change

Make it easy to build agentic AI applications on AWS using Bedrock AgentCore and Strands.

### Description of changes

- Add `py#strands-agent` generator which adds an agent to a python project, as well as vending all the relevant infrastructure and build targets for deploying it to AgentCore
- Update the default `computeType` for the MCP server generators to `BedrockAgentCoreRuntime`
- Fix for `py#mcp-server` where if open telemetry was pulled in by project dependencies with a different version than the AWS Distro for OpenTelemetry expected, instrumentation would fail. Fixed this by adding the distro as an explicit dependency which is then included in the bundle.
- Refactor agent core construct to support both MCP servers and agents
- Also revamped the local port assignment logic to support multiple ports per project. I didn't end up using this as I moved from a FastAPI based approach for building agents to using the Bedrock AgentCore Python SDK which hardcodes the port - kept the change though as it'll likely be useful for the future.

Created a sample project using the generator in conjunction with both ts and py MCP servers.

### Checklist

----


docs/src/content/docs/en/guides/py-strands-agent.mdx
```diff
@@ -0,0 +1,342 @@
...
+
+You can use this CDK construct in a CDK application:
+
+```ts {6}
+import { MyProjectAgent } from ':my-scope/common-constructs';
+
+export class ExampleStack extends Stack {
+  constructor(scope: Construct, id: string) {
+    // Add the agent to your stack
+    const agent = new MyProjectAgent(this, 'MyProjectAgent');
```
agdimech, line 231: Should probably mention `MyProjectAgent` is not a real name.

agdimech (approved):

PR #278, opened 2025-08-27
fix(license): include terraform files in default license config

cogwirrel:
### Reason for this change

Since we support Terraform projects it's useful to have those covered by the license generator by default.

### Description of changes

Add terraform comment syntax and include .tf files in default license config

Ran the generator in a terraform project.

NB the sync generator won't run unless there are other projects that have a `lint` target - I think we probably want to set up the standard `build` and `lint` targets for terraform projects as a separate change.

### Checklist

----


agdimech (approved):
Nice!

PR #300, opened 2025-09-22
feat(ts#smithy-api): add generator for building apis modelled using Smithy and implemented in ts

### LINKED ISSUE #151: Add Smithy API Generator
### Description

Add a generator for a Smithy-based API which:
- Creates an API model project which allows you to define operations and their inputs/outputs in Smithy
- Generates a project to define type-safe backend operations to run on AWS Lambda
- Generates a CDK construct for deploying the API

Additionally the api-connection generator should understand how to connect projects to Smithy APIs using type safe clients generated from the Smithy model

### Motivation

For those that are interested in model-driven development for APIs, using an IDL to define API operations.

### Suggested Implementation

_No response_

### Contribute to feature

cogwirrel:
### Reason for this change

Allow users to build APIs using [Smithy](https://smithy.io/) and the Smithy TypeScript Server SDK.

### Description of changes

- Add `smithy#project` generator for building Smithy model projects - hidden for now given this is built specifically for use with a TypeScript SSDK backend, but in theory could be generified a bit more and exposed for building "Smithy shape library" type projects
- Add `ts#smithy-api` generator which creates:
  - A model project which contains the Smithy definition and a Dockerfile for building the model and generating OpenAPI and the TypeScript SSDK
  - A backend project for implementing the Smithy API in TypeScript using the generated SSDK. The backend provides:
    - Local server with hot-reloading
    - Powertools metrics/tracer/logging built in
  - CDK / Terraform infrastructure for deploying the API - CDK construct provides type-safe integrations using the same code generation mechanism as FastAPI.
- Update `api-connection` to support connecting a react website to a Smithy API
  - Target project can be either the model or backend project
  - Uses the same generated client as FastAPI since we can generate it from the OpenAPI specification. It'd be possible to use a Smithy-generated client instead in future, but we can add this separately if there's demand!
  - Supports all the `serve-local` wiring and hot reloading

Additionally made a few misc related changes:

- Fix Smithy comment syntax in license generator as `/*` `*/` block comments aren't a thing
- Add Smithy syntax highlighting to the docs
- Order `generators.json` alphabetically in `ts#nx-generator` so we don't have to do it by hand
- Update OpenAPI client generator's pagination vendor extension (`x-cursor`) to accept an object value as well as a string/boolean, since Smithy can only add object vendor extensions natively. NB: it would be a nicer user experience if the Smithy `@paginated` trait told our generator that an operation is paginated, but that info is not available in the OpenAPI spec that Smithy generates, so would require some extra processing in the `build.Dockerfile`. We can consider this if there's demand!
- Refactor some of the code that the new generators share with FastAPI
- Add a generic method for adding a bundle target using [rolldown](https://rolldown.rs) - I'll look at migrating other generators to use this instead as it'll simplify things (rolldown config can define multiple bundle targets which are run in parallel so we don't need to manage individual ones in `project.json` for eg `ts#lambda-function` or `ts#mcp-server`)
- Refactor adding ignore patterns to eslint config
- Add helper for adding commands to targets for platform-agnostic filesystem operations (rm, mkdir, cp)

Manual testing, integ tests, unit tests

### Issue # (if applicable)

Fixes #151 

### Checklist

----


agdimech (approved):
Great stuff man!

agdimech (dismissed):

PR #303, opened 2025-09-25
feat(ts): use rolldown for bundling lambdas, mcp servers and trpc apis

cogwirrel:
### Reason for this change

Consistent bundling across all projects that's easier to configure.

### Description of changes

Move all bundle targets that used esbuild to use rolldown instead.

Rolldown supports creating multiple bundles in parallel via specifying an array in the config. We therefore use AST updates to add bundling configuration. This simplifies things from an nx perspective as we now only need 1 bundle target no matter how many bundles a project needs to create.

Manual test project, unit and integration tests.

### Checklist

----


agdimech (approved):
Nice improvements with the helpers :)

PR #376, opened 2025-12-10
feat(ts#sync): typescript path alias sync generator

### LINKED ISSUE #374: TypeScript path alias sync generator
### Description

Add a `ts#sync` [sync generator](https://nx.dev/docs/concepts/sync-generators) which hooks into the `compile` target and ensures that the relevant tsconfig path aliases from the base tsconfig are included.

### Motivation

It's quite common to add custom `compilerOptions.paths` to tsconfig in order to alias particular directories in projects. For example, inside a project I might like to add an alias such that `@/xxx` points to the root of that project so I don't need to figure out how many `../`s to use for imports.

We configure `compilerOptions.paths` in `tsconfig.base.json` so that projects can import from one another. Unfortunately, if `compilerOptions.paths` is specified in a project's `tsconfig.json`, this _overrides_ the paths inherited from the base config. Aliases for dependencies that are imported would need to again be specified in the project's tsconfig.

### Suggested Implementation

* `preset` generator updates `nx.json` to configure `ts#sync` generator for the `compile` target (see `license` generator implementation for example of `nx.json` updates)
* `ts#sync` generator is hidden, accepting no parameters, and:
  * Reads the paths from the base tsconfig in the workspace root (if any)
  * Loops through all projects using `getProjects(tree)` and:
    * Checks if there's a `tsconfig.json`, `tsconfig.lib.json`, `tsconfig.app.json` with `compilerOptions.paths` set
    * If there is, add any of the base tsconfig paths that are missing. If no compilerOptions.paths set, skip.

Include a section in the `ts#project` developer guide (just after talking about imports and the Nx sync generator) which explains our ts#sync generator and the problem it solves. Also include how to disable it (remove the config in `nx.json`)

### Contribute to feature

- [ ] I'd be willing to implement this feature

AlexTo:
### Reason for this change

It's quite common to add custom compilerOptions.paths to tsconfig in order to alias particular directories in projects. For example, inside a project I might like to add an alias such that @/xxx points to the root of that project so I don't need to figure out how many ../s to use for imports.

We configure compilerOptions.paths in tsconfig.base.json so that projects can import from one another. Unfortunately, if compilerOptions.paths is specified in a project's tsconfig.json, this overrides the paths inherited from the base config. Aliases for dependencies that are imported would need to again be specified in the project's tsconfig.

### Description of changes

Add a ts#sync [sync generator](https://nx.dev/docs/concepts/sync-generators) which hooks into the compile target and ensures that the relevant tsconfig path aliases from the base tsconfig are included.

Unit/Integ tests.

### Issue # (if applicable)

Closes #374.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/es/guides/typescript-project.mdx
+++ b/docs/src/content/docs/es/guides/typescript-project.mdx
@@ -3,8 +3,6 @@ title: "Proyectos de TypeScript"
 description: "Documentación de referencia para proyectos de TypeScript"
 ---
 
-
-
 import { FileTree } from '@astrojs/starlight/components';
 import RunGenerator from '@components/run-generator.astro';
 import InstallCommand from '@components/install-command.astro';
@@ -131,6 +129,10 @@ No, run the tasks without syncing the changes
 
 Selecciona `Yes` para permitir que Nx actualice tus referencias de proyecto.
 :::
+
+### Manteniendo los alias de rutas sincronizados
+
+Si añades entradas personalizadas de `compilerOptions.paths` en el `tsconfig.json` de un proyecto, TypeScript deja de heredar los alias del workspace definidos en `tsconfig.base.json`. El generador oculto `ts#sync` se ejecuta antes del objetivo `compile` (configurado en `nx.json`) para copiar cualquier alias base faltante en los archivos `tsconfig.json`, `tsconfig.lib.json` y `tsconfig.app.json` que ya declaren `paths`. Para desactivar esta funcionalidad, elimina `@aws/nx-plugin:ts#sync` de `targetDefaults.compile.syncGenerators` en `nx.json`.
 
 ### Dependencias
 
--- a/docs/src/content/docs/fr/guides/typescript-project.mdx
+++ b/docs/src/content/docs/fr/guides/typescript-project.mdx
@@ -3,8 +3,6 @@ title: "Projets TypeScript"
 description: "Documentation de référence pour les projets TypeScript"
 ---
 
-
-
 import { FileTree } from '@astrojs/starlight/components';
 import RunGenerator from '@components/run-generator.astro';
 import InstallCommand from '@components/install-command.astro';
@@ -131,6 +129,10 @@ No, run the tasks without syncing the changes
 
 Sélectionnez `Yes` pour permettre à Nx de mettre à jour vos références de projet.
 :::
+
+### Maintenir les alias de chemins synchronisés
+
+Si vous ajoutez des entrées `compilerOptions.paths` personnalisées dans le `tsconfig.json` d'un projet, TypeScript cesse d'hériter les alias du workspace définis dans `tsconfig.base.json`. Le générateur caché `ts#sync` s'exécute avant la cible `compile` (configurée dans `nx.json`) pour copier tous les alias de base manquants dans les fichiers `tsconfig.json`, `tsconfig.lib.json` et `tsconfig.app.json` qui déclarent déjà `paths`. Pour désactiver cette fonctionnalité, supprimez `@aws/nx-plugin:ts#sync` de `targetDefaults.compile.syncGenerators` dans `nx.json`.
 
 ### Dépendances
 
--- a/docs/src/content/docs/it/guides/typescript-project.mdx
+++ b/docs/src/content/docs/it/guides/typescript-project.mdx
@@ -3,8 +3,6 @@ title: "Progetti TypeScript"
 description: "Documentazione di riferimento per progetti TypeScript"
 ---
 
-
-
 import { FileTree } from '@astrojs/starlight/components';
 import RunGenerator from '@components/run-generator.astro';
 import InstallCommand from '@components/install-command.astro';
@@ -131,6 +129,10 @@ No, run the tasks without syncing the changes
 
 Seleziona `Yes` per consentire a Nx di aggiornare i riferimenti del progetto.
 :::
+
+### Mantenere gli alias di percorso sincronizzati
+
+Se aggiungi voci personalizzate `compilerOptions.paths` nel `tsconfig.json` di un progetto, TypeScript smette di ereditare gli alias del workspace definiti in `tsconfig.base.json`. Il generatore nascosto `ts#sync` viene eseguito prima del target `compile` (configurato in `nx.json`) per copiare eventuali alias di base mancanti nei file `tsconfig.json`, `tsconfig.lib.json` e `tsconfig.app.json` che già dichiarano `paths`. Per disattivare questa funzionalità, rimuovi `@aws/nx-plugin:ts#sync` da `targetDefaults.compile.syncGenerators` in `nx.json`.
 
 ### Dipendenze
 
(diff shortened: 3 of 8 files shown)


packages/nx-plugin/src/ts/sync/generator.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      },
+    });
+
+    await sharedConstructsGenerator(tree, { iacProvider: 'CDK' });
+
+    await syncGeneratorGenerator(tree);
+
+    expectHasMetricTags(tree, TS_SYNC_GENERATOR_INFO.metric);
+  });
+});
```
cogwirrel: Nice tests! Please could we add another test which checks that if the project's paths include the base paths already, that there also won't be any updates? :)

packages/nx-plugin/src/ts/sync/generator.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    const result = await syncGeneratorGenerator(tree);
+
+    expect(
+      readJson<Record<string, any>>(tree, 'packages/proj/tsconfig.json')
+        .compilerOptions.paths,
+    ).toBeUndefined();
+    expect(result).toEqual({});
+  });
+
+  it('should add generator metric to app.ts', async () => {
```
cogwirrel: Let's not add the generator metric - I think adding this will mean we stop the world on the first build as the changes it makes to common constructs will mean it's considered out of sync :)

packages/nx-plugin/src/ts/sync/generator.ts
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  return addedAliases;
+};
+
+/**
+ * Build the message to display when the sync generator would make changes to the tree
+ */
+const buildOutOfSyncMessage = (addedPaths: Record<string, string[]>): string =>
+  `TypeScript path aliases are out of sync with the base tsconfig. The following configs were updated:\n${Object.entries(
```
cogwirrel: Nit: let's just say `TypeScript path aliases are out of sync with the base tsconfig:`, as the message is shown to the user before they confirm to apply the changes.

packages/nx-plugin/src/ts/sync/generator.spec.ts
AlexTo: new test added

packages/nx-plugin/src/ts/sync/generator.spec.ts
AlexTo: generator metric removed

packages/nx-plugin/src/ts/sync/generator.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+const addBasePathsIfMissing = (
+  tree: Tree,
+  tsconfigPath: string,
+  basePaths: Record<string, string[]>,
+): string[] => {
+  const addedAliases: string[] = [];
+
+  updateJson(tree, tsconfigPath, (json) => {
```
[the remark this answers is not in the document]
AlexTo: noops return `{}` so are not considered changes. I added `readJson` to avoid unnecessary call to `updateJson`. Not related to this PR but I wanted to check if smoke tests run successfully on local but failed for some tests due to docker build is hard coded to `linux/arm64`. I temporarily remove the `platform` flag for smoke tests to work on my `linux/amd64` but maybe worth reviewing for a more cross platform approach?

packages/nx-plugin/src/ts/sync/generator.ts
```diff
@@ -0,0 +1,133 @@
...
+    const paths = json?.compilerOptions?.paths;
+
+    if (!paths) {
+      return json;
+    }
+
+    const missingEntries = Object.entries(basePaths).filter(
+      ([alias]) => !(alias in paths),
+    );
```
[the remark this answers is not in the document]
AlexTo: done

packages/nx-plugin/src/ts/sync/generator.ts
AlexTo: done

packages/nx-plugin/src/ts/sync/generator.ts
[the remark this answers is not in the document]
cogwirrel: Cool thanks - it's worth noting that it's not the presence of an `outOfSyncMessage` in the response which Nx uses to determine whether to pause and report that the workspace is out of sync, it's whether there are any changes to made to the `tree` https://nx.dev/docs/extending-nx/create-sync-generator#performance-and-dx-considerations - I'm not sure if it actually manifests but I guess anytime we call `updateJson` we risk a change being tracked in the tree :)

Interesting re the platform issue! Have you got a recent-ish version of docker with `buildx`? The hardcoded platforms are due to Bedrock AgentCore running ARM only - the intention is that docker does a cross-platform build to ensure the image is always built for the architecture AgentCore supports :)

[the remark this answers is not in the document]
AlexTo: Ah, got it, thanks for the clarification. Sorry for my confusion about how nx works; I’ll take a look at that page.
Also, false alarm on the cross-platform build. I thought we needed to explicitly use `docker buildx`, but it turns out that `docker build` works fine as well on my Ubuntu.

packages/nx-plugin/src/ts/sync/generator.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  return addedAliases;
+};
+
+/**
+ * Build the message to display when the sync generator would make changes to the tree
+ */
+const buildOutOfSyncMessage = (): string =>
+  `TypeScript path aliases are out of sync with the base tsconfig.`;
```
cogwirrel: Ooh sorry just noticed this while testing locally - sorry I should have been clearer - I think the list of files and paths are useful, just didn't want to imply that the changes had already been made in the message :)

cogwirrel: Would it be alright to add back the list of tsconfig paths that are out of sync? :)

AlexTo: Sure, should I change this message to "will be updated" instead ?
```
const buildOutOfSyncMessage = (addedPaths: Record<string, string[]>): string =>
  `TypeScript path aliases are out of sync with the base tsconfig. The following configs were updated:\n${Object.entries(
    addedPaths,
  )
    .map(
      ([config, aliases]) =>
        `${config}:\n${aliases.map((alias) => `- ${alias}`).join('\n')}`,
    )
    .join('\n\n')}`;
```

cogwirrel: Yeah that sounds good! :) Thank you!

cogwirrel:
Just did a bit of testing/playing around locally and there are a couple more cases I didn't think about:

- Changes to an existing path in tsconfig.base.json isn't sync'd - I think we also need to check if the values are the same for any matching ones from base and update those :)
- Removal of paths from tsconfig.base.json aren't removed from the project - tbh I don't think we can detect removed ones so I don't think we need to solve this

Mind updating to support changes to an existing alias? Maybe in the future we can think about a way to solve removal but for now I think it's ok to leave them for users to clean up if they delete projects.

packages/nx-plugin/src/ts/sync/generator.ts
AlexTo: done

cogwirrel (dismissed):
Thanks so much @AlexTo! 🚀

PR #404, opened 2026-01-30
fix(ts#infra): use --directory flag for checkov to support single stack

### LINKED ISSUE #403: (ts#infra): checkov fails when you remove application stage and have a single stack
### Current Behavior

The checkov command looks for `cdk.out/**/*.template.json` but the `.template.json` file is directly under the `cdk.out` directory when the CDK `App` only declares one or more `Stack`s at the top level rather than one or more `Stage`s.

### Expected Behavior

Checkov command shouldn't need to be updated if users just declare a top level stack.

### Steps to Reproduce

1. generate `ts#infra` project
2. remove `application-stage.ts`
3. update `main.ts` to declare `ApplicationStage` instead of `ApplicationStack`
4. build and observe checkov error

### Nx Report

```shell
n/a
```

### Failure Logs

```shell

```

### Additional Information

It would also be good to remove the pinned `pycares` version that was added due to #378 as it's most likely been fixed downstream.

littleKitchen:
## Summary

Fixes checkov failing when the CDK App declares stacks directly without stages.

## Problem

The checkov command used `--file dist/<path>/cdk.out/**/*.template.json` which only matches files in subdirectories, not files directly in `cdk.out/`.

**With Stages (worked):**
```
cdk.out/
  StageName/
    Stack.template.json  ✅ matched by **/*.template.json
```

**Without Stages (broken):**
```
cdk.out/
  Stack.template.json  ❌ not matched by **/*.template.json
```

## Solution

Changed from `--file` with glob pattern to `--directory` flag which scans all CloudFormation templates recursively. This matches the pattern already used in the Terraform generator.

**Before:**
```
--file dist/<path>/cdk.out/**/*.template.json
```

**After:**
```
--directory dist/<path>/cdk.out
```

## Testing

- [x] `pnpm nx test nx-plugin -- --run "infra/app/generator"` passes (17 tests)
- [x] Snapshot updated to reflect new command

## Related

- The Terraform generator (`terraform/project/generator.ts`) already uses `--directory` correctly

Closes #403

### DIFF
--- a/packages/nx-plugin/src/infra/app/__snapshots__/generator.spec.ts.snap
+++ b/packages/nx-plugin/src/infra/app/__snapshots__/generator.spec.ts.snap
@@ -84,7 +84,7 @@ exports[`infra generator > should configure Checkov target correctly > checkov-t
     "{workspaceRoot}/dist/{projectRoot}/cdk.out",
   ],
   "options": {
-    "command": "uvx --with pycares==4.11.0 checkov==3.2.500 --config-file packages/test/checkov.yml --file dist/packages/test/cdk.out/**/*.template.json",
+    "command": "uvx --with pycares==4.11.0 checkov==3.2.500 --config-file packages/test/checkov.yml --directory dist/packages/test/cdk.out --framework cloudformation",
   },
   "outputs": [
     "{workspaceRoot}/dist/{projectRoot}/checkov",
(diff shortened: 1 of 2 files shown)


cogwirrel:
Hey, thanks so much for this! :)

I think we might also need to add `--framework cloudformation` so that we don't run other scans against the `assets` directory (eg Docker etc).

Otherwise looks great! Thanks! 😄

cogwirrel (approved):
Thanks so much for contributing!

PR #414, opened 2026-02-09
feat: feat(ts#openapi): add OpenAPI 3.2 JSONL streaming support with itemSchema

### LINKED ISSUE #369: (py#strands-agent): sse streaming responses
### Description

Currently streaming responses for the agent is just raw text and doesn't conform to the SSE standard (ie `data: xxx`).

We should respond following this standard, and update the openapi client generator to support streaming in this manner (most likely following openapi 3.2's official support https://spec.openapis.org/oas/v3.2.0.html#complete-vs-streaming-content)

### Motivation

Follow recommended standard

### Suggested Implementation

_No response_

### Contribute to feature

seungholee-dev:
### Reason for this change

Add support for typed streaming responses using OpenAPI 3.2's itemSchema specification. This enables type-safe JSONL (JSON Lines) streaming for AI agent responses and similar use cases.

### Description of changes

- Add STREAMING_CONTENT_TYPES constant for application/jsonl and application/x-ndjson
- Implement detectOpenApi32Streaming() to detect itemSchema in OpenAPI spec
- Add JSONL line buffering logic in client.gen.ts.template for handling partial network chunks
- Generate typed streaming responses using $IO.fromJson() when model available

- Add JsonStreamingResponse class for OpenAPI 3.2 compliant streaming
- Add StreamChunk Pydantic model for typed response chunks
- Implement _process_item_schema() to convert x-itemSchema to standard itemSchema
- Update /invocations endpoint to use JSONL format instead of text/event-stream

- Add mockJsonlStreamingFetch helper for testing JSONL streaming
- Add tests for application/jsonl with itemSchema
- Add tests for application/x-ndjson with itemSchema
- Add backward compatibility test for x-streaming vendor extension

- Legacy x-streaming: true with text/event-stream still supported
- OpenAPI 3.0 specs without itemSchema work unchanged

I ran with `pnpm nx run @aws/nx-plugin:test -u`

### Issue # (if applicable)

Closes #369

### Checklist
- [V] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/awslabs/nx-plugin-for-aws/blob/main/DESIGN_GUIDELINES.md)

----

### DIFF
--- a/packages/nx-plugin/package.json
+++ b/packages/nx-plugin/package.json
@@ -42,7 +42,7 @@
     "lodash.orderby": "^4.6.0",
     "lodash.trim": "^4.5.1",
     "lodash.uniqby": "^4.7.0",
-    "minimatch": "^10.2.1",
+    "minimatch": "^10.2.3",
     "openapi-types": "^12.1.3",
     "pip-requirements-js": "^0.2.1",
     "typescript": "~5.9.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -367,13 +367,13 @@ importers:
         version: 22.5.1(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))
       '@nx/eslint':
         specifier: ~22.5.1
-        version: 22.5.1(@babel/traverse@7.28.4)(@swc/core@1.7.26(@swc/helpers@0.5.17))(@zkochan/js-yaml@0.0.7)(eslint@9.39.2(jiti@2.6.1))(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))
+        version: 22.5.1(@babel/traverse@7.28.4)(@swc/core@1.7.26(@swc/helpers@0.5.17))(@zkochan/js-yaml@0.0.7)(eslint@9.39.3(jiti@2.6.1))(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))
       '@nx/js':
         specifier: ~22.5.1
         version: 22.5.1(@babel/traverse@7.28.4)(@swc/core@1.7.26(@swc/helpers@0.5.17))(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))
       '@nx/react':
         specifier: ~22.5.1
-        version: 22.5.1(@babel/core@7.28.4)(@babel/traverse@7.28.4)(@swc/core@1.7.26(@swc/helpers@0.5.17))(@swc/helpers@0.5.17)(@types/babel__core@7.20.5)(@zkochan/js-yaml@0.0.7)(esbuild@0.27.3)(eslint@9.39.2(jiti@2.6.1))(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(typescript@5.9.3)(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))(vite@7.3.1(@types/node@24.10.1)(jiti@2.6.1)(less@4.1.3)(lightningcss@1.30.2)(sass@1.79.4)(stylus@0.64.0)(terser@5.44.0)(tsx@4.21.0)(yaml@2.8.1))(vitest@4.0.18)
+        version: 22.5.1(@babel/core@7.28.4)(@babel/traverse@7.28.4)(@swc/core@1.7.26(@swc/helpers@0.5.17))(@swc/helpers@0.5.17)(@types/babel__core@7.20.5)(@zkochan/js-yaml@0.0.7)(esbuild@0.27.3)(eslint@9.39.3(jiti@2.6.1))(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(typescript@5.9.3)(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))(vite@7.3.1(@types/node@24.10.1)(jiti@2.6.1)(less@4.1.3)(lightningcss@1.30.2)(sass@1.79.4)(stylus@0.64.0)(terser@5.44.0)(tsx@4.21.0)(yaml@2.8.1))(vitest@4.0.18)
       '@nx/vite':
         specifier: ~22.5.1
         version: 22.5.1(@babel/traverse@7.28.4)(@swc/core@1.7.26(@swc/helpers@0.5.17))(nx@22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17)))(typescript@5.9.3)(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))(vite@7.3.1(@types/node@24.10.1)(jiti@2.6.1)(less@4.1.3)(lightningcss@1.30.2)(sass@1.79.4)(stylus@0.64.0)(terser@5.44.0)(tsx@4.21.0)(yaml@2.8.1))(vitest@4.0.18)
@@ -408,8 +408,8 @@ importers:
         specifier: ^4.7.0
         version: 4.7.0
       minimatch:
-        specifier: ^10.2.1
-        version: 10.2.2
+        specifier: ^10.2.3
+        version: 10.2.3
       nx:
         specifier: ~22.5.1
         version: 22.5.1(@swc/core@1.7.26(@swc/helpers@0.5.17))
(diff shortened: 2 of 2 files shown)


packages/nx-plugin/src/py/strands-agent/files/app/init.py.template
```diff
@@ -11,6 +13,91 @@ class InternalServerErrorDetails(BaseModel):
...
+        """
+        return {
+            "description": description,
+            "content": {
+                "application/jsonl": {
+                    "schema": {
+                        "type": "string",
+                        "x-model-name": item_model.__name__,
+                    },
+                    "x-itemSchema": {"$ref": f"#/components/schemas/{item_model.__name__}"},
```
cogwirrel: I think we might need to rethink this a bit, as we're creating a ref here but not adding the schema to `components/schemas`. I just checked the code out and ran the `open-api#ts-hooks` generator and get this:

```
> nx g @aws/nx-plugin:open-api#ts-hooks --openApiSpecPath="dist/packages/py/openapi/openapi.json" --outputPath="packages/web/src/generated/api" --no-interactive

 NX  Generating @aws/nx-plugin:open-api#ts-hooks


 NX   Token "StreamChunk" does not exist.

MissingPointerError: Token "StreamChunk" does not exist.
```

From a bit of a play in a test project, it seems like we can do this:

```
    @staticmethod
    def openapi_response(
        item_model: Type[BaseModel],
        description: str = "Streaming response",
    ) -> dict[str, Any]:
        return {
            "description": description,
            "content": {
                "application/jsonl": {
                    "itemSchema": {"$ref": f"#/components/schemas/{item_model.__name__}"},
                }
            },
            "model": item_model
        }
```

And delete the `_process_item_schema` entirely :)

packages/nx-plugin/src/open-api/utils/codegen-data.ts
```diff
@@ -1113,21 +1114,85 @@ const getCursorOptions = (op: Operation) => {
...
+      result.typescriptType = jsonlStreaming.itemSchemaRef;
+      result.export = 'reference';
+    }
+  } else {
+    // Fall back to legacy x-streaming vendor extension
+    (op as any).isStreaming = !!(op as any).vendorExtensions?.[
+      VENDOR_EXTENSIONS.STREAMING
+    ];
+    (op as any).isJsonlStreaming = false;
+  }
```
cogwirrel: I think we can do this in a slightly simpler way :) (apologies this part of the codebase is quite complex!)

On line 131 we loop over `mediaTypes` within a response to build up our response data structures. We can add a check for the streaming media type to the end of that loop and then link to the right model for `itemSchema`, eg:

```ts
if (STREAMING_CONTENT_TYPES.has(mediaType) && 'itemSchema' in responseContent) {
  (response as any).isJsonlStreaming = true;
  (response as any).itemSchemaModel = await buildOrReferenceModel(spec, modelsByName, responseContent.itemSchema);
}
```

Then we can get rid of `detectOpenApi32Streaming`, and then the only update we need to make to `mutateOperationWithAdditionalData` is:

```ts
(op as any).isStreaming = !!(op as any).vendorExtensions?.[
  VENDOR_EXTENSIONS.STREAMING
] || op.responses.some(res => (res as any).isJsonlStreaming);
```

That then gives us the `itemSchemaModel` which we can use in `client.gen.ts.template` (see other comment there)

packages/nx-plugin/src/open-api/ts-client/files/client.gen.ts.template
```diff
@@ -433,11 +433,35 @@ export class <%- className %> {
...
     <%_ } else { _%>
     const reader = response.body?<% if (response.type !== 'binary') { %>.pipeThrough(new TextDecoderStream())<% } %>.getReader();
+    <%_ if (op.isJsonlStreaming) { _%>
+    <%_ const itemSchemaModel = models.find(m => m.name === op.streamingItemSchema); _%>
+    let buffer = '';
+    while (reader) {
+      const { value: chunk, done } = await reader.read();
+      if (done) {
+        if (buffer.trim()) {
+          yield <%- itemSchemaModel ? `$IO.${itemSchemaModel.typescriptName}.fromJson(JSON.parse(buffer))` : 'JSON.parse(buffer)' %>;
```
cogwirrel: We can reuse `renderStreamingResponseChunk` here which will support arrays if we make a small change to it :)

Here we'd do:

```
yield <%- renderStreamingResponseChunk(response.itemSchemaModel, 'buffer') %>;
```

And the same below but passing in `'line'`.

The slight change we'd need to make to `renderStreamingResponseChunk` would be to allow the variable name to be configured:

```ts
const renderStreamingResponseChunk = (response, chunkVariableName = 'value') => {
  return renderResponse(response, {
    asJson: `JSON.parse(${chunkVariableName})`,
    asText: chunkVariableName,
    asBlob: `new Blob([${chunkVariableName}])`,
  });
};
```

docs/src/content/docs/en/guides/py-strands-agent.mdx
```diff
@@ -298,41 +298,50 @@ It is also worth noting that depending on your use case you may need to implemen
...
+    "/invocations",
+    response_class=JsonStreamingResponse,
+    responses={200: JsonStreamingResponse.openapi_response(StreamChunk, "Stream of agent response chunks")},
+)
+async def invoke(input: InvokeInput) -> JsonStreamingResponse:
+    return JsonStreamingResponse(handle_invoke(input))
 ```
 
-You can define a Pydantic model to yield structured data instead. In order to do this, you will need to serialize the pydantic models yielded by `handle_invoke`:
+You can define your own Pydantic model to yield structured data. Simply update the `StreamChunk` model and the response configuration:
```
cogwirrel: Nit: ``You can customise the `StreamChunk` model to suit your needs:``

docs/src/content/docs/en/guides/py-strands-agent.mdx
```diff
@@ -298,41 +298,50 @@ It is also worth noting that depending on your use case you may need to implemen
...
+    responses={200: JsonStreamingResponse.openapi_response(StreamChunk, "Stream of response chunks")},
+)
+async def invoke(input: InvokeInput) -> JsonStreamingResponse:
+    return JsonStreamingResponse(handle_invoke(input))
 ```
 
-:::caution
-Note here that the vended server (and the above `serialize_stream` method) does not currently conform to the Server-Sent Events (SSE) standard, such that if desired the Agent can be invoked by our generated TypeScript client from a frontend, which doesn't currently expect streaming events in SSE format. Please +1 [this GitHub issue](https://github.com/awslabs/nx-plugin-for-aws/issues/369) if this affects you to help us prioritise addressing this.
-:::
+The `JsonStreamingResponse` class handles serialization automatically, converting each yielded Pydantic model to a JSON line.
```
cogwirrel: Let's update this to something like:

````mdx
:::note
Note that since FastAPI does not currently support generating an OpenAPI specification for streaming responses, we must explicitly set the response model for each streaming operation, ie:

```python
@app.post(
    "/invocations",
    response_class=JsonStreamingResponse,
    responses={200: JsonStreamingResponse.of(StreamChunk)},
)
async def invoke(input: InvokeInput) -> JsonStreamingResponse:
    return JsonStreamingResponse(handle_invoke(input))
```
:::

There is an [open feature request for native support in FastAPI](https://github.com/fastapi/fastapi/discussions/14362).
````

File: packages/nx-plugin/src/py/strands-agent/files/app/main.py.template [hunk 6]
```diff
@@ -1,34 +1,43 @@
...
             print(event)
-            content = event.get("event", {}).get("contentBlockDelta", {}).get("delta", {}).get("text")
-            if content is not None:
-                yield content
+            text = event.get("event", {}).get("contentBlockDelta", {}).get("delta", {}).get("text")
+            if text is not None:
+                yield StreamChunk(content=text)
             elif event.get("event", {}).get("messageStop") is not None:
-                yield "\n"
+                yield StreamChunk(content="\n")
```
cogwirrel: We'll need to update the CDK deploy end to end test (`cdk-deploy.spec.ts`) to expect responses as json lines :)

packages/nx-plugin/src/open-api/ts-client/generator.streaming.spec.ts
```diff
@@ -1044,4 +1045,240 @@ describe('openApiTsClientGenerator - streaming', () => {
...
+      // Verify the generated code has JSONL line buffering logic
+      expect(client).toContain('let buffer');
+      expect(client).toContain("split('\\n')");
+
+      // Test with JSONL mock
+      const mockFetch = mockJsonlStreamingFetch(200, [
+        '{"content":"Hello"}',
+        '{"content":" World"}',
+        '{"content":"!"}',
+      ]);
```
cogwirrel: Can we add a test with newlines inside our `content` and check that they don't break this?

packages/nx-plugin/src/open-api/utils/codegen-data.ts
```diff
@@ -1117,17 +1130,27 @@ const getCursorOptions = (op: Operation) => {
...
+      result.type = itemSchemaModel.name;
+      result.typescriptType = itemSchemaModel.name;
+      result.export = 'reference';
+    }
+  }
 
-  // Add infinite query details if applicable
   if (!isMutation) {
     mutateOperationWithInfiniteQueryDetails(op);
   }
```
cogwirrel: Nit: can we add back the deleted comments? And maybe add one above `const jsonlResponse = op.responses.find(....` that explains why this is needed, something along the lines of: `// Set the result type for the client method to our item schema so that it will resolve to AsyncIterableIterator<{itemSchemaModel.name}>`?

packages/nx-plugin/src/py/strands-agent/files/app/init.py.template
```diff
@@ -11,6 +13,89 @@ class InternalServerErrorDetails(BaseModel):
     detail: str
 
 
+class JsonStreamingResponse(StreamingResponse):
+    """
+    A streaming response that serializes items to JSON Lines format.
+
+    This response type is compatible with OpenAPI 3.2's streaming specification,
+    using application/jsonl media type with itemSchema for type definitions.
+    """
```
cogwirrel: Nit: let's update to:

```"""A streaming response that serializes items to JSON Lines format."""```

Unless things are really complex and nuance I generally prefer vending shorter comments in our generators so the code is less overwhelming 😄

seungholee-dev:
On it! :) 
Also, just to show JSONL works fine: 
image

PR #449, opened 2026-03-02
feat(ts#trpc-api): enhancements to single-function tRPC API

### LINKED ISSUE #298: Enhancements to single-function tRPC API.
### Description

While the default behavior for the TRPC API is to have one lambda per API endpoint, it's common to only have a single lambda proxy handler for the API, reducing the infrastructure overhead and centralizing the API. There is instructions on how to achieve this within the documentation by editing the generated API construct. But it would be great to: 

- Be able to vend this with an option without requiring additional changes to support this behavior.
- Fix the policy overload issue caused by this [code here](https://github.com/aws/aws-cdk/blob/eb8e74b8146d57a6f85484a2db053e1bfc27864e/packages/aws-cdk-lib/aws-apigateway/lib/integrations/lambda.ts#L64) where the single handler function policy becomes too large ones a set amount of API endpoints has been reached. 

### Motivation

I personally have a strong preference for these sorts of API's to be single proxy functions, and very much dislike having 50+ individual lambdas for a single API. I would appreciate it being officially supported.

### Suggested Implementation

_No response_

### Contribute to feature

- [ ] I'd be willing to implement this feature

AlexTo:
### Reason for this change

While the default behaviour for the TRPC API is to have one lambda per API endpoint, it's useful to provide an additional option to only have a single lambda proxy handler for the API, reducing the infrastructure overhead and centralising the API. 

### Description of changes

Add an integrationPattern option with two supported values: `Isolated` and `Shared `.
* `Isolated` mode provisions one Lambda function per API endpoint.
* `Shared` mode provisions a single Lambda function as a router handler for all endpoints.
  *  Set `scopePermissionToRoute` to `false` and `scopePermissionToMethod` to `false` for HTTP API and REST API respectively to fix the policy overload issue .

* Unit tests
* Generated 4 combinations: Rest/Isolated, Rest/Shared, HTTP/Isolated, HTTP/Shared
* Deployed to AWS and verified in API Gateway that all the integrations for the 4 APIs are correct and a single policy statement is generated when integrationPattern `Shared ` is selected.

### Issue # (if applicable)

Closes #298.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/utils/api-constructs/files/cdk/app/apis/http/__apiNameKebabCase__.ts.template
+++ b/packages/nx-plugin/src/utils/api-constructs/files/cdk/app/apis/http/__apiNameKebabCase__.ts.template
@@ -106,7 +106,7 @@ export class <%= apiNameClassName %><
       <%_ } else { _%>
       operations: OPERATION_DETAILS,
       <%_ } _%>
-      defaultIntegrationOptions: {
+      defaultIntegrationOptions: <FunctionProps>{
         <%_ if (['trpc', 'smithy'].includes(backend.type)) { _%>
         runtime: Runtime.NODEJS_LATEST,
         handler: 'index.handler',
@@ -143,7 +143,7 @@ export class <%= apiNameClassName %><
           AWS_LAMBDA_EXEC_WRAPPER: '/opt/bootstrap',
           <%_ } _%>
         },
-      } satisfies FunctionProps,
+      },
       buildDefaultIntegration: (op, props: FunctionProps) => {
         <%_ const lambdaTarget = backend.type === 'fastapi' ? 'handler.currentVersion' : 'handler'; _%>
         <%_ const integrationOptions = backend.integrationPattern === 'shared' ? ', { scopePermissionToRoute: false }' : ''; _%>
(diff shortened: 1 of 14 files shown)


packages/nx-plugin/src/utils/api-constructs/files/cdk/app/apis/http/__apiNameKebabCase__.ts.template
```diff
@@ -157,6 +191,7 @@ export class <%= apiNameClassName %><
           integration: new HttpLambdaIntegration(`<%= apiNameClassName %>${op}Integration`, handler),
           <%_ } _%>
         };
+        <%_ } _%>
       },
```
cogwirrel: I think we can simplify this a little, something like:

```ejs
        <%_ const integrationOptions = backend.integrationPattern === 'shared' ? ', { scopePermissionToRoute: false }' : '' _%>
        <%_ const lambdaTarget = backend.type === 'fastapi' ? 'handler.currentVersion' : 'handler' _%>
        const handler = new Function(scope, `<%= apiNameClassName %>${op}Handler`, props);
        <%_ if (backend.type === 'fastapi') { _%>
        const stack = Stack.of(scope);
        handler.addLayers(
          LayerVersion.fromLayerVersionArn(
            scope,
            `<%= apiNameClassName %>${op}LWALayer`,
            `arn:aws:lambda:${stack.region}:753240598075:layer:LambdaAdapterLayerX86:24`,
          ),
        );
        <%_ } _%>
        return {
          handler,
          integration: new HttpLambdaIntegration(`<%= apiNameClassName %>${op}Integration`, <%- lambdaTarget %><%- integrationOptions %>),
        };
        <%_ } _%>
```

cogwirrel (dismissed):
Fantastic @AlexTo! Thanks so much! 🙌

PR #552, opened 2026-04-06
feat(ts#rdb): add relational database support

### LINKED ISSUE #485: (ts#rdb): new generator to vend packages for relational database support
### Description

Currently, nx-plugin-for-aws provides generators for vending APIs, but it does not yet support generators for vending databases.
There are reference implementations available, such as the Prisma + Lambda CDK example:
https://github.com/aws-samples/prisma-lambda-cdk
However, adapting these examples to our framework comes with a few gotchas:

* Ensuring the ORM client is automatically regenerated during the build process
* Managing infrastructure required for database migrations
* Enabling APIs to properly connect to RDB packages
* Defining Nx targets that make it easy to work with one or multiple databases

### Motivation

Make it easier to quickly setup and use relational databases

### Suggested Implementation

Build a ts#rdb generator to support the following options

* Database: Postgres / MySQL
* Engine: Aurora Serverless v2 / DSQL
* ORM Framework: Prisma / MikroORM ...

And a connection api -> ts#rdb if extra setup is required from the API side. At the beginning, maybe start with only Aurora Serverless + Prisma

### Contribute to feature

AlexTo:
### Reason for this change

Currently, nx-plugin-for-aws provides generators for vending APIs, but it does not yet support generators for vending databases.
There are reference implementations available, such as the Prisma + Lambda CDK example:
https://github.com/aws-samples/prisma-lambda-cdk
However, adapting these examples to our framework comes with a few gotchas:

* Ensuring the ORM client is automatically regenerated during the build process
* Managing infrastructure required for database migrations
* Enabling APIs to connect to RDB packages
* Enabling the deployed APIs to connect to the database either directly to the cluster or via RDS proxy
* Managing credentials
* Defining Nx targets that make it easy to work with one or multiple databases for local development with no or minimal manual setup by the user
* Meeting AWS requirements for using SSL with Aurora

### Description of changes

* Create a new `ts#rdb` generator to generate a Typescript database project
* Provide documentation on how to use the generator

* Unit tests
* Manual testing
  * Generate an RDB package using CDK/Terraform with Aurora PostgreSQL, along with a tRPC API package. Deploy it to AWS and verify that the resources are created correctly, migrations run successfully, and the API can connect to the database.
  * Generate an RDB package using CDK/Terraform with Aurora MySQL, along with a tRPC API package. Deploy it to AWS and verify that the resources are created correctly, migrations run successfully, and the API can connect to the database.

### Issue # (if applicable)

Closes #485 .

### Checklist

----


docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+This generator creates a new relational database project backed by [Amazon Aurora](https://aws.amazon.com/rds/aurora/) PostgreSQL. It generates the application code and infrastructure needed to provision and manage a database using AWS CDK or Terraform.
+
+The generated infrastructure includes:
+
+- Reusable Aurora construct or module
+- Application-specific database construct or module  
+- IAM-authenticated application database user handler
+- Runtime configuration registration in AWS AppConfig
+- Migration handler for automatic Prisma migration deployment
```
cogwirrel: Nit: Can we make this a single paragraph rather than bullets? And it's probably a bit too specific in those bullet points about what's generated - this part should ideally be a short paragraph about the generator at a high level, mentioning the key services/frameworks (in this case Aurora and Prisma) and key features (eg declarative schema definition, migrations and ORM client)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+    - create-db-user-handler.ts Lambda handler used to create the application database user during deployment
+    - migration-handler.ts Lambda handler used to run database migrations during deployment
+  - .gitignore Git ignore entries including generated Prisma client output
+  - Dockerfile Container image definition for the migration handler
+  - project.json Project configuration and build targets
+  - prisma.config.ts Configuration for Prisma CLI
+  - tsconfig.json Base TypeScript configuration for source and tests
+  - tsconfig.lib.json TypeScript configuration for source code
+  - tsconfig.spec.json TypeScript configuration for tests
+  - eslint.config.mjs Configuration for ESLint
```
cogwirrel: Nit: can omit these :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+  </TabItem>
+  <TabItem label="bun" icon="seti:bun">
+
+```bash
+bun nx run <your-db-package-name>:generate
+```
+
+  </TabItem>
+</Tabs>
```
cogwirrel: You can use `<NxCommands commands=["generate <your-db-project-name>"] />` :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+  </TabItem>
+  <TabItem label="bun" icon="seti:bun">
+
+```bash
+bunx --bun prisma migrate dev --create-only
+```
+
+  </TabItem>
+</Tabs>
```
cogwirrel: You can use `<PackageManagerExecCommand commands=["prisma migrate dev --create-only"] />` :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+  </TabItem>
+  <TabItem label="bun" icon="seti:bun">
+
+```bash
+bunx --bun prisma migrate deploy
+```
+
+  </TabItem>
+</Tabs>
```
cogwirrel: Same re `PackageManagerExecCommand` :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+- Retrieves database configuration from AWS AppConfig using `RUNTIME_CONFIG_APP_ID` environment variable
+- Generates temporary authentication tokens via AWS RDS Signer for IAM authentication
+- Manages SSL/TLS connections with certificate validation
+- Handles connection pooling through persistent database connection pools
+
+### Creating Migrations
+
+:::note[Run Prisma from the Package Directory]
+When using the Prisma CLI, run commands from your generated database package directory (e.g., `packages/postgres`) so Prisma can find `prisma/schema.prisma` and `prisma.config.ts`.
+:::
```
cogwirrel: Let's put this note under the first set of prisma commands, then if we add a `prisma` target to the generator as follows:

```
"prisma": {
  "executor": "nx:run-commands",
  "options": {
    "cwd": "{projectRoot}",
    "command": "prisma"
  }
}
```

Then you can include in this note that you can also run from the root with `<NxCommands commands=["run <project>:prisma migrate dev --create-only"] />`

packages/nx-plugin/generators.json
```diff
@@ -265,6 +265,12 @@
       "description": "provides React integration to a React website",
       "metric": "g10",
       "hidden": true
+    },
+    "ts#rdb": {
+      "factory": "./src/ts/rdb/generator",
+      "schema": "./src/ts/rdb/schema.json",
+      "description": "Create a relational database project",
+      "metric": "g34"
```
cogwirrel: Might need to rebase on the latest and double check the number, I think this might need to be `g38` now haha :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+```
+
+  </TabItem>
+</Tabs>
+
+:::note[Automatic AWS Migrations]
+For AWS deployments, you don't need to run `prisma migrate deploy` manually. The generated infrastructure automatically creates the application database user and runs migrations during deployment.
+:::
+
+## Connecting from a tRPC API
```
cogwirrel, line 174: Can we add a short sentence here saying something like:

> While this section describes how to connect to your database from a [tRPC API](<link to generator guide>), it also serves as a reference for use in any other TypeScript project.

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+### Using the Prisma Client in Handlers
+
+Import `getPrisma` from your database package and call it inside your handler to get a type-safe Prisma client:
+
+```ts title="packages/api/src/procedures/list-users.ts"
```
cogwirrel: You can add `{8}` after the title to highlight the `getPrisma` line :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+export const listUsers = dbProcedure
+  .output(ListUsersOutputSchema)
+  .query(async ({ ctx: { db } }) => {
+    return db.user.findMany({ orderBy: { id: 'asc' } });
+  });
+```
+
+:::note
+See [Implementing Custom Middleware](./trpc#implementing-custom-middleware) for more details on the middleware pattern.
```
cogwirrel: Love the above section on middleware, awesome!!

I don't think this link will work when deployed, will need say:

>See the ``<Link path="/guides/trpc">`ts#trpc-api` generator guide</Link>`` for more details on the middleware pattern.

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+## Deploying to AWS
+
+The relational database generator creates CDK or Terraform infrastructure based on your selected `iacProvider`.
+
+<Infrastructure>
+<Fragment slot="cdk">
+The CDK construct is created in `common/constructs`. Example usage:
+
+```ts title="packages/infra/src/stacks/application-stack.ts"
+import { Postgres } from ':my-scope/common-constructs';
```
cogwirrel: Let's name this `MyDatabase` as it's based on the name of the database :)

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+    });
+
+    const db = new Postgres(this, 'Db', {
+      vpc,
+      vpcSubnets: {
+        subnetType: SubnetType.PRIVATE_ISOLATED,
+      }
+    });
+  }
+}
```
cogwirrel, line 275: I wonder if we split this into two, starting with:

```ts
new MyDatabase(this, 'Db', {
  vpc,
  vpcSubnets: { subnetType: SubnetType.PRIVATE_ISOLATED },
});
```

Then the text below about what's provisioned, followed up by:

> Your VPC will need xxx subnets, and can be created if needed as follows:

`<vpc code>`

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+
+  tags = local.common_tags
+}
+```
+
+This provisions an Aurora cluster with RDS Proxy, admin credentials, create-db-user Lambda, runtime config registration, migration Lambda, and container registry resources.
+
+The generated infrastructure creates two database users:
+- **Admin user** - Created during cluster provisioning with credentials stored in AWS Secrets Manager
+- **Application user** - Created via a Lambda function with IAM authentication enabled and full privileges on the application database
```
cogwirrel, line 303: Same for terraform can we also add 

> Your VPC will need xxx subnets, and can be created if needed as follows:

`<vpc terraform code>`

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+</Fragment>
+</Infrastructure>
+
+The application user is automatically created with a random name and IAM authentication — `getPrisma()` is already configured to authenticate as this user using short-lived RDS tokens, so your application code never handles database passwords.
+
+### Connecting the API Construct to the Database
```
cogwirrel: Nit: `Connecting an API to the Database`

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+      vpcSubnets: { subnetType: SubnetType.PRIVATE_WITH_EGRESS },
+    })
+    .build(),
+});
+
+Object.entries(api.integrations).forEach(([operation, integration]) => {
+  db.allowDefaultPortFrom(integration.handler, `Allow ${operation} to connect to the database`);
+  db.grantConnect(integration.handler);
+});
+```
```
cogwirrel, line 380: Perhaps add a :::note that this example grants every handler in your API access, but if only some handlers need access it's better to configure individually.

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+:::note[Additional Cost]
+RDS Proxy is enabled by default but incurs additional charges on top of the Aurora cluster cost. See [AWS RDS Proxy pricing](https://aws.amazon.com/rds/proxy/pricing/) for details.
+:::
+
+The generated infrastructure includes an RDS Proxy by default, which sits between your application and the Aurora cluster. RDS Proxy provides several benefits:
+
+- **Connection pooling** - Maintains a pool of database connections that can be shared across application instances, reducing the overhead of establishing new connections
+- **Connection resilience** - Automatically handles failovers and reconnects during Aurora instance replacements or maintenance
+- **IAM authentication** - Supports IAM-based database authentication, eliminating the need to manage database credentials in your application code
+- **Improved security** - Enforces TLS encryption for all connections
```
cogwirrel: I wonder if we restructure to something like:

```mdx
The generated infrastructure includes an RDS Proxy by default, which sits between your application and the Aurora cluster. RDS Proxy provides several benefits:

- **Connection pooling** - Maintains a pool of database connections that can be shared across application instances, reducing the overhead of establishing new connections
- **Connection resilience** - Automatically handles failovers and reconnects during Aurora instance replacements or maintenance
- **IAM authentication** - Supports IAM-based database authentication, eliminating the need to manage database credentials in your application code
- **Improved security** - Enforces TLS encryption for all connections

By default, RDS Proxy is enabled. The generated runtime client (`getPrisma()`) automatically connects through the proxy endpoint.

:::note[Additional Cost]
RDS Proxy is enabled by default but incurs additional charges on top of the Aurora cluster cost. See [AWS RDS Proxy pricing](https://aws.amazon.com/rds/proxy/pricing/) for details. If you would prefer to disable the proxy, see the section below.
:::

#### Disable RDS Proxy

You can disable the RDS proxy as follows:

<Infrastructure>
...
```

e2e/src/smoke-tests/rdb.spec.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { existsSync, rmSync } from 'fs';
+import { ensureDirSync } from 'fs-extra';
+import { buildCreateNxWorkspaceCommand, runCLI, tmpProjPath } from '../utils';
+
+describe('smoke test - rdb', () => {
```
cogwirrel, line 9: Nice! Please could you add this to pr.yml and ci.yml? :)

packages/nx-plugin/src/ts/rdb/files/src/utils.ts.template
```diff
@@ -0,0 +1,99 @@
...
+const getSecretValue = async (secretArn: string): Promise<string> => {
+  const region = process.env.AWS_REGION;
+
+  if (!region) {
+    throw new Error('Missing required environment variable AWS_REGION.');
+  }
+
+  const client = new SecretsManagerClient({
+    region,
+  });
```
cogwirrel: I think it's equivalent to just create the client without specifying a region? :)

packages/nx-plugin/src/ts/rdb/files/src/utils.ts.template
```diff
@@ -0,0 +1,99 @@
...
+
+  return data.SecretString;
+};
+
+export const getDatabaseSecret = async (): Promise<{
+  dbname: string;
+  username: string;
+  password: string;
+  host: string;
+  port: number;
```
cogwirrel: Let's define a type for this as it's referenced below too :)

packages/nx-plugin/src/ts/rdb/generator.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { Tree } from '@nx/devkit';
+import { tsRdbGenerator, TS_RDB_GENERATOR_INFO } from './generator';
+import { createTreeUsingTsSolutionSetup } from '../../utils/test';
+import { METRICS_ASPECT_FILE_PATH } from '../../utils/metrics';
+import { sharedConstructsGenerator } from '../../utils/shared-constructs';
+import { readProjectConfigurationUnqualified } from '../../utils/nx';
+
+const expectHasMetricTags = (tree: Tree, ...metrics: string[]) => {
```
cogwirrel: Can we import this from `utils/metrics.spec.ts`? :)

packages/nx-plugin/src/ts/rdb/generator.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+    );
+    expect(tree.read('packages/db/Dockerfile', 'utf-8')).toContain(
+      'FROM public.ecr.aws/lambda/nodejs:24',
+    );
+    expect(tree.read('packages/db/Dockerfile', 'utf-8')).toContain(
+      'ARG AWS_REGION',
+    );
+    expect(tree.read('packages/db/Dockerfile', 'utf-8')).toContain(
+      'RUN npm install prisma@7.7.0',
+    );
```
cogwirrel: Best not to hardcode versions in unit tests as it'll break our update-versions workflow :) This'll be addressed by the snapshot comment above though :)

packages/nx-plugin/src/ts/rdb/generator.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+    expect(packageJson.dependencies['@prisma/client']).toBe('7.7.0');
+    expect(packageJson.dependencies.pg).toBe('8.20.0');
+    expect(packageJson.dependencies.mariadb).toBeUndefined();
+    expect(packageJson.dependencies['@prisma/adapter-mariadb']).toBeUndefined();
+    expect(packageJson.devDependencies['@types/aws-lambda']).toBe('8.10.161');
+    expect(packageJson.devDependencies['@types/pg']).toBe('8.20.0');
+    expect(packageJson.devDependencies.prisma).toBe('7.7.0');
+    expect(packageJson.devDependencies.ncp).toBe('2.0.0');
+    expect(packageJson.devDependencies.rimraf).toBe('6.1.3');
+    expect(packageJson.devDependencies['make-dir-cli']).toBe('4.0.0');
```
cogwirrel: Likewise we should just check that the dependencies are defined/undefined rather than pinning versions in the tests :)

packages/nx-plugin/src/ts/rdb/generator.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+    expect(auroraConstruct).toContain('this.grantConnect(migrationHandler);');
+    expect(dbConstruct).toContain('export class Db');
+    expect(dbConstruct).toContain(
+      "import { DB_PACKAGE_NAME } from ':proj/db';",
+    );
+    expect(dbConstruct).toContain("databaseName: 'databaseName'");
+    expect(dbConstruct).toContain("adminUser: 'databaseUser'");
+    expect(dbConstruct).toContain('runtimeConfigKey: DB_PACKAGE_NAME');
+    expect(dbConstruct).toContain('findWorkspaceRoot');
+    expect(dbConstruct).toContain('dist/packages/db/bundle/create-db-user');
```
cogwirrel: Let's snapshot these 2 files instead of these assertions :)

packages/nx-plugin/src/ts/rdb/generator.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+      "import { Signer } from '@aws-sdk/rds-signer';",
+    );
+    expect(prismaFile).toContain("import { Pool } from 'pg';");
+    expect(prismaFile).toContain("export const DB_PACKAGE_NAME = 'Db';");
+    expect(prismaFile).toContain('password: async () => {');
+    expect(prismaFile).toContain('await getDatabaseConfig(DB_PACKAGE_NAME)');
+    expect(prismaFile).toContain('allowExitOnIdle: true,');
+    expect(migrationHandler).toContain(
+      "import { Signer } from '@aws-sdk/rds-signer';",
+    );
```
cogwirrel: Likewise instead of these insertions let's snapshot the relevant files/directories :) ie:

```ts
snapshotTreeDir(tree, 'path/to/project/prisma');
snapshotTreeDir(tree, 'path/to/project/src');
expect(tree.read('path/to/project/Dockerfile', 'utf-8')).toMatchSnapshot();
expect(tree.read('path/to/project/prisma.config.ts', 'utf-8')).toMatchSnapshot();
```

Same in the other tests :)

e2e/src/smoke-tests/rdb.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+          redirectStderr: true,
+        },
+      );
+      const projectRoot = `${targetDir}/e2e-test`;
+      const opts = { cwd: projectRoot, env: { NX_DAEMON: 'false' } };
+
+      await runCLI(
+        `generate @aws/nx-plugin:ts#rdb --name=my-db --service=Aurora --engine=Postgres --databaseUser=admin --databaseName=mydb --ormFramework=Prisma --iacProvider=${iacProvider} --no-interactive`,
+        opts,
+      );
```
cogwirrel: Can we also add this to `smoke-test.ts` and update the `application-stack.ts.template` that's used for the CDK deploy test? That way we'll test that the CDK construct deploys successfully and runs migrations etc every release :)

packages/nx-plugin/src/ts/rdb/generator.ts
```diff
@@ -0,0 +1,203 @@
...
+          'pg',
+        ]);
+
+  addDependenciesToPackageJson(
+    tree,
+    runtimeDependencies,
+    options.engine === 'MySQL'
+      ? withVersions(['prisma', '@types/aws-lambda'])
+      : withVersions(['prisma', '@types/pg', '@types/aws-lambda']),
+  );
```
cogwirrel: Let's simplify so we declare the shared deps once, ie:

```ts
addDependenciesToPackageJson(
    tree,
   [
      '@aws-lambda-powertools/parameters',
      '@aws-sdk/client-appconfigdata',
      '@aws-sdk/client-secrets-manager',
      '@aws-sdk/rds-signer',
      '@prisma/client',
      ...(options.engine === 'MySQL' ? ['@prisma/adapter-mariadb', 'mariadb'] : ['@prisma/adapter-pg', 'pg'])
    ],
    ...
  );

```

packages/nx-plugin/src/ts/rdb/schema.json
```diff
@@ -0,0 +1,78 @@
...
+      "type": "string",
+      "description": "Name of the database project to generate",
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      },
+      "x-priority": "important",
+      "x-prompt": "What name would you like your database project to have? i.e: MyDb"
+    },
+    "directory": {
```
cogwirrel, line 18: Let's add `subDirectory` for consistency with other generators:

```
    "subDirectory": {
      "type": "string",
      "description": "The sub directory the project is placed in. By default this is the project name.",
      "x-prompt": "Which sub directory do you want to create the project in? (By default this is the project name)"
    },
```

packages/nx-plugin/src/ts/rdb/schema.json
```diff
@@ -0,0 +1,78 @@
...
+      "default": "Inherit",
+      "x-prompt": "Which provider would you like to manage your infrastructure? (default: Inherit)"
+    }
+  },
+  "required": [
+    "name",
+    "service",
+    "engine",
+    "databaseUser",
+    "databaseName",
```
cogwirrel: Can these two be optional (and not "important") and we choose sensible defaults? eg does it make sense to default to `databaseName = name` and `databaseUser = 'admin'` or similar?

That way you can just do `nx g @aws/nx-plugin:ts#rdb my-db --no-interactive` and all the defaults are used so it's super easy to use :)

packages/nx-plugin/src/utils/rdb-constructs/files/cdk/core/rdb/aurora.ts.template
```diff
@@ -0,0 +1,353 @@
...
+      adminUser: this.adminUser,
+      dbUser: this.dbUser,
+      region: Stack.of(this).region,
+    });
+
+    // TODO: Adding `prisma` package increases the image size by > 240MB,
+    // as the Prisma CLI bundles multiple engine binaries (schema-engine, libquery-engine,
+    // query-engine, prisma-fmt) for the target platform.
+    // This exceed the size limits of the standard NodeJsFunction, so we have use DockerImageFunction instead.
+    // https://github.com/prisma/prisma/issues/29394
```
cogwirrel: Let's get rid of this comment, no need to vend the TODO to everyone :) Docker image function is fine for this use case, especially as we're using the prisma cli to perform the migration :)

packages/nx-plugin/src/utils/rdb-constructs/files/cdk/core/rdb/aurora.ts.template
```diff
@@ -0,0 +1,353 @@
...
+      'Allow the migration handler to connect to the database',
+    );
+    this.grantSecretRead(migrationHandler);
+<%_ } else { _%>
+    this.allowDefaultPortFrom(
+      migrationHandler,
+      'Allow the migration handler to connect to the database',
+    );
+    this.grantConnect(migrationHandler);
+<%_ } _%>
```
cogwirrel: Nit: can call `this.allowDefaultPortFrom` in both cases :)

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -0,0 +1,676 @@
...
+  triggers = {
+    bundle_hash    = local.migration_bundle_hash
+    repository_url = aws_ecr_repository.migration_handler.repository_url
+  }
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      aws ecr get-login-password --region ${data.aws_region.current.region} | docker login --username AWS --password-stdin ${self.triggers.repository_url}
+      docker build --platform linux/arm64 --provenance=false --build-arg AWS_REGION=${data.aws_region.current.region} -t ${self.triggers.repository_url}:${self.triggers.bundle_hash} ${local.migration_bundle_path}
+      docker push ${self.triggers.repository_url}:${self.triggers.bundle_hash}
```
cogwirrel: For consistency with agentcore I wonder if we should move the docker build into a `docker` target in the db project's `project.json`, and ensure `docker` depends on `bundle`?

Worth taking a look at `runtime.tf.template` :)

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -0,0 +1,676 @@
...
+  #checkov:skip=CKV_AWS_116:Dead Letter Queue not required for the create-db-user handler
+  #checkov:skip=CKV_AWS_115:Concurrent execution limit not required for the create-db-user handler
+  #checkov:skip=CKV_AWS_173:Lambda environment variables encrypted by managed key
+  #checkov:skip=CKV_AWS_272:Code signing not configured as deployment packages are built and deployed within the same pipeline
+  filename         = data.archive_file.create_db_user_zip.output_path
+  function_name    = local.create_db_user_function_name
+  role             = aws_iam_role.create_db_user.arn
+  handler          = "index.handler"
+  source_code_hash = data.archive_file.create_db_user_zip.output_base64sha256
+  runtime          = "nodejs22.x"
```
cogwirrel: Should this be node 24 to match the current latest used in CDK? :)

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/core/rdb/aurora/aurora.tf.template
```diff
@@ -0,0 +1,587 @@
...
+}
+
+resource "random_password" "master_password" {
+  length           = 32
+  special          = true
+  override_special = "!#$%&*()-_=+[]{}<>:?"
+}
+
+resource "aws_secretsmanager_secret" "credentials" {
+  #checkov:skip=CKV2_AWS_57:Secret rotation requires a VPC-hosted rotation Lambda; configure rotation post-deployment based on your network topology
```
cogwirrel: Should we mention this in the docs? Perhaps we have a short `Credential Rotation` section under `Infrastructure` and mention that it's set up by default for CDK but for terraform it's not currently configured for you.

packages/nx-plugin/src/utils/rdb-constructs/files/cdk/core/rdb/aurora.ts.template
```diff
@@ -0,0 +1,353 @@
...
+    const key = new Key(this, 'EncryptionKey', { enableKeyRotation: true });
+    this.cluster = new DatabaseCluster(this, 'DatabaseCluster', {
+      ...clusterProps,
+      vpc,
+      vpcSubnets,
+      engine: engine.clusterEngine(engineVersion ?? engine.defaultVersion),
+      writer: writer ?? ClusterInstance.serverlessV2('writer'),
+      credentials: Credentials.fromGeneratedSecret(adminUser, {
+        encryptionKey: key,
+      }),
```
cogwirrel, line 225: Can we add automatic rotation by default? :) Possibly with a flag `enableCredentialRotation` like for the proxy?

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds-readme.html#rotating-credentials

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/core/rdb/aurora/aurora.tf.template
```diff
@@ -0,0 +1,587 @@
...
+    ]
+  })
+
+  tags = merge(var.tags, {
+    Name = "${var.name}-aurora-proxy"
+  })
+}
+
+resource "aws_security_group" "proxy" {
+  #checkov:skip=CKV2_AWS_5:Security group is attached to aws_db_proxy.aurora; Checkov cannot resolve count-conditional cross-resource references
```
cogwirrel, line 470: Is this definitely the case? Annoying limitation if so!

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,684 @@
...
+</Fragment>
+</Infrastructure>
+
+For more details, see the AWS Lambda [SSL/TLS requirements for Amazon RDS connections](https://docs.aws.amazon.com/lambda/latest/dg/services-rds.html#services-rds-tls) and the Amazon RDS Proxy [TLS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy.howitworks.html#rds-proxy-security.tls). When using RDS Proxy, you do not need to configure the RDS CA bundle in your Lambda function.
+
+## Local Development
+
+### Using a Local Database
+
+Start a local Postgres instance (e.g. with Docker) and set `DATABASE_URL` when running Prisma CLI commands from the generated database package:
```
cogwirrel: Is there a good guide to link to here or a command we can add to show users how to do this? Or even better if there's a reliable docker one-liner we could add it as a `serve-local` target? :)

cogwirrel (commented):
Fantastic work @AlexTo ! 🙌

You've designed this really nicely and the pattern fits well in the Nx Plugin, setting up a nice type-safe way to interact with relational databases!

packages/nx-plugin/src/utils/rdb-constructs/files/cdk/core/rdb/aurora.ts.template
AlexTo: oh not the same, Prisma CLI for MySQL currently can't connect to Proxy due to not being able to use `mysql_clear_password` which is required for IAM. The `if` block for MySQL allows connection to the cluster only. I added the note about it in `migration-handler.ts`

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done, doc is updated to reflect that the user should use `nx run <db-package>:prisma` and not worry about the current working dir

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: ah nice

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: done

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: nice, much more organised

e2e/src/smoke-tests/rdb.spec.ts
AlexTo: done

e2e/src/smoke-tests/rdb.spec.ts
AlexTo: done. However, it does not run the migration because the generated code does not generate the initial migration for the `ExampleTable`. I leave the migration to be handled manually by the user because to generate the migration, Prisma CLI needs access to a database which still depends on the user to set it up manually :( But I think maybe we can generate the initial migration for the e2e test to be complete? Although I decided not to generate that as most users will just delete the ExampleTable and create their own tables.

AlexTo: ah I mean, the migration will still be triggered but it will just not find any migration to apply

packages/nx-plugin/src/ts/rdb/files/src/utils.ts.template
AlexTo: true, not sure why I did that :D

packages/nx-plugin/src/ts/rdb/files/src/utils.ts.template
AlexTo: Done. Note the overlapping of the 2 types `DatabaseConfig` and `DatabaseSecret`, because `DatabaseSecret` is the structured generated by Aurora for the admin user. So `username` in `DatabaseSecret` refers to the adminUser. But in our AppConfig, we have `adminUser` as well as `dbUser` that we generate for IAM.

packages/nx-plugin/src/ts/rdb/generator.spec.ts
AlexTo: done

packages/nx-plugin/generators.json
AlexTo: done

packages/nx-plugin/src/ts/rdb/generator.spec.ts
AlexTo: done

packages/nx-plugin/src/ts/rdb/generator.spec.ts
AlexTo: done

packages/nx-plugin/src/ts/rdb/generator.spec.ts
AlexTo: done

packages/nx-plugin/src/ts/rdb/generator.spec.ts
AlexTo: done

packages/nx-plugin/src/utils/rdb-constructs/files/cdk/core/rdb/aurora.ts.template
AlexTo: done

packages/nx-plugin/src/ts/rdb/generator.ts
AlexTo: done

packages/nx-plugin/src/ts/rdb/schema.json
AlexTo: done

packages/nx-plugin/src/ts/rdb/schema.json
AlexTo: done

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
AlexTo: done, docker build target is generated for Terraform only

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: added serve-local

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/core/rdb/aurora/aurora.tf.template
AlexTo: I think it's this one https://github.com/bridgecrewio/checkov/issues/7358 when the resource uses the conditional count and the proxy references the security group using index

packages/nx-plugin/src/utils/rdb-constructs/files/cdk/core/rdb/aurora.ts.template
AlexTo: done

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/core/rdb/aurora/aurora.tf.template
AlexTo: updated

AlexTo: also setup in Terraform

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,748 @@
...
+
+## Local Development
+
+### Running Prisma Commands
+
+Use the `prisma` target to run any Prisma CLI command against the local database. It automatically starts the database via `serve-local` before running, so there is no need to set up a local database separately.
+
+<NxCommands commands={['run <project>:prisma migrate dev']} />
+
+<NxCommands commands={['run <project>:prisma migrate deploy']} />
```
cogwirrel: Should we remove this given we stated above that the custom resource does the deployment? :)

packages/nx-plugin/src/ts/rdb/files/scripts/docker-start.ts.template
```diff
@@ -0,0 +1,80 @@
...
+      `MYSQL_ROOT_PASSWORD=${dbPassword}`,
+    ],
+    ExposedPorts: { '3306/tcp': {} },
+    HostConfig: {
+      AutoRemove: true,
+      PortBindings: { '3306/tcp': [{ HostPort: hostPort }] },
+      Binds: [`${containerName}-data:/var/lib/mysql`],
+    },
+<%_ } else { _%>
+    Image: 'postgres',
```
cogwirrel: Should we worry about versions here matching what's deployed? Might just be a note in the docs!

Also can we use ECR images eg public.ecr.aws/docker/library/postgres:latest? https://gallery.ecr.aws/docker/library/postgres

packages/nx-plugin/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "module": "commonjs",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "moduleResolution": "bundler"
```
cogwirrel: ooh what's this for?

packages/nx-plugin/tsconfig.lib.json
```diff
@@ -5,7 +5,12 @@
     "declaration": true,
     "types": ["node"]
   },
-  "include": ["src/**/*.ts", "bin/**/*.ts", "sdk/**/*.ts"],
+  "include": [
+    "src/**/*.ts",
+    "bin/**/*.ts",
+    "sdk/**/*.ts",
+    "src/ts/rdb/files/scripts/wait-for-db.ts.template"
```
cogwirrel: and this - I think all templates are copied by the `compile` target? :)

cogwirrel (commented):
Thanks so much @AlexTo ! Just a few very minor comments but otherwise approved!

docs/src/content/docs/en/guides/ts-rdb.mdx
```diff
@@ -0,0 +1,748 @@
...
+  engine_version = "8.0.mysql_aurora.3.12.0"
+}
+```
+</Fragment>
+</Infrastructure>
+</OptionFilter>
+
+### Deletion Protection
+
+Configure lifecycle settings to protect against accidental deletion.
```
cogwirrel: Looks like deletion protection is on by default? (which it should be!) Worth mentioning and then updating the example to say how to disable it?

e2e/src/smoke-tests/rdb.spec.ts
cogwirrel: cool that's fine :) it at least checks the flow

docs/src/content/docs/en/guides/ts-rdb.mdx
AlexTo: oh this is for local development. For e.g, when you `git pull` migrations created by other developers, you can deploy to your local db with `migrate deploy`

packages/nx-plugin/src/ts/rdb/files/scripts/docker-start.ts.template
AlexTo: It makes sense to use a matching local version. We're pinning a default version for Aurora
```
defaultVersion = AuroraPostgresEngineVersion.VER_17_7,
```
 so I think we can default to the same version from `ecr.aws` for local Docker and put a note in the docs that it's recommended to match the local version with the version on Aurora.

packages/nx-plugin/tsconfig.json
AlexTo: must be Claude doing something sneaky behind my back

packages/nx-plugin/tsconfig.lib.json
AlexTo: same as above

cogwirrel (approved):
Great work Alex! Excited to make use of this new generator!! 🙌

PR #586, opened 2026-04-18
chore(open-api#py-client): add hidden Python httpx client generator

nx-plugin-for-aws:
> **Note:** the shared-pipeline fixes here that also affect the TypeScript generators landed separately in #1178, and this branch is rebased onto it. What remains in the shared files is the Python-only half — the per-media-type conflict check, the Python name escaping, and the `all-of` shape keying that needs the `effectiveProperties` flattening this branch adds.

### Reason for this change

Future connection generators (React → FastAPI Python client, Python client → tRPC API over OpenAPI, etc.) need a Python client generator symmetric to `open-api#ts-client`. This PR lands the foundation as a **hidden** generator so it can evolve without release-note surface until a connection generator consumes it.

### Description of changes

Adds `open-api#py-client` (hidden) that emits:

- `types.py` — pydantic v2 models for every schema, per-operation error wrapper classes, per-operation request `TypedDict`s.
- `errors.py` — the exception hierarchy, shared by both clients.
- `client.py` — sync client using `httpx.Client` (when `clientType` is `sync` or `both`).
- `async_client.py` — async client using `httpx.AsyncClient` (when `clientType` is `async` or `both`).

Both clients render from a single template, `files/client/__clientModuleName__.py.template`: `generateFiles` substitutes `__clientModuleName__` from the context, so one source emits both modules, and an `isAsync` flag selects the keywords through a flavour table the preamble derives once. See "One template, two clients" below.

`__init__.py` exports the clients, `ApiError` and `types`; the per-operation exception subclasses stay importable from `.errors` for narrowing. The client is added to `.gitignore`, as the React connection generator already does for the TypeScript client.

The shape mirrors `open-api#ts-client` end-to-end:

- One class per spec. Methods are **keyword-only** with required-first argument ordering.
- **Tag grouping**: operations with `tags` are exposed as `api.pet.add_pet(...)` rather than flat methods.
- **Object-body flattening**: when the request body is an object that won't collide with path/query/header params, its fields become top-level kwargs.
- **Positional single-arg** for list/dict bodies (`api.user.create_users_with_list_input([User(...), ...])`).
- **Typed per-op exceptions** (`GetPetByIdApiError`) with `.error` narrowable to a discriminated union (`GetPetById404Error | GetPetById5XXError | GetPetByIdDefaultError`).
- **Streaming** for `application/jsonl` + `itemSchema`, `application/x-ndjson` and `x-streaming`, returning `Iterator[T]` / `AsyncIterator[T]`.
- **Cookies**, **deprecation warnings**, `Field(frozen=True)` for readOnly, `Field(description=...)` from spec, multipart/form-data + binary request bodies, urlencoded bodies, matrix/label path styles, `allowReserved`, `deepObject`, content-based parameters, tagged discriminated unions.

### Pluggability and extensible auth

The extension point is the caller's own `httpx` client, which is the Python counterpart of the custom `fetch` the ts-client accepts and the idiom the plugin already uses elsewhere (`SigV4HTTPXAuth`, `SessionHeaderAuth`):

```python
from myapi import MyApi, MyApiConfig
from myapp.auth.sigv4 import SigV4HTTPXAuth

with httpx.Client(auth=SigV4HTTPXAuth(credentials, "execute-api", region)) as http:
    api = MyApi(MyApiConfig(url=url, httpx_client=http))
    pet = api.pet.add_pet(name="rex")
```

An `httpx.Auth`, event hooks, a custom transport, timeouts, retries, proxies and the client's own headers and query params all apply, on every operation including multipart and streaming ones. A supplied client is never closed by the generated one, so it can be shared. Each `Config` field carries a docstring saying so, and the whole surface is pinned by tests.

### Shared codegen changes

To keep the Python generator as thin as the TypeScript one, language-agnostic concerns live in `codegen-data.ts`:

- `op.requestShape` — neutral description of operation inputs with a `source` tag so language templates translate once.
- `op.errorShape` — neutral error taxonomy (one entry per non-success response), and `op.successResponses` for the responses an operation returns rather than raises for.
- `pythonMethodName` / `pythonTagNames` — member names kept clear of the client's own members and of each other.
- `toPythonType` / `toPythonAnnotation` / `qualifyPythonType` / `toPythonClassName` / `toPythonLiteral` in `languages.ts`, emitting PEP-585 generics and `Literal[...]` enums.
- Models gain `pythonType` / `pythonAnnotation` / `effectiveProperties` (recursive allOf flattening) / `isInlinedByAllOf` / `referencedCollectionKind`.

Only Python-specific concerns stay in `open-api/py-client/generator.ts`, which also calls `annotatePythonData` — so a TypeScript consumer of the same spec never computes Python names and types it does not read. `open-api#ts-client` output is unchanged apart from the lower-case wildcard response-code fix noted below, which has its own regression test.

### Documentation

Adds `docs/.../guides/open-api-py-client.mdx` covering usage, types, errors, streaming, async and authentication. Every example in it is executed against a live FastAPI server as part of validation. The page is deliberately **absent from the sidebar** until a connection generator vends the client.

**Unit tests** — 342 py-client tests, all behavioural suites that invoke the client and assert both the recorded request and the returned value. A mock is a specification, not a catch-all: every request is checked against the routes the spec declares, and each request assertion goes through a helper that fails unless exactly one request was made. Injecting a wrong URL into the emitted `_url` fails 107 tests and survives `-u`; making the client send every request twice fails 46.

Every generated client is also type checked with `ty` — the checker the plugin's Python projects run — resolving imports against the worker's environment so the pinned `httpx`/`pydantic` versions are the ones checked against. `PythonVerifier.typeCheckUsage` type checks a module written against the compiled client and returns `ty`'s diagnostics, which is how the type-safety spec asserts each of 15 misuses is rejected. A separate suite lints whatever the generator emitted with the exact `[tool.ruff.lint].select` a generated Python project vends, and a parity suite normalises the async keywords out of both clients and asserts the remainder is identical, so a fix applied to one template and not the other fails the build.

**End-to-end against a real server** — a FastAPI app covering the shapes the client must handle (tags, aliases, nested models, enums, arrays, cookies, dates, discriminated unions, 204/404/500, text, binary, dicts, multipart, urlencoded, jsonl streaming) is served over a socket and its own OpenAPI document used to generate the client. 40 assertions pass over real HTTP, including a SigV4 `httpx.Auth` signing GET, POST, multipart and streaming requests.

**Other specs** — generated and imported clients for `py#fast-api` and `ts#smithy-api` project specs, the full Swagger petstore v3, petstore-expanded, uspto, api-with-examples, link-example, webhook-example, plus adversarial specs (non-identifier tag names, reserved operationIds, pydantic protected names, nested allOf, recursive refs, alias cycles, enum values needing escaping, primitive stream items).

This surfaced defects fixed here, each with a regression test:

- **Requests** — `spaceDelimited`/`pipeDelimited` ignored their delimiter; a query and header parameter sharing a name decided each other's serialisation; array path params and cookie values serialised via Python's `repr` (`['a', 'b']`, `True`); an object multipart part raised `TypeError` and an array of binaries was sent urlencoded as `b'...'`; a config header comma-joined with the operation's own; `cookies=` warned on every request under httpx 0.28; an `allowReserved` query string was replaced by the client's params.
- **Responses** — an array of integers parsed as one float while annotated `list[int]`; a declared `number`/`date` was returned raw, so dates stayed strings and integers ints; a nullable body raised instead of yielding `None`; a `+json` type with parameters was read as text; a `text/plain` enum was JSON-decoded; a FastAPI-shaped discriminated union (tag defaulted, so not `required`) degraded to a plain union and dropped the tag, so the server rejected it; an error body not matching its declared schema raised `ValidationError` over the error itself.
- **Structure** — sync and async defined separate exception classes, so `except ApiError` from one missed the other; an operation named `close` replaced the method that releases the client; a non-identifier tag emitted a `SyntaxError`; nested allOf emitted a nameless field; alias emission order raised `NameError`; a primitive stream item called `int.model_validate_json`; two properties whose Python names collide silently dropped one.

### One template, two clients

The two client templates were ~970 lines each and ~97% identical, and they had drifted three times — a tagged async stream that failed to forward `aclose()` after the fix landed only in sync, a multipart comment the async side was missing, and two JSDoc comments present on one side only. Every fix had to be written twice and nothing made the second one mandatory, so they are now one template.

Drift is prevented structurally rather than by convention: there is one copy of every client line, so a fix has no second site to miss. The body branches on the flavour only where the two genuinely differ in structure rather than spelling — the two docstrings, and the stream delegate whose sync `yield from` and async `return` are the only forms that propagate closing to the generator that opened the response. `generator.template-invariants.spec.ts` pins that surface: writing an async keyword or an inline `isAsync` branch into the Python body fails the build, which is the path back to duplication. I verified it by injecting one.

The parity test previously compared the two generated clients, which is no longer meaningful once they share a source, so it asserts each flavour's branches directly instead. Reintroducing the `aclose()` leak fails six of its tests plus the stream-abandonment test that measures real connection release.

Output is unchanged, verified three ways:

- **503 generated files byte-identical** through the real generator — 37 specs x 3 `clientType` values, including four branch specs that are the corpus's only coverage of the stream delegate, whole-body multipart and `allowReserved` streams.
- **74/74 raw EJS renders** match the previous two templates *before* any formatting, so byte-identity is a render-level invariant rather than something ruff happens to repair. (The sync `_stream_with_url` signature is pre-wrapped to the spelling ruff produces, in its own output-neutral commit, which is what makes this hold.)
- **No snapshot moved**, and the merged template is confirmed to ship in the compiled package.

Broadening the parity corpus to reach those branches also surfaced two latent defects, fixed here: the missing async multipart comment, and a streaming operation with a void response annotated `Iterator[None]` while yielding the response's raw text lines — which `ty` rejects.

### Fixes from adversarial review

A second review round reproduced each of these by running code; every one has a regression test that fails without the fix.

The harness accepted any request a client made — a mock carried no method or path — which is why several of the rest survived. That is fixed first, and the defects it had been hiding:

- A primitive body under a non-JSON media type was JSON-encoded, putting `"hello"` on the wire under a `text/plain` header.
- A declared `2XX`, or a second concrete 2xx, was raised as an error, so an operation declaring `200` and `204` raised on a 204. A `default`-only operation did the reverse and parsed a 500 as success. `if True:` no longer guards a branch, and the unreachable raise after a `default` is gone.
- Containers in a path, header or cookie serialised via Python's `repr` (`tags=['a', 'b']`) rather than RFC 6570 expansion.
- A tag or operationId with no alphanumerics snake_cased to the empty string, emitting `self.: _Namespace = ...` — a module that does not parse.
- A tagged async streaming operation delegated with `async for … yield`, which never forwards `aclose()`, so breaking out of the loop held a pooled connection until GC. The sync side was already correct.

Generated output now passes the lint rules the plugin itself vends: running the vended selection over a petstore client reported 217 violations (PEP 604 unions, redundant forward-ref quotes, an unconditional `Never` import, over-long docstrings). Tag methods — the form the guide teaches — now carry the operation's docstring.

Shared codegen: the dedup suffix for two inputs reaching one Python name was not re-checked for its own collision (`SyntaxError: duplicate argument`); a spec declaring both `5xx` and `5XX` silently lost one and now throws; a hoisted `oneOf` member of an `allOf` was marked inlined so its union vanished; `effectiveProperties` took the first branch's `isRequired` although `allOf` is a conjunction.

Response handling: a nullable object body raised on JSON `null`; a response declaring JSON alongside another media type, or an empty body, raised a bare `JSONDecodeError`; `x-streaming` with a void response always raised; `_is_file_part` treated any tuple as a file part; a CRLF JSONL stream left a trailing carriage return.

Vended output no longer carries helpers a client never calls (`_send_with_url` appeared in 21 snapshots and was called in none), and the guide's examples no longer reference undefined names. The eight stale `clientType` translations that still said `client.gen.py` are deleted so the translation workflow regenerates them.

### Independent testing against apps the suite did not author

A real-world FastAPI document exposed a defect the whole in-repo corpus missed: a slurping EJS close tag ate the alias line's own newline, so a documented module-level alias landed on the same line as its docstring and `types.py` did not parse. The fixture gap was the cause — across 36 specs, none produced a documented module-level alias, so the emission path had no coverage. That document is now a fixture (`vessel-registry-spec.json`) with 14 tests; reintroducing the tag fails all 14. The Python verifier already rejected the bad output with a precise `SyntaxError`, so the machinery was sound and is unchanged.

To stop relying on specs written by the same hand as the generator, subagents built **126 FastAPI applications** and drove generated clients against them over real HTTP — **206 specs** in total. The sweep now generates 178 and **fails to parse none**; the 28 rejections are all deliberate, each naming the ambiguity and the operation to edit (duplicate operationIds within a tag, a schema-name normalisation collision, a non-discriminated `anyOf`, property-name conflicts, a `date-time` vs `string` conflict). The OK count fell from 188 as those fixes landed, which is the intended direction: specs that previously produced silently-wrong clients now fail loudly.

Ten further defects, each with a regression test:

- **Request bodies** — a `bytes` body was sent as `repr(bytes)`, so three bytes went out as fifteen: FastAPI writes it as `{type: string, contentMediaType}`, and the promotion to binary was scoped to form media types. That promotion also walked only one level of `properties`, so `list[UploadFile]` was typed `list[str]` and sent as text parts with no filename (server 422); it now recurses through `items`, `anyOf`/`oneOf`/`allOf` and `prefixItems`. An enum body under a non-JSON media type was JSON-encoded (`red` arriving as `"red"`) because an enum is a single wire value but not `isPrimitive`. A required body admitting null sent zero bytes, indistinguishable from no body, since `json=None` makes httpx omit it entirely; it now sends `null` with the declared Content-Type.
- **Names** — an operationId spelt with no alphanumerics escaped to the empty string, so its error class was emitted as bare `ApiError`, redefining the base class it derives from: every class after it inherited from the redefinition and `except ApiError` caught the wrong type. A leading-digit operationId emitted `class 42ApiError`. An empty object schema became the self-referential `Empty = dict[str, Empty]`, a `NameError` on import that a parse-only check does not catch. Property names that do not snake_case collided instead of encoding to distinct identifiers.
- **Tags** — two tags differing only in punctuation merge into one namespace downstream, and deduplicating by tag cannot separate them, so an operation was silently dropped. The duplicate-operationId check is now keyed on the normalised tag and reports the id as written with its method and path; the camelCased id it used to report is empty for these names.

Two of these are latent `ts-client` bugs, fixed in the shared pipeline and verified against that generator directly:

- `discriminator.mapping` values are bare reference strings, not `{$ref}` objects, and the parser expected objects. Fixing it took `ts-client` from 0 to 4 working tag dispatches on the same spec.
- A structured body under a media type that is neither JSON nor a form was JSON-encoded while the declared Content-Type was asserted, so an object went out as JSON bytes labelled `text/plain` — a wire form no conforming server accepts. It is now rejected at generation.

Three attempted fixes overreached and were reverted rather than shipped: rewriting nullable composites globally dropped `| null` from `ts-client` (9 failures), a multi-media-type rejection broke JSONL streaming (20 failures, since resolved by exempting streaming content types), and a tag-name conflict check rejected valid specs where the operationId check alone is sufficient.

A third round — six agents over distinct areas (auth and header handling, schema composition, RFC 6570 parameter serialisation, response dispatch, naming, request bodies) — found twelve more, each fixed with a regression test. A self-review of the preceding fixes came first and is in its own commit.

- **Cookies** — `_cookie_value` escaped `= # % < > ^ ` { | }`, all legal RFC 6265 cookie-octets. Servers do not URL-decode cookies, so every base64-padded token (a JWT, a Fernet session, a CSRF token) arrived corrupted and a value already carrying a percent sequence was double-encoded. Separately, operation cookies went to httpx's `cookies=`, which applies them through `http.cookiejar` — and that skips a request already carrying a `Cookie` header, so a caller holding cookie auth as a header (which the `httpx.Auth` extension point the guide documents encourages) silently lost every operation cookie, required ones included. They are now merged into one header alongside the caller's jar, their client-level header and the config header.
- **Content-Type** — an operation declaring a `Content-Type` parameter of its own lost the body's declared media type, since the parameter sits in `header_params` holding `None` when the caller omits it and `setdefault` read the key as already set; the request went out with no Content-Type at all. A differently-cased parameter did not match the exact key, so both were sent — a duplicate RFC 9110 disallows. A whole binary body under `multipart/form-data` wrote that header into a `header_params` that was never declared, raising `NameError` on every call: invisible to `ast.parse`, caught only by the type check.
- **Names that break the import** — a property named after a type the annotations refer to (`str`, `int`, `bool`, `bytes`, `list`, `dict`, `datetime`) made `types.py` unimportable. The class body binds the name to the field default, and `from __future__ import annotations` means pydantic evaluates every annotation on that class in that namespace, so one such field makes its *siblings* unresolvable. A `"` or newline in a wire name was interpolated raw into a Python string literal, closing it early. A spec title that class-cases to nothing emitted `class :`; one spelt `None` emitted a keyword; one spelt `ApiError` bound the name twice in `__init__.py`, so the base exception won by import order and the client class became unreachable from the package root. A digit-leading operationId emitted `class 42Request(TypedDict)`.
- **Names that silently retype** — the classes derived per operation share `types.py` with those declared for schemas, and were minted without consulting them. A spec declaring `GetThing404Error`, `GetThingError` or `GetThingRequestQueryParameters` had that class emitted twice; Python keeps the last, so every reference resolved to the generated wrapper and the operation could never validate its own payload. Hoisting had the same gap one level down: the disambiguated name was not itself checked, so a declared `OuterInner1` was replaced by the schema hoisted out of `Outer.inner`. Neither is visible to `ast.parse` or to the import — only a duplicate-name sweep finds them.
- **Response keys** — every key under `responses` became a status branch whatever it was. A `x-vendor-note` extension, which OpenAPI permits there, emitted a class name containing hyphens and the package could not be imported; an alphabetic key like `unknown` parsed, imported and type checked, then emitted `if response.status_code == unknown:` and raised `NameError` only once that status arrived. Extensions are now dropped and anything that is not a status code, a wildcard range or `default` is rejected.

The parameter-position model names are built in the shared pipeline, so that half of the collision fix applies to `open-api#ts-client` too.

The self-review also found four faults in the preceding commits: three helpers had been inserted directly beneath an existing doc comment, orphaning it onto the new function while the one it described lost it; one docstring had become false as its function's callers widened; and a dead disjunct in the enum-body guard. More usefully, `PYTHON_CLIENT_MEMBERS` — the set that stops an operation *replacing* one of the client's own helpers — was hand-maintained against the template with nothing checking the two agreed, where a missed name breaks every operation that calls that helper. A template invariant now derives the expected set from the template; it caught the next helper added, one commit later.

Five findings from this round are deliberately not addressed here, being design questions rather than escaping bugs: `allOf` combined with `oneOf` at one level silently drops the union (identically in `ts-client`); a non-string discriminator tag degrades to an untagged union; an inheritance-base `discriminator` is ignored by the Python template although the TypeScript one consumes it; exploded-object parameter styles render `key,value` where the specification's own table gives `key=value`; and a collection declared only under a non-JSON media type is still read with `response.json()`.

### Testing against 515 published OpenAPI documents

A fourth round drove the generator over **515 real OpenAPI 3.x documents** — 497 distinct providers from APIs.guru plus 19 hand-picked (Stripe, GitHub, Kubernetes, Twilio, Cloudflare at 12.9 MB and 3,407 operations, OpenAI, Discord, Adyen, Jira, Box, Asana, DigitalOcean, Sentry). Six agents fetched and drove them in batches; every spec was generated, parsed, **imported for real** in a venv, and swept for duplicate top-level declarations and for `types.X` references the emitted `types.py` never declares.

Final result: **404 generate, parse and import cleanly (78.4%)**; 111 are refused with an actionable message; **zero fail to parse, zero fail to import, zero declare anything twice, zero leave a dangling type reference**. The 111 refusals span 55 distinct causes, each naming the schema or operation at fault — genuinely conflicting schemas per media type (19), an `allOf` of real non-objects (8), property names that collide after normalisation (8), schema names that collapse onto one identifier (11), unions of multiple array types that cannot be told apart at runtime (6).

Eleven further defects, each with a regression test. Four are of a kind no amount of parsing or importing reveals:

- A named scalar reached through `items` was inlined while the array wrapping it was substituted without following through, leaving `items` pointing at a schema the same pass deleted. The client was annotated `list[types.IndexName]` with nothing of that name declared — `AttributeError` on every call, after the module parsed, imported and type checked.
- A parameter named `types` shadowed the module every annotation qualifies a model through, so the method's own return annotation resolved to the caller's value. The request went out correctly and the call then failed.
- A schema named `Exception` made `types.py` unimportable: a class shadows a builtin only from its own definition onwards, and `from __future__ import annotations` defers evaluation to module scope, so an annotation above it resolved to the builtin. Whether it happens depends on where the schema sorts. All 75 class-cased builtins are now escaped, checked against `dir(builtins)` by a test.
- An inline `anyOf` under an array parameter's `items` had no name hoisted for the element, so it rendered as `U` — the anonymous schema's name escapes to nothing — and the request model could not be constructed at all.

Five were valid documents the generator refused:

- An `allOf` member that constrains nothing (`{description}`, `{example}`, `{title}`, or `{type: object}` with no properties) was counted as a non-object and failed the whole document. It is how OpenAPI 3.0 annotates a `$ref`, whose siblings are ignored there, so it is the commonest `allOf` idiom in published specs — 30 of 108 rejections. A single-member `allOf` around a scalar is likewise unambiguous. A member declaring `required` is *not* empty, though: it raises a sibling branch's property to mandatory and nothing carries that across, so those keep failing rather than quietly making a required field optional.
- The per-media-type conflict check compared schemas as written, but hoisting rewrites only the preferred media type and does so at any depth — so a `$ref` was compared against the schema it was hoisted from and never matched. One array of objects served as both JSON and XML was refused. A media type that declares no schema is unconstrained rather than a second shape, and no longer counts as one.
- `application/json; charset=utf-8` matched neither hoisting branch, since the JSON case compared exactly while the `+json` fallback stripped parameters. A media type's parameters do not identify it (RFC 9110 §8.3).
- `marshallingKey` keyed an object member on its name, but the normaliser hoists each inline schema to a distinct one, so two union members with byte-identical properties were rejected as declaring "different types".
- A `$ref` whose target omits `type` beside `properties` was treated as a non-object, since hoisting required the explicit `type` the parser already infers.

And one silently wrong value: hoisting a nullable inline object property replaced it with a bare `$ref` and moved `nullable` onto the hoisted definition, where it applied to every other user of that schema and to this property not at all. A required nullable object therefore rejected the null its own spec permits — what a real API returns for an absent sub-resource. This one reaches `open-api#ts-client` too, which now emits the `| null` those properties ask for.

Two fixes in this round were wrong first and corrected against the corpus. Treating a `{required: [...]}` member as constraining nothing made a mandatory field optional, which is worse than the rejection it replaced. And resolving references by expanding every occurrence turned the two largest specs from generating into `RangeError: Invalid string length` and a heap exhaustion — each schema is now expanded once and every level reduced to a digest, with a test that builds a 40-level, 60-way shared document and fails without it.

Most of this round's fixes are in the shared pipeline, so `open-api#ts-client` gains them as well.
### Issue # (if applicable)

N/A. Scoped as `feat`: the generator is hidden from the CLI, but `openApiPyClientGenerator` is exported through the published `./sdk/*` map, so it is new public API and should carry its own minor bump rather than riding a patch derived from unrelated commits.

### Checklist

----


nx-plugin-for-aws:
Thanks — this was a genuinely useful review. I reproduced every claim through the generator before acting on it, which changed two verdicts and turned one "minor" into a must-fix. Rebased onto main (`45465f0c` clears the `local-dev` smoke failure, confirming it was upstream).

## Fixed

**#1 backslash escaping** — confirmed exactly as described (`SyntaxError: unterminated string literal`). Descriptions now go through `toPythonLiteral`, which already escapes backslash-first. Also covers `Field(description=...)` and operation docstrings in both clients.

**Lowercase `5xx` range codes** (filed under #11 as minor) — this is the same severity as #1: it emits `SyntaxError: invalid decimal literal`. Fixed in the shared normaliser, since `open-api#ts-client` had the identical bug for the same spec.

**#3 response-code ordering** — you recommended a comment; I made the invariant hold by construction instead (concrete codes → ranges → `default`, which renders as an unconditional branch). One sort key, and a test that declares `default` first to prove ordering no longer depends on spec key enumeration.

**#9 ts-client acceptance change** — agreed, and this was a regression I introduced. `assertNoClashingPythonNames` now runs in the Python generator, so a TypeScript user never sees a Python diagnostic. Verified: the clashing spec generates a TS client again and still fails loudly for Python.

**#11 object query params** — confirmed `filter={'a': 'x', 'b': 1}` on the wire. Worth noting ts-client emits `filter=[object Object]` for the same spec, so this was shared brokenness rather than a py-only defect. Python now does correct form/explode expansion (`?a=x&b=1`).

**#11 `allowReserved` with `#`** — confirmed `q=a#frag&p=keep`, i.e. `p` is silently dropped into the fragment. `#` is now encoded while other reserved characters stay literal, matching ts-client.

**#11 required+nullable** — fixed, but the report's framing was slightly off: explicit `None` *does* reach the wire (`{"note":null}`) and omitting it is already a `TypeError`, so nothing was lost on the wire. The real defect was `default=None` weakening pydantic-side validation. Fixed and described accurately.

**#4 idempotency test** — added, covering the whole tree plus the `.gitignore` entry and metric that a second run appends to rather than overwrites.

**#5 vacuous `omit_content_type_header` test** — fixed the way its own trailing comment described. Confirmed it's no longer vacuous by disabling the flag and watching it fail.

All six correctness fixes are pinned in a new `generator.hostile-specs.spec.ts`.

## Not fixed, with reasons

**#2 recursive collection aliases** — confirmed, but it is **pre-existing and shared, not introduced here**: for `A: array of B`, `B: array of A`, ts-client emits `Promise<Array<B>>` with no `B` defined, and `tsc` rejects it (`error TS2304: Cannot find name 'B'`). Root cause is the normaliser's "inline non-object refs" pass, which inlines then deletes both schemas but can't fully inline a cycle. I attempted a fix and it moved the failure into `toTypeScriptType`, which recurses infinitely on a cyclic alias — properly supporting these means changing the released ts-client's core type renderer. That's the wrong scope for a hidden-generator PR, so I'm leaving it as a shared follow-up rather than half-fixing it.

**#10 flattening-predicate duplication** — documented rather than migrated, as you suggested as the ideal follow-up. The known divergences (discriminated bodies, `additionalProperties`/union/`patternProperties` bodies, a property named `body`, optional bodies, primitive JSON bodies, `+json` media-type preference) are now spelled out in the generator's doc comment. Migrating ts-client onto `requestShape` means touching the most battle-tested generator here for no user-visible gain in this PR.

**#8 worker stdout hardening** — real hazard, but it only bites a contributor who adds a stray `print()` to a template, and the blast radius is the test harness, never shipped code. Same for the per-worker tempdir. Cheap later.

**#6, #7 and the remaining #11 divergences** — test-coverage gaps and cosmetic parity differences on a hidden generator. Tracked, not fixed.

**Part 2 findings** — all three non-blocking, as noted: the `except (A, B)` formatter quirk is pre-existing on main; mypy `--strict`'s `no-any-return` is cosmetic and flagged by neither pyright nor the `ty` gate this suite uses; the orphaned `async_client.py` after switching `clientType` sits in a gitignored directory.

## Verification

3941 unit tests pass. 40 live-HTTP assertions against a real FastAPI server still pass, as do all 13 documented examples. `ty` is clean across 14 specs including every new hostile one. ts-client is provably untouched: zero file diffs against main and all 228 of its tests pass.

packages/nx-plugin/src/utils/test/py.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+    // Use the same pinned versions the generator declares so the test
+    // environment matches what consumers get at runtime. `ty` joins them so it
+    // resolves the client's imports against that same environment — it is the
+    // checker the plugin's Python projects run.
+    const deps = withPyVersions(
+      OPEN_API_PY_CLIENT_DEPENDENCIES,
+      declaredNames(OPEN_API_PY_CLIENT_DEPENDENCIES.py),
+    )
+      .concat(`ty${PY_VERSIONS.ty}`)
+      .flatMap((spec) => ['--with', spec]);
```
cogwirrel: dependencies should be provided when creating the verifier - follow typescriptverifier

nx-plugin-for-aws: Done — the constructor now takes fully-qualified specifiers, defaulting to the set a generated client needs so existing call sites are unchanged:

```ts
new PythonVerifier()                                  // the py-client's deps
new PythonVerifier(['httpx==0.28.1'])                 // just what you need
new PythonVerifier([])                                // nothing
```

`ty` is still always appended, since the worker type checks with it rather than it being a dependency of the code under test.

The default set moved to `python-dependencies.ts` (a non-test module) so the vitest `globalSetup` that warms uv's cache can import it too — previously both derived the list independently, so the cache being warmed wasn't provably the one the verifier hits.

Covered in `python-verifier.spec.ts` in both directions: a dependency passed in resolves, and one not passed in fails to import.

nx-plugin-for-aws:
## Adversarial review + end-to-end test of `open-api#py-client`

Reviewed at `5302927` with all 44 CI checks green. Findings below were each reproduced by running code — generating a client and inspecting the actual request/response, or executing the emitted Python. Ordered by the sequence I'd suggest fixing them in, since #1 is why several of the others survived.

Two claims raised during review were **refuted** by testing and are *not* defects — noted at the bottom so they don't get "fixed".

---

## 1. BLOCKING — the suite does not verify request URLs or methods

`normaliseMock` (`generator.utils.spec.ts:167`) returns `[{ response: mock }]` with no `method`/`path`, so `_MockMatch.matches()` (`python-worker/worker.py:43`) returns `True` for any request and `NoMockMatched` is unreachable. `expectRequestTarget` exists but is called **once in the whole suite** (`generator.wire-format.spec.ts:161`).

Reproduced in an isolated worktree by making every generated client request the wrong path:

```diff
- return self._base_url + path
+ return self._base_url + "/WRONG" + path        # client.py.template:315
```

Then the workflow CLAUDE.md documents:

```
$ npx vitest run --root packages/nx-plugin src/open-api/py-client -u
Snapshots  20 updated
Tests      2 failed | 230 passed (232)
```

12 committed `.snap` files silently absorbed `/WRONG`. Re-running after that leaves only `generator.wire-format.spec.ts`'s two tests failing. A contributor following "update snapshots if there are failures" would commit a client that requests wrong URLs with an almost-green suite.

The PR description's "fails loudly on an unmatched mock" is not accurate as written.

**Fix:** make `normaliseMock` default to constraining `method` + `path` so a mock is a specification, not a catch-all. Related: 29 `calls?.[0]` sites across 5 files with only 2 `calls.length` assertions — a client sending every request twice passes all of them. And `generator.content-type.spec.ts:92`'s `?? ''` makes a negative assertion unable to distinguish "header omitted" from "no request sent".

## 2. BLOCKING — non-JSON request bodies are JSON-encoded

`client.py.template:649` (async `:647`) is a catch-all emitting `{"json": body}` for **every** media type except multipart / urlencoded / binary:

```python
body = self._dump(body)
request_kwargs: dict[str, Any] = {"json": body}
if body is not None and not self._config.omit_content_type_header:
    header_params.setdefault("Content-Type", "text/plain")
```

For `requestBody: { content: { 'text/plain': { schema: { type: 'string' } } } }`, `send_text("hello world")` puts this on the wire:

```
Content-Type: text/plain
Body bytes  : b'"hello world"'      <- quotes included, 13 bytes
```

`application/xml` object bodies are likewise sent as JSON under an XML content type. ts-client handles this with an explicit primitive branch (`client.gen.ts.template:764`, `String(input)`); py-client has none.

**Fix:** mirror ts-client — route a primitive non-JSON body through `{"content": str(body)}`, parallel to the existing `bodyIsUrlEncoded && isPrimitive` branch at `:641`.

## 3. BLOCKING — `2XX` and `default` responses are raised as errors

`op.result` (`codegen-data.ts:606`) takes the lowest concrete 2xx; `buildErrorShape` (`:304`) then makes **every** other response an error entry. A spec declaring both `200` and `2XX` generates:

```python
if response.status_code == 200:
    return response.json()
if 200 <= response.status_code < 300:
    raise GetXApiError(response.status_code, ...)   # a 201 raises
```

Same for a declared `204` alongside `200`. Separately, a `default`-only operation renders `if True: return response.json()` — every status including 500 parsed as success.

**Fix:** treat `2XX` and `default` as success-eligible fallbacks *after* the concrete codes, and don't let a bare `default` swallow non-2xx. Concrete→wildcard→`default` ordering is otherwise correct and verified.

## 4. BLOCKING — containers serialise as Python `repr`; empty identifiers emit `SyntaxError`

`_scalar` (`:354`) falls through to `str(self._dump(value))`, so:

| Input | Emitted |
|---|---|
| cookie `tags=["a","b"]` | `Cookie: tags=['a', 'b']` |
| object path param | `/p/{'x': 1, 'y': 2}` |
| object header | `x-ctx: {'a': 'b'}` |
| nested object in urlencoded body | `nested=%7B%27a%27%3A+%27v%27%7D` |

`_query` handles objects correctly (`:378`), so the gap is confined to `_url` / `_headers` / `_cookies`. RFC 6570 simple expansion wants `x,1,y,2`; cookies want the `form` delimiter + percent-encoding (ts-client does this at `client.gen.ts.template:701`).

Also: `snakeCase` strips all non-alphanumerics, so a CJK tag (`ペット`) or operationId (`取得`) yields the empty string and the template emits `self.: _Namespace = ...` / `def (` — a hard `SyntaxError`, i.e. an un-importable package with no generation-time error. A leading-digit tag (`123-numeric-start`) gives `self.123_numeric_start` likewise. Guard in `uniquePythonMemberName` / `annotatePythonMemberNames`: synthesise a fallback name or throw naming the offending tag.

## 5. BLOCKING — async tagged streaming leaks the connection on early exit

The two namespace delegations are not equivalent:

```python
# sync/client.py.template:774
yield from self._parent._stream_it(...)

# async/async_client.py.template:772
async for item in self._parent._stream_it(...):
    yield item
```

`yield from` propagates `close()` into the inner generator; `async for … yield item` does **not** propagate `aclose()`. Measured on the real shape:

```
SYNC  inner closed after close():   True
ASYNC inner closed after aclose():  False
ASYNC inner closed after gc:        True    # non-deterministic
```

So `break`-ing out of an `async for` on a **tagged** streaming operation holds an httpx pool connection until GC. Untagged async is correct, which is why this hid.

**Fix:** delegate with a plain `def` returning the inner async generator (which forwards `aclose()`), or wrap in `try/finally` closing it. Relatedly the public return type is `AsyncIterator[T]`, which has no `aclose`, so a caller cannot close it in a type-safe way — consider `AsyncGenerator[T, None]`.

This is the second demonstrated consequence of the two 795-line templates being ~95% identical. Normalising the async keywords shows they have **already** drifted in ways unrelated to async (an `isPositionalBody` doc comment and a multipart Content-Type comment exist only in sync). Please share the body, or add a test asserting the two normalise to the same text so drift fails the build.

## 6. BLOCKING — generated clients fail the lint rules the plugin itself vends

`DEFAULT_RUFF_SELECT` (`utils/format.ts:316`) is `['E','F','UP','B','SIM','I']` and `py#project` targets `>=3.14`. Running that exact selection over the rendered tags client:

```
5   UP045   non-pep604-annotation-optional     (Optional[X] -> X | None)
3   UP037   quoted-annotation
```

Petstore is far worse. Separately `errors.py.template:5` imports `Never` unconditionally but only uses it when an operation has **zero** error responses — a plain **F401** on any real spec, confirmed with ruff.

Every other Python template in the repo already uses `X | None`. As-is, a user generates a client into a `py#project` and their own `lint` target goes red on code they didn't write.

**Fix:** emit `X | None` / `A | B`, drop the redundant quotes under `from __future__ import annotations`, make the `Never` import conditional — and run the vended ruff selection over a rendered client in the suite so it can't regress.

## 7. `if True:` — 38 times in the petstore client

`client.py.template:695`:

```python
if True:
    raise AddPetApiError(
        response.status_code,
        self._error(types.AddPetDefaultError, response.status_code, None),
    )
raise AddPetApiError(response.status_code, None)     # unreachable
```

ts-client suppresses the conditional for `default` (`client.gen.ts.template:802`) and omits the trailing throw when a `default` exists. Please match.

## 8. The public methods customers call have no docstrings

`_add_pet` (private) carries `"""Add a new pet to the store."""`; `api.pet.add_pet` — the form the guide teaches, and what autocomplete shows — has none (`generator.petstore.spec.ts.snap:2289`). The tag-namespace builder (`:771` / async `:769`) string-replaces into `signature(op)` and never emits the docstring. ts-client emits `docString(op, '    ')` on every tag member (`client.gen.ts.template:863`).

## 9. Shared-codegen bugs

- **`codegen-data.ts:504`** — the dedup suffix isn't re-checked for collision, unlike the sibling `uniquePythonMemberName` (`:402`) which loops correctly. Header `fooBarHeader` + query `fooBar` + header `foo_bar` → `SyntaxError: duplicate argument 'foo_bar_header'`, and the header dict binds the wrong value.
- **`normalise.ts:441`** — `5xx`→`5XX` upper-cases unconditionally, so a spec declaring both loses one bucket silently and the surviving *name* can be attached to the other's *schema*. The repo's own precedent (`normalise.ts:51`) throws on such conflicts. The underlying case-insensitivity fix is correct and genuinely repairs previously-invalid ts-client output — but it's an unpinned behaviour change to a released generator, so please add a ts-client regression test for a lower-case wildcard.
- **`codegen-data.ts:160`** — `effectiveProperties` only recurses `all-of` / reads `interface` properties, so an `allOf[{own}, oneOf[A,B]]` member is marked `isInlinedByAllOf` (templates skip it) *and* dropped by the `!prop.name` guard: the union constraint vanishes. TypeScript keeps it as `ParentAllOf & ParentAllOf1`. Either don't set `isInlinedByAllOf` for non-interface members, or fail fast.
- `effectiveProperties` override precedence is first-wins; per JSON Schema `allOf` is a conjunction. At minimum OR `isRequired` across branches rather than taking the first.

## 10. Other correctness (medium)

- Nullable **object** response bodies raise `ValidationError` on JSON `null` — `isNullable` is never propagated to response models (`codegen-data.ts:255/275/287` set it only for request inputs), so the `isModelClass(resp) && !resp.isNullable` guard at `:86` never fires. Nullable *primitives* work.
- A `200` declaring both `application/json` and `text/plain` always emits `response.json()` (`isJsonWire`, `:44`, is compile-time only) — a `text/plain` response throws a raw `JSONDecodeError` out of the client. An empty body on a declared-JSON response does the same.
- `x-streaming` with a void response emits `TypeAdapter(None).validate_json(line)`, which always raises. `resultTypeSync` special-cases `void`; `streamChunkParseExpr` doesn't.
- `_is_file_part` (`:466`) treats any `tuple` as a file part, so a `prefixItems`-typed field passed as `(1, "x")` — honouring the emitted `tuple[int, str]` annotation — crashes httpx with `TypeError: expected str, bytes or os.PathLike object, not int`.
- A malformed line mid-JSONL-stream raises raw `pydantic.ValidationError`, outside the client's own exception hierarchy.
- Minor: CRLF JSONL leaves a trailing `\r` on each item (`iter_text()` doesn't translate newlines); empty non-exploded array query param emits a bare `key=`.

## 11. Dead code in vended output

`_send_with_url` / `_stream_with_url` and the `contextlib` import are emitted **unconditionally** but only called on the `allowReserved` path — so they're dead in most clients (`_stream_with_url` appears in 21 snapshots, called in 0). `hasAllowReservedParams` is already computed at `:203` and gates the `_query_string` helper; gate these too.

Also unused: `PYTHON_BUILTIN` (both templates, `:2-5`), `renderClassModel` (`types.py.template:111`, an empty stub whose comment says "Used inline below" — it isn't), `stripQuotes` (`:35`), `ErrorShapeEntry.statusCodes` + its producer `statusCodesFor` (zero consumers; allocates a 100-element array per wildcard bucket), `RequestInput.fromFlattenedBody` (written, never read).

## 12. Layering

`requestShape`, `errorShape`, `effectiveProperties`, `isInlinedByAllOf`, `referencedCollectionKind` have **exactly one consumer** — the py-client templates — and `canFlattenBodyIntoRequest` (`:212`) encodes Python's flattening rules, which this PR's own docblock documents as *diverging* from TypeScript's on five counts. A genuinely neutral abstraction wouldn't have language-divergent semantics baked in.

The PR correctly pulled `assertNoClashingPythonNames` out of the shared pipeline for precisely this reason, but `annotatePythonData` still runs unconditionally inside `buildOpenApiCodeGenData` (`:134`) for every ts-client invocation, and `languages.ts` more than doubled (235→498 lines, now ~60% Python). Either export `annotatePythonData` and call it from the py-client generator (same shape as the assertion), or gate it — and consider keeping the Python-shaped pieces in `py-client/` until a second language actually consumes them.

The `PythonType` tree replacing string-matching is a real improvement; this is only about where it lives and when it runs.

## 13. Docs, schema, release plumbing

- **Stale translations, user-visible.** All 8 non-`en` values in `schema-translations.json:130-137` say `client.gen.py` / `async_client.gen.py`; the generator emits `client.py`. `generator-parameters.astro:67` prefers the translation over the schema description, so the rendered JP options table shows the wrong filename while EN is right. **Root cause worth fixing, not just the symptom:** `4c38db0b5` updated `.en` in the same commit as the rename, so `translate.ts:611`'s drift check (`existing.en !== description`) sees no drift and will *never* re-translate these. Delete the 8 locale keys and let the workflow regenerate.
- **`open-api-py-client.mdx:171`** uses an undefined `MyAuth()` in the flagship pluggability example; `url` is an undefined bare name at `:158`, `:171`, `:232`. 18/19 examples ran verbatim against a live FastAPI server — these are the gap. `:136`'s `types.GetPetById5XXError` also only exists for a spec literally declaring `5XX`, so it won't typecheck against the FastAPI-derived spec the surrounding narrative implies.
- The page has **zero inbound links** anywhere in `docs/src` but *is* in `sitemap-0.xml` for all 9 locales, so it's search-indexable while unlisted. Docs build is clean (1513 pages, link validator passes). Since the repo has no `draft:` convention, either `sidebar: { hidden: true }` (Starlight supports it, and it marks the omission as deliberate) or hold the guide back with the connection generator that vends the client — the latter also takes ~2,300 translated lines out of a 22k-line diff.
- **`chore` is the wrong type.** `sdk/open-api.ts:12` exports `openApiPyClientGenerator` through the published `./sdk/*` map, so a consumer can call it today. With `specifierSource: "conventional-commits"`, `chore` yields no bump and new public API ships in a patch derived from unrelated commits. Precedent leans `feat`: `ts#sync` (`af04dc444`), `license#sync` (`635c4da55`), `open-api#ts-metadata` (`8fcfa0477`) were all `feat` despite being hidden; only `internal#test-matrix` used `chore`, and it isn't SDK-exported. Either `feat`, or drop the SDK export until consumed.
- Not in **either** e2e matrix (`e2e/src/smoke-tests/generator-matrix.ts`, `internal/test-matrix/generator.ts`); CONTRIBUTING requires both.
- `clientType` lacks the `x-prompt` nearly every other user-facing enum option carries — fine while hidden, flagging for the unhiding since the guide advertises the Nx Console UI flow.

## 14. Test coverage gaps

- **Async is barely exercised:** 98 sync behavioural invocations vs **6** async; 7 sync streaming vs **1**; 20 `client.py` snapshots vs **1** `async_client.py`; **zero** `errors.py` snapshots. Finding #5 lives exactly in that gap.
- No behavioural spec ever passes `clientType` — all three values are only file-existence checked.
- Zero coverage for: stream abandonment (`GeneratorExit`/`islice`), concurrency, `TimeoutException`, unicode payloads, binary responses (the harness supports `bytes_b64`; no spec uses it).
- Verified uncaught: changing `buffer += chunk` to `buffer = chunk` (`:700`) passes the entire suite, because the mock returns a single chunk — multi-chunk JSONL boundary corruption is invisible.
- **Test count:** I measure **232** tests for py-client + harness in ~34s, of which ~167 are py-client's own. The rest are parasitic re-registration — `generator.spec.ts` imports `utils/metrics.spec` (+14 CDK/Terraform metric tests) and `petstore.spec.ts` pulls in `ts.spec` (+4 TypeScript tests), so the run prints `✓ should not throw for valid typescript with a dependency` inside a py-client file. I could not reconstruct the claimed 361/376 by any counting method. Worth importing the helpers from non-`.spec` modules.
- **Worker hygiene:** 24 `uv` workers spawn (one per spec file); teardown is per-suite `afterAll`, so a vitest hard crash leaks them. My e2e run left ~200 orphaned `/tmp/pyclient-*` dirs (`worker.py:132` `mkdtemp`, last one never cleaned). `warm-python-cache.ts` swallowing all errors is right for a missing `uv`, but means a broken cache degrades silently to 24 racing cold-cache workers — the original CI failure mode.

---

## Verified working (no action needed)

Worth recording so it isn't re-litigated. **Wire format:** `spaceDelimited`/`pipeDelimited`/csv/multi, matrix + label path styles with and without explode, `deepObject` including nested objects and arrays, `allowReserved` keeping reserved chars literal while `#` is correctly percent-encoded, `allowReserved` folding the caller's client-level `params` in without loss while still running auth + event hooks, ISO-8601 dates, `true`/`false` bool spellings, `False`/`0` correctly **not** dropped (`is None`, not truthiness), query vs header collection formats tracked separately, content-based (`content:`) JSON params. **Bodies:** multipart via `files=` with per-part content types from `encoding`, binary parts, list fields as repeated parts, urlencoded flat objects with arrays as repeated keys and `None` omitted, primitive urlencoded sent verbatim, `omit_content_type_header`, and body-derived Content-Type correctly winning over a config header without comma-joining. **Responses:** concrete→wildcard→`default` order, `+json` suffix, `; charset=` stripped, `TypeAdapter` vs `model_validate` dispatch, binary → `.content`, tuples positional. **Errors:** single shared `ApiError` base in `errors.py` — confirmed `client.ApiError is async_client.ApiError`, and `except ApiError` catches both flavours. **Lifecycle:** `_owns_client` guards `close()`/`aclose()`; a supplied client survives and stays usable. **Naming:** keywords, pydantic `model_*`, and client members (`close`, `url`, …) all escaped; docstring and enum escaping correct. **Thread safety:** no shared mutable state; `field(default_factory=dict)` used correctly.

Also verified: **idempotency is real** (3× re-run byte-identical; `.gitignore` and metric each stay at one entry — `generator.spec.ts:139` is a genuinely good test); **`ty` type checking is real** (a deliberately wrong annotation fails the suite; its cache keys on file *content*, not path, and only stores the fingerprint after a clean check, so no false pass; a missing binary surfaces as a failure); **ts-client output otherwise unchanged** (352 tests green, no snapshot moved, no ts-client file touched); `hidden: true` + metric `g72` unique and next-free; MCP correctly excludes the generator from both `list-generators` and `generator-guide`; licence headers correct everywhere including their deliberate absence from templates; translations structurally sound (14 code blocks, byte-identical bodies, bot-authored per `translate-docs.yml`); `formatFilesInSubtree` last, Tree-only, no direct `fs`.

The `PythonVerifier` / `ty` integration, `typeCheckUsage` asserting 15 misuses are actually rejected, and "compile whatever the generator emitted rather than a hardcoded file list" are all better than the TS side. The defect list in the description is real and each item has a regression test.

## Refuted — please do NOT "fix" these

1. **The urlencoded / multipart regression assertions at `generator.spec-compliance.spec.ts:72` and `generator.content-type.spec.ts:151` are not dead.** I forced the urlencoded branch to fall through to the JSON path; the assertion **fired**, and `expect(res.calls?.[0]?.body).toBe('username=alice')` caught it too.
2. **`_owns_client` inversion is not uncaught** — `generator.pluggability.spec.ts` catches it ("leaves a caller-supplied httpx client open").
3. `except json.JSONDecodeError, KeyError:` in the FastAPI template is **valid** [PEP 758](https://peps.python.org/pep-0758/) under the `>=3.14` target, and `ruff format` rewrites *to* that form. Not a bug.
4. The async client exposing `aclose()` rather than `close()` is correct httpx convention.

## Suggested order

1 (harness first — it's why 2/3/4/5 survived) → 2 → 3 → 4 → 5 (ideally by de-duplicating the templates) → 6 → 7, 8 → 9 → 10 → 11, then matrices / `feat` / translations / docs.

<sub>Reviewed with a 6-agent fan-out: cogwirrel-review-style research over ~500 PRs, template correctness, shared-codegen regression risk, test-suite mutation testing, a live FastAPI end-to-end run, and docs/plumbing. Every finding above was verified against code at `5302927`; mutation experiments ran in a throwaway worktree and the tree was left clean.</sub>

PR #620, opened 2026-04-28
fix(open-api#ts-client): correct Date/Blob checks and const schema handling

nx-plugin-for-aws:
### Reason for this change

Stress-testing the ts-client generator against a real FastAPI service surfaced two correctness bugs:

1. **Dead `typeof` checks for Date/Blob primitives.** The composite `toJson`/`fromJson` loops emitted `if (typeof model === "Date") { return model; }`, but `typeof` never evaluates to `'Date'` or `'Blob'`. For composite aliases containing a `date-time` branch this meant the Date branch was never taken — Dates were either returned unserialised (breaking JSON bodies) or deserialised incorrectly.
2. **Phantom `_String`/`Kind` references from OpenAPI 3.1 `const`.** FastAPI emits `Literal['X']` as `{type: 'string', const: 'X'}`. On discriminated oneOf unions `hey-api-openapi-ts` synthesised a reference to a never-declared type (e.g. `_String`, or a name derived from `title`), breaking `tsc --strict --noEmit` with `Cannot find name '_String'`.

### Description of changes

- `client.gen.ts.template` — in the composite `toJson`/`fromJson` primitive loops, branch on `primitive.typescriptType`:
  - `Date` → `if (model instanceof Date) { return model.toISOString()[.slice(0,10)]; }` / `if (typeof json === 'string' || json instanceof Date) { return new Date(json); }`
  - `Blob` → `if (model instanceof Blob) { return model; }` / `if (json instanceof Blob) { return json; }`
  - everything else keeps the existing `typeof` check.
- `normalise.ts` — new `rewriteConstToEnum` pass at the top of `normaliseOpenApiSpecForCodeGen`. Walks the whole spec; for any object where `'const' in obj && !('enum' in obj)`, sets `obj.enum = [obj.const]` and deletes `obj.const`. The two forms are semantically equivalent in OpenAPI 3.1 and the enum form is handled correctly by the rest of the pipeline.
- `languages.ts` — defence-in-depth: in `toTypeScriptType`'s `reference`/default branch, short-circuit to `toTypescriptPrimitive(property)` when `PRIMITIVE_TYPES.has(property.type)` so a reference that resolves to a primitive can never produce a phantom named type.

- Added two regression tests in `generator.composite-types.spec.ts`:
  - A composite alias with a `date-time` branch — asserts the generated code uses `instanceof Date`, no `typeof model === "Date"` remains, and round-trips a `Date` through `$IO.MaybeDate.toJson`/`fromJson` plus an end-to-end client call.
  - A FastAPI-style discriminated oneOf with `const` — asserts no `_String` phantom references leak through, `validateTypeScript` passes in strict mode, and snapshots both `types.gen.ts` and `client.gen.ts`.
- `pnpm nx run @aws/nx-plugin:test` — 1736 tests pass, existing composite / fast-api snapshots do not regress.
- `pnpm nx run @aws/nx-plugin:lint` — clean.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/open-api/ts-client/__snapshots__/generator.composite-types.spec.ts.snap
+++ b/packages/nx-plugin/src/open-api/ts-client/__snapshots__/generator.composite-types.spec.ts.snap
@@ -354,6 +354,254 @@ export class TestApi {
 "
 `;
 
+exports[`openApiTsClientGenerator - composite schemas > should handle FastAPI-style discriminated unions with string const (Literal) 1`] = `
+"export type Circle = {
+  kind: string;
+  radius: number;
+};
+export type Kind = 'circle';
+export type Kind1 = 'square';
+export type ShapesOut = {
+  canonical: ShapesOutCanonical;
+};
+export type ShapesOutCanonical = Circle | Square;
+export type Square = {
+  kind: string;
+  side: number;
+};
+export type GetShapesError = never;
+"
+`;
+
+exports[`openApiTsClientGenerator - composite schemas > should handle FastAPI-style discriminated unions with string const (Literal) 2`] = `
+"import type {
+  Circle,
+  ShapesOut,
+  ShapesOutCanonical,
+  Square,
+} from './types.gen.js';
+
+/**
+ * Utility for serialisation and deserialisation of API types.
+ */
+export class $IO {
+  protected static $mapValues = (data: any, fn: (item: any) => any) =>
+    Object.fromEntries(Object.entries(data).map(([k, v]) => [k, fn(v)]));
+
+  public static Circle = {
+    toJson: (model: Circle): any => {
+      if (model === undefined || model === null) {
+        return model;
+      }
+      return {
+        ...(model.kind === undefined
+          ? {}
+          : {
+              kind: model.kind,
+            }),
+        ...(model.radius === undefined
+          ? {}
+          : {
+              radius: model.radius,
+            }),
+      };
+    },
+    fromJson: (json: any): Circle => {
+      if (json === undefined || json === null) {
+        return json;
+      }
+      return {
+        kind: json['kind'],
+        radius: json['radius'],
+      };
+    },
+  };
+
+  public static ShapesOut = {
+    toJson: (model: ShapesOut): any => {
+      if (model === undefined || model === null) {
+        return model;
+      }
+      return {
+        ...(model.canonical === undefined
+          ? {}
+          : {
+              canonical: $IO.ShapesOutCanonical.toJson(model.canonical),
+            }),
+      };
+    },
+    fromJson: (json: any): ShapesOut => {
+      if (json === undefined || json === null) {
+        return json;
+      }
+      return {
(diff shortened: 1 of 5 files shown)


nx-plugin-for-aws:
Great set of fixes — I independently found the same three issues when stress-testing a FastAPI against the generators and landed on essentially the same shape of fixes (template branch, const→enum normaliser, `languages.ts` defence-in-depth). Two small comments on `rewriteConstToEnum`:

**1. `walk` mutates in-place; consider using the `cloneDeepWith` idiom the rest of this file uses.**

The existing spec transforms in `normalise.ts` — [line 55](https://github.com/awslabs/nx-plugin-for-aws/blob/main/packages/nx-plugin/src/open-api/utils/normalise.ts#L55) (schema-name rewriting) and [line 462](https://github.com/awslabs/nx-plugin-for-aws/blob/main/packages/nx-plugin/src/open-api/utils/normalise.ts#L462) (ref inlining) — use `cloneDeepWith(spec, (v) => ...)`, which reads as "return a new value for this node or undefined to continue recursing". The new `walk(obj)` recursive mutator diverges from that pattern.

It's safe in practice because `normaliseOpenApiSpecForCodeGen` already clones `inSpec` via `cloneDeepWith(inSpec)` at the top (line 288), so the caller's spec isn't affected. But the mutation is still surprising at the call site — anyone reading `spec = rewriteConstToEnum(spec)` would assume (reasonably) that it returns a new object. A `cloneDeepWith` customiser would both match the house style and remove the potential for surprise:

```ts
const rewriteConstToEnum = (spec: Spec): Spec =>
  cloneDeepWith(spec, (v) => {
    if (
      v && typeof v === 'object' && !Array.isArray(v) &&
      'const' in v && !('enum' in v)
    ) {
      const { const: c, ...rest } = v;
      return cloneDeepWith({ ...rest, enum: [c] });
    }
  });
```

(Alternative: fold the const rewrite into the existing ref-inlining pass at line 462 as a sibling branch, so there's one less whole-spec walk.)

**2. The regression test `should handle FastAPI-style discriminated unions with string const (Literal)` is more lenient than it could be.**

```ts
expect(types).toMatch(/kind:\s*('circle'|"circle"|string|Kind\b)/);
```

Looking at the snapshot, what actually lands is `kind: string` — the `rewriteConstToEnum` ends up hoisting `Kind = 'circle'` and `Kind1 = 'square'` aliases (visible in the snapshot) but the property type lines still reference bare `string` via the `languages.ts` defence-in-depth. That means:
- The orphaned `Kind`/`Kind1` aliases are dead code.
- The property type lost the literal information (`string` instead of `'circle'`).

Neither is a regression — the client compiles and works — but it's worth asserting specifically in the test which path is taken, rather than accepting any of four possibilities. If the intent is "const should produce a Literal type", the test should fail if the output degrades to bare `string`; if the intent is "just compile cleanly", that's worth stating in the assertion. Same for `kind?: string` vs `kind?: 'circle' | Kind`.

Happy to take either direction — just flagging the ambiguity.

cogwirrel (approved):

cogwirrel (approved):

PR #660, opened 2026-05-12
feat(agentcore-gateway): new generator for AgentCore Gateway

nx-plugin-for-aws:
### Reason for this change

Amazon Bedrock AgentCore Gateway lets agents consume tools from multiple MCP servers (and other backends) through a unified MCP endpoint with centralised authentication, optional Cedar-based authorization, and observability. The plugin previously had no generator for this — users wiring agents to multiple MCP servers had to reach for the MCP connection per server, which doesn't scale and doesn't give them policy-based access control.

This PR adds end-to-end support for AgentCore Gateway, including optional per-gateway Cedar policies, IAM SigV4 outbound to MCP server targets, IAM inbound from agents, and a local gateway for fast iteration without AWS.

### Description of changes

Four new generators, composed via the top-level `connection` generator:

1. **`agentcore-gateway`** scaffolds a gateway project at `packages/<name>/` containing `serve-local.ts` — a **local gateway** (express + MCP SDK aggregator) run by the `serve-local` target on its own assigned port, exposing every attached MCP server's tools as `<target>___<tool>` to match the deployed gateway's namespacing. Options:
   - `cedarPolicy` (default `true`): attaches a `PolicyEngine` in `ENFORCE` mode with one `Policy` per `.cedar` file in `policies/` (nested folders supported; policy names get a unique suffix since they are unique account-wide). Policies are EJS templates rendered at synth/plan time with `gatewayArn` and `accountId` in scope (`ejs` is added as a workspace devDependency) — by the CDK construct directly, or by a vended `render-cedar.cjs` invoked via the `external` data source in Terraform. The docs and vended README show how to add custom template variables (e.g. exact agent role ARNs).
   - `infra` (`agentcore` | `none`, default `agentcore`): with `none` no infrastructure is generated; re-running with `agentcore` adds it idempotently, matching other infra-vending generators.
   - CDK uses the stable `aws-cdk-lib/aws-bedrockagentcore` Gateway L2; Terraform uses the native `aws` provider resources (`aws_bedrockagentcore_gateway` / `_policy_engine` / `_policy`, provider >= 6.50).
   - The gateway role's policy-evaluation grant is scoped to the policy engine ARN plus this account/region's gateway ARNs (authorization actions are evaluated against both; the gateway's own ARN can't be referenced without a circular dependency).
   - CDK vends a shared `AgentCoreGateway` construct in `core/agentcore-gateway` (gateway, target registration, readiness probe, Cedar policy loading via an optional `cedarPolicyPath` + `cedarPolicyVariables`); per-gateway constructs are thin subclasses.
   - Fresh-deploy reliability: Cedar policies referencing target actions depend on the targets (Terraform exposes `policy_dependencies`), and a readiness probe (CDK: trigger Lambda vended as a file asset; Terraform: `uv run --with boto3` local-exec in the runtime module) invokes each target MCP runtime with an MCP `initialize` through its cold start (3 min max, 5s interval) before targets are created — `GetAgentRuntimeEndpoint` reports READY ~13s before the runtime actually serves, so invoke-probing is required.
2. **`agentcore-gateway#mcp-connection`** wires a gateway to an MCP server: IAM SigV4 outbound (`iamCredentialProvider.service="bedrock-agentcore"`) when deployed, and registers the server in the gateway's `serve-local.ts` (via GritQL) + serve-local dependency chain for local development.
3. **`ts#agent#gateway-connection`** emits a per-gateway MCP client — SigV4-signed to the deployed gateway URL (resolved from runtime config), or plain HTTP to the local gateway when `SERVE_LOCAL=true` — and AST-patches the agent to use it.
4. **`py#agent#gateway-connection`** the Python mirror.

Also:
- Extends runtime-config with an `agentcore.gateways` namespace and `getConnectedGatewayUrl` / `get_connected_gateway_url` helpers (TS + Py)
- Registers all four connection edges in the connection dispatcher; gateways are identified by project-level generator metadata (`addGeneratorMetadata` / `readAgentCoreGatewayMetadata`)
- Shared `utils/agent-connection` helpers wire MCP clients into agents, used by both the mcp-connection and gateway-connection generators (TS + Py)
- Ships a default account-scoped `permit-all.cedar` so Cedar gateways boot with an explicit-permit posture
- `addMcpServer(server)` on the vended construct infers the gateway target name from the construct id; CFN child ids derive from target names so renames replace cleanly
- CDK imports in the vended construct are namespaced so a gateway named `gateway` compiles (no clash with the CDK `Gateway` class)
- `<OptionFilter>` now supports boolean options (used by the guide's Cedar sections)

**Unit tests**: cover scaffolding (across `cedarPolicy` and `infra` variants, including the infra=none → agentcore upgrade), CDK/Terraform template emission, project metadata, the runtime-config gateway-url helper, serve-local.ts GritQL registration and idempotency, and validation failure paths for each connection generator. The full plugin suite passes.

**End-to-end (deployed, CDK)**: the package-manager smoke tests run the full generator matrix and `build --all`; `cdk-deploy` deploys the gateway and invokes tools through it. Additionally verified manually in fresh workspaces: CDK deploy to us-west-2 with a Cedar gateway (including a nested `policies/team-a/` policy), invoking tools on both TS and Python deployed agents through the live gateway with the scoped IAM grant.

**End-to-end (deployed, Terraform)**: verified manually in a Terraform workspace deployed to us-west-2 — gateway + Cedar policy engine + EJS-rendered policies (including a nested folder) + both MCP server targets reach READY, and tools were invoked on the deployed agent through the live gateway endpoint.

**End-to-end (local)**: the serve-local smoke test runs a gateway-fronted agent alongside the direct MCP connections in the same project (Nx dedupes the shared continuous targets); the agent's serve-local starts the local gateway and both MCP servers, and gateway-prefixed tool calls route to the correct upstream server — all without AWS.

### Issue # (if applicable)

N/A

### Checklist

----


nx-plugin-for-aws:
Pushed fixes addressing the review issues from end-to-end testing:

- Empty interface lint failure on the generated `agentcore-gateway-mcp-client.ts` (Bug #1)
- Default `permit-all.cedar` now scopes to the deploying AWS account so the AgentCore policy validator accepts it on update (Bug #5) — adds `${account_id}` placeholder substitution alongside the existing `${gateway_arn}` in CDK (`Stack.of(this).account` via `Lazy.string`) and Terraform (`data.aws_caller_identity.current.account_id`)
- Local `SERVE_LOCAL` multiplex now mirrors the deployed gateway's `<target>___<tool>` tool-name prefixing so 1-to-many topologies don't collide on tool names (Bug #3) — TS uses a transport-level `PrefixingTransport` wrapping `StreamableHTTPClientTransport`; Py uses Strands' built-in `prefix=` constructor arg
- `CfnGatewayTarget` now depends on the gateway role's policy update so CFN doesn't probe the new target before its IAM permission is in place, avoiding "DeleteGatewayTarget can't be performed when in Creating state" rollback failures (Bug #9)
- Back-propagation no longer leaves blank lines between entries in the gateway client files (Bug #6)
- Terraform policy resource names now PascalCase-normalize matching the CDK template, satisfying the AgentCore `name` regex regardless of whether the `.cedar` file uses kebab- or snake-case (Bug #8)
- Rewrote the "Cedar Policies" guide section as **Writing Policies** with placeholders, validator gotchas, and common patterns (Bug #4)

E2E generator matrix: added `agentcore-gateway` and all four connection edges (`ts/py#strands-agent → agentcore-gateway`, `agentcore-gateway → ts/py#mcp-server`); CDK and Terraform deploy fixtures now provision the gateway with multiple MCP server targets fronting both agent runtimes.

End-to-end re-tested in a fresh workspace (TS + Py agents × 2 gateways × TS + Py MCP servers): full build passes; deploy succeeded in 100s; redeploy with new Cedar policy succeeded in 22s (Bug #5 verified — no longer fails with `Resource stabilization failed`); deployed gateway tool calls work; Cedar enforcement still denies admin caller when policy is restricted to the agent role; local SERVE_LOCAL multiplex now produces deployed-style `<target>___<tool>` names (Bug #3 verified on TS and Py).

Unit tests: 1895 pass.

docs/src/content/docs/en/guides/connection/agentcore-gateway-mcp.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+The Gateway must have `protocol: MCP` and the MCP server must have `auth: IAM` — the generator validates both. Non-IAM MCP servers cannot be attached because the Gateway signs outbound traffic with SigV4.
+
+## Usage
+
+### Run the Generator
+
+<RunGenerator generator="connection" />
+
+Select the Gateway project as the source and the MCP server project as the target. If either project contains multiple components, specify `sourceComponent` and `targetComponent` to disambiguate.
```
cogwirrel: Source cannot have components? Just mention target component

docs/src/content/docs/en/guides/connection/agentcore-gateway-mcp.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+### Options
+
+<GeneratorParameters generator="connection" />
+
+## Generator Output
+
+The generator does **not** emit new source files. Instead, it wires existing projects together:
```
cogwirrel: Make more consistent with other generators and include modified files in file tree

docs/src/content/docs/en/guides/connection/agentcore-gateway-mcp.mdx
```diff
@@ -0,0 +1,113 @@
...
+<Infrastructure>
+<Fragment slot="cdk">
+In the stack where you instantiate the Gateway, register the MCP server as a target:
+
+```ts title="packages/infra/src/stacks/application-stack.ts" {5-8}
+const mcpServer = new MyMcpServer(this, 'MyMcpServer');
+const gateway = new MyGateway(this, 'MyGateway');
+
+// Register the MCP server as a target of the Gateway
+gateway.addMcpServerTarget('MyMcpServer', {
```
cogwirrel: Add a method on the gateway for adding an mcp server by passing it straight in, inferring the name from the name of the mcp server By default

docs/src/content/docs/en/guides/connection/agentcore-gateway-mcp.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+<NxCommands commands={["<gateway-name>-serve-local"]} />
+
+will start every attached MCP server on its assigned local port. Agents connected to the Gateway via the <Link path="guides/connection/ts-strands-agent-gateway">TS</Link> or <Link path="guides/connection/py-strands-agent-gateway">Python</Link> gateway-connection generators will multiplex directly to the MCP servers using their client's local fallback path.
+
+:::caution[Local fidelity gap]
+Local development **bypasses the Gateway entirely**. The generated gateway client detects `SERVE_LOCAL=true` and connects directly to each attached MCP server over plain HTTP. As a consequence:
+
+- **Cedar policies are not evaluated locally.** Every agent sees every tool on every attached MCP server. Test Cedar policies against a deployed Gateway (see the `GATEWAY_URL_OVERRIDE` escape hatch in the agent-connection guides).
+- **Tool names are not prefixed locally.** When running against a deployed Gateway, tool names are prefixed with `<target-name>___`. Local multiplexing preserves the tool's original name.
```
cogwirrel: We should prefix locally too if possible

docs/src/content/docs/en/guides/connection/py-strands-agent-gateway.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+Additionally, the generator:
+
+- AST-patches your agent's `agent.py` to import the Gateway client, resolve its return shape, and register the client(s) in `tools`
+- Adds `agent_connection` as a workspace dependency of the agent
+- Wires the agent's `<agent>-serve-local` target to depend on the Gateway's `<gateway>-serve-local` aggregator
+
+## Using the connected Gateway
+
+The generated Gateway client exposes a single `create(session_id)` factory that returns either a single `MCPClient` (deployed) **or** a `list[MCPClient]` (local multiplex across attached MCP servers).
```
cogwirrel: There have been some session id changes, rebase and adjust to be consistent

docs/src/content/docs/en/guides/connection/py-strands-agent-gateway.mdx
```diff
@@ -0,0 +1,186 @@
...
+import contextlib
+from contextlib import contextmanager, ExitStack
+from strands import Agent
+
+from my_scope_agent_connection import MyGatewayClient
+
+@contextmanager
+def get_agent(session_id: str):
+    my_gateway = MyGatewayClient.create(session_id=session_id)
+    my_gateway_clients = my_gateway if isinstance(my_gateway, list) else [my_gateway]
```
cogwirrel: This is too hacky and not type safe

docs/src/content/docs/en/guides/connection/py-agent-gateway.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+## Local Development
+
+Run the agent locally with:
+
+<NxCommands commands={["<agent-name>-serve-local <project-name>"]} />
+
+This starts the agent with `SERVE_LOCAL=true` plus every MCP server attached to the Gateway. The generated Gateway client detects the flag and returns a list of `MCPClient`s pointed at each attached MCP server's local port.
+
+### `GATEWAY_URL_OVERRIDE`
```
cogwirrel: No - everything should have a serve and serve-local. Serve should not have local env set and needs runtime config app config id env variable only.

docs/src/content/docs/en/guides/connection/ts-strands-agent-gateway.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+The generator exposes a single high-level `create(sessionId)` factory on the generated Gateway client:
+
+```ts
+const myGateway = await MyGatewayClient.create(sessionId);
+```
+
+When deployed (`SERVE_LOCAL` unset), `create` returns a **single** `McpClient` pointed at the Gateway's MCP endpoint and authenticated with SigV4.
+
+When `SERVE_LOCAL=true`, `create` returns an **array** of `McpClient`s — one per MCP server in the client's `ATTACHED_MCP_SERVERS` list. Strands' `Agent.tools` accepts either shape, so the patched `agent.ts` works uniformly in both modes.
```
cogwirrel: Too hanky

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+import { FileTree } from '@astrojs/starlight/components';
+import RunGenerator from '@components/run-generator.astro';
+import GeneratorParameters from '@components/generator-parameters.astro';
+import Infrastructure from '@components/infrastructure.astro';
+import Snippet from '@components/snippet.astro';
+import Link from '@components/link.astro';
+import NxCommands from '@components/nx-commands.astro';
+
+Generate an [Amazon Bedrock AgentCore Gateway](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway.html) project. An AgentCore Gateway is a managed entry point that aggregates one or more MCP server targets behind a single MCP endpoint, evaluates every tool call against a [Cedar](https://www.cedarpolicy.com/) policy engine, and signs outbound traffic to MCP servers with IAM SigV4.
```
cogwirrel: Cedar policy link should be to relevant aws docs gateway policies

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+## Why use a Gateway?
```
cogwirrel: Remove this section, let’s be consistent with other docs

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+- **Aggregation**: Expose many MCP servers as a single MCP endpoint. Agents connect to one URL and see the union of all tools.
+- **Policy enforcement**: Each inbound `tools/list` and `tools/call` is evaluated by a Cedar `PolicyEngine` running in `ENFORCE` mode. Policies can permit or forbid tool invocations based on the caller identity (the IAM principal) and the tool being invoked.
+- **IAM SigV4 outbound**: The Gateway signs outbound calls to your MCP server runtimes using AWS IAM credentials, so the MCP servers never need to accept unauthenticated traffic.
+- **IAM inbound**: Callers authenticate to the Gateway with AWS IAM (SigV4). Together with Cedar, this gives you identity-aware tool filtering out of the box.
+
+## Prerequisites
+
+Before running this generator, ensure your workspace has an infrastructure project. If you haven't already, run one of:
```
cogwirrel: Why? Not needed

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+### Placeholder substitution
+
+The generator substitutes these placeholders at synth/plan time so policies stay portable across accounts and gateway redeploys:
+
+| Placeholder      | Substituted with                              |
+| ---------------- | --------------------------------------------- |
+| `${gateway_arn}` | The deployed Gateway's ARN                    |
+| `${account_id}`  | The AWS account this Gateway is deployed into |
+
+Always reference these placeholders rather than hard-coding values.
```
cogwirrel: Document how to add your own if possible

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+permit (
+  principal is AgentCore::IamEntity,
+  action,
+  resource == AgentCore::Gateway::"${gateway_arn}"
+) when {
+  principal.id like "arn:aws:*::${account_id}:*"
+};
+```
+
+This is a safe development starting point. To lock things down further, replace this with one or more narrower policies (see patterns below). Always keep at least one matching `permit` in the policy set, otherwise default-deny will block every call.
```
cogwirrel: Remove first sentence here

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+principal.id like "arn:aws:sts::*:assumed-role/*MyAgentExecutionRole*"
+```
+
+Tool invocations reach the policy engine as actions with this shape:
+
+```
+AgentCore::Action::"<target-name>___<tool-name>"
+```
+
+where `<target-name>` is the Gateway target name passed to `gateway.addMcpServerTarget(...)`, `<tool-name>` is the MCP tool's name, and the separator is **three underscores**. Cedar does not support wildcards on actions — match exact actions, or omit `action ==` to match all actions.
```
cogwirrel: Say ___ in backticks instead of three underscores

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+The resource is always the Gateway itself:
+
+```cedar
+resource == AgentCore::Gateway::"${gateway_arn}"
+```
+
+### Validator gotchas
```
cogwirrel: Reword this section more professionally

docs/src/content/docs/en/guides/connection.mdx
```diff
@@ -121,6 +121,12 @@ The Connection generator supports the following connections:
   />
 </CardGrid>
 
+### AgentCore Gateway Connections
```
cogwirrel: Add to the card grid. Use existing agentcore icon for gateway

nx-plugin-for-aws:
Pushed `b29b6497` addressing the review feedback. Per-comment summary:

| Comment | Resolution |
|---|---|
| "Source cannot have components — just mention target component" | Dropped `sourceComponent` from the disambiguation note in the gateway-mcp guide. |
| "Make more consistent with other generators and include modified files in file tree" | The generator-output `<FileTree>` now lists the actual files modified (gateway `project.json` + the per-Gateway `<gateway>-client.ts` / `_client.py` whose `ATTACHED_MCP_SERVERS` lists are updated). |
| "Add a method on the gateway for adding an mcp server by passing it straight in, inferring the name from the name of the mcp server" | New `gateway.addMcpServer(mcpServer)` helper on the generated CDK construct. The `gatewayTargetName` defaults to the construct's `node.id` in kebab-case (e.g. `MyMcpServer` → `my-mcp-server`). The lower-level `addMcpServerTarget(id, { gatewayTargetName, mcpServerRuntimeArn })` remains for callers that hold an ARN rather than a construct. The `agentcore-gateway#mcp-connection` printout and the docs use the new helper. |
| "We should prefix locally too if possible" | Already implemented in this PR's earlier round of fixes — the local multiplex client wraps the transport so tool names appear as `<target-name>___<tool-name>` to match the deployed Gateway. The doc now explicitly states "tool-name prefixing is preserved" for local mode. |
| "There have been some session id changes, rebase and adjust to be consistent" | Rebased onto latest main. The Gateway clients (TS + Py) read the request-scoped session via the same context mechanism the rest of the stack now uses (TS `withSessionId.forSession`, Py `session_id_context` middleware) — `create()` no longer takes a `session_id`/`sessionId` parameter. The docs were updated to match. |
| "This is too hacky and not type safe" (Py `get_agent` plumbing) | Replaced the `MCPClient \| list[MCPClient]` shape with a single context-manageable `MCPClient \| AgentCoreGatewayLocalMultiplexClient`. The new local-mode client implements `__enter__` / `__exit__` / `list_tools_sync` itself, so the patched `agent.py` is now a one-line `with (gateway,):` mirroring the `mcp-connection` shape. |
| "everything should have a serve and serve-local. Serve should not have local env set and needs runtime config app config id env variable only." | The Gateway project now generates both `<gateway>-serve` and `<gateway>-serve-local`. Each agent-side gateway-connection wires `<agent>-serve` → `<gateway>-serve` and `<agent>-serve-local` → `<gateway>-serve-local`. Neither aggregator sets `SERVE_LOCAL`. |
| "Too hanky" (TS `get_agent` plumbing) | Same root cause — the doc described `create(sessionId)` which the actual API never had. Doc rewritten to match. The patched `agent.ts` is now `const myGateway = await MyGatewayClient.create();` followed by `tools: [myGateway, ...]`, which is the same shape as `mcp-connection`. |
| "Cedar policy link should be to relevant aws docs gateway policies" | Replaced the `cedarpolicy.com` link with [`docs.aws.amazon.com/.../policy.html`](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/policy.html). |
| "Remove this section, let's be consistent with other docs" (Why use a Gateway?) | Removed. |
| "Why? Not needed" (Prerequisites with infrastructure-project step) | Removed. |
| "Document how to add your own if possible" | New "Adding a policy" subsection at the top of "Writing Policies" with a step-by-step worked example. |
| "Remove first sentence here" (after default permit-all) | Removed. |
| "Say `___` in backticks instead of three underscores" | Done — both in the policy guide and in the local-fidelity note. |
| "Reword this section more professionally" (Validator gotchas / Common patterns) | Renamed to "Validation considerations", reworded throughout, and replaced the duplicated "Common patterns" section with a single targeted example. |
| "Add to the card grid. Use existing agentcore icon for gateway" | Added `agentcore` to the `ConnectionCard` logo set (sourced from the existing `bedrock-agentcore.svg`) and added three card entries for the Gateway connections. |

Plus: replaced the `setInterval` keep-alive hack on the Gateway's `serve` / `serve-local` aggregator targets with `nx:noop` + `dependsOn` (Nx keeps the dependencies' `continuous: true` processes alive across the non-continuous aggregator). Verified locally that `<agent>-serve-local` boots both the TS and Py MCP servers plus the agent itself with this shape.

nx-plugin-for-aws:
Tested this PR end-to-end in a fresh workspace (`pnpm create @aws/nx-workspace@next` + linked local plugin), both locally via `serve-local` and deployed to AWS via CDK. Pushed three commits addressing what surfaced:

### End-to-end validation

**Local (`SERVE_LOCAL=true`)**: Generated a TS agent + TS MCP server (`divide`), a Py agent + Py MCP server (`add`), a `cedar#agentcore-gateway`, and wired all four connection edges. `my-agent-serve-local` booted the gateway aggregator + both MCP servers; both agents listed gateway-prefixed tools (`ts-mcp___divide`, `py-mcp___add`) and routed calls to the correct upstream server (84÷4=21 via TS, 19+23=42 via Py).

**Deployed (us-west-2)**: Deployed the gateway (Cedar PolicyEngine in ENFORCE + permit-all policy), both MCP servers and both agents via CDK. Invoked both deployed agents through the live gateway — tool listing and invocation through the gateway's SigV4 path worked for both runtimes. Stack torn down after testing.

### Fixes pushed

1. **`3e7c1504` — order-independence for `ATTACHED_MCP_SERVERS`**: connecting an agent to the gateway *after* MCP servers were attached left the generated client's local multiplex list empty (back-propagation only ran in the gateway→mcp direction). The agent-side generators now seed the list from the gateway's serve-local aggregator dependencies. Unit tests added for both TS and Py.

2. **`5dae744b` — double `Target` suffix**: `addMcpServer` produced CFN child ids like `TsMcpTargetTarget` (it appended `Target` and `addMcpServerTarget` appended it again). Verified the fix with a fresh deploy. Also trimmed the vended construct's comment blocks to keep generated code lean.

3. **`1f5b195e` — docs build fix (cause of the failing Build check)**: the automated translation commits had introduced duplicated trailing fragments in several translated mdx files (es/fr/it/jp/ko/pt/vi/zh `fastapi`, `react-website`, `ts-smithy-api`, dungeon-game), which fail MDX parsing. `docs:build` now passes.

Full unit suite passes (109 files / 2113 tests), lint clean, `build --all` green locally.

nx-plugin-for-aws:
Merged latest `main` (which added the `idempotency` smoke test) and pushed `b64f27ec`: the `cedar#agentcore-gateway` generator previously threw `A project already exists` on re-run, which failed the new idempotency smoke test. It now preserves the project on re-run — user-edited `.cedar` policies are kept, the serve/serve-local aggregators retain `dependsOn` wiring added by `mcp-connection`, and component metadata is not duplicated. Verified locally by re-running the full gateway generator sequence (gateway + all four connection edges) in a committed workspace: zero git diff on the second pass. Unit test added covering re-run behaviour.

nx-plugin-for-aws:
All CI checks are green ✅ (Build, all 22 smoke tests including the new `idempotency` suite, Windows smoke tests, CodeQL, Package).

One more fix was needed after the merge from `main`: `254d718c` — the Python agent connection generators' "already wired" GritQL guards lacked the `language python` prefix, so they parsed patterns as JavaScript and never matched. Re-running a connection nested a second `with` block inside the first. Guards now use the python prefix, and the subsequent-connection creation-line insert anchors on the single `def get_agent` body so it inserts exactly once. Verified locally: re-running the full gateway + MCP + connection sequence in a committed workspace yields zero git diff.

nx-plugin-for-aws:
Pushed `3e2cbe64` simplifying the vended code surface based on a developer-experience review (net −149 lines of vended/template code):

- **Single client type in both modes (TS)**: the local multiplex client is now an `McpClient` subclass that fans out to every attached MCP server, so `<Gateway>Client.create()` returns a plain `McpClient` in both deployed and local modes — no more `McpClient | McpClient[]` union. This matches the Python mirror's design.
- **`SERVE_LOCAL` is the only mode switch**: `GATEWAY_URL_OVERRIDE` is removed. To exercise Cedar policies from a local machine, run the agent's `serve` target, which resolves the deployed gateway URL from runtime config. Docs updated accordingly.
- **Removed unused `withoutAuth` factories** from both core gateway clients (inbound auth is IAM-only in v1).
- **Target ids derived from target names**: `addMcpServerTarget` now takes just props and derives the CFN child id from `gatewayTargetName`, so renaming a target changes both together and replaces the resource cleanly (previously the id was caller-supplied and a rename under an unchanged name failed with AlreadyExists).
- **Keep-alive is a vended `serve.ts` run via tsx** that logs its purpose once, replacing the opaque inline `node -e "setInterval(...)"` in project.json.
- **Comment trims across all vended templates** (clients, CDK construct, `permit-all.cedar`) — detail now lives in `policies/README.md` and the docs site.

Re-verified end-to-end in a fresh workspace: `serve-local` multiplexing for both TS and Python agents (both MCP servers reachable through the single client), full CDK deploy to us-west-2 with live gateway invocation from both deployed agents, idempotent re-run of the full generator sequence (zero git diff), full unit suite (2236 tests), and `docs:build`. Test stack torn down.

nx-plugin-for-aws:
Pushed `5c94eb6d` addressing the latest round of feedback:

1. **Generator renamed to `agentcore-gateway`** — Cedar is now an optional feature, not the namespace. New `cedarPolicy` option (default `true`); with `false` the gateway deploys without a policy engine and no `policies/` directory is vended.
2. **Local development now runs a real local gateway.** `serve.ts` is a small MCP aggregator (express + MCP SDK, run via tsx on the gateway's own assigned port) that proxies every attached MCP server with `<target>___<tool>` prefixing. The TS and Python multiplex client templates are deleted — both agent clients are now a pure URL switch (plain HTTP to the local gateway under `SERVE_LOCAL`, SigV4 to the deployed gateway otherwise). The mcp-connection generator registers servers in `serve.ts` rather than patching agent-side client files.
3. **Migrated off `@aws-cdk/aws-bedrock-agentcore-alpha`** to the stable `aws-cdk-lib/aws-bedrockagentcore` Gateway L2 (the alpha module is deprecated). The policy engine attaches via the underlying `CfnGateway` (the stable L2 doesn't expose it yet) with an explicit role-grant dependency — without it, gateway creation races the IAM policy update and fails with `Access denied while calling GetPolicyEngine`.
4. **Gateways can be named `gateway`** — CDK imports in the vended construct are namespaced (`import * as agentcore`), fixing the `Individual declarations in merged declaration 'Gateway'` compile errors. Regression test added; `IMcpServerConstruct` moved to a shared file so multiple gateways don't produce duplicate exports.
5. **Cedar policies are EJS templates** (`<%= gatewayArn %>`, `<%= accountId %>`), rendered by the CDK construct via an `ejs` devDependency added to the workspace (only when `cedarPolicy` is enabled), or by `replace()` in the Terraform module.

Re-verified end-to-end in a fresh workspace: local serve-local through the local gateway for both TS and Python agents; CDK deploy to us-west-2 with both a Cedar gateway and a `cedarPolicy=false` gateway named `gateway` side by side (tool listing + invocation verified on both); idempotent generator re-runs (zero git diff); full unit suite (2241 tests); docs build. Test stacks torn down. PR title/description updated to match.

nx-plugin-for-aws:
Pushed `ec64db6e` with the latest feedback round:

1. **Local gateway entry point renamed to `serve-local.ts`.**
2. **New `infra` option** (`agentcore` | `none`, default `agentcore`): `none` skips all infrastructure; re-running with `agentcore` adds the construct/module idempotently, matching other infra-vending generators. Covered by unit tests including the upgrade path.
3. **Terraform now renders Cedar policies via ejs too**: a vended `render-cedar.cjs` invoked through the `external` data source with `gatewayArn`/`accountId` in scope — same EJS tokens as the CDK construct, so policies are portable across IaC providers.
4. **Terraform gateway resource migrated to `awscc_bedrockagentcore_gateway`** — the native `aws` provider resource doesn't support `policy_engine_configuration`. Deploying end to end also surfaced: awscc policy engine names reject hyphens (now underscore-separated), the awscc policy resource takes `policy_engine_id` with a structured `cedar` definition, and a policies-dir relative path that was one directory short.

**Terraform e2e verified**: fresh `--iac=terraform` workspace, full generator matrix, `terraform apply` to us-west-2 — gateway, policy engine, EJS-rendered `PermitAll` policy, and both MCP server targets all reach READY; tools listed and invoked through the live gateway endpoint (`ts-mcp___divide` → 21, `py-mcp___add` → 42). Stack destroyed afterwards. CDK path unchanged and still green (63 generator unit tests pass). PR description updated to reflect the current state.

docs/src/content/docs/en/guides/connection/ts-agent-gateway.mdx
```diff
@@ -0,0 +1,150 @@
...
+  return new Agent({
+    systemPrompt: '...',
+    tools: [myGateway],
+  });
+};
+```
+
+`create()` always returns a single `McpClient`. When deployed (`SERVE_LOCAL` unset), it points at the Gateway's MCP endpoint and authenticates with SigV4. When `SERVE_LOCAL=true`, it points at the local gateway started by the Gateway project's `serve-local` target, so the same `agent.ts` works uniformly in both modes.
+
+The session ID is propagated to downstream MCP servers automatically: the underlying transport reads the request-scoped session via the [Strands TS session context](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability.html) and forwards it to the Gateway as the `X-Amzn-Bedrock-AgentCore-Runtime-Session-Id` header.
```
cogwirrel: should be:

>The session ID is propagated to downstream MCP servers automatically via the `X-Amzn-Bedrock-AgentCore-Runtime-Session-Id` header.

docs/src/content/docs/en/guides/connection/ts-agent-gateway.mdx
```diff
@@ -0,0 +1,150 @@
...
+export const getAgent = async () => {
+  const myGateway = await MyGatewayClient.create();
+  return new Agent({
+    systemPrompt: '...',
+    tools: [myGateway],
+  });
+};
+```
+
+`create()` always returns a single `McpClient`. When deployed (`SERVE_LOCAL` unset), it points at the Gateway's MCP endpoint and authenticates with SigV4. When `SERVE_LOCAL=true`, it points at the local gateway started by the Gateway project's `serve-local` target, so the same `agent.ts` works uniformly in both modes.
```
cogwirrel: Delete "`create()` always returns a single `McpClient`."

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,212 @@
...
+<GeneratorParameters generator="agentcore-gateway" />
+
+## Generator Output
+
+The generator creates a new project at `packages/<name>/`, plus a CDK construct or Terraform module for the infrastructure:
+
+<FileTree>
+  - packages/\<name>/
+    - policies/ Cedar policy source files (omitted when `cedarPolicy: false`)
+      - permit-all.cedar Default Cedar policy that permits everything
```
cogwirrel: "that permits authenticated callers from within the same AWS account"

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,212 @@
...
+- An `AgentCore::PolicyEngine` running in `ENFORCE` mode, attached to the Gateway (omitted when `cedarPolicy: false`)
+- One `AgentCore::Policy` per `.cedar` file in `policies/`
+
+The Gateway URL is automatically registered in the `agentcore.gateways.<ClassName>` namespace of <Link path="guides/runtime-config">Runtime Configuration</Link> so agents can discover it at runtime.
+
+## Writing Policies
+
+:::note
+This section applies when the gateway is generated with `cedarPolicy: true` (the default). With `cedarPolicy: false` the gateway performs no policy evaluation — any caller with IAM access can invoke any tool.
+:::
```
cogwirrel: Use `<OptionFilter>` here per CONTRIBUTING.md

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,212 @@
...
+### Template variables
+
+Policies are [EJS](https://ejs.co/) templates rendered at synth/plan time so they stay portable across accounts and Gateway redeploys:
+
+| Variable             | Substituted with                              |
+| -------------------- | --------------------------------------------- |
+| `<%= gatewayArn %>`  | The deployed Gateway's ARN                    |
+| `<%= accountId %>`   | The AWS account this Gateway is deployed into |
+
+Always reference these variables rather than hard-coding values.
```
cogwirrel, line 111: Point to where to edit to add variables, eg to avoid hardcoding the role arn for the example above

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+```
+arn:aws:sts::<account-id>:assumed-role/<role-name>/<session-name>
+```
+
+Match these with `like` and a wildcard:
+
+```cedar
+principal.id like "arn:aws:sts::*:assumed-role/*MyAgentExecutionRole*"
+```
```
cogwirrel: remove this and instead recommend to pass exact role arns as template variables

e2e/src/smoke-tests/generator-matrix.ts
```diff
@@ -220,7 +220,36 @@ export const runGeneratorMatrix = async (opts: RunCliOpts) => {
...
 
-  // Website -> strands agent connections (TypeScript HTTP, TypeScript AG-UI, Python HTTP, Python AG-UI/CopilotKit)
+  // AgentCore Gateway + the four connection edges (ts/py agent -> gateway,
+  // gateway -> ts/py mcp-server) so each smoke test exercises a deployable
+  // gateway with multiple MCP server targets fronting both agent runtimes.
+  // Wired after the agent <-> MCP and a2a connections so the gateway client
+  // is the last entry prepended to the agent's tools array — placing it
+  // earlier surfaces a pre-existing GritQL/prettier interaction in the
+  // shared mcp-connection prepend pattern when subsequent appends run on
+  // the resulting multi-line array.
```
cogwirrel: Double check if this comment is actually relevant - if so we should fix the underlying issue

e2e/src/smoke-tests/serve-local.spec.ts
```diff
@@ -216,6 +216,36 @@ describe('smoke test - serve-local', { timeout: 20 * 60 * 1000 }, () => {
         `generate @aws/nx-plugin:ts#rdb --name=postgres-db --infra=aurora --engine=postgres --framework=prisma --no-interactive`,
         opts,
       );
+      // Isolated AgentCore Gateway topology: a dedicated agent plus dedicated
+      // TypeScript and Python MCP servers, all wired only to the gateway. Kept
+      // separate from the agents/MCP servers above so the gateway's serve-local
+      // chain never shares (and re-starts) an MCP server that another agent's
+      // chain also starts. The two MCP servers exercise the gateway's core job:
+      // aggregating multiple (heterogeneous) MCP servers behind one endpoint.
```
cogwirrel: check if this really needs to be separate - nx should dedupe the continuous targets

packages/nx-plugin/src/agentcore-gateway/files/cedar/policies/README.md
```diff
@@ -0,0 +1,43 @@
...
+allowed.
+
+## Template variables
+
+Policies are [EJS](https://ejs.co/) templates rendered at synth/plan time
+with the following variables in scope:
+
+- `<%%= gatewayArn %>` — this gateway's ARN
+- `<%%= accountId %>` — the AWS account this gateway is deployed into
+
```
cogwirrel, line 27: point to where variables can be added based on whether infra was in tf or cdk

packages/nx-plugin/src/agentcore-gateway/files/project/serve-local.ts.template
```diff
@@ -0,0 +1,126 @@
+// Local gateway for <%- nameClassName %>: a single MCP endpoint aggregating
+// every attached MCP server, with tools exposed as `<target>___<tool>` to
+// match the deployed AgentCore Gateway. Unlike the real gateway, Cedar
+// policies are not evaluated and there is no semantic tool search.
```
cogwirrel: move this comment to a /** style comment above createServer

packages/nx-plugin/src/agentcore-gateway/mcp-connection/generator.ts
```diff
@@ -0,0 +1,143 @@
...
+    const content = tree.read(serveTsPath)!.toString();
+    if (!content.includes(`'${mcpComponentName}'`)) {
+      const entry = `  { name: '${mcpComponentName}', url: 'http://localhost:${mcpPort}/mcp' },`;
+      const updated = content.replace(
+        /(const ATTACHED_MCP_SERVERS = \[)([\s\S]*?)(\];)/m,
+        (_match, open, inner, close) =>
+          `${open}${inner.replace(/\s+$/, '')}\n${entry}\n${close}`,
+      );
+      if (updated !== content) tree.write(serveTsPath, updated);
+    }
```
cogwirrel: gritql must be used here instead of regex.

packages/nx-plugin/src/agentcore-gateway/mcp-connection/generator.ts
```diff
@@ -0,0 +1,143 @@
...
+    `✔ Connected ${mcpClassName} MCP server to ${gatewayClassName} gateway.`,
+  );
+  console.log('');
+  console.log('Add the following to the stack that instantiates your gateway:');
+  console.log('');
+  console.log(
+    `  const ${mcpVarName} = new ${mcpClassName}(this, '${mcpClassName}');`,
+  );
+  console.log(`  ${gatewayVarName}.addMcpServer(${mcpVarName});`);
+  console.log('');
```
cogwirrel: remove this - the documentation in the generator guide is enough.

packages/nx-plugin/src/agentcore-gateway/generator.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  addComponentGeneratorMetadata(
+    tree,
+    fullyQualifiedName,
+    AGENTCORE_GATEWAY_GENERATOR_INFO,
+    '.',
+    name,
+    { rc: nameClassName, protocol, auth, port },
+  );
```
cogwirrel: use addGeneratorMetadata since this is a project not a component

packages/nx-plugin/src/py/agent/gateway-connection/generator.ts
```diff
@@ -0,0 +1,305 @@
...
+  console.log('');
+  console.log(
+    `✔ Connected ${agentComponent.name} agent to ${gatewayClassName} gateway.`,
+  );
+  console.log('');
+  console.log('Add to the stack that instantiates your agent + gateway:');
+  console.log(
+    `  ${gatewayClassName.charAt(0).toLowerCase()}${gatewayClassName.slice(1)}.grantInvokeAccess(agent);`,
+  );
+  console.log('');
```
cogwirrel: likewise remove this

packages/nx-plugin/src/py/agent/gateway-connection/generator.ts
```diff
@@ -0,0 +1,305 @@
...
+
+  await formatFilesInSubtree(tree);
+  return () => {
+    installPackagesTask(tree);
+  };
+};
+
+const addGatewayToolsToAgent = async (
```
cogwirrel: can we reuse the existing helpers for adding tools from the agent -> mcp server connection generator? refactor to common utility if so.

packages/nx-plugin/src/ts/agent/gateway-connection/generator.ts
```diff
@@ -0,0 +1,217 @@
...
+      agentFilePath,
+      [clientClassName],
+      `:${npmScope}/agent-connection`,
+    );
+
+    const clientCreationStmt = `const ${clientVarName} = await ${clientClassName}.create();`;
+
+    // Wrap / prepend client creation inside the arrow function that builds
+    // the Agent — same pattern as mcp-connection / a2a-connection.
+    await applyGritQL(
```
cogwirrel: same here can we reuse the mcp connection helpers for typescript and refactor?

packages/nx-plugin/src/ts/agent/gateway-connection/generator.ts
```diff
@@ -0,0 +1,217 @@
...
+  console.log(
+    `✔ Connected ${agentComponent.name} agent to ${gatewayClassName} gateway.`,
+  );
+  console.log('');
+  console.log('Add to the stack that instantiates your agent + gateway:');
+  console.log('');
+  console.log(
+    `  ${gatewayClassName.charAt(0).toLowerCase()}${gatewayClassName.slice(1)}.grantInvokeAccess(agent);`,
+  );
+  console.log('');
```
cogwirrel: remove

packages/nx-plugin/src/utils/agent-core-constructs/files/cdk/app/agentcore-gateway/__nameKebabCase__/__nameKebabCase__.ts.template
```diff
@@ -0,0 +1,241 @@
...
+    // policy update — AgentCore validates GetPolicyEngine access at creation.
+    const policyGrant = this.gateway.role.addToPrincipalPolicy(
+      new iam.PolicyStatement({
+        actions: [
+          'bedrock-agentcore:GetPolicyEngine',
+          'bedrock-agentcore:AuthorizeAction',
+          'bedrock-agentcore:PartiallyAuthorizeActions',
+          'bedrock-agentcore:CheckAuthorizePermissions',
+        ],
+        resources: ['*'],
```
cogwirrel: can this be scoped to the policy engine? ensure you fully test if so

packages/nx-plugin/src/utils/agent-core-constructs/files/cdk/app/agentcore-gateway/__nameKebabCase__/__nameKebabCase__.ts.template
```diff
@@ -0,0 +1,241 @@
...
+
+    // Each .cedar file under <%- projectDirectory %>/policies/ becomes a
+    // Policy. Files are EJS templates rendered at synth time with
+    // `gatewayArn` and `accountId` in scope.
+    const policiesDir = path.join(
+      findWorkspaceRoot(url.fileURLToPath(new URL(import.meta.url))),
+      '<%- projectDirectory %>/policies',
+    );
+    if (fs.existsSync(policiesDir)) {
+      for (const file of fs.readdirSync(policiesDir)) {
```
cogwirrel: Make sure we support nested folders within the policies directory.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/app/agentcore-gateway/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -0,0 +1,218 @@
...
+resource "awscc_bedrockagentcore_policy" "policies" {
+  for_each = fileset(local.policies_dir, "*.cedar")
+
+  policy_engine_id = awscc_bedrockagentcore_policy_engine.this.policy_engine_id
+  name = join("", [
+    for word in split("-", replace(trimsuffix(each.value, ".cedar"), "_", "-")) :
+    "${upper(substr(word, 0, 1))}${substr(word, 1, length(word))}"
+  ])
+  description     = "Policy loaded from ${each.value}"
+  validation_mode = "IGNORE_ALL_FINDINGS"
```
cogwirrel, line 202: Shouldn't this be ENFORCE?

packages/nx-plugin/generators.json
```diff
@@ -3,6 +3,19 @@
   "name": "@aws/nx-plugin",
   "version": "0.0.1",
   "generators": {
+    "agentcore-gateway": {
+      "factory": "./src/agentcore-gateway/generator",
+      "schema": "./src/agentcore-gateway/schema.json",
+      "description": "Generate an AgentCore Gateway project",
+      "metric": "g43"
```
cogwirrel: double check the metrics are unique as other generators have been added since this pr started

docs/src/content/docs/en/guides/connection/ts-agent-gateway.mdx
nx-plugin-for-aws: Done — replaced with the suggested wording.

docs/src/content/docs/en/guides/connection/ts-agent-gateway.mdx
nx-plugin-for-aws: Done — deleted.

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Done — now reads "permits authenticated callers from within the same AWS account".

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Done — the Writing Policies section is now wrapped in `<OptionFilter when={{ cedarPolicy: true }}>`. This needed a small extension to the option-filter machinery (boolean options are now filterable alongside enums, offering true/false in the filter bar).

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Done — added an "Adding your own variables" subsection showing exactly where to extend the render context for both CDK (the `ejs.render` call in the construct) and Terraform (the `query` of the `rendered_policies` data source).

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Done — removed the wildcard role-ARN guidance; the examples now pass exact role ARNs as template variables (e.g. `<%= tsAgentRoleArn %>`) per the new "Adding your own variables" section.

e2e/src/smoke-tests/generator-matrix.ts
nx-plugin-for-aws: Verified it no longer reproduces (tested gateway-before-mcp ordering in a throwaway spec) — removed the comment and the special-case ordering.

e2e/src/smoke-tests/serve-local.spec.ts
nx-plugin-for-aws: Confirmed — Nx dedupes the shared continuous targets (verified a single `nx run` per target when the gateway's serve-local chain overlaps the MCP servers'). Removed the isolated topology; the gateway agent now shares ts-project's MCP servers.

packages/nx-plugin/src/agentcore-gateway/files/cedar/policies/README.md
nx-plugin-for-aws: Done — the README now points at the CDK construct's `ejs.render` call and the Terraform module's `rendered_policies` data source respectively.

packages/nx-plugin/src/agentcore-gateway/files/project/serve-local.ts.template
nx-plugin-for-aws: Done — moved to a /** */ comment above `createServer`.

packages/nx-plugin/src/agentcore-gateway/mcp-connection/generator.ts
nx-plugin-for-aws: Done — registration in serve-local.ts now uses GritQL (with an idempotency guard and a regression test that the rewrite never produces dangling commas).

packages/nx-plugin/src/agentcore-gateway/mcp-connection/generator.ts
nx-plugin-for-aws: Done — removed.

packages/nx-plugin/src/agentcore-gateway/generator.ts
nx-plugin-for-aws: Done — the gateway uses `addGeneratorMetadata` (project-level), with a `readAgentCoreGatewayMetadata` helper used by the connection generators, and the connection dispatcher matches on project metadata.

packages/nx-plugin/src/py/agent/gateway-connection/generator.ts
nx-plugin-for-aws: Done — removed.

packages/nx-plugin/src/py/agent/gateway-connection/generator.ts
nx-plugin-for-aws: Done — extracted shared helpers in `utils/agent-connection` (`addPythonClientToAgent` / `addTypeScriptClientToAgent`); both the mcp-connection and gateway-connection generators (ts + py) now use them.

packages/nx-plugin/src/ts/agent/gateway-connection/generator.ts
nx-plugin-for-aws: Done — same shared helpers as the Python side; the ts mcp-connection and gateway-connection generators both use `addTypeScriptClientToAgent`.

packages/nx-plugin/src/ts/agent/gateway-connection/generator.ts
nx-plugin-for-aws: Done — removed.

packages/nx-plugin/src/utils/agent-core-constructs/files/cdk/app/agentcore-gateway/__nameKebabCase__/__nameKebabCase__.ts.template
nx-plugin-for-aws: Done, with one nuance found through deployed testing: the authorization actions are evaluated against both the policy engine ARN and the gateway's own ARN (scoping to the engine alone fails gateway creation with AccessDenied on the gateway ARN, and vice versa). The gateway ARN can't be referenced without a circular dependency, so the statement is scoped to the engine ARN plus this account/region's `gateway/*`. Verified by full deploys of both CDK and Terraform workspaces, invoking deployed TS + Python agents through the live gateway.

packages/nx-plugin/src/utils/agent-core-constructs/files/cdk/app/agentcore-gateway/__nameKebabCase__/__nameKebabCase__.ts.template
nx-plugin-for-aws: Done — CDK uses `fs.readdirSync(..., { recursive: true })` and Terraform `fileset(..., "**/*.cedar")`; folder segments fold into the policy name (e.g. `team-a/permit-team.cedar` → `TeamAPermitTeam`). Deploy-tested a nested policy in both CDK and Terraform. This also surfaced that policy names must be unique account-wide, so both templates now add a unique suffix.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/app/agentcore-gateway/__nameKebabCase__/__nameKebabCase__.tf.template
nx-plugin-for-aws: ENFORCE is the policy *engine* mode, which is set on the gateway's `policy_engine_configuration`. `validation_mode` on the Policy resource is a create-time analyzer setting whose only values are FAIL_ON_ANY_FINDINGS | IGNORE_ALL_FINDINGS — the analyzer flags legitimate scoped policies as findings, so we deploy regardless and rely on the engine's ENFORCE mode. Added a comment in both templates clarifying this.

packages/nx-plugin/generators.json
nx-plugin-for-aws: Good catch — g43–g46 had been taken by generators merged since this PR started. Renumbered the gateway generators to g48–g51.

cogwirrel (dismissed):

cogwirrel (dismissed):

nx-plugin-for-aws:
## 📋 Comprehensive E2E Verification — AgentCore Gateway

Ran an exhaustive matrix — CDK/Terraform × Cedar on/off × local/deployed × TS/Python agents & MCPs × **all 3 agent protocols** (http/a2a/ag-ui) × topologies T1–T6 × policy styles (permit-all / scoped-permit / forbid / nested) × scale up to 4 MCPs and 3 gateways per workspace. 13 scenarios total; every live deploy was invoked and torn down (cleanup verified twice — no leftover AWS resources).

### ✅ Verified working
- **All four generators + every connection edge** generate, build, and compose — including A2A and AG-UI agents as gateway clients (not just HTTP), verified across all 6 protocol×language combinations.
- **`serve-local` aggregation** registers every attached MCP with `<target>___<tool>` namespacing — and **actually boots and routes** locally (mock-LLM agent drove `ts-mcp___divide`→3 and `py-mcp___add`→8 through the local gateway).
- **Scale/topology:** 1 gateway→4 heterogeneous MCPs, 1 agent→3 gateways, multi-agent→shared gateway, chained agent→A2A-agent→gateway→same MCP (deployed on Terraform). Distinct ports, no collisions.
- **CDK & Terraform** both deploy gateways live (us-west-2). Terraform's `render-cedar.cjs` EJS data source and CDK's synth-time render both substitute `gatewayArn`/`accountId`/custom vars and discover nested `policies/**/*.cedar`.
- **infra:none→agentcore upgrade** idempotent on both IaC providers; custom directory/subDirectory, hyphenated names, and `gatewayTargetName` override all work.
- **Cedar enforcement — definitively proven** (deterministic, LLM bypassed: two IAM roles calling the gateway MCP endpoint directly via SigV4):

  | Caller | `tools/list` | `ts-mcp___divide` | `py-mcp___add` |
  |---|---|---|---|
  | full role (permit-all) | both tools | ✅ → 3 | ✅ → 8 |
  | restricted role (`forbid` divide) | **divide hidden** | ❌ `Tool Execution Denied: Tool call not allowed due to policy enforcement` | ✅ → 8 |

  Confirms `forbid` enforces at **both** layers (discovery filtering + call denial) and is action-scoped.

### 🐞 Finding 1 — action-scoped Cedar policies fail on a *fresh* deploy (CDK)
A three-phase experiment isolated it: permit-all deploys clean; adding a `forbid ... action == "ts-mcp___divide"` and **redeploying** succeeds (`UPDATE_COMPLETE`, policy ACTIVE); but a **fresh single-shot** deploy with that policy present fails — `unrecognized action "ts-mcp___divide" (NotStabilized)`. The synthesized template shows the `CfnPolicy` has **`DependsOn: null`** — no dependency on the `CfnGatewayTarget`s that register those actions, so on a first deploy the policy is created before its target's actions exist. `permit-all.cedar` references no action, which is why it's never hit this. **Fix:** add a dependency from each action-referencing policy to the relevant target (or create policies via a custom resource after targets are READY). **Workaround:** two-phase deploy.

### 🐞 Finding 2 — gateway *target* can lose a cold-start race on fresh deploy (CDK & TF)
Observed intermittently (~1 in 8 fresh gateway deploys): a `CfnGatewayTarget` fails with `Failed to connect and fetch tools from the provided MCP target server`. AgentCore probes the MCP runtime (`tools/list`) during target creation, but the runtime's CFN `CREATE_COMPLETE` doesn't mean the container is warm/serving — so the probe can race the cold start and fail the whole deploy. Both findings **leave the stack wedged** (`ROLLBACK_FAILED`, target stuck in `CREATING`), needing a manual `delete-stack` before retry. **Fix:** retry the tool-discovery probe with backoff, or gate target creation on runtime readiness; at minimum document that a fresh gateway+MCP deploy may need a re-run.

### 📝 Finding 3 — stale PR description
The Terraform module uses the **native `aws` provider** (`aws_bedrockagentcore_gateway`/`_policy_engine`/`_policy`), not `awscc_bedrockagentcore_gateway` as described. (The CDK "stable Gateway L2" claim *is* accurate.)

### Scope notes (honest gaps)
- **TS HTTP-agent live invocation through the gateway is deploy-verified but not invocation-verified by my testing** — my harness couldn't reproduce the WebSocket SigV4 handshake the vended `AgentCoreTrpcClient` uses (repeated 404/403). The Python agent invoked live successfully, the local runtime routed both tools, and CI's own `invokeTrpcAgentCoreAgent` exercises the TS path — so this is a harness limitation, not evidence of a defect.
- Teardown was clean in every scenario; an ACTIVE Cedar policy did **not** block policy-engine deletion.

**Net:** the generators work correctly across all tested IaC/Cedar/protocol/topology/scale permutations and Cedar enforcement is proven end-to-end. Two real fresh-deploy reliability bugs (action-scoped policy ordering; target cold-start race) are worth fixing before release, plus a one-line description correction.

---
<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

nx-plugin-for-aws:
Thanks — all three findings verified and addressed in dcafe79a:

**Finding 1 (action-scoped policy fails fresh deploy):** Reproduced exactly as described (`unrecognized action ... NotStabilized` on a fresh single-shot CDK deploy with a `forbid ts-mcp___divide` policy). The CDK construct now adds a dependency from every Cedar policy to every gateway target, so policies are created after the targets have registered their actions. The Terraform module gets a `policy_dependencies` variable for the same purpose (documented in the mcp-connection guide's Terraform example), since targets there are user-wired in the root module.

**Finding 2 (target cold-start race):** The root cause is that target creation fetches the runtime's tools while the MCP runtime container may still be cold. The CDK construct now vends a readiness probe — a trigger Lambda that invokes each target runtime with an MCP `initialize` request, retrying through the cold start — and targets depend on it. In Terraform, the agent-core runtime module probes MCP runtimes via the AWS CLI (`null_resource` + `local-exec`) and routes the runtime ARN output through the probe, so gateway targets implicitly wait for readiness.

**Finding 3 (stale description):** PR description updated — the Terraform module uses the native `aws` provider resources (>= 6.50), not awscc.

Verified with fresh single-shot deploys of both CDK and Terraform workspaces (us-west-2), each including the action-scoped `forbid` policy alongside permit-all and a nested-folder policy: all policies ACTIVE on first deploy, `ts-mcp___divide` denied by policy, `py-mcp___add` returns 42. Full unit suite (2244) and docs build green; stacks torn down.

cogwirrel (dismissed):

nx-plugin-for-aws:
## ✅ Re-test on the updated branch (HEAD `6564d97`) — both fixes verified

Re-ran the full suite against the updated branch, which added `dcafe79a` (policy→target dependency + readiness probe) and `0c2ad603` (shared core CDK construct). Every deployed scenario was run on a **fresh single-shot deploy** (no two-phase workaround), invoked, and torn down — cleanup verified clean.

### Finding 1 (action-scoped policy ordering) — **FIXED** ✅
Deploying with a `forbid ... action == "ts-mcp___divide"` policy present **from the start** now succeeds on a fresh single-shot deploy on both **CDK** and **Terraform** — policy reaches `ACTIVE`, no `unrecognized action` / `NotStabilized`. CDK deploy timeline confirms the ordering: targets `CREATE_COMPLETE` → policies `CREATE_COMPLETE`. (CDK: `policy.node.addDependency(target)`; Terraform: `policy_dependencies` + `terraform_data` anchor.)

### Finding 2 (gateway-target cold-start race) — **FIXED** ✅
The readiness probe runs **before** target creation and warms each MCP runtime via an MCP `initialize` (CDK: `TargetReadinessProbe` TriggerFunction; Terraform: `null_resource.runtime_ready` uv/boto3 probe routing the runtime ARN). Targets reached `READY` on the first attempt with no `failed to fetch tools` / stabilization errors across CDK and Terraform deploys.

### Cedar enforcement — **proven end-to-end** ✅
Deterministic test (LLM bypassed — two IAM roles calling the gateway MCP endpoint directly via SigV4), using the new `cedarPolicyVariables` prop per the updated docs. Verified the *deployed* policy rendered correctly (`principal == ...assumed-role/r4-restricted`):

| Caller | `tools/list` | `ts-mcp___divide` | `py-mcp___add` |
|---|---|---|---|
| full (permit-all) | both tools | ✅ → 3 | ✅ → 8 |
| restricted (`forbid` divide) | **divide hidden** | ❌ `Tool Execution Denied: Tool call not allowed due to policy enforcement` | ✅ → 8 |

`forbid` enforces at both layers (discovery filtering + call denial), action-scoped.

### TS HTTP agent over WebSocket — **works** ✅
Invoking the deployed TS HTTP agent via the vended `AgentCoreTrpcClient.withIamAuth(...).invoke.subscribe(...)` over WebSocket succeeds — the agent streams a response. (My earlier "TS agent 404/403" notes were a harness mistake — hand-rolled WS SigV4 instead of the vended client — not a defect.)

### No regressions
The shared-core-construct refactor introduced no regressions: full multi-protocol mesh (http/a2a/ag-ui) builds, the new `core/agentcore-gateway/` construct + `readiness-probe` are vended correctly, and all **62** gateway/connection unit tests pass.

### 📝 One new docs gap — Terraform MCP-connection
On Terraform, the gateway's IAM role needs `bedrock-agentcore:InvokeAgentRuntime*` on each MCP runtime (passed via the gateway module's `additional_iam_policy_statements`) — otherwise gateway-target creation fails with `Authorization error when sending message`. The [`agentcore-gateway-mcp` Terraform example](https://github.com/awslabs/nx-plugin-for-aws/blob/feat/agentcore-gateway/docs/src/content/docs/en/guides/connection/agentcore-gateway-mcp.mdx) shows the `aws_bedrockagentcore_gateway_target` resource and `policy_dependencies` but **omits this grant**, so following it literally fails at deploy. (The CDK path wires the grant automatically inside `addMcpServer`, so it's unaffected.) Suggest adding the `invoke_<mcp>_server_statement` locals + `additional_iam_policy_statements` to the docs example (pattern exists in `e2e/src/files/terraform-deploy/main.tf.template`).

**Net:** both fresh-deploy reliability bugs from the prior review are fixed and verified on CDK and Terraform; Cedar enforcement and TS WebSocket invocation are confirmed working. The only remaining item is the Terraform docs example gap above.

---
<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

cogwirrel (approved):

PR #767, opened 2026-06-07
perf(ci): speed up PR workflow init via cache dedup and tool gating

nx-plugin-for-aws:
### Reason for this change

The PR workflow takes ~1 hour wall-clock. Every one of its 24 jobs (Build + 20 Linux + 3 Windows smoke tests) runs the shared `init-monorepo` composite action, which takes **~100s** per job. Profiling a recent run (#765) showed the bulk of that time is environment setup, not the repo's own `pnpm i` (only ~13s):

| init-monorepo step | Time |
|---|---|
| pnpm/action-setup (self-installer) | 16s |
| setup-node — restores 478MB pnpm store (`cache: 'pnpm'`) | **22s** |
| setup-qemu | 13s |
| setup-buildx | 5s |
| **actions/cache — restores the SAME 478MB pnpm store** | **19s** |
| `pnpm i --frozen-lockfile` | 13s |

Two clear inefficiencies:
1. The pnpm store was being **cached and restored twice** — once by `setup-node`'s `cache: 'pnpm'` and again by the explicit `actions/cache` step (identical 478MB content, confirmed in the cache list). This costs a redundant ~19s restore + post-job save per job, and doubles pnpm-store storage. The repo is pinned at the **10GB Actions cache ceiling**, so this duplication accelerates cache eviction → occasional cold installs.
2. The **Build job builds no containers** (verified: zero `docker build`/QEMU activity in its log; the docker/terraform/uv keywords there are only snapshot test names), yet still pays ~18s for QEMU + Buildx setup.

### Description of changes

- **De-duplicate the pnpm store cache**: removed `cache: 'pnpm'` from `setup-node`. The explicit `actions/cache` step remains as the single source of truth — it additionally has `restore-keys` for a warm fallback when the lockfile changes, which `setup-node` lacked.
- **Enable uv's built-in cache** (`enable-cache: true`) so the Python-heavy smoke tests stop re-downloading wheels every run.
- **Gate container tooling**: added a `container-tools` input (default `true`) to `init-monorepo` guarding QEMU + Docker Buildx. Set to `false` for the Build jobs in both `pr.yml` and `ci.yml`, which compile and unit-test only. Smoke test jobs keep the default and still build ARM containers.

YAML validated locally. Will observe the actual `init-monorepo` step timings on this PR's own workflow run and compare against the baseline above, iterating if needed.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/.github/actions/init-monorepo/action.yml
+++ b/.github/actions/init-monorepo/action.yml
@@ -5,6 +5,10 @@ inputs:
   aws-docker-login-role-arn:
     description: 'ARN of the AWS role to assume for Docker login'
     required: false
+  container-tools:
+    description: 'Set up QEMU and Docker Buildx for multi-arch container builds. Disable for jobs that do not build containers.'
+    required: false
+    default: 'true'
 
 runs:
   using: 'composite'
@@ -31,18 +35,18 @@ runs:
       with:
         node-version: 22
         registry-url: 'https://registry.npmjs.org'
-        cache: 'pnpm'
     - name: Install the latest version of uv
       uses: astral-sh/setup-uv@v5
       with:
         version: 'latest'
+        enable-cache: true
     - name: Set up QEMU
-      if: ${{ runner.os != 'Windows' }}
+      if: ${{ runner.os != 'Windows' && inputs.container-tools == 'true' }}
       uses: docker/setup-qemu-action@v3
       with:
         image: public.ecr.aws/eks-distro-build-tooling/binfmt-misc:qemu-v7.0.0
     - name: Set up Docker Buildx
-      if: ${{ runner.os != 'Windows' }}
+      if: ${{ runner.os != 'Windows' && inputs.container-tools == 'true' }}
       uses: docker/setup-buildx-action@v3
     - name: Set up WSL2 Docker for Linux builds (Windows)
       if: ${{ runner.os == 'Windows' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
       - uses: ./.github/actions/init-monorepo
         with:
           aws-docker-login-role-arn: ${{ secrets.AWS_DOCKER_LOGIN_ROLE_ARN }}
+          container-tools: 'false'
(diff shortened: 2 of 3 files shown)


nx-plugin-for-aws:
Measured the effect on this PR's own workflow run (run `27106472836`) against the baseline:

| Job | `init-monorepo` before | after |
|---|---|---|
| **Build** | ~103s | **64s** |
| **Smoke tests** (avg) | ~100s | **~84s** |

Verified from logs:
- Build job now does a **single** pnpm-store restore (`Cache restored from key: Linux-pnpm-store-...`) — the duplicate `node-cache-...pnpm` restore from `setup-node`'s `cache: 'pnpm'` is gone.
- QEMU + Buildx setup steps are **skipped** on the Build job (only the action repo is downloaded by the runner, which is unavoidable and cheap).
- Smoke test jobs keep container tools and still build ARM containers; their gain comes purely from the cache dedup.

The per-job init saving is ~16–39s across all 24 jobs. The duplicate pnpm-store cache also freed ~5GB against the repo's 10GB Actions cache ceiling, which should reduce eviction-driven cold installs.

Follow-up candidates (separate PRs) profiled while investigating: cross-run npm/yarn/bun download caches for the smoke tests (generated-project installs from the public registry dominate each ~13min smoke job), and cutting the ~16s pnpm self-installer via corepack.

nx-plugin-for-aws:
✅ Full run is green — all 28 checks pass.

**Total wall-clock: 52 min vs ~60 min baseline (~13% faster).**

One useful observation from this run: the critical path has shifted. The slowest jobs are now the **Windows smoke tests** (`pnpm-11` 38m, `dungeon-adventure` 34m), which are dominated by the WSL2 Docker setup (`wsl --install Ubuntu-24.04` + Docker engine) in `init-monorepo`. Linux smoke jobs are 2–16m. So the biggest remaining lever for wall-clock is the Windows WSL2 setup — worth a dedicated follow-up (cache/snapshot the WSL distro, or skip it for Windows smoke tests that don't build Linux containers).

cogwirrel (approved):

Issue #124, opened 2025-03-27
Make generators idempotent


agdimech:
### Description

Make generators idempotent

### Motivation

Re-running generators should result in updated files (where applicable)

### Contribute to feature

- [ ] I'd be willing to implement this feature

cogwirrel:
One part #650

cogwirrel:
Impotency e2e test ensures generators run twice with no drift https://github.com/awslabs/nx-plugin-for-aws/pull/783

PR #772, opened 2026-06-08
fix(generators): make IaC and project-creator generators idempotent on re-run

nx-plugin-for-aws:
### Reason for this change

Re-running several generators with the same options threw a hard `A project already exists` error (from `addProjectConfiguration`/`libraryGenerator`) instead of being a clean no-op. This is one of the gaps flagged in #124: re-running a generator with the same options should be a graceful no-op (or skip with a log), not a failure.

### Description of changes

Added existing-project guards to the four generators that called `addProjectConfiguration` (directly or via `tsProjectGenerator`) unconditionally. First-run output is unchanged (no snapshots changed); a different name still creates a new independent project.

- **ts#infra** (`infra/app/generator.ts`) — guard `tsProjectGenerator` behind a `readProjectConfiguration` existence check (mirrors the ts#dynamodb / trpc#backend pattern). Downstream `updateJson`/`generateFiles`/`addDependencyToTargetIfNotPresent` were already re-run safe.
- **terraform#project** (`terraform/project/generator.ts`) — when the project already exists, use `updateProjectConfiguration` instead of `addProjectConfiguration`; otherwise add as before. The nx.json plugin add and `generateFiles` (Overwrite) were already safe.
- **smithy#project** (`smithy/project/generator.ts`) — guard `addProjectConfiguration` behind an existence check (the only standalone project generator missing the pattern).
- **ts#astro-docs** (`ts/astro-docs/generator.ts`) — return early with an informative `logger.info` if the project already exists, rather than throwing. Since ts#docs delegates to ts#astro-docs, this fixes both.

Added `should be idempotent when re-run with same options` spec tests to each generator (run twice on the same tree, assert no throw, same project count, same file content) plus a `different name creates an independent project` test. `pnpm nx run @aws/nx-plugin:test` passes (2018 tests) with no snapshot changes, confirming first-run output is byte-identical. `build` and `lint` pass.

### Issue # (if applicable)

Relates to #124.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/infra/app/generator.ts
+++ b/packages/nx-plugin/src/infra/app/generator.ts
@@ -49,7 +49,18 @@ export async function tsInfraGenerator(
   schema: TsInfraGeneratorSchema,
 ): Promise<GeneratorCallback> {
   const lib = getTsLibDetails(tree, schema);
-  await tsProjectGenerator(tree, schema);
+
+  let projectExists: boolean;
+  try {
+    readProjectConfiguration(tree, lib.fullyQualifiedName);
+    projectExists = true;
+  } catch {
+    projectExists = false;
+  }
+
+  if (!projectExists) {
+    await tsProjectGenerator(tree, schema);
+  }
 
   // CDK shells out to a container engine to build image assets. Default
   // (docker) needs no env override; finch is set via CDK_DOCKER per
--- a/packages/nx-plugin/src/infra/app/generator.ts
+++ b/packages/nx-plugin/src/infra/app/generator.ts
@@ -26,6 +26,7 @@ import {
   addGeneratorMetadata,
   getGeneratorInfo,
   type NxGeneratorInfo,
+  projectExists,
 } from '../../utils/nx';
 import { sortObjectKeys } from '../../utils/object';
 import { getPackageManagerDisplayCommands } from '../../utils/pkg-manager';
@@ -50,15 +51,7 @@ export async function tsInfraGenerator(
 ): Promise<GeneratorCallback> {
   const lib = getTsLibDetails(tree, schema);
 
-  let projectExists: boolean;
-  try {
-    readProjectConfiguration(tree, lib.fullyQualifiedName);
-    projectExists = true;
-  } catch {
-    projectExists = false;
-  }
-
-  if (!projectExists) {
+  if (!projectExists(tree, lib.fullyQualifiedName)) {
     await tsProjectGenerator(tree, schema);
   }
 
--- a/packages/nx-plugin/src/terraform/project/generator.ts
+++ b/packages/nx-plugin/src/terraform/project/generator.ts
@@ -12,7 +12,6 @@ import {
   joinPathFragments,
   OverwriteStrategy,
   readNxJson,
-  readProjectConfiguration,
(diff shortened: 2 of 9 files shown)


packages/nx-plugin/src/infra/app/generator.ts
```diff
@@ -49,7 +49,18 @@ export async function tsInfraGenerator(
...
   const lib = getTsLibDetails(tree, schema);
-  await tsProjectGenerator(tree, schema);
+
+  let projectExists: boolean;
+  try {
+    readProjectConfiguration(tree, lib.fullyQualifiedName);
+    projectExists = true;
+  } catch {
+    projectExists = false;
+  }
```
cogwirrel: this should be a helper

packages/nx-plugin/src/ts/astro-docs/generator.ts
```diff
@@ -44,6 +46,13 @@ export const tsAstroDocsGenerator = async (
     schema.subDirectory || docsNameKebabCase,
   );
 
+  if (getProjects(tree).has(fullyQualifiedName)) {
+    logger.info(
+      `Project ${fullyQualifiedName} already exists, skipping ts#astro-docs generator.`,
+    );
+    return () => {};
+  }
```
cogwirrel: Shouldn't return here- should skip creating the project config if it exists, but should still continue the rest of the generator to eg add a blog if run first with includeBlog=false then again with includeBlog=true, etc

packages/nx-plugin/src/infra/app/generator.ts
nx-plugin-for-aws: Done — extracted the project-exists try/catch into a `projectExists` helper in `utils/nx.ts` and used it here as well as in `terraform#project` and `smithy#project`.

packages/nx-plugin/src/ts/astro-docs/generator.ts
nx-plugin-for-aws: Good call — changed it to skip only the project configuration creation and continue the rest of the generator. Re-running after escalating options (eg `noBlog` false → true) now adds the blog cleanly. Added an escalation test covering this.

packages/nx-plugin/src/terraform/project/generator.ts
```diff
@@ -229,15 +231,24 @@ export async function terraformProjectGenerator(
...
+  };
+
+  if (projectExists(tree, lib.fullyQualifiedName)) {
+    updateProjectConfiguration(tree, lib.fullyQualifiedName, {
+      ...projectConfiguration,
+      name: lib.fullyQualifiedName,
+    });
+  } else {
+    addProjectConfiguration(tree, lib.fullyQualifiedName, projectConfiguration);
+  }
```
cogwirrel: Don't overwrite project.json when the project exists, just skip project.json updates

nx-plugin-for-aws: Done — on re-run it now skips the project configuration write entirely (rather than overwriting via `updateProjectConfiguration`), so existing project.json customisations are preserved. Added a test covering this.

cogwirrel (approved):

PR #774, opened 2026-06-08
fix(generators): make port assignment and core generators idempotent on re-run

nx-plugin-for-aws:
### Reason for this change

Re-running a generator with the same options (or escalating `--infra none` to a real infra option later) was not a clean no-op. Generators duplicated ports, dependencies, command lines and config, and one generator could wipe user content. This addresses the core idempotency gaps tracked in #124.

### Description of changes

- **`utils/port.ts` (root cause):** `assignPort` previously counted a project's own already-assigned ports as taken on a re-run, allocating a fresh port and growing `metadata.ports`. Added an optional `reusePort` argument so callers can return an already-assigned port without registering a duplicate, plus an exported `getExistingProjectPort` helper. `assignSharedPort` reuses a project's own port if it already has one.
- **Single-port generators** (ts#trpc-api, ts#smithy-api, py#fast-api, ts#rdb): pass `getExistingProjectPort(project)` so a re-run reuses the existing port.
- **Multi-component generators** (ts#agent, py#agent, ts#mcp-server, py#mcp-server): added `getExistingComponentPort` (in `utils/nx.ts`) which looks up the port previously recorded for the same component (by generator id + name). Each generator reuses its component's port on re-run while still assigning fresh ports for brand-new components on the same project.
- **`connection/agent-serve-local.ts`:** added a `not contains` GritQL guard so the `runtimeConfig.<ns>.<Agent>` override is only injected when absent, preventing duplicate assignment lines on re-run (ts#agent#react-connection / py#agent HTTP path).
- **py#fast-api:** the `cp run.sh` bundle command is now only appended if not already present.
- **dynamodb-constructs / rdb-constructs:** use `addDependencyToTargetIfNotPresent` to dedup the `<project>:build` entry on the shared-constructs `build.dependsOn`.
- **ts#rdb:** the bundle target transform now falls back to the existing rolldown command when `options.command` is absent on a re-run, so the rolldown invocation is no longer lost.
- **ts#nx-plugin:** `index.ts` is only blanked when it still holds the default ts#lib template, preserving user content. Extracted as `blankNxPluginIndexIfTemplate`.

First-run output is unchanged (no snapshot changes); only re-run behaviour differs.

- Added `should be idempotent when re-run with same options` tests for the port util and for trpc backend, py#fast-api, smithy-api, ts#agent, py#agent, ts#mcp-server, py#mcp-server, ts#rdb and ts#dynamodb, asserting port metadata does not grow, dependencies/commands are not duplicated and project.json is byte-identical across runs. The ts#rdb test also asserts the rolldown command survives. Added a guard test for ts#nx-plugin index.ts.
- `pnpm nx run @aws/nx-plugin:test -u`: 2023 tests pass, no snapshot changes.
- `pnpm nx run-many --target build --all` passes.
- `pnpm nx run-many --target lint --all --fix` passes.

### Issue # (if applicable)

Relates to #124.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/py/agent/generator.ts
+++ b/packages/nx-plugin/src/py/agent/generator.ts
@@ -32,6 +32,7 @@ import { getNpmScope } from '../../utils/npm-scope';
 import {
   addComponentGeneratorMetadata,
   addDependencyToTargetIfNotPresent,
+  getExistingComponentPort,
   getGeneratorInfo,
   type NxGeneratorInfo,
   readProjectConfigurationUnqualified,
@@ -241,7 +242,17 @@ export const pyAgentGenerator = async (
   // A2A servers use port 9000 as per the Strands A2A SDK default and AgentCore A2A contract.
   // HTTP and AG-UI agents use port 8081+ to avoid conflict with VS Code server on 8080.
   const localDevPortStart = protocol === 'a2a' ? 9000 : 8081;
-  const localDevPort = assignPort(tree, project, localDevPortStart);
+  const localDevPort = assignPort(
+    tree,
+    project,
+    localDevPortStart,
+    getExistingComponentPort(
+      tree,
+      project.name,
+      PY_AGENT_GENERATOR_INFO,
+      agentTargetPrefix,
+    ),
+  );
 
   // All protocols use fastapi dev for hot reload:
   // - HTTP: FastAPI app directly defined in init.py
--- a/packages/nx-plugin/src/py/agent/generator.ts
+++ b/packages/nx-plugin/src/py/agent/generator.ts
@@ -32,7 +32,6 @@ import { getNpmScope } from '../../utils/npm-scope';
 import {
   addComponentGeneratorMetadata,
   addDependencyToTargetIfNotPresent,
-  getExistingComponentPort,
   getGeneratorInfo,
   type NxGeneratorInfo,
   readProjectConfigurationUnqualified,
@@ -242,17 +241,9 @@ export const pyAgentGenerator = async (
   // A2A servers use port 9000 as per the Strands A2A SDK default and AgentCore A2A contract.
   // HTTP and AG-UI agents use port 8081+ to avoid conflict with VS Code server on 8080.
   const localDevPortStart = protocol === 'a2a' ? 9000 : 8081;
-  const localDevPort = assignPort(
-    tree,
-    project,
-    localDevPortStart,
-    getExistingComponentPort(
-      tree,
-      project.name,
-      PY_AGENT_GENERATOR_INFO,
-      agentTargetPrefix,
-    ),
-  );
+  const localDevPort = assignPort(tree, project, localDevPortStart, {
+    component: { info: PY_AGENT_GENERATOR_INFO, name: agentTargetPrefix },
+  });
 
   // All protocols use fastapi dev for hot reload:
   // - HTTP: FastAPI app directly defined in init.py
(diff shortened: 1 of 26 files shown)


packages/nx-plugin/src/py/agent/generator.ts
```diff
@@ -241,7 +242,17 @@ export const pyAgentGenerator = async (
   // A2A servers use port 9000 as per the Strands A2A SDK default and AgentCore A2A contract.
   // HTTP and AG-UI agents use port 8081+ to avoid conflict with VS Code server on 8080.
   const localDevPortStart = protocol === 'a2a' ? 9000 : 8081;
-  const localDevPort = assignPort(tree, project, localDevPortStart);
+  const localDevPort = assignPort(
+    tree,
+    project,
+    localDevPortStart,
+    getExistingComponentPort(
```
cogwirrel: This should be built into assignPort

packages/nx-plugin/src/ts/nx-plugin/generator.ts
```diff
@@ -88,8 +100,12 @@ export const tsNxPluginGenerator = async (
...
   updateProjectConfiguration(tree, fullyQualifiedName, project);
 
-  // Remove the hello world example from index.ts
-  tree.write(joinPathFragments(project.sourceRoot, 'index.ts'), '');
+  // Remove the hello world example from index.ts on first scaffold only,
+  // preserving any user content when the generator is re-run.
+  blankNxPluginIndexIfTemplate(
+    tree,
+    joinPathFragments(project.sourceRoot, 'index.ts'),
+  );
```
cogwirrel: remove this and leave index.ts as-is from the ts#project generator

packages/nx-plugin/src/ts/rdb/generator.ts
cogwirrel: Better to instead leave `commands` as-is if it already exists

packages/nx-plugin/src/utils/port.ts
```diff
@@ -15,6 +32,12 @@ export const assignSharedPort = (
...
 ): number => {
+  // Reuse the project's own port if it already has one (idempotent on re-run)
+  const ownPort = getExistingProjectPort(project);
+  if (ownPort !== undefined) {
+    return ownPort;
+  }
+
   const existingPort = [...getProjects(tree).values()].find(
     (p) => (p.metadata as any)?.generator === generatorId,
   )?.metadata as any;
```
cogwirrel: Are you checking for existing port twice in this function?

nx-plugin-for-aws: Good catch — `assignSharedPort` now checks the project's own port first, then does the sibling lookup with the current project excluded, so there's no longer an overlapping double-check.

packages/nx-plugin/src/py/agent/generator.ts
nx-plugin-for-aws: Done — folded the lookup into `assignPort` via an optional `component` option. `assignPort` now reuses a project's own port (or a component's recorded port) on re-run, so call sites no longer pass a precomputed reuse port. This applies across all the agent/mcp-server/backend generators.

packages/nx-plugin/src/ts/nx-plugin/generator.ts
nx-plugin-for-aws: Done — removed the index.ts blanking entirely; index.ts is now left exactly as the ts#project generator created it. Updated the test accordingly.

packages/nx-plugin/src/ts/rdb/generator.ts
nx-plugin-for-aws: Done — the transform now only runs when `commands` isn't already present, leaving the bundle target as-is on a re-run.

cogwirrel (approved):

PR #779, opened 2026-06-09
fix(ts#react-website): make generator idempotent (preserve user files on re-run)

nx-plugin-for-aws:
### Reason for this change

A new e2e idempotency smoke test runs the full generator matrix, commits, re-runs the same matrix and asserts no git diff. It found that re-running `ts#website` (alias of `ts#react-website`) is destructive.

The generator emitted user-owned source under `src/` with `OverwriteStrategy.Overwrite`. On a second run this regenerated those files back to the bare scaffold, wiping everything later generators (connection, `ts#react-website#auth` / cognito-auth, runtime-config) had added:

- `src/main.tsx` lost all connection providers, the `<CognitoAuth>` wrapper, `RuntimeConfigProvider`, `useRuntimeConfig`/`useAuth`, the `RouterProviderContext` type members and the router `context`.
- `src/components/AppLayout/index.tsx` lost the auth `useAuth()` hook and the sign-out `utilities={[...]}` menu.
- `src/styles.css` flipped quote style (`@import 'tailwindcss';` to `@import "tailwindcss";`) when regenerated and reformatted.

### Description of changes

Generate the user-editable React source with `OverwriteStrategy.KeepExisting` instead of `Overwrite`, for all four `generateFiles` calls that write into the website `src/` (app common, app per-UX, tanstack-router common, tanstack-router per-UX). This covers `main.tsx`, `styles.css`, `config.ts`, `AppLayout`, `__root.tsx`, the routes and the TanStack-managed route tree.

First-run output is unchanged: on a fresh project the `src` directory is deleted and freshly scaffolded before generation, so `KeepExisting` writes the same files as before. Only re-run behaviour changes - existing files are now left alone. Framework-owned config (project targets, tsconfig, vite config via guarded GritQL transforms) continues to update in place.

### Description of how you validated changes

- Added a unit test `should preserve user edits to main.tsx, AppLayout and styles.css on re-run` that runs the generator, writes marker content into `main.tsx`, `AppLayout/index.tsx`, `styles.css`, the routes and `config.ts`, re-runs with the same options and asserts the markers are preserved. Verified this test FAILS on the previous (Overwrite) generator and PASSES with the fix.
- `pnpm nx run @aws/nx-plugin:build` (runs the full unit test suite): 2038 tests pass, no first-run snapshot changes.
- `pnpm nx run-many --target build --all` and `pnpm nx run-many --target lint --all --fix` pass.

### Issue # (if applicable)

Relates to #124.

### Checklist
[...]

PR #778, opened 2026-06-09
test(e2e): add idempotency smoke test

nx-plugin-for-aws:
### Reason for this change

Issue #124 requires generators to be idempotent — re-running with the same options must not overwrite user-touched files, duplicate wiring, or otherwise mutate the workspace. We had per-generator unit tests but no end-to-end check that the **whole matrix** is idempotent together.

### Description of changes

Adds `e2e/src/smoke-tests/idempotency.spec.ts`: it runs the full generator matrix (`runGeneratorMatrix` plus the CDK `ts#infra` projects and a terraform project), `git init` + commits the result, runs the **exact same matrix again**, and asserts `git status --porcelain` is empty. A non-empty status (modified/added/deleted tracked files) means a generator mutated the workspace on re-run and is therefore not idempotent — the failure surfaces the offending files and the diff.

Registered as `idempotency` in the CI and PR smoke-test matrices.

### Description of how you validated changes

Ran locally. The test **works and currently fails**, surfacing real non-idempotency bugs (react-website overwriting user files, website auth throwing on re-run, shared-config key/order/dup churn, terraform metric-tag growth). Companion PRs fix each of these; this PR adds the guard rail. Once the fixes land, the matrix re-run produces no diff.

### Issue # (if applicable)

Relates to #124.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/awslabs/nx-plugin-for-aws/blob/main/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #780, opened 2026-06-09
fix(ts#react-website#auth): make generator idempotent on re-run

nx-plugin-for-aws:
### Reason for this change

A new e2e idempotency smoke test (running the full generator matrix, committing, then re-running the same matrix and asserting no git diff) revealed that the `ts#react-website#auth` / `ts#website#auth` generator was not idempotent. Re-running it on a project that already had auth configured threw a hard error:

> This generator has already been run on \<project\>.

Per #124, every generator must be idempotent — a same-options re-run must be a clean no-op, not a hard error. The throw existed only because the downstream menu AST mutators (GritQL transforms) were not individually idempotent and would otherwise append a duplicate user menu / state declarations on a second run.

### Description of changes

- Removed the hard throw guard (the `components/CognitoAuth/index.tsx` existence check) so a re-run naturally no-ops.
- Added idempotency guards (`where { ... <: not contains <stable marker introduced by the transform> }`) to the four previously-unguarded menu GritQL transforms so re-applying them is a no-op:
  - `cloudscape-auth-menu.grit` — guarded on `iconName: 'user-profile-active'`
  - `none-auth-menu.grit` — guarded on the `signout-link` button
  - `shadcn-auth-menu.grit` — guarded on `aria-label="Open user menu"`
  - `shadcn-state-declarations.grit` — guarded on the `menuOpen`/`setMenuOpen` `useState` declaration

All other steps (generateFiles with `KeepExisting`, import helpers, router-context wiring, component metadata, identity infra) were already individually idempotent, so dropping the throw leaves the workspace byte-identical on re-run. First-run output is unchanged.

- Replaced the `should not be able to run the generator multiple times` test (which asserted a throw) with `should be idempotent when re-run with same options` in `generator.spec.ts`, asserting no throw, byte-identical output, and the wrapper appears exactly once.
- Added per-ux idempotency tests in `generator.ux-provider.spec.ts` for cloudscape, none and shadcn — each runs the generator twice and asserts the menu utilities / state declarations / user menu and `useAuth()` wiring appear exactly once and the file is byte-identical across runs.
- `pnpm nx run @aws/nx-plugin:test -u`: 2040 tests pass. Only snapshot changes were (a) removal of the obsolete throw snapshot and (b) the "None" ux `should update AppLayout` snapshot losing a previously-duplicated user-greeting block — the test's input AppLayout already contained an auth menu, so the old snapshot had captured the buggy double-append that the guard now correctly prevents. Genuine clean first-run snapshots (Cloudscape/Shadcn) are unchanged.
- `pnpm nx run-many --target build --all` passes.
- `pnpm nx run-many --target lint --all --fix` passes.

### Issue # (if applicable)

Relates to #124.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/ts/react-website/cognito-auth/__snapshots__/generator.spec.ts.snap
+++ b/packages/nx-plugin/src/ts/react-website/cognito-auth/__snapshots__/generator.spec.ts.snap
@@ -281,8 +281,6 @@ export class UserIdentity extends Construct {
 "
 `;
 
-exports[`cognito-auth generator > should not be able to run the generator multiple times 1`] = `[Error: This generator has already been run on test-project.]`;
-
 exports[`cognito-auth generator > should update main.tsx when RuntimeConfigProvider exists > main-tsx-with-runtime-config 1`] = `
 "import CognitoAuth from './components/CognitoAuth';
 import { RuntimeConfigProvider } from './components/RuntimeConfig';
--- a/packages/nx-plugin/src/ts/react-website/cognito-auth/__snapshots__/generator.ux-provider.spec.ts.snap
+++ b/packages/nx-plugin/src/ts/react-website/cognito-auth/__snapshots__/generator.ux-provider.spec.ts.snap
@@ -286,26 +286,6 @@ const AppLayout: React.FC<React.PropsWithChildren> = ({ children }) => {
               Sign out
             </button>
           </div>
-          <div className="user-greeting">
-            <span>Hi, {\`\${user?.profile?.['cognito:username']}\`}</span>
-            <button
-              type="button"
-              className="signout-link"
-              onClick={() => {
-                removeUser();
-                signoutRedirect({
-                  post_logout_redirect_uri: window.location.origin,
-                  extraQueryParams: {
-                    redirect_uri: window.location.origin,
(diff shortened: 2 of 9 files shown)


nx-plugin-for-aws:
All checks are green except **Smoke Tests - finch**, which is failing for an environmental reason unrelated to this change:

```
Error: installFinch requires sudo, dpkg, and apt-get — only Debian/Ubuntu hosts are supported
  ❯ installFinch src/smoke-tests/finch-install.ts:173:11
```

The finch smoke test installs the upstream Finch `.deb` via `apt-get`/`dpkg`, but the runner is an Amazon Linux (CodeBuild) host without those tools. This same failure occurs at ~1m40s on the other concurrent idempotency PRs (#778, #779, #781), confirming it is a runner-image limitation rather than a regression from this PR. The `serve-local` failure on the first run was a flaky Python-MCP port collision (`Errno 98 Address already in use` → `UND_ERR_HEADERS_TIMEOUT`) and passed cleanly on re-run.

This PR touches only the `ts#react-website#auth` (cognito-auth) generator and its grit transforms — nothing related to container engines.

nx-plugin-for-aws:
Superseded by #783, which consolidates the idempotency smoke test and all the fixes into a single PR.

PR #782, opened 2026-06-09
fix(dev): add jiti to root devDependencies for pnpm link workflow

AlexTo:
### Reason for this change

When developing locally using `pnpm link dist/packages/nx-plugin`, running any generator fails with:

```Error: Cannot find module 'jiti'```

This happens because the linked dist package resolves its runtime dependencies from the workspace root `node_modules`, not from inside the dist directory. Since jiti was only declared as a dependency in `packages/nx-plugin/package.json` (included in the published package), it is absent from the root workspace `node_modules` and therefore unavailable during the pnpm link dev workflow.

### Description of changes

Add jiti to the root `package.json` devDependencies 

### Checklist

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "fs-extra": "11.3.5",
     "github-slugger": "2.0.0",
     "husky": "9.1.7",
+    "jiti": "2.7.0",
     "js-yaml": "4.2.0",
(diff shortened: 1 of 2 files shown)


cogwirrel (approved):
Great catch, thank you!

PR #784, opened 2026-06-09
feat(generators): harden security defaults for scaffolded apps

nx-plugin-for-aws:
### Reason for this change

A threat-model audit of applications scaffolded by the generators (analysing real generated CDK and Terraform output) surfaced several security best-practice gaps in generated defaults. This PR addresses the subset of improvements that strengthen the production-grade foundations the generators provide while keeping the developer experience. Improvements carrying a genuine DX/behaviour tradeoff are documented separately for maintainer review rather than included here.

### Description of changes

- **`ts#website` (static-website, CDK + Terraform):** add a CloudFront `ResponseHeadersPolicy` / `aws_cloudfront_response_headers_policy` setting `Strict-Transport-Security` (with `preload`), `X-Content-Type-Options`, `X-Frame-Options: DENY` and `Referrer-Policy` on all responses. No default `Content-Security-Policy` (it is application-specific; documented how to add one). `runtime-config.json` is served with `Cache-Control: no-cache` so config changes are never served stale from the edge.
- **`ts#react-website#auth` / identity construct (CDK + Terraform):** drop the `https://<region>.console.aws.amazon.com` OAuth callback/logout URL from the generated Cognito User Pool client — it is unused by the generated app's auth flow and unnecessarily widened the redirect allowlist. Docs recommend removing the `localhost` URLs for production stages.
- **`ts#rdb` (Aurora, CDK):** enable Performance Insights on the writer instance by default (KMS-encrypted), and add an opt-in `enableCloudwatchLogs` prop (default `false`) that exports the correct Aurora log types (`audit`/`error`/`general`/`slowquery` for MySQL, `postgresql` for PostgreSQL). This matches the existing Terraform module's behaviour. Log export is opt-in because enabling it also turns on verbose statement logging (PostgreSQL `log_statement=all`), which can log sensitive data and adds cost.
- **`py#agent` (HTTP):** stop printing the raw model event stream (`print(event)`) and the raw request/exception (`print(request)`/`print(err)`) to stdout; unhandled errors are logged via `logging.exception` (message + traceback, without dumping the request).
- **Docs:** updated the website, auth and rdb guides (English only).

Unit/CI:
- `pnpm nx run @aws/nx-plugin:test -u` — all tests pass; snapshots updated. `lint` passes. Branch rebased onto latest `main` (incl. #785) so the diff contains only the intended changes.

Real deployments (provisioned in AWS, verified live, then torn down):
- **CDK website + auth** (`cdk deploy`): live `curl` confirmed `strict-transport-security: max-age=63072000; includeSubDomains; preload`, `x-frame-options: DENY`, `x-content-type-options: nosniff`, `referrer-policy: strict-origin-when-cross-origin`, and `runtime-config.json` → `cache-control: no-cache`; `describe-user-pool-client` confirmed the AWS Console URL is absent (sandbox stage retains localhost + CloudFront URLs).
- **CDK Aurora rdb** (`cdk deploy` with `enableCloudwatchLogs: true`): live cluster showed `EnabledCloudwatchLogsExports: ["postgresql"]`, Performance Insights enabled (customer-KMS encrypted), and the `/aws/rds/cluster/.../postgresql` log group created. The migration trigger ran `prisma migrate deploy` against the live cluster successfully (stack `CREATE_COMPLETE`).
- **Terraform website + auth** (`terraform apply`): live `curl` confirmed the same security headers (incl. HSTS `preload`) and `runtime-config.json` `Cache-Control: no-cache`; the deployed Cognito client callback URLs were verified.
- **`py#agent` logging:** generated output matches the change; an isolated runtime test confirmed the stream loop emits no stdout (no raw event dump) and the exception handler logs via `logging.exception` with full traceback while not leaking the request body; `py_compile` passes. (The Terraform Aurora template is unchanged by this PR, so no separate TF-rdb deploy was needed.)

### Issue # (if applicable)

N/A

### Checklist

----


packages/nx-plugin/src/py/agent/files/http/init.py.template
```diff
@@ -72,8 +73,7 @@ app.add_middleware(ExceptionMiddleware, handlers=app.exception_handlers)
 
 @app.exception_handler(Exception)
 async def unhandled_exception_handler(request, err):
-    print(request)
-    print(err)
+    logging.exception("Unhandled exception while handling request")
```
cogwirrel: We should still log the exception

nx-plugin-for-aws:
Pushed two follow-up commits addressing review feedback and deploy-testing:

- **`5283f79`** — gate Cognito localhost callback URLs to development stages (CDK `UserIdentity` gains `allowLocalhostCallbackUrls`, defaulting to true for `sandbox`/`dev`/`test` stage names). Verified against a **real `cdk deploy`**: sandbox stage → localhost present; prod stage synth → localhost excluded (CloudFront URL only).
- **`38662fe`** — revert the same gating for **Terraform** only. Deploy-testing surfaced that the Terraform flow creates the user pool client *before* the CloudFront callback URL is appended (via the `add-callback-url` module, with `ignore_changes` on `callback_urls`). Removing localhost for production left an empty callback list at create time, which Cognito rejects (`InvalidOAuthFlowException: CallbackUrls can not be empty`). CDK doesn't have this issue. For Terraform the localhost URLs are retained and documented for manual removal in production.

All changes deploy-tested end-to-end in both CDK and Terraform example workspaces (real `cdk deploy` and `terraform apply`), and resources torn down afterward. Live verification confirmed: CloudFront security headers (incl. HSTS `preload`), `runtime-config.json` `Cache-Control: no-cache`, and stage-gated callback URLs.

nx-plugin-for-aws:
Per review feedback, reverted the automatic stage-inference approach for callback URLs entirely. The `UserIdentity` construct (CDK) and Terraform identity module are now **unchanged from before** — the generated client still includes the localhost callback URLs as it always did. The only change is **documentation**: the auth guide now recommends removing the `http://localhost` callback/logout URLs for production stages (or using a separate dev client). History was cleaned up to drop the intermediate inference commits.

cogwirrel (approved):

PR #787, opened 2026-06-10
docs(pdk-migration): update guide for @aws/nx-plugin 1.0.0-rc.33

nx-plugin-for-aws:
### Reason for this change

The PDK migration guide was pinned to `@aws/nx-plugin@0.50.0` and used an old workspace-creation command and pre-1.0 generator names. This validates and updates the guide against the latest `1.0.0-rc.33` release.

I performed a full real migration of the shopping-list tutorial app end-to-end on `1.0.0-rc.33` — including a live AWS deploy, an authenticated end-to-end API test, and teardown — to confirm every step.

### Description of changes

- **Workspace creation**: replaced the hand-written `create-nx-workspace --preset` Tabs with the documented `<CreateNxWorkspaceCommand>` component, pinned to `1.0.0-rc.33` via the `tag` prop (removed now-unused `Tabs`/`TabItem` import).
- **Version pin** updated from `0.50.0` to `1.0.0-rc.33` in the point-in-time callout.
- **API generator**: use the documented `ts#api` entry point with `framework: 'smithy'` (instead of `ts#smithy-api` directly), and lowercase `auth: 'iam'` (the enum is now lowercase).
- **Website generators**: use the documented `ts#website` (`framework: 'react'`, `ux: 'cloudscape'` — the default UX provider changed to shadcn in #854 and the shopping list app uses CloudScape) and `ts#website#auth` entry points (instead of `ts#react-website` / `ts#react-website#auth`). Same fix applied to the Type Safe API FAQ snippet.
- **Echo operation**: added a step to delete the scaffolded example `Echo` operation handler (`backend/src/operations/echo.ts`), which otherwise breaks the build once the operation is removed from the model.
- **`@aws-northstar/ui` + `ace-builds`**: added a manual step to exclude `ace-builds/webpack-resolver` from the Vite bundle. Northstar bundles a code editor that uses a webpack-specific import which the current Vite/Rolldown bundler can't resolve; the shopping-list app doesn't use that component, so we externalise it. This is documented as a manual tutorial step under "Add AWS Northstar Dependency" since Northstar isn't part of the website generator. (Re-verified still required on `1.0.0-rc.33`.)
- **Local development**: the guide now uses the `dev` target (the `serve-local` target was removed in favour of `dev` in #806). The manual `tsx` install step was dropped since `ts#smithy-api` vends `tsx` as a dev dependency as of #786.

- Ran the entire migration on a fresh `pnpm create @aws/nx-workspace@1.0.0-rc.33` workspace using the exact documented commands.
- `nx run-many --target build` passes for all projects.
- Verified `nx dev website` runs both the website (4200) and Smithy API (3001) local servers.
- Deployed to AWS, verified the deployed API works with IAM SigV4 auth (GET → POST → GET → DELETE → GET), verified the CloudFront website serves, then destroyed the stack.
- Built the docs site (`nx build docs`) — all pages render without errors.

### Issue # (if applicable)

N/A

### Checklist

----


cogwirrel:
closed in favour of #950

PR #802, opened 2026-06-13
feat(agentcore-gateway): support connecting gateways to other gateways

nx-plugin-for-aws:
### Reason for this change

AgentCore Gateways could aggregate MCP servers, but not other Gateways. Composing gateways hierarchically (e.g. a team-level gateway fronting several domain gateways, each fronting its own MCP servers) had no first-class support — there was no `addGateway(...)` construct method and no connection generator for the `agentcore-gateway -> agentcore-gateway` edge.

### Description of changes

Adds an `agentcore-gateway -> agentcore-gateway` connection so a Gateway can aggregate another Gateway's tools behind its own MCP endpoint. Tools surface through the chain as `<gatewayTarget>___<target>___<tool>` (each gateway adds one prefix), and both gateways evaluate their own Cedar policies — the outer authorizes the original caller for the prefixed action, the inner authorizes the outer gateway's role for the inner action.

- **New `agentcore-gateway#gateway-connection` generator** (hidden, dispatched via the `connection` generator). It chains the source gateway's `serve-local` onto the target gateway's `serve-local`, and registers the target gateway in the source's `serve-local.ts` so the local gateway aggregates its tools. Self-connections and cycles (direct or transitive) are rejected up front, since a cycle would recurse infinitely on `tools/list`.
- **Shared `AgentCoreGateway` CDK construct** gains `addGateway(gateway)` / `addGatewayTarget(props)`: registers the target gateway's MCP endpoint as a gateway target, grants the source gateway's role `bedrock-agentcore:InvokeGateway`, and orders target creation after the target gateway and its own targets (AgentCore fetches a target's tools at creation time). Vended app gateway constructs now expose a `gatewayName` used as the default target name. The existing MCP-target path is refactored to share a private `addTarget(...)` helper.
- **Refactor**: the shared `serve-local` wiring is extracted from the `mcp-connection` generator into `attach-upstream.ts`, reused by both connection generators.
- **Terraform**: documented the analogous `aws_bedrockagentcore_gateway_target` + `InvokeGateway` wiring in the new connection guide; the deployed gateway endpoint is used directly as the target endpoint.

- **Unit tests**: new `gateway-connection/generator.spec.ts` (serve-local wiring, three-gateway chains, idempotency, self-connection + direct/transitive cycle rejection, non-gateway/protocol/auth validation, metric tags); new `resolveConnection` case for `agentcore-gateway -> agentcore-gateway`; new assertions in the gateway generator spec for `addGateway`/`addGatewayTarget`/`gatewayName`. Full `@aws/nx-plugin:test` suite passes.
- **Local**: generated two gateways + an MCP server in a test workspace, connected gateway→gateway, and verified the `serve-local` chain (outer :8100 → inner :8101 → MCP :8000) aggregates tools as `inner-gateway___my-mcp___divide` and routes a `callTool` end-to-end. Added a chained-gateways case to the `serve-local` e2e smoke test (passes locally).
- **Deployed**: deployed outer→inner→MCP to AWS via CDK and confirmed listing/calling `inner-gateway___my-mcp___divide` over SigV4 returns the correct result through both gateways (Cedar ENFORCE at each hop), then tore the stack down. Extended the `cdk-deploy` and `terraform-deploy` smoke tests + generator matrix with a parent gateway and an `invokeAgentCoreGatewayTool` helper.
- **Docs**: new `connection/agentcore-gateway-gateway` guide (CDK + Terraform + Cedar-across-chained-gateways + local dev), linked from the connection index, gateway guide next steps, and sidebar.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/agentcore-gateway/gateway-connection/generator.ts
+++ b/packages/nx-plugin/src/agentcore-gateway/gateway-connection/generator.ts
@@ -64,11 +64,15 @@ export const agentcoreGatewayGatewayConnectionGenerator = async (
         `Gateway '${gateway.name}' has protocol='${gateway.protocol}'. Gateway->Gateway connections are only supported between MCP-protocol gateways.`,
       );
     }
-    if (gateway.auth !== 'iam') {
-      throw new Error(
-        `Gateway '${gateway.name}' uses auth='${gateway.auth}'. Gateway->Gateway connections currently require both gateways to use IAM authentication.`,
-      );
-    }
+  }
+
+  // The source gateway invokes the target signing with its own IAM role, so
+  // only the target's inbound auth must be IAM. The source's inbound auth (how
+  // its own callers reach it) is irrelevant to this hop.
+  if (targetGateway.auth !== 'iam') {
+    throw new Error(
+      `Gateway '${targetGateway.name}' uses auth='${targetGateway.auth}'. Gateway->Gateway connections currently require the target gateway to use IAM authentication.`,
+    );
   }
 
   assertNoCycle(tree, sourceProject.name, targetProject.name);
--- a/packages/nx-plugin/src/agentcore-gateway/gateway-connection/generator.ts
+++ b/packages/nx-plugin/src/agentcore-gateway/gateway-connection/generator.ts
@@ -1,0 +1,149 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+  type GeneratorCallback,
+  installPackagesTask,
+  type Tree,
+} from '@nx/devkit';
+import { formatFilesInSubtree } from '../../utils/format';
+import { addGeneratorMetricsIfApplicable } from '../../utils/metrics';
+import { kebabCase } from '../../utils/names';
+import {
+  getGeneratorInfo,
+  type NxGeneratorInfo,
+  readProjectConfigurationUnqualified,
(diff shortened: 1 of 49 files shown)


packages/nx-plugin/src/agentcore-gateway/gateway-connection/generator.ts
```diff
@@ -0,0 +1,149 @@
...
+      throw new Error(
+        `Gateway '${gateway.name}' has protocol='${gateway.protocol}'. Gateway->Gateway connections are only supported between MCP-protocol gateways.`,
+      );
+    }
+    if (gateway.auth !== 'iam') {
+      throw new Error(
+        `Gateway '${gateway.name}' uses auth='${gateway.auth}'. Gateway->Gateway connections currently require both gateways to use IAM authentication.`,
+      );
+    }
+  }
```
cogwirrel, line 76: Why does source auth matter?

nx-plugin-for-aws: Good catch — the source gateway's auth doesn't matter for this hop. The source invokes the target signing with its own execution role (the `GATEWAY_IAM_ROLE` credential provider), regardless of how its own callers authenticate to it. Only the target's inbound auth needs to be IAM. Updated the generator to validate only the target (matching how `mcp-connection` only checks the target MCP server), kept the protocol check on both (both must be MCP), updated the docs, and added a test that a non-IAM source gateway is allowed. Pushed in 57112b36.

cogwirrel (approved):

PR #804, opened 2026-06-14
feat(py#dynamodb): add generator for DynamoDB single table design with PynamoDB

AlexTo:
### Reason for this change

Add a `py#dynamodb` generator for DynamoDB single table design with PynamoDB, along with connection generators for wiring it up to Python Agent, FastAPI, and MCP Server projects.

### Description of changes

* Added `py#dynamodb` generator and connections using similar patterns as `ts#dynamodb`
* All configuration that was previously passed as CLI arguments on each target command (image name, container name, port, table name etc...), and as constants in `constants.ts` and `constants.py` has been consolidated into `config.json` (replaced `dynamodb.config.json`) , which the scripts now read at runtime. This gives cleaner target commands and standardises how configuration is read between languages.
* Moved the three local dev scripts: `pull-image`, `start-container`, and `create-local-table`, out of each project's scripts/ directory and into a shared common location (`common/ts-dynamodb/scripts` or `common/py-dynamodb/scripts`). Since these scripts are identical across all DynamoDB projects, they are now vended once and are referenced from each project's targets.
* Minor improvement: put switching logic for container engines in the script instead of in the template at generation time and set `containerEngine` in `config.json` to allow switching `containerEngine` after the package has been generated.

* Unit tests
* Manually vended the projects and connections, tested `serve-local` and deployed with CDK/Terraform

### Checklist

----


docs/src/content/docs/en/guides/py-dynamodb.mdx
```diff
@@ -0,0 +1,170 @@
...
+</FileTree>
+
+The local development scripts are shared across all Python DynamoDB projects and generated once into:
+
+<FileTree>
+  - packages/common/py-dynamodb/scripts
+    - create_local_table.py Creates the DynamoDB table in the local DynamoDB Local instance
+    - pull_image.py Pulls the DynamoDB Local image
+    - start_container.py Starts the DynamoDB Local container
+</FileTree>
```
cogwirrel, line 48: I wonder if we can instead refactor the `common/scripts` project out of `ts#infra` (vended for the stage config feature), and reuse the same typescript scripts for both ts and py dynamodb generators?

It seems like the scripts do the same thing and every workspace always needs a `package.json` etc so can always make use of typescript scripts with `tsx` :)

docs/src/content/docs/en/guides/py-dynamodb.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+Add or update entity files under `<name>/entities/`, using the generated example entity as a starting point:
+
+```python title="packages/my_table/my_table/entities/example.py"
+from datetime import UTC, datetime
+from typing import Iterator
+from pynamodb.attributes import UnicodeAttribute
+from .base import BaseModel
+
+class ExampleModel(BaseModel, discriminator='ExampleModel'):
```
cogwirrel, line 73: How does the `DescriminatorAttribute` fit into this? Is that handled by the `BaseModel`? Might be good to go into a little bit more detail in the docs about how to use this as the pynamodb docs seem quite light on polymorphism :)

docs/src/content/docs/en/guides/py-dynamodb.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+<FileTree>
+  - \<name>
+    - \_\_init\_\_.py Package exports
+    - client.py DynamoDB client and table name resolution
+    - entities
+      - base.py Base PynamoDB model with GSI declarations
+      - example.py Example entity definition
+      - \_\_init\_\_.py Entity exports
+  - config.json Table configuration including GSI definitions and local development settings
```
cogwirrel, line 37: Nice I like the consolidation here! :)

packages/nx-plugin/src/py/dynamodb/files/__name__/entities/base.py.template
```diff
@@ -0,0 +1,57 @@
...
+    Uses PynamoDB polymorphism to support multiple entity types in a single table.
+    See https://pynamodb.readthedocs.io/en/stable/polymorphism.html
+    """
+
+    class Meta:
+        table_name: str
+        region = os.environ.get("AWS_REGION", 'us-east-1')
+        host = f"http://localhost:{_serve_local()['port']}" if is_local() else None
+        aws_access_key_id = '0123456789' if is_local() else None
+        aws_secret_access_key = 'ABCDEFGHIK' if is_local() else None
```
cogwirrel: can we say `'UNUSED'` or something better than these placeholders? :)

packages/nx-plugin/src/py/dynamodb/files/__name__/client.py.template
```diff
@@ -0,0 +1,37 @@
...
+import os
+from typing import Any, cast
+
+# Generated: must match runtimeConfigKey in config.json
+runtime_config_key = '<%= runtimeConfigKey %>'
+
+def _serve_local() -> dict:
+    config_path = os.path.join(os.path.dirname(__file__), '..', 'config.json')
+    with open(config_path) as f:
+        return json.load(f)['serveLocal']
```
cogwirrel, line 13: Worth a comment noting that this must be guarded by an `is_local` call as `config.json` won't be included in the deployed bundle :)

packages/nx-plugin/src/utils/ts-dynamodb-scripts/files/create-local-table.ts.template
```diff
@@ -1,33 +1,39 @@
...
+  ProjectionType,
   ResourceInUseException,
+  TableStatus,
   UpdateTableCommand,
-} from '@aws-sdk/client-dynamodb';
-import { readFileSync } from 'fs';
-import { join, dirname } from 'path';
-import { fileURLToPath } from 'url';
+} from "@aws-sdk/client-dynamodb";
+import { AttributeType } from "aws-cdk-lib/aws-dynamodb";
```
cogwirrel: I think better not to depend on aws-cdk-lib here as it's quite a large dependency that may not be used otherwise (ie terraform case)

cogwirrel (commented):
Thanks @AlexTo ! Looks great! Will be awesome to have a python option for this with local dev built in! 😄

docs/src/content/docs/en/guides/py-dynamodb.mdx
AlexTo: oh I don't get it? dynamodb can be vended without/before `ts#infra` right, or am I missing something?

cogwirrel: So iirc as part of the “enableStageConfig” feature the ts infra generator creates a common/scripts project :) we can do the same in both ts/py dynamodb (creating the project if it doesn’t exist just like common/constructs), then order doesn’t matter :)

And the same typescript scripts can go in common/scripts for both ts and py dynamodb so we don’t need to maintain a Python equivalent :)

AlexTo: Done. Split the infra code from `shared-scripts.ts` into `shared-infra-scripts.ts` (used by `ts#infra`) and add another `shared-dynamodb-scripts.ts` (used by `*#dynamodb`)

The generated folder structure is now as follows

image

`py#dynamodb` and `ts#dynamodb` share the same scripts, only differ by `config.json` and the entities.

docs/src/content/docs/en/guides/py-dynamodb.mdx
AlexTo: Done, I added more documentation about access pattern first guideline, common patterns such as `one-to-many`, `many-to-many` in the context of `PynamoDB`

packages/nx-plugin/src/py/dynamodb/files/__name__/entities/base.py.template
AlexTo: Done

packages/nx-plugin/src/py/dynamodb/files/__name__/client.py.template
AlexTo: Comment added to both ts and py.

packages/nx-plugin/src/utils/ts-dynamodb-scripts/files/create-local-table.ts.template
AlexTo: I accidentally used VS-Code intellisense to add `AttributeType` thinking it would come from `@aws-sdk/client-dynamodb` but it brought in `aws-cdk-lib` :D. Switched to `ScalarAttributeType` from `@aws-sdk/client-dynamodb` instead.

docs/src/content/docs/en/guides/py-dynamodb.mdx
cogwirrel: Perfect, thank you!

docs/src/content/docs/en/guides/py-dynamodb.mdx
cogwirrel: The updated docs are really helpful, much clearer, thank you! :)

packages/nx-plugin/src/py/dynamodb/files/__name__/client.py.template
```diff
@@ -0,0 +1,38 @@
+import json
+import os
+from typing import Any, cast
+
+# Generated: must match runtimeConfigKey in config.json
+runtime_config_key = '<%= runtimeConfigKey %>'
+
+# config.json is only present in local dev (not included in the deployed bundle); the call to _serve_local must be guarded by an is_local() call.
```
cogwirrel: Looks like this fails linting and is the cause of the smoke test failures :)

cogwirrel (commented):
Looks great, thank you! Just that lint issue to fix and we should be good to merge! 😄 

Thanks so much for all your contributions!! 🙌

packages/nx-plugin/src/py/dynamodb/files/__name__/client.py.template
nx-plugin-for-aws: Fixed in 1422f18 (pushed to this branch on Jack's instruction). The comment was a single 145-char line, exceeding the project's ruff `line-length = 120` (E501), so a freshly generated project failed its own lint target. I split it across two lines:

```python
# config.json is only present in local dev (not in the deployed bundle),
# so calls to _serve_local must be guarded by is_local().
```

Verified the generated project now lints/builds clean and updated the affected generator snapshots (all 37 `py#dynamodb` tests pass).

cogwirrel (approved):

AlexTo:
you're welcome @cogwirrel :D

Issue #718, opened 2026-05-28
[TRACKING] Roadmap to v1.0


nx-plugin-for-aws:
### Overview

This issue tracks the work required to release `@aws/nx-plugin` v1.0 — the first stable release signaling production-readiness and a reliable upgrade story.

### What v1.0 means

- **Migrations support**: Users can run `nx migrate @aws/nx-plugin` to upgrade safely
- **Idempotent generators**: Re-running generators is safe and predictable
- **License compliance**: Generated projects have verified dependency license compatibility

---

#### 1. Nx Migrations

Establish the infrastructure for automated code migrations when users upgrade.

- [x] #703
- [x] #704
- [x] #705
- [x] #808
- [x] #988
- [x] #814
- [x] #815

#### 2. Idempotency

Ensure generators behave predictably when re-run.

- [x] #124

#### 3. Adding to Existing Projects

Enable adoption in existing Nx and non-Nx workspaces.

- [x] #153
- [x] #493
- [x] #706

#### 4. License Dependency Checks

Ensure all vended dependencies are license-compliant.

- [x] #707

#### 5. Naming Consolidation

Simplify and future-proof generator naming.

- [ ] #708
- [ ] #709
- [ ] #710
- [ ] #711
- [ ] #712
- [x] #721
- [ ] #722
- [x] #723

#### 6. Documentation

- [ ] #713

#### 7. Security

- [ ] #717
- [ ] #715
- [ ] #716

#### 8. Finch Support

- [ ] #160
- [ ] #714

#### 9. Testing

- [x] #748

#### 10. Release Cutover

Done last, once every workstream above is complete.

- [x] #737

---

### Definition of Done

All issues above are closed, CI is green, documentation is published, and the release blog post is live.

nx-plugin-for-aws:
Nx Migrations workstream: strategy finalized around Nx 23's agentic `nx migrate` (deterministic / prompt / hybrid migration kinds). Work order locked in: **#703 (foundations: migrations.json wiring + version stamping + contributing guide) → #704 (scaffold generator) → #705 (e2e migration smoke tests) → #808 (single-source upgrade guides + MCP tool + agent prompts)**. Details in the latest comments on each issue; #703 PR in progress.

cogwirrel:
Released! https://github.com/awslabs/nx-plugin-for-aws/actions/runs/34074691205/job/101621123935

PR #806, opened 2026-06-17
feat(nx-plugin): migrate to Nx 23 and standardize on dev target

nx-plugin-for-aws:
### Reason for this change

Two related changes for the Nx 23 line:

1. **Migrate the plugin and all generated workspaces to Nx 23.**
2. **Remove the confusing `serve-local` target and commit to a single `dev` target** for local development. Previously every generator vended both a `serve-local` runner and a thin `dev` alias, plus a `serve-local` Vite mode and `SERVE_LOCAL` env var — three near-identical concepts that were easy to confuse.

### Description of changes

**Nx 23 migration** (first commit)
- Bump Nx and `@nx/*` packages to 23, update generated workspace dependencies and config accordingly.

**Standardize on `dev`** (second commit)
- Component generators (`ts#agent`, `py#agent`, `ts#mcp-server`, `py#mcp-server`, `agentcore-gateway`) now author a `<name>-dev` runner directly; the project-level `dev` target aggregates each component's `<name>-dev` (so `nx run <project>:dev` starts everything, and `<name>-dev` runs a single component).
- Standalone-project generators (`ts#rdb`, `ts#dynamodb`, `py#dynamodb`, `py#fast-api`, `trpc#backend`, `smithy#ts#api`) author `dev` as the runner.
- Rename the website Vite mode `serve-local` -> `local-dev` and the `SERVE_LOCAL` env var -> `LOCAL_DEV` in all vended code.
- Rename serve-local-named files/symbols to local-dev: connection helpers (`addTargetToLocalDev`, etc.), the gateway's `local-dev.ts`, the `config.json` `localDev` key, and the e2e smoke spec.
- Docs: recommend the project-level `dev` target; for component guides, note `<name>-dev` disambiguates a single component when several are generated.

- `pnpm nx run @aws/nx-plugin:compile` — clean.
- Full plugin unit suite (2383 tests) passes; snapshots regenerated.
- `pnpm nx run-many --target build --all` and docs site build (1262 pages, all internal links valid).
- E2e smoke spec renamed/updated; CI smoke matrix entry `serve-local` -> `local-dev`.

### Checklist

----


nx-plugin-for-aws:
### Follow-up: custom local generators fail under Nx 23 native strip-types (to revisit)

While testing this migration end-to-end I found one issue we should fix in a follow-up (noting it here so we don't lose it):

**Running a generator from a generated `ts#nx-plugin` project fails.** e.g. after `ts#nx-plugin` + `ts#nx-generator`:

```
nx g @my-scope/plugin:my#generator --exampleOption=test
```

```
NX   ⨯ Unable to compile TypeScript:
error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0.
NX  Failed to load .../generator.ts under Node's native TypeScript stripping.
```

**Root cause:** the scaffolded `generator.ts` imports its schema type with a *value* import:

```ts
import { MyGeneratorGeneratorSchema } from './schema';
```

but `schema` only exists as `schema.d.ts` (type-only). Under Nx 23's new default (Node native TypeScript stripping), Node can't tell the import is type-only, tries to resolve `./schema` as a runtime module, fails, and falls back to ts-node — which TypeScript 6.0.3 then rejects because ts-node defaults to the deprecated `moduleResolution: node10` (TS5107). The result is that local generators don't run at all.

**Verified fix:** emit a type-only import in the `ts#nx-generator` template:

```ts
import type { MyGeneratorGeneratorSchema } from './schema';
```

With `import type`, strip-types elides the import and the generator runs cleanly with no swc/ts-node fallback required (confirmed locally with `NX_PREFER_NODE_STRIP_TYPES=true` and no `@swc-node/register`/`@swc/core` installed). Installing the swc toolchain is an alternative workaround, but `import type` is the cleaner fix and keeps the generated plugin dependency-light.

We'll tackle this in a separate change.

nx-plugin-for-aws:
### Update: custom-generator fix is now in this PR

Following the investigation in my earlier comment, I dug into how Nx 23 loads local generators (`handle-import.js` + `register.js`): it loads a workspace's `.ts` generator via Node's **native type stripping**, treating it as **ESM or CJS per the nearest `package.json` `"type"`**, and only falls back to swc/ts-node if that load *throws*. Since generated workspaces are `type: module`, the vended generator runs as **ESM**, which exposed two issues in the `ts#nx-generator` `local` template:

1. **Type-only names in a value import** (`GeneratorCallback`, `Tree` from `@nx/devkit`) → ESM linker error → ts-node fallback → `TS5107` under TypeScript 6.0.
2. **`__dirname`** is undefined in ESM scope → `ReferenceError` at execution time (fails even with swc installed, because the module already loaded via strip-types).

Fixed in this PR, scoped to the **vended `local` template** (the only surface that runs as ESM in user workspaces):
- Split type-only imports into `import type { ... }`.
- `__dirname` → `import.meta.dirname`.

Verified end-to-end: the generated generator now runs under pure native strip-types with **no `@swc-node/register` / `@swc/core` dependency**.

The plugin's own ~40 generators are compiled to **CommonJS** and loaded as `.js` (never strip-typed), so they're unaffected. A full ESM migration of the plugin itself (and an audit of other vended `.ts` templates) is tracked separately in #814, linked to the v1 roadmap (#718).

nx-plugin-for-aws:
### Correction: reverted the generator ESM change — the original template was not actually broken

After the build/smoke failures on the previous push, I dug deeper and the earlier "custom generator fails under strip-types" finding turned out to be a **false alarm caused by my local test environment**, not a real defect. Reverting in 8b65bd05.

What actually happens in a real generated workspace:

- Nx 23 loads a local `.ts` generator via native strip-types, and **falls back to swc/ts-node when the load throws** (e.g. the type-in-value-import linkage error from `@nx/devkit`).
- `@swc-node/register` + `@swc/core` are **transitively present** in every generated workspace (peer deps of `@nx/devkit`, `@nx/js`, `@nx/react`, …). So the fallback succeeds, the file is compiled ESM→CJS, and `__dirname` resolves. The original template both **compiles** (the plugin is CJS via `tsc`) and **runs** correctly — and this is exactly what the green smoke run on the pre-existing branch tip already exercised (the matrix generates *and runs* `plugin:my#generator`).

My local repro had failed only because I had manually `pnpm remove`d swc, which doesn't reflect a real workspace.

Why the proposed change was wrong:

- `__dirname` → `import.meta.dirname` **breaks the plugin's `tsc` build** (`TS1343`: `import.meta` is illegal under the plugin's `module: commonjs`, which Nx still requires — see #494 / nrwl/nx#15682). That's what failed every smoke test.
- `import type` alone (keeping `__dirname`) **compiles but fails at runtime**: removing the value-import linkage error means strip-types no longer falls back to swc, so the file runs as ESM where `__dirname` is undefined. The original value-import + `__dirname` only works *because* the two together trigger the CJS fallback.

So the template is correct as-is for the CJS-plugin model. The genuine forward-looking work — migrating the plugin and its generators to ESM so `import.meta` becomes valid — remains tracked in #814 (no longer described as fixed here).

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -228,9 +228,11 @@ This denies the Python agent role from calling `ts-mcp___divide` while leaving t
 
 ## Local Development
 
-The generator adds a `<name>-serve-local` target to the Gateway project, which runs `serve-local.ts`: a local gateway exposing a single MCP endpoint that aggregates every attached MCP server (connected via the <Link path="guides/connection/agentcore-gateway-mcp">`agentcore-gateway#mcp-connection` generator</Link>), with tools prefixed `<target>___<tool>` to match the deployed Gateway. Running it starts the local gateway and all attached MCP servers together.
+The generator adds a `<name>-dev` target to the Gateway project, which runs `local-dev.ts`: a local gateway exposing a single MCP endpoint that aggregates every attached MCP server (connected via the <Link path="guides/connection/agentcore-gateway-mcp">`agentcore-gateway#mcp-connection` generator</Link>), with tools prefixed `<target>___<tool>` to match the deployed Gateway. Running it starts the local gateway and all attached MCP servers together. Use the project's `dev` target:
 
-<NxCommands commands={["<name>-serve-local"]} />
+<NxCommands commands={["dev <name>"]} />
+
+If you have generated multiple components into the same project, run a single Gateway on its own with its `<name>-dev` target.
```
cogwirrel: Delete this line

docs/src/content/docs/en/guides/local-development.mdx
```diff
@@ -65,10 +65,23 @@ local: Local {
...
+### Projects with multiple components
+
+Some project types can hold multiple components in a single project (for example a TypeScript or Python project containing several agents and MCP servers). For these:
+
+- The project-level `dev` target starts **all** of the project's components together. Each component is added to `dev` as it is generated:
+
+<NxCommands commands={['dev my-project']} />
+- Each component also exposes a `<component-name>-dev` target, so you can run a single component on its own:
+
+<NxCommands commands={['run my-project:my-component-dev']} />
```
cogwirrel: Use “xxxx-dev project” format rather than nx run

docs/src/content/docs/en/guides/local-development.mdx
```diff
@@ -65,10 +65,23 @@ local: Local {
...
+
+Because Nx identifies tasks by `project:target`, any component `dev` target shared between components (such as a common MCP server connected to two agents) runs exactly once when you start the project-level `dev`.
```
cogwirrel: Delete this line

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -407,9 +407,9 @@ This command uses `uv run` to execute your Agent using the [Bedrock AgentCore Py
 
 The generator configures a `<your-agent-name>-chat` Nx target that drops you into an interactive terminal chat with your agent.
 
-The chat target runs standalone. By default it connects to your locally running agent, so start `<your-agent-name>-serve-local` first (in a separate terminal):
+The chat target runs standalone. By default it connects to your locally running agent, so start the agent's `<your-agent-name>-dev` target first (in a separate terminal):
 
-<NxCommands commands={['run your-project:agent-serve-local']} />
+<NxCommands commands={['run your-project:agent-dev']} />
```
cogwirrel: Use “agent-dev project” syntax

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Done — deleted this line in 4dd567b6.

docs/src/content/docs/en/guides/local-development.mdx
nx-plugin-for-aws: Done — switched to the `my-component-dev my-project` format in 4dd567b6.

docs/src/content/docs/en/guides/local-development.mdx
nx-plugin-for-aws: Done — deleted this line in 4dd567b6.

docs/src/content/docs/en/guides/py-agent.mdx
nx-plugin-for-aws: Done — switched to the `agent-dev your-project` format in 4dd567b6 (also applied the same fix to the equivalent ts-agent guide).

cogwirrel (dismissed):

cogwirrel (dismissed):

cogwirrel (approved):

PR #803, opened 2026-06-14
feat(ts#agent,py#agent): standalone agent-chat with local + deployed support for all protocols

nx-plugin-for-aws:
### Reason for this change

The `agent-chat` target only let you chat with a *locally-running* agent: it depended on `serve-local` (auto-starting the agent), and for A2A / AG-UI it invoked `agent-chat-cli` directly as a binary against the local URL. There was no way to chat with a **deployed** agent on Bedrock AgentCore, and the HTTP-only chat script meant the three protocols behaved inconsistently.

### Description of changes

Makes `agent-chat` a standalone target for **every** protocol (`http`, `a2a`, `ag-ui`) across both TypeScript and Python:

- The target no longer `dependsOn` `serve-local`. You run `serve-local` yourself (or point at a deployed agent). Python HTTP still depends on its `generate-client` target.
- Every protocol now vends a small `scripts/<agent>/chat.ts` plus a shared `agentcore.ts` helper (new shared generator util `utils/agent-chat/`).
- By default chat connects to the local `serve-local` server (`process.env.URL`).
- When `RUNTIME_CONFIG_APP_ID` is set, chat resolves the deployed agent's runtime ARN from AppConfig and connects to Bedrock AgentCore, authenticating with **SigV4** (IAM) or a **Cognito bearer token** via `AGENT_ACCESS_TOKEN`.
- Bumps `agent-chat-cli` `0.2.0 → 0.3.0`, which adds `A2AChatAdapter({ clientFactory })` and `AGUIChatAdapter({ fetch })` injection points — these enable authenticated calls to deployed agents from a single unified script per protocol.
- Updates the `ts#agent` / `py#agent` guides and the dungeon-adventure tutorial (run `serve-local` first, then `agent-chat`; documents deployed-agent chat with IAM/Cognito).

### Description of how you validated changes

- Updated and ran the `ts#agent` / `py#agent` generator unit tests; full `@aws/nx-plugin` suite green (2258 tests).
- Added `agent-chat` connection assertions to the `serve-local` e2e for all six agent variants (a new `chatConnects` helper drives each target headlessly and asserts the CLI connects).
[...]

PR #807, opened 2026-06-17
feat(py#agent): add langchain framework for AG-UI agents

dsturnbull:
### Reason for this change

The `py#agent` generator only emits Strands agents. Teams wanting a LangChain/LangGraph agent have to generate a Strands one and hand-replace its guts, which drifts on every regenerate — and there's no way to wire MCP servers or A2A agents into a LangChain agent via the `connection` generator.

The `framework` schema option already existed but was inert (enum was `"strands"` only, and `generator.ts` never read it). This change makes it real.

### Description of changes

Add `"langchain"` to the `framework` enum so `framework=langchain` + `protocol=ag-ui` emits a LangChain/LangGraph AG-UI agent (`create_agent` + `ChatBedrockConverse` + `InMemorySaver`, served via `ag_ui_langgraph`), and teach the `connection` generators to wire MCP servers and A2A agents into it. The default model is `global.anthropic.claude-haiku-4-5-20251001-v1:0` (overridable via `MODEL_ID`).

Notable decisions:

- **`framework` is the first real framework branch point.** Strands was keyed only on `protocol`; this branches on `framework` for deps, template directory, and license exceptions. Default stays `"strands"` — existing behaviour unchanged.
- **Templates split into `files/common/<framework>/` and `files/ag-ui/<framework>/`.** The Strands templates move by pure git rename, so Strands output is byte-identical and the snapshot diff is additive only.
- **`langchain` is restricted to `protocol=ag-ui`** (fail-fast throw). The `http`/`a2a` server templates are Strands-specific by construction.
- **MCP connection:** tools load at agent-construction time, which runs inside uvicorn's event loop — so the loader uses a worker-thread loop rather than `asyncio.run()` (which raises in that context). Uses `langchain-mcp-adapters`; no Strands pulled in.
- **A2A connection:** reuses the Strands `A2AAgent` purely as transport (it carries no model), wrapped as a `langchain_core` tool. A LangChain agent with an A2A connection therefore still pulls `strands-agents[a2a]` for that transport only.
- **Gateway connection** fails fast for LangChain (it'd need its own gateway client); use a direct MCP connection or a Strands agent for now.
- **License:** the `ag-ui-langgraph` exception uses the hyphenated name pip-licenses actually reports (an underscore silently never matches). `jsonpatch`/`jsonpointer` (unconditional `langchain-core` deps, BSD-3-Clause but with only free-text "Modified BSD License" metadata) get per-package exceptions rather than a broad allowlist alias.

Docs (`en/guides/py-agent.mdx`): framework-neutral intro, dropped the "What is Strands" section, added a "LangChain supports AG-UI only" note and a short LangChain aside (not full tabs). Non-English locales are regenerated by the translation bot in a separate commit.

The Dockerfile, CDK wiring, nx targets, ports, and AG-UI edge contract are unchanged from the Strands path.

- **Unit tests:** langchain cases across the `py#agent` generator and the MCP/A2A/gateway connection specs (the connection ones assert the gritql transforms produce correct `create_agent` output, no Strands leakage, and the gateway fail-fast), plus a `known-exceptions.spec.ts` pinning the license exception names. Existing Strands/snapshot tests unchanged.
- **Smoke tests:** the generator matrix scaffolds a langchain agent with MCP + A2A connections; `serve-local` connects the langchain agent to MCP (so it boots the MCP loader under uvicorn) and exercises it with the mock model.
- **Manual local verification (not CI):** generated an MCP-connected LangChain agent and ran `typecheck`, `license-check`, and `serve-local` against live Bedrock with the MCP tool firing. This surfaced three defects now fixed here: the `ChatBedrockConverse` `model=`/`model_id=` alias (runs but fails `ty`), the `asyncio.run()`-under-a-running-loop crash, and the hyphenated license-exception name.

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----


packages/nx-plugin/src/py/agent/files/ag-ui/langchain/main.py.template
```diff
@@ -0,0 +1,88 @@
...
+
+    session_id = request.headers.get(SESSION_ID_HEADER) or get_current_session_id()
+
+    async def event_generator():
+        # Set any per-request context HERE, inside the generator: this is the context that
+        # drives agui_agent.run() and the LangGraph tool execution. ChatBedrockConverse has
+        # no native async, so a synchronous tool runs via run_in_executor wrapped in
+        # copy_context().run — context set here, before run(), reaches the tool. Re-bind the
+        # session id here too, since the streaming body runs in its own task outside the
+        # middleware's contextmanager.
```
cogwirrel: Please could we make this and the other vended comments in main.py/agent.py a little more succinct? :)

packages/nx-plugin/src/py/agent/files/common/langchain/agent.py.template
```diff
@@ -0,0 +1,40 @@
...
+
+from langchain.agents import create_agent
+from langchain_aws import ChatBedrockConverse
+from langchain_core.tools import tool
+from langgraph.checkpoint.memory import InMemorySaver
+
+# Pass region_name explicitly so an ambient AWS_REGION does not silently shadow the
+# region the agent should run in. The Bedrock provider otherwise defaults to us-east-1.
+REGION = os.environ.get("AWS_REGION", "us-east-1")
+# ChatBedrockConverse requires a model id (unlike Strands, which picks a Bedrock default).
```
cogwirrel: No need to mention differences to strands etc :)

packages/nx-plugin/src/py/agent/files/common/langchain/agent.py.template
```diff
@@ -0,0 +1,40 @@
...
+# Override with the MODEL_ID env var to use a different model / inference profile.
+MODEL_ID = os.environ.get("MODEL_ID", "anthropic.claude-3-5-haiku-20241022-v1:0")
```
cogwirrel: Should we pick something more modern like haiku 4.5?

packages/nx-plugin/src/py/agent/generator.ts
```diff
@@ -92,6 +95,19 @@ export const pyAgentGenerator = async (
...
   const protocol = options.protocol ?? 'http';
+  const framework = options.framework ?? 'strands';
+
+  // The langchain framework currently ships only an AG-UI server template. The
+  // http and a2a server templates are Strands-specific by construction (http's
+  // main.py consumes a contextmanager yielding a strands.Agent; a2a uses
+  // strands.multiagent.a2a.A2AServer), and create_agent returns a compiled
+  // LangGraph graph that is incompatible with those. Fail fast rather than emit
+  // a Strands server over langchain dependencies.
+  if (framework === 'langchain' && protocol !== 'ag-ui') {
```
cogwirrel, line 107: Please could we update the `py#agent` generator docs to mention that langchain only supports ag-ui? 

And worth a general pass of the docs page (and other related pages) as it's quite strands specific at the moment, eg I think we can get rid of the "What is strands" section and use tabs or optionfilter for "working with your agent" to have docs for both strands and langgraph (if you use tabs, extra points for nice svg icons for the framework logos haha)

cogwirrel (commented):
Thanks so much for this! This was certainly the intention of the `--framework` option! 🙌 Really appreciate the contribution!!

A few minor comments inline, and some more below:

1. I think you'll need to rebase and update the snapshots for the build to pass as we updated some versions recently :)

1. I wonder if you tested out the connection generator for langchain py#agent -> mcp server and a2a agent? I think worth adding some tests:
   1. unit tests for the agent -> mcp and agent -> agent generators to make sure the gritql updates work on the langchain agent.py
   1. update the generator matrix for the smoke tests to include a langchain agent with connections to an mcp server and a2a agent
   1.  update the serve-local test to also include connecting to a langchain agent (https://github.com/awslabs/nx-plugin-for-aws/pull/803 is now merged which tests that the `agent-chat` cli works against each agent language/protocol) - you'll need to follow a similar pattern of swapping out for the mock openai model :)

1. For the license-check you can add another skip for agui like https://github.com/awslabs/nx-plugin-for-aws/blob/main/e2e/src/smoke-tests/license-check.spec.ts#L139 - let's chat about the other jsonpatch transitive deps :)

packages/nx-plugin/src/py/agent/files/ag-ui/langchain/main.py.template
dsturnbull: Trimmed — thanks!

packages/nx-plugin/src/py/agent/files/common/langchain/agent.py.template
dsturnbull: Done, dropped the Strands comparisons.

packages/nx-plugin/src/py/agent/generator.ts
dsturnbull: Done the de-strands-ify pass (framework-neutral intro, dropped "What is Strands", added a "LangChain supports AG-UI only" note). One deviation: I scoped "Working with Your Agent" to a short LangChain aside rather than full framework tabs — it was getting heavy and the detail lives in the LangChain docs. Happy to do the tabs if you'd rather.

dsturnbull:
> Thanks so much for this! This was certainly the intention of the `--framework` option! 🙌 Really appreciate the contribution!!
> 
> A few minor comments inline, and some more below:
> 
> 1. I think you'll need to rebase and update the snapshots for the build to pass as we updated some versions recently :)

Rebased, regenerated, thanks for pointing it out :)

> 2. I wonder if you tested out the connection generator for langchain py#agent -> mcp server and a2a agent? I think worth adding some tests:
>    
>    1. unit tests for the agent -> mcp and agent -> agent generators to make sure the gritql updates work on the langchain agent.py
>    2. update the generator matrix for the smoke tests to include a langchain agent with connections to an mcp server and a2a agent
>    3. update the serve-local test to also include connecting to a langchain agent (#803 is now merged which tests that the `agent-chat` cli works against each agent language/protocol) - you'll need to follow a similar pattern of swapping out for the mock openai model :)
> 3. For the license-check you can add another skip for agui like https://github.com/awslabs/nx-plugin-for-aws/blob/main/e2e/src/smoke-tests/license-check.spec.ts#L139 - let's chat about the other jsonpatch transitive deps :)

- Unit tests for the langchain agent→mcp and agent→a2a transforms (and gateway → fail-fast, since it'd need its own client -- any tips?).
- Added the langchain agent + mcp/a2a connections to the generator matrix.
- serve-local now connects the langchain agent to MCP and exercises it with the mock model — that's what caught the `asyncio.run()`-under-uvicorn crash.

I added the `ag-ui-langgraph` skip (had to use the hyphenated name; the underscore silently never matches pip-licenses).

On jsonpatch/jsonpointer: they're unconditional `langchain-core` deps, genuinely BSD-3-Clause (LICENSE files match the SPDX template), but their wheels only carry free-text "Modified BSD License" metadata. I went with per-package exceptions rather than a broad "Modified BSD License" allowlist alias, so we don't auto-bless anything else self-labelling that way. Open to your preference here :)

packages/nx-plugin/src/py/agent/files/common/langchain/agent.py.template
dsturnbull: Default is now `global.anthropic.claude-haiku-4-5-20251001-v1:0`.

cogwirrel (dismissed):

cogwirrel:
Looks great, thanks @dsturnbull!

Might need a little bit of iteration on the build and tests but overall approach looks great!

> Unit tests for the langchain agent→mcp and agent→a2a transforms (and gateway → fail-fast, since it'd need its own client -- any tips?).

Mind raising a GitHub issue to track `langchain -> gateway`? In there we should also mention the general issue of our MCP clients being agent framework specific (I think this is the case for both ts and py) and to refactor so that we vend an agnostic client + framework specific wrappers if necessary :)

nx-plugin-for-aws:
Heads up on the `ag-ui-langgraph` license exemption added here (`AG_UI_LANGGRAPH_EXCEPTIONS`) — the upstream missing-license-metadata issue has just been fixed. As of today, **`ag-ui-langgraph==0.0.42`** (published 2026-06-19) ships proper license metadata in its wheel:

```
License-Expression: MIT
License-File: LICENSE
ag_ui_langgraph-0.0.42.dist-info/licenses/LICENSE
```

The `0.0.41` pin in this PR still lacks it, but bumping to `0.0.42` should let the dependency checker resolve the license natively and drop the `ag-ui-langgraph` exemption (the `jsonpatch`/`jsonpointer` BSD-3-Clause free-text exceptions are unrelated and would still be needed).

This mirrors the `ag-ui-strands` situation, which had the same problem — see #817 (now merged) for the bump-and-remove-exemption pattern, in case it's useful to fold the same cleanup into this PR.

PR #834, opened 2026-06-23
feat(connection): guard against unconsidered connection permutations

nx-plugin-for-aws:
### Reason for this change

A common pitfall when contributing to the plugin is not being aware of the permutations of connections that need to be considered. A connection's behaviour depends on the options on *each* side — an agent's `framework`/`protocol`/`auth`, an MCP server's `auth`, a database's `engine`, and so on. Adding a new option value (e.g. a new agent framework) silently introduces a whole new set of source × target permutations that the connection generators have never been asked to handle. Nothing today signals that those permutations were missed, so it's easy to ship a connection that crashes or vends broken code for a combination nobody thought about.

The requirement: for every permutation, the `connection` generator must **either** vend valid, buildable code **or** throw a clear "not supported" error — never crash or silently emit broken code.

### Description of changes

Adds a permutation guard centred on the connection generator that makes "I forgot to consider that combination" a **build failure** rather than a runtime surprise. It derives everything from existing sources of truth, so it grows automatically as options are added:

- **`connection/permutations.ts`** — reads each connection type's enum dimensions live from its `schema.json`. The enums *are* the source of truth; there's no second list of options to keep in sync. Also provides full-cartesian enumeration and a bounded sampler.
- **`connection/connection-expectations.ts`** — declares, for every supported `source -> target` routing pair (type-bound to `ConnectionKey`), a function classifying each permutation as `supported` or `unsupported` (with the `RegExp` the thrown error must match). This is where contributors record *intent*.
- **`connection/permutations.spec.ts`** (unit tripwire) — holds a committed snapshot of every connection type's schema enums and fails the build if the live schema drifts from it, and asserts every cell of the full source × target cartesian is classified. Adding an enum value fails this test until both the snapshot and the expectations are updated.
- **`connection/permutations-behaviour.spec.ts`** (unit) — runs the **real** `connection` generator on an in-memory tree for every permutation marked `unsupported`, asserting it throws the declared error.
- **`e2e/.../connection-permutations.spec.ts`** (e2e) — drives the actual CLI and a real build: supported agent variants (cognito auth, AG-UI/A2A protocols) must build; unsupported ones must fail at the CLI. Cases derive from the same expectation table, so it auto-extends.
- Exports `SUPPORTED_CONNECTIONS` and `ConnectionKey` from the connection generator (previously module-private) so the guard can bind to them.
- Documents the guard and the "what to do when this build fails" workflow in **CONTRIBUTING.md**, next to the Generator Idempotency section.

The concrete `py#agent` example from the issue is fully covered: adding a new `framework` value immediately fails `permutations.spec.ts` for every agent-involving pair, forcing each new permutation to be classified as supported (and wired up) or unsupported (with a clear error).

- `pnpm nx run @aws/nx-plugin:test` — full suite green (119 files, 2623 tests).
- New connection tests: 43 (exhaustiveness/tripwire) + 225 (behavioural) pass.
- Verified the tripwire actually trips: temporarily added `langgraph` to the `py#agent` framework enum and confirmed `permutations.spec.ts` fails with a message pointing at `connection-expectations.ts` + `KNOWN_DIMENSIONS`; reverted.
- `pnpm nx run @aws/nx-plugin:compile`, e2e `build` and `lint` targets all pass.

### Issue # (if applicable)

N/A.

### Checklist

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,6 +102,33 @@ it('should be idempotent when re-run with same options', async () => {
 - **Component generators**: run twice with the same inputs and assert the tree is unchanged after the second run (no duplicate wiring), and that adding a second differently-named component leaves the first intact.
 - **Guarded refusals**: assert the generator throws the expected error on re-run.
 
+### Connection Permutations
+
+A common pitfall when working on connections is forgetting that a connection has to behave for _every_ combination of its source and target options, not just the defaults. The agent generators are the clearest example: a `py#agent` can connect to MCP servers, gateways, other agents and databases, and each side carries options — `framework`, `protocol`, `auth` — whose enum values multiply into many connection permutations. Adding a new value (say a new agent `framework`) silently introduces a whole new set of permutations that the connection generators have never been asked to handle.
+
+#### The principle
+
+> **Every connection permutation must be consciously considered.** For each combination of source and target options, the `connection` generator must _either_ vend valid, buildable code _or_ throw a clear "not supported" error. It must never crash unexpectedly or silently vend broken code.
+
+#### The guard
+
+The permutation guard lives next to the connection generator and makes "I forgot to consider that combination" a build failure rather than a runtime surprise:
+
+- **`connection/permutations.ts`** reads the enum dimensions of each connection-participating generator live from its `schema.json`. The enums _are_ the source of truth — there is no second list of options to keep in sync.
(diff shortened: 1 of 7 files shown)


nx-plugin-for-aws:
Closing — wrong approach. The committed snapshot + expectation table still relies on the contributor remembering to update them, which defeats the purpose. Replacing with an automatic, enum-driven test that exercises real connections for every agent framework so a newly-added framework that forgets connections fails CI on its own.

PR #835, opened 2026-06-23
test(e2e): browser-driven website integration tests for APIs and agents

nx-plugin-for-aws:
### Reason for this change

The `serve-local`, `cdk-deploy` and `terraform-deploy` e2e tests verified that a generated website served HTML and that its backends were independently reachable, but **never exercised the website's actual connection to its APIs and agents**. A user-facing regression — the website's vended client failing to reach a connected API/agent (wrong URL, broken auth, runtime-config mismatch) — would pass CI.

While building this out, exactly such a bug surfaced: the Python HTTP agent's `serve-local` runtime-config override pointed at `http://localhost:PORT/invocations`, but the website's generated OpenAPI client appends the `/invocations` operation path itself, so a browser request from the website hit `/invocations/invocations` → 404. The deployed path was unaffected (its base URL is derived from the AgentCore ARN, with no suffix).

### Description of changes

Adds a shared `e2e/src/smoke-tests/website-integration.ts` helper that runs a **real browser (Playwright) integration test**:

- Writes an example page into the generated website that uses the website's **own vended hooks/clients** (API hooks, agent hooks, the SigV4 auth hook) to invoke every connected API (discovered generically from `runtime-config`) and agent (TS/Python HTTP + AG-UI). No test code is vended to users — the page is added only in the test workspace, and it exercises the same clients a real app would.
- Installs Chromium on demand (idempotent), so no CI workflow change is needed.
- For deployed sites, creates a Cognito test user (MFA off) and completes the **hosted-UI login** before driving the page; SigV4 signing then happens through the website's own auth hook. Serve-local needs no auth.
- Drives the page in a browser and asserts every API and agent responded.

Wired into all three smoke tests so they stay in lock-step as the generators evolve:

- **serve-local**: connect the website to all 4 agent protocols (TS/Python HTTP + AG-UI) so every protocol is exercised locally; replaces the old "Website cascade" test with the browser-driven one (which also guards the Python-HTTP fix).
- **cdk-deploy / terraform-deploy**: deploy + grant the website's authenticated Identity Pool role invoke access to the connected Smithy API and the HTTP + AG-UI agents, and drive the deployed CloudFront site post-login.

**Bug fix**: the Python HTTP agent `serve-local` runtime-config override now omits the trailing `/invocations`, matching the existing chat-target convention (HTTP → bare base URL, AG-UI → `/invocations`).

- Built a representative example workspace (website + tRPC/FastAPI APIs + TS/Python HTTP + AG-UI agents + connections), then drove the real serve-local website in headless Chromium. Confirmed all APIs and all 4 agent protocols invoke successfully through the website's vended clients, and reproduced + fixed the Python-HTTP `/invocations` 404.
- Added a unit-test assertion guarding the fix. Unit tests pass.
- Real `cdk-deploy` and `terraform-deploy` e2e runs to AWS are in progress to validate the deployed (Cognito + SigV4) path end-to-end.

### Issue # (if applicable)

N/A

### Checklist

----


e2e/src/smoke-tests/website-integration.ts
```diff
@@ -0,0 +1,551 @@
...
+ */
+export const writeIntegrationTestPage = (
+  websiteRoot: string,
+  apiNames: string[],
+  agents: AgentSpec[],
+): string => {
+  const routesDir = join(websiteRoot, 'src', 'routes');
+  mkdirSync(routesDir, { recursive: true });
+
+  // Each agent contributes an import, a hook call, and an invocation. The
```
cogwirrel: Can we simplify the construction of this file with generateFiles from nx devkit? Create and flush an fstree?

e2e/src/smoke-tests/website-integration.ts
```diff
@@ -0,0 +1,551 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Page = any;
+interface IntegrationDeps {
+  chromium: { launch: () => Promise<Browser> };
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  cognito: Record<string, any>;
+}
+
+/**
+ * Install playwright and the Cognito SDK (plus the Chromium browser) into an
```
cogwirrel: We can’t just add these to monorepo dev deps?

e2e/src/smoke-tests/website-integration.ts
```diff
@@ -0,0 +1,551 @@
...
+    // wired to / granted). Every generated API exposes GET /echo; tRPC wraps
+    // the reply in result.data.message, REST and Smithy return it at the top.
+    const apis: Record<string, string> = runtimeConfig.apis ?? {};
+    const connectedApis: string[] = ${JSON.stringify(apiNames)};
+    for (const name of connectedApis) {
+      const baseUrl = apis[name];
+      await record('api:' + name, async () => {
+        if (!baseUrl) throw new Error('missing runtime-config api: ' + name);
+        const trimmed = baseUrl.replace(/\\/$/, '');
+        const trpc = await sigv4.fetch(
```
cogwirrel: We need to use the vended clients from the connection generators to invoke the APIs

nx-plugin-for-aws: Done in 5f8db80. The page now invokes each connected API through its vended client — tRPC via `useMyApiClient().echo.query({ message })` and OpenAPI REST/Smithy via `use<Api>Client().echo({ message })` — instead of hand-rolled SigV4 fetches. APIs are passed as typed `ApiSpec[]` (`{ kind: 'trpc' | 'openapi', className }`). Validated end-to-end: the generated page type-checks and `nx build website` passes against the real generated clients.

e2e/src/smoke-tests/website-integration.ts
nx-plugin-for-aws: Done in 5f8db80. The page is now rendered from an EJS template (`e2e/src/files/website-integration/routes/integration-test.tsx.template`) via `generateFiles` into an `FsTree`, then `flushChanges` writes it to disk — same machinery the generators use. Only the hooks for the connected APIs/agents are emitted, so the page still satisfies the website's `noUnusedLocals` build without the previous conditional string-assembly.

e2e/src/smoke-tests/website-integration.ts
nx-plugin-for-aws: I dug into this and you're right that publishing isn't a concern — root `devDependencies` don't propagate to the published `@aws/nx-plugin` package, which declares its own deps. The blocker is purely a lockfile-resolution side effect in this monorepo:

This workspace currently resolves `@nx/react@22.7.5` into **three** peer variants in `pnpm-lock.yaml` — differing by whether their transitive `@nx/vite` links against vite 7 vs vite 8 (we have both majors present). Adding **any** new root dep (I tested playwright alone, the Cognito SDK alone, and both) makes pnpm re-resolve and (a) dedupe the vite-7 variant away, and (b) bump `@emnapi/core` 1.10→1.11 across the whole `@nx/*` tree. A no-op `pnpm install` with no package.json change is stable.

That re-resolution regenerates **15 `ts#react-website` snapshots** — a mix of legitimate vite-8 alignment (dropping `nxViteTsPaths`/`nxCopyAssetsPlugin`, matching what a real generated workspace actually gets) and unrelated churn from the emnapi bump (`.prettierrc`/`.prettierignore` reformat, `<!doctype>`→`<!DOCTYPE>`, tsconfig `extends`/`outDir`). I couldn't suppress the dedupe/emnapi bump (tried `dedupe-peer-dependents=false` and pinning `@emnapi` via overrides — neither held).

So adding the deps can't be a clean, single-concern change — it drags ~15 unrelated react-website snapshot updates into this e2e PR. To keep the lockfile and generated-output snapshots byte-identical to main, I've left playwright + the Cognito SDK installed at runtime into the gitignored `e2e/.integration-deps` (loaded via a dedicated `require`). Happy to split the vite-8 snapshot correction into its own PR if you'd like it fixed independently — it looks like a latent staleness on main.

nx-plugin-for-aws: Follow-up: per your point about wanting weekly updates, I've moved both versions into the plugin's `versions.ts` (`TS_VERSIONS`) in ff80861, so the weekly `update-versions` workflow now bumps playwright and `@aws-sdk/client-cognito-identity-provider` alongside everything else. The e2e helper imports them from there (`TS_VERSIONS.playwright` etc.) rather than hardcoding. They're not added to any generator's `withVersions(...)` call, so they aren't vended into generated projects — they only feed the isolated `e2e/.integration-deps` install, which still keeps the root lockfile (and the react-website vite snapshots) byte-identical to main.

nx-plugin-for-aws: Resolved properly in c503835 — you were right on both counts. Added playwright + `@aws-sdk/client-cognito-identity-provider` to the workspace devDependencies and now import them directly (dropped the isolated `.integration-deps` install entirely). Two clarifications from digging in:

- The react-website snapshot updates this triggers are a **fix, not churn**: adding the deps re-resolves the `@nx/react`/`@nx/vite` peer variant to the **vite 8** one a real generated workspace actually gets. The committed snapshot encoded a stale vite-7 resolution. The regenerated snapshot drops `nxViteTsPaths`/`nxCopyAssetsPlugin` and the `prettier` devDependency — aligning with the biome migration. Only one snapshot file changed (verified on a clean `node_modules`).
- No `versions.ts` entry needed: the weekly `update-versions` workflow runs `npm-check-updates` over the root `package.json` directly, so the new devDeps get bumped like everything else.

cogwirrel (approved):

PR #836, opened 2026-06-23
feat(py#agent): support langchain agents + mcp/a2a/gateway connections

dsturnbull:
### Reason for this change

The `py#agent` generator only emits Strands agents, and the `connection` generators only wire connections into Strands agents. Teams wanting a LangChain/LangGraph agent have to generate a Strands one and hand-replace its guts, which drifts on every regenerate, and there is no supported way to wire an MCP server, an A2A agent or an AgentCore Gateway into a LangChain agent.

Since #818 and #822 made the MCP and gateway connection clients framework-agnostic (a shared transport, signer and endpoint resolver under generic `AgentCore*` names, with a thin per-framework Layer-2 client carrying the framework suffix), adding a second framework is now the additive change those refactors were designed for. This adds LangChain on that seam, for Python.

### Description of changes

Add `langchain` as a framework to `py#agent` and wire MCP, A2A and AgentCore Gateway connections into it, reusing the framework-agnostic Layers 0 and 1 untouched and adding only the LangChain-specific Layer-2.

**Python (`py#agent`): full support.** `framework=langchain` + `protocol=ag-ui` emits a LangChain/LangGraph AG-UI agent (`create_agent` + `ChatBedrockConverse` + `InMemorySaver`, served via `ag_ui_langgraph`). MCP, A2A and AgentCore Gateway connections are all supported. Default model is `global.anthropic.claude-haiku-4-5-20251001-v1:0` (overridable via `MODEL_ID`). `langchain` is restricted to `protocol=ag-ui` (fail-fast for `http`/`a2a`, which are Strands-specific server shapes).

- **MCP / Gateway** use LangChain Layer-2 clients (`langchain-mcp-adapters`); each returned tool opens a fresh MCP session per call so the tools stay usable after agent construction.
- **A2A** reuses the framework-agnostic Strands `A2AAgent` purely as transport (it carries no model), wrapped as a `langchain_core` `@tool` and spread into `create_agent`. The agent's own framework stays LangChain.

**TypeScript: no langchain.** There is no LangChain framework for TypeScript agents — `ts#agent` is Strands-only. (An earlier revision of this PR added TypeScript langchain connection clients; those have been removed.)

[...]

PR #841, opened 2026-06-24
feat(py#agent): support http and a2a protocols for langchain agents

nx-plugin-for-aws:
### Reason for this change

#836 added the `langchain` framework to `py#agent` but restricted it to the `ag-ui` protocol. Teams wanting a LangChain agent over plain HTTP (JSONL streaming, like the Strands HTTP agent) or exposed as an A2A agent had no supported path. This adds the remaining two protocols so `langchain` reaches parity with `strands` across all three.

### Description of changes

Add the `http` and `a2a` server templates for LangChain and lift the ag-ui-only guard.

- **HTTP** (`files/http-langchain/main.py`): reuses the framework-agnostic `init.py` (the FastAPI `app` + `JsonStreamingResponse` the OpenAPI client is generated from) and streams the compiled `create_agent` graph via `graph.astream(stream_mode="messages")`, yielding `StreamChunk(content=...)` JSONL — the **same wire shape as the Strands HTTP agent**, so the generated typed client, chat CLI, and `react` connection work unchanged.
- **A2A** (`files/a2a-langchain/main.py`): drives the graph from an `a2a-sdk` `AgentExecutor` served by `A2AStarletteApplication`, wrapped in the same outer FastAPI (`/ping` + session middleware) as the Strands A2A agent and advertising `AGENTCORE_RUNTIME_URL` in its agent card. Pulls `a2a-sdk[http-server]`, no Strands.
- **generator.ts**: lift the guard, select per-framework protocol dirs (`http-langchain`/`a2a-langchain`), and make the langchain deps protocol-conditional. Session forwarding reuses the framework-agnostic `session_context`; the langgraph checkpointer is keyed by the AgentCore session via `thread_id`.
- **DynamoDB connection**: already works for a LangChain agent unchanged — `py#agent -> py#dynamodb` only adds a workspace dependency + serve-local chain (no `agent.py` transform).

- **Unit tests**: `py#agent` generator cases for langchain `http` (reuses `init.py`, streams JSONL, langchain deps, no strands/ag-ui) and `a2a` (a2a-sdk `AgentExecutor` + `A2AStarletteApplication`, `a2a-sdk` dep, no strands). Snapshots updated.
- **e2e**: langchain http + a2a agents added across the generator matrix, serve-local (JSONL invoke + A2A card/stream + chat) and the CDK/Terraform deploy specs (deployed + invoked over each protocol); plus a langchain→dynamodb connection.
- **Manual, against live Bedrock** in a sample workspace: the langchain HTTP agent streamed JSONL `{"content": ...}` with its tool firing (12−7=5); the langchain A2A agent returned the delegated answer over the standard A2A client (20−8=12); the dynamodb connection wired the workspace dep + serve-local chain and type-checked.

### Issue # (if applicable)

Follow-up to #836.

### Checklist

### DIFF
--- a/packages/nx-plugin/src/py/agent/files/a2a-langchain/main.py.template
+++ b/packages/nx-plugin/src/py/agent/files/a2a-langchain/main.py.template
@@ -6,9 +6,9 @@ from a2a.server.agent_execution import AgentExecutor, RequestContext
 from a2a.server.apps import A2AStarletteApplication
 from a2a.server.events import EventQueue
 from a2a.server.request_handlers import DefaultRequestHandler
-from a2a.server.tasks import InMemoryTaskStore
-from a2a.types import AgentCapabilities, AgentCard
-from a2a.utils.message import new_agent_text_message
+from a2a.server.tasks import InMemoryTaskStore, TaskUpdater
+from a2a.types import AgentCapabilities, AgentCard, Part, TextPart
+from a2a.utils import new_task
 from fastapi import FastAPI, Request
 from starlette.middleware.base import BaseHTTPMiddleware
 
@@ -28,14 +28,22 @@ _graph = get_agent()
 
 class _GraphAgentExecutor(AgentExecutor):
     async def execute(self, context: RequestContext, event_queue: EventQueue) -> None:
+        if context.message is None:
+            raise ValueError("A2A request is missing a message")
+        task = context.current_task or new_task(context.message)
+        await event_queue.enqueue_event(task)
+        updater = TaskUpdater(event_queue, task.id, task.context_id)
+        await updater.start_work()
         # The session bound by the middleware drives the LangGraph thread, keeping
         # checkpointed conversation state separate per caller session.
-        session_id = get_current_session_id() or context.context_id or str(uuid.uuid4())
+        session_id = get_current_session_id() or task.context_id
         result = await _graph.ainvoke(
             {"messages": [{"role": "user", "content": context.get_user_input()}]},
             {"configurable": {"thread_id": session_id}},
         )
-        await event_queue.enqueue_event(new_agent_text_message(result["messages"][-1].content))
+        reply = result["messages"][-1].content
+        await updater.add_artifact([Part(TextPart(kind="text", text=reply))])
+        await updater.complete()
 
     async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
         raise NotImplementedError("Cancellation is not supported.")
@@ -53,7 +61,7 @@ _agent_card = AgentCard(
     version="1.0.0",
     default_input_modes=DEFAULT_MODES,
     default_output_modes=DEFAULT_MODES,
-    capabilities=AgentCapabilities(streaming=False),
+    capabilities=AgentCapabilities(streaming=True),
     skills=[],
 )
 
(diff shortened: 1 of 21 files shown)


packages/nx-plugin/src/py/agent/files/http-langchain/main.py.template
```diff
@@ -0,0 +1,79 @@
...
+    return "".join(
+        part.get("text", "")
+        for part in chunk.content
+        if isinstance(part, dict) and part.get("type") == "text"
+    )
+
+
+async def handle_invoke(input: InvokeInput):
+    """Streaming handler for agent invocation"""
+    config = {"configurable": {"thread_id": get_current_session_id() or "default"}}
```
dsturnbull: Session memory bleeds across requests. `get_current_session_id()` is called inside `handle_invoke`, which is the `JsonStreamingResponse` body generator. Starlette's `BaseHTTPMiddleware` exits its `with session_id_context(...)` block when `call_next` returns, but the streaming body runs after that, so the contextvar has already been reset to `None` (`session_id_context` does `_session_id_var.reset(token)` in its `finally`). Result: `thread_id` is always `"default"`, so every HTTP request shares one LangGraph checkpointer thread and conversation memory leaks between sessions.

The merged `ag-ui-langchain/main.py` template hit and documented exactly this ("Re-bind the session: the streaming body runs outside the middleware"). The fix is to capture the session in the route handler (still inside middleware scope) and re-bind inside the generator:

```python
async def handle_invoke(input: InvokeInput, session_id: str):
    with session_id_context(session_id):
        config = {"configurable": {"thread_id": session_id}}
        async for token, _meta in _graph.astream(...):
            ...

@app.post("/invocations", ...)
async def invoke(input: InvokeInput) -> JsonStreamingResponse:
    session_id = get_current_session_id() or str(uuid.uuid4())
    return JsonStreamingResponse(handle_invoke(input, session_id))
```

packages/nx-plugin/src/py/agent/files/a2a-langchain/main.py.template
```diff
@@ -0,0 +1,94 @@
...
+        if isinstance(part, dict) and part.get("type") == "text"
+    )
+
+
+class <%= agentNameClassName %>Executor(AgentExecutor):
+    async def execute(self, context: RequestContext, event_queue: EventQueue) -> None:
+        updater = TaskUpdater(event_queue, context.task_id, context.context_id)
+        await updater.submit()
+        await updater.start_work()
+        config = {"configurable": {"thread_id": context.context_id or "default"}}
```
dsturnbull: The AgentCore session header is plumbed in (the `_SessionIdMiddleware` binds `session_id_context`) but the executor never reads it: `thread_id` is keyed on `context.context_id` (the A2A protocol's own conversation id), so `session_id_context` is bound and then unused, and the `x-amzn-bedrock-agentcore-runtime-session-id` header does not drive checkpointer memory. Is that intended? If the A2A `context_id` is the canonical thread key, consider dropping the now-dead `session_id_context` import + middleware here; if the AgentCore session should win, key the thread on `get_current_session_id() or context.context_id`.

dsturnbull (dismissed):
Nice one!

cogwirrel (approved):

PR #844, opened 2026-06-26
fix(preset): remove Prettier in favour of Biome

nx-plugin-for-aws:
### Reason for this change

The base Nx workspace created by `create-nx-workspace` scaffolds a Prettier config (`.prettierrc`, `.prettierignore`) and a `prettier` devDependency. The preset generator adds Biome for formatting and linting, but left Prettier in place. As a result, generated workspaces shipped with **both** formatters — an unused `prettier` dependency competing with Biome.

This was confirmed by generating a fresh workspace with the current (1.0 RC) plugin: after running the preset, `biome.json` is present but `.prettierrc`, `.prettierignore`, and `"prettier"` in `package.json` remained.

### Description of changes

In the preset generator, after writing `biome.json`:
- Delete `.prettierrc` and `.prettierignore` if present.
- Remove `prettier` from the workspace `dependencies`/`devDependencies`.

- Added a unit test that seeds a Prettier config + `prettier` devDependency and asserts the preset removes both while keeping `biome.json`.
- Updated the preset snapshot (the `.prettierrc` snapshot is removed).
- Verified end-to-end against a freshly generated 1.0 RC workspace: running the updated preset removes `prettier 3.8.4` from the install and deletes the config files.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/preset/__snapshots__/generator.spec.ts.snap
+++ b/packages/nx-plugin/src/preset/__snapshots__/generator.spec.ts.snap
@@ -57,8 +57,6 @@ exports[`preset generator > should run successfully > .mcp.json 1`] = `
 "
 `;
 
-exports[`preset generator > should run successfully > .prettierrc 1`] = `"{"singleQuote":true}"`;
-
 exports[`preset generator > should run successfully > .vscode/mcp.json 1`] = `
 "{
   "servers": {
--- a/packages/nx-plugin/src/preset/generator.spec.ts
+++ b/packages/nx-plugin/src/preset/generator.spec.ts
@@ -2,7 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-import { readJson, readNxJson, type Tree } from '@nx/devkit';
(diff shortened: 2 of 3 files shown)


nx-plugin-for-aws:
Closing — this fixes a non-problem.

I verified against a **real** workspace generated via `@aws/nx-workspace` (the actual user flow, using the locally-built plugin through the e2e verdaccio harness). The generated workspace has **no prettier at all**:
- `.prettierrc`: absent
- `.prettierignore`: absent
- `prettier` in any `package.json`: absent
- `biome.json`: present, formatter enabled

The prettier artifacts that motivated this PR came from a flawed reproduction:
1. I'd generated a workspace with `create-nx-workspace --preset=ts`/`--preset=apps` — **built-in Nx presets** that force prettier — which is not the AWS flow.
2. The preset unit test's `.prettierrc` came from `createTreeWithEmptyWorkspace()` (an Nx *testing* helper that seeds it), not from any real generator.

For a third-party preset like `@aws/nx-plugin`, Nx's `new` generator defaults `formatter: none` and skips the built-in `generatePreset` path that adds prettier; neither `create-nx-workspace` nor `@aws/create-nx-workspace` injects `--formatter`; and our preset already calls `initGenerator(formatter: 'none')`. So no prettier is ever introduced.

No code change needed.

PR #866, opened 2026-06-30
ci: cache npm/yarn downloads and shallow-clone smoke tests

nx-plugin-for-aws:
### Reason for this change

Smoke test jobs are the long pole of PR/CI runs. Investigating step-level timings on a recent run, the slowest matrix jobs (notably `npm` at ~16m) are dominated by **repeated package downloads**: smoke tests generate a fresh workspace and install it with the target package manager, hitting the public registry **uncached on every run**. The pnpm store is already cached in `init-monorepo` (logs show `reused 1916` of 1919), but npm and yarn get no cache and re-download every tarball.

Two avenues, both low-risk:
1. Cache the npm and yarn download directories like pnpm's store already is.
2. Smoke test jobs check out with `fetch-depth: 0` (full history) but run a **fixed** test rather than `nx affected`, so they don't need history.

### Description of changes

- **`init-monorepo` action**: add `actions/cache` entries for the npm cache dir (`npm config get cache`) and the yarn classic + berry global cache dirs, keyed on `packages/nx-plugin/src/utils/versions.ts` (the file that pins what the generators install into test workspaces) with a prefix `restore-keys` fallback for broad hits.
- **`pr.yml` / `ci.yml`**: switch the Linux and Windows smoke test job checkouts to `fetch-depth: 1`. The package/build/release jobs keep `fetch-depth: 0` (nx-set-shas and release tagging need history).

Note: bun is intentionally **not** cached — the e2e `global-setup.ts` disables bun's install cache (`[install.cache] disable = true`), so a cache would never be populated.

- Validated all three YAML files parse (`js-yaml`).
- Confirmed `fetch-depth` changes are scoped correctly: only the four smoke test checkouts changed to `1`; package/build/release remain `0`.
- Cache effectiveness will be confirmed by watching this PR's smoke test timings (first run primes the cache; the effect compounds on subsequent runs).

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/.github/actions/init-monorepo/action.yml
+++ b/.github/actions/init-monorepo/action.yml
@@ -131,6 +131,32 @@ runs:
         key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
         restore-keys: |
           ${{ runner.os }}-pnpm-store-
+    # Smoke tests generate workspaces that install with npm and yarn, hitting
+    # the public registry uncached on every run. Persist their download caches
+    # (keyed on the versions the generators pin) to avoid re-downloading.
+    - name: Get npm/yarn cache directories
+      id: pm-cache
+      shell: bash
+      run: |
+        echo "NPM_PATH=$(npm config get cache)" >> $GITHUB_OUTPUT
+        echo "YARN_CLASSIC_PATH=${HOME}/.cache/yarn" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


nx-plugin-for-aws:
Closing this — profiling the slowest smoke job (npm, ~17min) showed the bottleneck is the generator matrix (92% of the job: ~89 sequential generations with ~29 installs), not registry download speed. The hot install path is pnpm (already cached), so download-caching npm/yarn barely applies. The real win is reducing the number of installs (batch to one install at the end) and parallelising the matrix. Pursuing that instead.

PR #884, opened 2026-07-03
feat(ts#website#auth): use L2 UserPoolDomain construct

drskur:
### Reason for this change

The L1 `CfnUserPoolDomain` only exposes raw CloudFormation attributes.
Switching to the L2 `UserPoolDomain` gives us helpers like `baseUrl()` and lets us configure the various options for new app clients in a more type-friendly way going forward. This is also groundwork for dcr-proxy.

### Description of changes

Upgraded the user pool domain in the generated `user-identity.ts` template from the L1 `CfnUserPoolDomain` to the L2 `UserPoolDomain` construct:

- `userPoolDomain` field is now typed as `UserPoolDomain` (was `CfnUserPoolDomain`); `createManagedLoginBranding` accepts the L2 type accordingly.
- `createUserPoolDomain` now builds a `UserPoolDomain` with  `cognitoDomain.domainPrefix` set to the derived prefix, and `managedLoginVersion` set to the `ManagedLoginVersion.NEWER_MANAGED_LOGIN` enum (replacing the raw `2`).
- Since the L2 `UserPoolDomain` wraps the same underlying `CfnUserPoolDomain`, the synthesized resource is identical — this is a construct-level change only, with no CloudFormation output diff.
- The L2 construct additionally exposes `baseUrl()` for downstream consumers.

No impact on the connection flow: `cognitoProps` still carries only `region` / `identityPoolId` / `userPoolId` / `userPoolWebClientId` and never referenced the domain.

Updated the `cognito-auth` generator unit tests: the derived-domain spec regex now matches the new `domainPrefix` field (previously `domain`), and the generator snapshot was regenerated to reflect the L2 construct usage. Ran the test suite with snapshot updates and verified the build passes.

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/ts/react-website/cognito-auth/__snapshots__/generator.spec.ts.snap
+++ b/packages/nx-plugin/src/ts/react-website/cognito-auth/__snapshots__/generator.spec.ts.snap
@@ -101,13 +101,14 @@ import {
 import {
   AccountRecovery,
   CfnManagedLoginBranding,
-  CfnUserPoolDomain,
   FeaturePlan,
+  ManagedLoginVersion,
   Mfa,
   OAuthScope,
   StandardThreatProtectionMode,
   UserPool,
   UserPoolClient,
+  UserPoolDomain,
 } from 'aws-cdk-lib/aws-cognito';
 import {
   CfnLoggingConfiguration,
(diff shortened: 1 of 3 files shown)


cogwirrel (approved):
Looks great! Thank you! :)

PR #886, opened 2026-07-04
feat(connection): more complete, correct OpenAPI client generation with lighter installs

nx-plugin-for-aws:
### Reason for this change

The OpenAPI code generators (`open-api#ts-client`, `open-api#ts-hooks`, `open-api#ts-metadata`) built their code generation data structure from `@hey-api/openapi-ts`'s legacy parser. That parser was removed upstream, so the dependency was pinned and could not be updated, and the data structure it produced only ever "nearly" fit our needs — requiring extensive `as any` workarounds to patch and reshape it downstream. Several parts of the OpenAPI 3.0/3.1 spec were also handled approximately or incorrectly as a result.

`@hey-api/openapi-ts` was also a **runtime dependency** of the published `@aws/nx-plugin` package, so its transitive closure (a config loader, a Handlebars templating engine, file watchers, a tar extractor, etc.) was installed by every consumer of the plugin, despite us only using its parser.

### What this means for customers

Generated API clients, hooks and metadata now handle more of the OpenAPI 3.0/3.1 spec correctly, and the plugin installs lighter.

**More correct & complete generated clients**

- **Polymorphism** — discriminated `oneOf`/`anyOf` unions and inheritance-style base schemas (subtypes composing a base via `allOf`) now marshal to the matching subtype. Subtype fields round-trip correctly instead of being dropped, and fields no longer leak between variants (previously a value could come back with spurious/`Invalid Date` fields from a non-matching branch).
- **`deepObject` query parameters** serialise as `filter[prop]=value` pairs — including nested objects/arrays — instead of `filter=[object Object]`.
- **`spaceDelimited` / `pipeDelimited` query arrays** serialise with the correct delimiter (and compile — they previously emitted collection formats the runtime neither typed nor honoured).
- **Object query/path parameters** are now fully typed and marshalled, so nested values such as dates convert correctly rather than being passed through untouched.
- **Broader spec coverage** — `$ref` path items, vendored `+json` media types, empty request bodies, and OpenAPI 3.1 `const` / multi-type (`type: [...]`) schemas are all handled.
- **Clear errors on ambiguous specs** — colliding schema, property or parameter names (e.g. two that normalise to the same identifier, or a query and header sharing a name) now fail with an actionable message instead of emitting code that silently misbehaves or fails to compile.

**Lighter install footprint**

- **28 fewer packages** in the plugin's install closure (the transitive closure of `@hey-api/openapi-ts`), and **~8.9 MiB less unpacked** on disk — fetched and extracted on every clean install of the plugin.
- Retires `tar@6.2.1`, which npm now flags as deprecated for known vulnerabilities.

### Description of changes

Replaces the `@hey-api/openapi-ts` parser with an in-house OpenAPI parser (`open-api/utils/parser.ts`) that produces the code generation data structure directly, then extends it to close the spec-coverage gaps above.

- **In-house parser** — `buildClientData` walks a normalised spec (schemas, operations, parameters, responses) into the models/services structure the generators consume.
- **Fully-typed data structure** — `Model`, `Operation`, `Service` and `CodeGenData` are proper interfaces, with no `as any` casts in the parser or codegen-data pipeline (the hey-api structure previously required extensive ones).
- **Correctness fixes** — discriminator dispatch (composite + inheritance base), `deepObject` serialisation, collection-format delimiters, object-parameter typing/marshalling, `$ref` path items, vendored `+json` hoisting, nested `const`/multi-type rewrites, and name-clash validation.
- **Dependency removed** — `@hey-api/openapi-ts` is dropped from `package.json`; `LICENSE-THIRD-PARTY` and the lockfile are regenerated accordingly.

The generator test suites render clients/hooks/metadata from a wide range of specs (enums, arrays, dictionaries, additional/pattern properties, `allOf`/`oneOf`/`anyOf` composites, discriminated unions and inheritance bases, recursive refs, streaming, form data, duplicate/colliding names, void-vs-any responses, OpenAPI 3.1 null/const/multi-type, deepObject and collection-format serialisation) and snapshot the output, with runtime round-trip assertions against a mock `fetch`.

- All `@aws/nx-plugin` tests pass, including new coverage for every behaviour above.
- Snapshots were updated where output intentionally changed (discriminators, deepObject, collection-format delimiters, typed object parameters); output is unchanged for specs that don't exercise those features.
- `tsc` typechecks cleanly, with no `as any` in the parser or codegen-data pipeline.

### Follow-ups (not in this PR)

- Emit discriminated unions as true TypeScript *tagged* unions (literal discriminator tags for `switch` narrowing) — currently a usable but untagged union.
- Auto-disambiguate colliding property/parameter names rather than erroring (the wire and TypeScript names are already tracked separately, so this is safe).

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "@commitlint/cli": "20.5.3",
     "@commitlint/config-conventional": "20.5.3",
     "@getgrit/gritql": "0.0.3",
-    "@hey-api/openapi-ts": "0.64.13",
     "@iarna/toml": "2.2.5",
     "@middy/core": "7.7.0",
     "@modelcontextprotocol/inspector": "0.19.0",
--- a/packages/nx-plugin/LICENSE-THIRD-PARTY
+++ b/packages/nx-plugin/LICENSE-THIRD-PARTY
@@ -3659,60 +3659,6 @@ USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 ---
 
-The following software may be included in this product: @hey-api/json-schema-ref-parser (1.0.3)
-This software contains the following license and notice below:
-
-MIT License
-
-Copyright (c) Hey API
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
-
----
-
-The following software may be included in this product: @hey-api/openapi-ts (0.64.13)
-This software contains the following license and notice below:
-
-MIT License
-
-Copyright (c) Hey API
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
-
----
-
 The following software may be included in this product: @hono/node-server (1.19.9)
 This software contains the following license and notice below:
 
@@ -3912,35 +3858,6 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
-
----
-
-The following software may be included in this product: @jsdevtools/ono (7.1.3)
-This software contains the following license and notice below:
-
-The MIT License (MIT)
-
-Copyright (c) 2015 James Messinger
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in
-all copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
-
-.
 
 ---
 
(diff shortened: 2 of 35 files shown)


nx-plugin-for-aws:
Pushed follow-up fixes for edge cases surfaced by differentially testing the generators against a large corpus (real FastAPI/Smithy specs, ~60 real-world APIs from APIs.guru, and the openapi-generator / oapi-codegen / OpenAPI-Initiative test-fixture suites), diffing baseline `main` vs this branch and compiling every generated client under `strict`/`noImplicitAny`.

**Correctness fixes:**
- **boolean & numeric enums** now render as their primitive (`boolean`/`number`) instead of `string` (affected e.g. Slack's `ok`, Stripe's `deleted`).
- **multi-type schemas** (`type: ['integer','string']`) render a proper union instead of collapsing to the first type; `[T, 'null']` becomes nullable `T`.
- **path-level (shared) parameters** are inherited by every operation on the path (operation-level overrides by name+location) — previously dropped entirely (big impact on e.g. Google APIs' `alt`/`fields`/`key`).
- **`requestBody.description`** renders as the body parameter's doc comment.

**Robustness fixes (previously crashed or emitted non-compiling code):**
- guard operations whose only responses are errors (no 2XX/default) — was a hard crash.
- sanitize identifiers (deburr + strip non-identifier chars) so operationIds/names like `getFoo≠Bar` / `café` yield valid TS; fall back to a safe identifier when a name (e.g. `_`) reduces to empty.
- nested-array marshalling annotates the inner `.map()` param (`Date[][]`, `Model[][]` failed `noImplicitAny`).
- `$IO.$mapValues` is now `public` so dictionary responses compile.
- hoisted inline-schema names are normalised so snake_case property names resolve.

All existing tests pass with snapshots updated only for the `$mapValues` visibility change; added `generator.regressions.spec.ts` (21 cases) covering every fix. A known remaining limitation (unchanged from `main`): `allOf`/`oneOf`/`anyOf` composing bare primitives/enums is still rejected as not runtime-distinguishable.

cogwirrel (dismissed):

cogwirrel (dismissed):

nx-plugin-for-aws:
Ran another round of differential testing with adversarial edge-case specs (this branch vs the hey-api-based generator on `main`), which surfaced and fixed four gaps in `7e84ca2`:

1. **`$ref` path items** (`paths: {"/x": {"$ref": "#/components/pathItems/..."}}`) — operations inside referenced path items were never visited by normalisation, so they had no operationId or hoisted schemas and the generated client did not compile (main produced compilable but unmarshalled output). Path item refs are now inlined per-path before normalisation, so paths sharing one path item each get their own operation.
2. **Vendored `+json` media types** (`application/vnd.api+json`, `application/problem+json`) — inline request/response schemas were only hoisted for exactly `application/json`, so vendored JSON payloads fell back to raw `response.text()` handling (broken on both branches; a runtime crash here since responses lack a `text` path for objects). Hoisting now prefers `application/json` and falls back to any `+json` type.
3. **Empty `requestBody.content` map** — generated a phantom body parameter and an empty `Content-Type: ` header (same behaviour on main). An empty content map now means "no body".
4. **Rewrites nested inside rewritten schemas** — `lodash.cloneDeepWith` does not revisit values returned from the customizer, so a `const` or multi-type schema nested inside another rewritten schema (e.g. multi-type items of a `type: ["array", "null"]` array, which collapsed `number | string` items to `number`) was silently skipped. The `rewriteConstToEnum`/`rewriteMultiTypeToAnyOf` rewrites now recurse with themselves (matching the existing pattern in `rewriteCompositeSiblingProperties`).

All fixes have regression tests: generator-level tests with snapshots + compile + runtime invocation in `generator.edge-cases.spec.ts` (sections G–J), and unit tests in `normalise.spec.ts` / `codegen-data.spec.ts`. Full suite: 2472 tests pass.

Also probed (no issues found, behaviour matches main): discriminator mappings with normalised schema names, implicit discriminators, cookie params, boolean/numeric enums, recursive schemas, components-level requestBodies/responses refs, oneOf of dictionaries, readOnly props, 2XX range responses, binary bodies, urlencoded/multipart bodies, deepObject query params, and op-level overrides of path-level `$ref` parameters.

nx-plugin-for-aws:
## Round 2: name-clash audit + serialisation fixes

A second differential-testing pass (ts-client, ts-hooks, ts-metadata vs the old hey-api parser), focused on every site where a name is defined and re-used. **No regressions from the parser swap were found** — the branch actually fixes two cases that were broken on `main` (camelCase-colliding tags, jsonl streaming). The following latent issues (all pre-existing on `main` too) are now fixed with unit + snapshot tests:

**Name clashes**
- **Schema names** — two schemas normalising to the same identifier (`Foo-Bar` + `Foo.Bar` → `FooBar`) silently overwrote each other. Now throws an actionable error.
- **Properties / same-position params** — `foo-bar` + `foo_bar` → `fooBar`, or two query params → one TS name, emitted a duplicate-member type that didn't compile. Now validated and throws.
- **Query + header param sharing a name** — valid OpenAPI, but collided as a duplicate key in one `collectionFormats` object. Now split into `queryCollectionFormats` / `headerCollectionFormats`; still works.

**Serialisation**
- **spaceDelimited / pipeDelimited query arrays** — emitted `ssv`/`tsv`, which the runtime helpers neither typed (`'multi'|'csv'`) nor honoured, so the client didn't compile and arrays were comma-joined. `$url`/`$headers` now join by the correct delimiter (space / pipe); `pipeDelimited` maps to `pipes`.
- **Infinite-query cursor that is a `$ref`** (e.g. an enum) — referenced in the options-proxy cursor type but never imported, so the proxy didn't compile. The cursor model is now imported.

All 2480 `@aws/nx-plugin` tests pass; snapshot changes are limited to the `$url`/`$headers`/collection-format template updates.

### Follow-up (not in this PR)
Discriminated `oneOf` `fromJson` merges **all** branches instead of selecting by `discriminator.propertyName`, so a deserialised value gains spurious fields from the non-matching branch (e.g. a required `date-time` becomes `Invalid Date`). This is pre-existing on `main` and a larger change; worth a dedicated follow-up.

nx-plugin-for-aws:
## Follow-up: discriminated oneOf/anyOf marshalling

Fixes the deferred item from the round-2 comment above.

Previously, a discriminated `oneOf`/`anyOf` merged **every** branch in its generated `toJson`/`fromJson`, so a deserialised value carried fields from the non-matching branches — e.g. a `Dog` gained a spurious `napAt: Invalid Date` from the `Cat` branch. (The parser already knew the discriminator, but only used it to collapse the discriminator property to its primitive type.)

**Fix:**
- The parser now resolves full discriminator metadata onto the composite model: `propertyName` plus the value → composed-model `mapping`, from an explicit `mapping` or implicitly from member schema names.
- The client template emits a `switch` on the discriminator that marshals directly to the matching branch, **falling through to the existing merge for unknown values** — so non-discriminated composites stay byte-identical (only the two discriminated-union snapshots changed, and both diffs are purely additive).
- `allOf` is a conjunction, not a choice, so it is never discriminated.

**Tests:** parser unit tests (explicit / implicit / none) + ts-client runtime round-trip tests asserting the non-matching branch's fields don't leak. All 2485 `@aws/nx-plugin` tests pass.

nx-plugin-for-aws:
## Round 3 + full discriminator support

**Round-3 adversarial testing** (two subagents attacking the newly-added discriminator/clash/collection-format code + realistic spec structures) surfaced one real bug in last round's code, now fixed:
- **Implicit discriminator over inline union members** emitted dispatch cases on synthetic *hoisted* schema names (`case 'ShapeOneOf':`) that can never appear on the wire. `buildDiscriminator` now excludes `x-aws-nx-hoisted` members, so such unions fall back to the merge.
- Also added regression coverage for a **discriminator whose wire name differs from its TS name** (`pet-type` vs `petType`) — verified correct in both marshalling directions.
- Investigated **inheritance-style discriminators** (base carries the discriminator, subtypes `allOf`-compose it, response typed as the base) — confirmed byte-identical to `main` (a pre-existing limitation, not a PR regression). This motivated the feature below.

**Full discriminator support** — extends dispatch to the inheritance pattern (common in Java/Smithy/Swagger output):
- A base object schema carrying a discriminator whose subtypes `allOf`-compose it now marshals via runtime dispatch to the matching subtype, so a value typed as the base **round-trips subtype-only fields** instead of silently dropping them (previously a `Cat` through a `Pet`-typed response lost `lastNapAt`).
- The base **type** stays its declared interface — I verified with `tsc` that making it a union (`Pet = Cat | Dog` while `Cat = Pet & CatAllOf`) is a circular type alias (`TS2456`) that won't compile. So "full support" is sound runtime marshalling, not an unsound type change.
- **Cycle-safe:** the base emits its own-fields body as non-dispatching `$toJsonBase`/`$fromJsonBase`, and subtypes compose *that* (not the dispatching `toJson`/`fromJson`) — so `base → subtype → base` can't recurse infinitely.

Non-discriminated composites and every existing fixture are **byte-identical** (the new paths only activate when a discriminator base exists). Covered by parser unit tests + ts-client runtime round-trip tests. All **2491** `@aws/nx-plugin` tests pass.

nx-plugin-for-aws:
## deepObject query parameter serialisation

Closes the one serialisation gap identified in the spec-coverage review: an object query parameter with `style: deepObject` previously fell through to array handling and serialised as `filter=[object Object]`.

It now serialises as `filter[prop]=value` pairs per the OpenAPI spec (brackets unencoded, matching the convention servers expect), with `undefined`/`null` members omitted. `getCollectionFormat` maps the `deepObject` style to a new collection format, and the shared `$url` helper expands such values into bracketed pairs. Covered by ts-client runtime tests. Only the shared `$url` helper + `collectionFormats` union type change in generated output (client snapshots updated). All 2499 `@aws/nx-plugin` tests pass.

nx-plugin-for-aws:
## deepObject done properly (follow-up)

Two gaps remained after the initial deepObject commit; both fixed:

1. **Object params were typed `unknown` and never marshalled.** Inline object query/path parameter schemas were not hoisted (only composite ones were), so their members were left `unknown` — a nested `Date` in a `deepObject` param went out via `.toString()`. Inline object param schemas are now hoisted to named models exactly like request bodies, so the parameter is fully typed and its members are deep-marshalled (nested dates → ISO). This also improves nullable object params (`unknown | null` → a proper named model).

2. **Nested deepObject emitted `[object Object]`.** The `$url` deepObject branch only handled one level. It now recurses via a `$deepObject` helper (`filter[page][size]=10`, arrays included), matching common client behaviour. The OpenAPI spec leaves nested deepObject undefined, so recursing is a strict improvement over the broken output.

Note: this repo only ever used `@hey-api/openapi-ts` as a *parser* — its client runtime (`@hey-api/client-fetch`) was never a dependency, so the `$url` serialisation has always been this repo's own template. hey-api's client (as a reference) serialises deepObject the same recursive way.

Covered by ts-client runtime tests (nested objects, nested dates, null/undefined omission) + normalise unit tests. All 2502 `@aws/nx-plugin` tests pass; generated-output changes are limited to the shared `$url`/`$deepObject` helper and newly-typed object params.

cogwirrel (approved):

PR #888, opened 2026-07-04
test(dungeon-adventure): defer per-generator installs and enable the Nx daemon

nx-plugin-for-aws:
### Reason for this change

The **Windows dungeon-adventure smoke test** is the single longest job in the PR workflow and therefore gates overall PR wall-clock. Across the four most recent PR runs it took **37–40 min**, while the *same* test on Linux finishes in **~8.5 min**. Since all jobs run in parallel, this one job determines how long a PR takes to go green.

We can't move it to the faster CodeBuild Windows runners (no Docker there, and it needs to stay a faithful end-to-end tutorial test), and sharding would fracture the tutorial's integrity. This PR attempts to remove wasted work *inside* the job instead.

### Description of changes

Two changes to `e2e/src/smoke-tests/dungeon-adventure.spec.ts`:

- **Defer per-generator installs.** Every module-1 generator runs with `--prefer-install-dependencies=false`, and the full accumulated dependency set is installed once via `runInstall` before the first sync/build — mirroring the generic smoke test (`smoke-test.ts`).
- **Enable the Nx daemon** (`NX_DAEMON=true`) for the generate/build phase, to keep the project graph and TypeScript state warm across the tutorial's four repeated sync/lint/build waves.

Pushed and measured the real Windows CI timing against the baseline (run [`28704975842`](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/28704975842)).

> [!WARNING]
> **Result: no measurable speedup.** The changes are green and correct, but the `Windows Smoke Tests - dungeon-adventure` job did not get faster.

**Before → After (Windows dungeon-adventure):**

| Phase | Baseline (#887) | This PR | Δ |
|-------|----------------|---------|---|
| Generation (workspace + generators + install) | 10.5 min | 10.2 min | ~0 |
| Build / validate (4 waves incl. Docker) | 25.6 min | 27.9 min | +2.3 (within run-to-run noise) |
| **Test step total** | **36.2 min** | **38.1 min** | **no improvement** |
| Full job (queue + init + test) | ~38.8 min (4-run avg) | 41.7 min | no improvement |

**Why it didn't help (from the completed job log):**

- **`preferInstallDependencies=false` was largely ineffective here.** The generation phase still shows installs firing (`40s / 35s / 9s / 47s / 1m9s`). `installDependencies` (`utils/install.ts`) runs anyway whenever a graph-critical dependency isn't yet linked, and this test's generator mix trips that condition on nearly every step — so the batch-once deferral mostly doesn't happen. Generation stayed ~10.5 min either way.
- **The daemon engaged (`NX_DAEMON: true` confirmed) but didn't move the needle.** The build/validate phase is dominated by actual `tsc` / `vite` / **Docker image** builds (e.g. `inventory:mcp-server-docker` = `added 278 packages in 2m`), none of which the daemon accelerates. The graph computation it keeps warm is a small fraction of the phase.

### Recommendation

Since neither change delivers a speedup, and the daemon flip reverses a deliberate prior decision (`f2b5939d "test: disable daemon for e2e tests"`), the leading option is to **close/revert this PR** and pursue the real cost centres instead:

- The 4 sequential full-workspace **build waves** (~26 min) are the dominant cost — reducing intermediate rebuilds is where the time is.
- The **Docker image builds** inside the test are a large fixed cost on the slow GitHub-hosted `windows-latest` runner.

Keeping the PR is low-risk (green, tutorial-faithful) but adds complexity for no benefit.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/e2e/src/smoke-tests/dungeon-adventure.spec.ts
+++ b/e2e/src/smoke-tests/dungeon-adventure.spec.ts
@@ -14,6 +14,7 @@ import {
   createTestWorkspace,
   getDungeonAdventureElectroDbDependencies,
   runCLI,
+  runInstall,
   tmpProjPath,
 } from '../utils';
 import { startLlmMock } from '../utils/llm-mock';
@@ -186,7 +187,11 @@ describe('smoke test - dungeon-adventure', () => {
   const pkgMgr = 'pnpm';
   const targetDir = `${tmpProjPath()}/dungeon-adventure-${pkgMgr}`;
   const projectRoot = `${targetDir}/dungeon-adventure`;
-  const opts = { cwd: projectRoot, env: { NX_DAEMON: 'false' } };
+  // Keep the Nx daemon on so the project graph and TypeScript state stay warm
+  // across the tutorial's repeated sync/lint/build invocations. This matters
+  // most on Windows, where the dungeon-adventure smoke test is the CI critical
+  // path and a cold graph per command dominates its runtime.
+  const opts = { cwd: projectRoot, env: { NX_DAEMON: 'true' } };
   const runningProcesses: ChildProcess[] = [];
   let llmMock: MockServer | undefined;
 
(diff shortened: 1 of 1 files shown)


nx-plugin-for-aws:
Closing: measured on CI (run 28704975842) and the changes produced no speedup — the daemon doesn't accelerate the tsc/vite/Docker-dominated build phase, and the per-generator install deferral is a no-op here because graph-critical deps force installs anyway. Pursuing build-wave reduction instead.

PR #892, opened 2026-07-05
fix(ts#rdb): connect postgres migrations to the cluster directly and stop container on log-stream exit

AlexTo:
### Reason for this change

- Postgres migrations connected through the RDS Proxy using the app user's IAM credentials, while MySQL migrations already connected directly to the cluster with the admin secret. This inconsistency had no real justification.
- `dbUser` should only ever be used by the API, not by migrations. Because it also had to run Postgres migrations, it was granted far more privileges than the API itself needs (`ALL PRIVILEGES` on database/tables/sequences/functions, plus schema `CREATE`).
- Migrations now always run as admin via `DATABASE_SECRET_ARN`, so `dbUser` no longer needs DDL-level access and can be scoped down to basic CRUD.
- The `logs.on("exit")` handler in `start-container.ts.template` was introduced before `cleanup()` existed, just to exit the script if the container died or crashed. It was never updated to actually clean up the container. Since the `logs -f` child shares the terminal's signal group, a Ctrl+C can also reach it directly, racing with the parent's `SIGINT` handler. If `logs.on("exit")` won that race it exited the script directly, leaving the container running.

### Description of changes

- Unified migration handler (CDK + Terraform) to always connect to the cluster directly via the admin secret, regardless of engine.
- Trimmed the API `dbUser`'s database grants down to basic CRUD.
- `logs.on("exit")` now calls `cleanup()` instead of `process.exit(code)`.

Unit tests.

### Issue # (if applicable)

N/A

### Checklist

----


cogwirrel (approved):
Thanks Alex! Nice to see the privilege reduction and great simplification too :)

AlexTo:
you're welcome !

PR #908, opened 2026-07-08
fix(utils#api-constructs): add VPC vars to Terraform API Lambda module

AlexTo:
### Reason for this change

The Terraform module generated for REST/HTTP APIs (used by `py#fast-api`, `ts#trpc-api` and `ts#smithy-api`) had no way to deploy the API's Lambda function into a VPC. This is needed, for example, to reach an RDS database from a `fast-api`/`trpc`/`smithy` backend. The CDK construct already supports this generically via `FunctionProps` (`vpc`/`vpcSubnets`/`securityGroups`), so this was a Terraform-only gap.

### Description of changes

- Added optional `vpc_id` and `subnet_ids` variables to the REST and HTTP API Terraform templates (`utils/api-constructs/files/terraform/app/apis/{rest,http}/__apiNameKebabCase__/__apiNameKebabCase__.tf.template`).
- When `vpc_id` is set, the module creates and manages its own security group for the API Lambda (mirroring the pattern already used by the `rdb-constructs` Terraform module), attaches a `vpc_config` block, and conditionally attaches `AWSLambdaVPCAccessExecutionRole` to the Lambda's execution role.
- Added a `security_group_id` output so other modules (e.g. a database module) can add ingress rules for the API Lambda.
- Left VPC config fully optional/backward compatible — Lambdas remain outside a VPC by default.

- Updated and reviewed the affected generator snapshot tests (`py/fast-api`, `trpc/backend`, `smithy/ts/api`) which cover the rendered Terraform output for both REST and HTTP APIs.
- Ran the full `@aws/nx-plugin:test` suite (triggered via the pre-commit hook) — all green.

### Checklist
- [x] My code adheres to the CONTRIBUTING GUIDE and DESIGN GUIDELINES

----

### DIFF
--- a/docs/src/content/docs/en/snippets/connection/rdb-api-infrastructure.mdx
+++ b/docs/src/content/docs/en/snippets/connection/rdb-api-infrastructure.mdx
@@ -51,6 +51,7 @@ module "my_database" {
 
 module "api" {
   source             = "..."
+  enable_vpc         = true
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnet_ids
 
--- a/docs/src/content/docs/es/snippets/connection/rdb-api-infrastructure.mdx
+++ b/docs/src/content/docs/es/snippets/connection/rdb-api-infrastructure.mdx
@@ -51,6 +51,7 @@ module "my_database" {
 
 module "api" {
   source             = "..."
+  enable_vpc         = true
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnet_ids
 
--- a/docs/src/content/docs/fr/snippets/connection/rdb-api-infrastructure.mdx
+++ b/docs/src/content/docs/fr/snippets/connection/rdb-api-infrastructure.mdx
@@ -51,6 +51,7 @@ module "my_database" {
 
 module "api" {
   source             = "..."
+  enable_vpc         = true
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnet_ids
 
--- a/docs/src/content/docs/it/snippets/connection/rdb-api-infrastructure.mdx
+++ b/docs/src/content/docs/it/snippets/connection/rdb-api-infrastructure.mdx
@@ -51,6 +51,7 @@ module "my_database" {
 
 module "api" {
   source             = "..."
+  enable_vpc         = true
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnet_ids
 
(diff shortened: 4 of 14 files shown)


nx-plugin-for-aws:
### Plan-time `count` failure when the VPC is created in the same apply

Deployed the branch against a Terraform workspace with a REST + HTTP tRPC API (both with and without a VPC). Without a VPC everything works; passing an existing `vpc_id` works. But passing a `vpc_id` from a VPC created **in the same apply** — the PR's motivating case (an API Lambda reaching an RDS DB where the `rdb` module stands up the VPC) — fails at plan:

```
Error: Invalid count argument

  on ../../common/terraform/src/app/apis/my-api/my-api.tf line 145, in resource "aws_security_group" "api_lambda":
 145:   count = var.vpc_id != null ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so OpenTofu cannot predict how many instances will be created.
```

This is a hard Terraform/OpenTofu rule: `count` and `for_each` must be known at plan time. `var.vpc_id != null` is only plan-known if `vpc_id` is — and it isn't when it's a computed attribute of a not-yet-created VPC, so the comparison evaluates to unknown and planning fails. No expression rewrite on `vpc_id != null` fixes this; anything derived from a computed value stays unknown.

#### Fix: gate `count`/`for_each` on a plan-known boolean

Add an `enable_vpc` bool and gate on that instead of on the computed id. `vpc_id`/`subnet_ids` stay as-is — they're only used as resource *arguments*, which are allowed to be unknown at plan time. This mirrors how `rdb-constructs` already gates its optional resources (`enable_rds_proxy`, `enable_credential_rotation`) on booleans rather than on ids.

```hcl
variable "enable_vpc" {
  description = "Deploy the Lambda function into a VPC. Set alongside vpc_id/subnet_ids."
  type        = bool
  default     = false
}

variable "vpc_id" {
  description = "VPC ID to deploy the Lambda into. Required when enable_vpc is true."
  type        = string
  default     = null

  validation {
    condition     = !var.enable_vpc || (var.vpc_id != null && var.subnet_ids != null)
    error_message = "vpc_id and subnet_ids must be set when enable_vpc is true."
  }
}
```

Then swap every gate from `var.vpc_id != null` to `var.enable_vpc`:

```hcl
resource "aws_security_group" "api_lambda" {
  count = var.enable_vpc ? 1 : 0
  # ...
}

resource "aws_vpc_security_group_egress_rule" "api_lambda_https" {
  count = var.enable_vpc ? 1 : 0
  # ...
}

dynamic "vpc_config" {
  for_each = var.enable_vpc ? [1] : []
  # ...
}

resource "aws_iam_role_policy_attachment" "lambda_vpc_access" {
  count = var.enable_vpc ? 1 : 0
  # ...
}

output "security_group_id" {
  value = var.enable_vpc ? aws_security_group.api_lambda[0].id : null
}
```

`enable_vpc` is a literal the caller passes, so `count` is always resolvable even while `vpc_id` is still being computed in the same apply, and the `aws_security_group.api_lambda[0]` references stay valid. Keeps the feature optional/backward-compatible (no VPC by default) while making single-apply VPC creation work.

Applies to both the `rest` and `http` templates.

cogwirrel (approved):
Thanks @AlexTo ! 🙌

PR #922, opened 2026-07-10
fix(open-api#ts-client): urlencoded bodies, tuples, JSON primitive responses, union marshalling

nx-plugin-for-aws:
### Reason for this change

Exhaustively testing FastAPI's feature surface against the generated TypeScript client (in an example workspace, against a live backend) surfaced four wire-level gaps where valid FastAPI endpoints produced broken clients:

1. **`Form(...)` endpoints were uncallable.** An `application/x-www-form-urlencoded` request body was `JSON.stringify`-ed, so FastAPI rejected every call with a 422 (`Field required`).
2. **Tuples lost their types and their dates.** Pydantic emits OpenAPI 3.1 `prefixItems` for `tuple[A, B]` fields; the generator collapsed these to `Array<unknown>` with no marshalling, so e.g. a `tuple[str, datetime]` element stayed a raw string.
3. **Primitive JSON responses were corrupted.** FastAPI returns primitives JSON-encoded on the `application/json` wire, but the client read them as raw text — a `str` return kept its quotes (`"hello"`), and a `datetime` return became `Invalid Date`.
4. **Non-discriminated unions leaked fields across branches.** FastAPI's plain `Union[A, B]` return emits `anyOf` with no discriminator; `fromJson` merged every branch and converted required properties unconditionally, so a non-matching branch fabricated values (e.g. a spurious `sentAt: Invalid Date`). This replaces #917 with a deterministic approach — no runtime branch inference.

### Description of changes

- **urlencoded bodies**: a `$urlEncodedForm` helper (built on `URLSearchParams`; arrays as repeated keys, undefined fields omitted) is emitted when an operation declares a urlencoded body, mirroring the multipart `$formData` pattern.
- **Tuples**: `prefixItems` is now a first-class `tuple` model kind — typed positionally (`[number, string]`), marshalled element-wise (dates inside tuples round-trip), inline object members hoisted to named models. A tuple with additional `items` beyond the prefix degrades to a plain array.
- **JSON primitive responses**: response deserialisation is keyed on the declared media type — primitives on a JSON wire use `response.json()`; `text/plain` and streaming chunk handling keep raw-text semantics.
- **Union marshalling (deterministic)**: `fromJson` now spreads *every* property presence-guarded (mirroring `toJson`), so an absent field is never converted and composing non-discriminated union members leaks nothing — same output for the same payload, no shape-guessing. When two union members claim the same wire property with different marshalling semantics (e.g. `date-time` vs plain string), no deterministic conversion exists, so the generator fails fast at generation time with an error asking for a discriminator (e.g. Pydantic `Field(discriminator=...)`).
- **Binary detection scoped to form bodies**: FastAPI emits the identical `{type: string, contentMediaType: ...}` schema for multipart file fields (binary) and `bytes` in a JSON body (base64 string). Normalisation now rewrites only form-body binary string fields to `format: binary` (→ `Blob`); JSON `bytes` fields stay strings, and the parser trusts only `format: binary`.

Notably *not* included per prior feedback: no runtime branch inference for unions (non-deterministic), and no default `python-multipart` dependency (users add it when they use `Form`/`UploadFile`).

- New `generator.fast-api-gaps.spec.ts` covering urlencoded runtime behaviour (encoding, array fields, undefined omission), tuple typing + element-wise date marshalling both directions, open-tuple degradation, and JSON-vs-text primitive response handling; a reworked composite test asserts non-discriminated unions no longer leak fields, plus a test asserting the generation-time error for genuinely ambiguous unions. Full suite (2537 tests) and build green.
- End-to-end in a fresh workspace with locally built packages against a live FastAPI backend: urlencoded login, tuples with revived dates, JSON string/datetime/untyped responses, base64 `bytes` round-trip, `Union[EmailEvent, SmsEvent]` marshalling each payload with only its own keys and valid Dates in both directions, and an ambiguous `Scheduled | Named` union failing generation with the actionable error.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/open-api/ts-client/__snapshots__/generator.additional-properties.spec.ts.snap
+++ b/packages/nx-plugin/src/open-api/ts-client/__snapshots__/generator.additional-properties.spec.ts.snap
@@ -108,17 +108,13 @@ export class $IO {
           ),
           (additionalProperty) => additionalProperty,
         ),
-        ...(json['id'] === undefined
-          ? {}
-          : {
-              id: json['id'],
-            }),
+        id: json['id'],
         ...(json['timestamp'] === undefined
           ? {}
           : {
               timestamp: new Date(json['timestamp']),
             }),
-      } as GetDynamicPropertiesAny200Response;
+      };
     },
   };
 
@@ -167,17 +163,13 @@ export class $IO {
               additionalProperty,
             ),
         ),
-        ...(json['id'] === undefined
-          ? {}
-          : {
-              id: json['id'],
-            }),
+        id: json['id'],
         ...(json['version'] === undefined
           ? {}
           : {
               version: json['version'],
             }),
-      } as GetDynamicPropertiesNested200Response;
+      };
     },
   };
 
@@ -204,17 +196,13 @@ export class $IO {
         return json;
       }
       return {
-        ...(json['name'] === undefined
-          ? {}
-          : {
-              name: json['name'],
-            }),
+        name: json['name'],
         ...(json['config'] === undefined
           ? {}
           : {
               config: json['config'],
             }),
-      } as GetDynamicPropertiesNested200ResponseValue;
+      };
     },
   };
 
@@ -257,17 +245,9 @@ export class $IO {
           ),
           (additionalProperty) => additionalProperty,
         ),
-        ...(json['id'] === undefined
-          ? {}
-          : {
-              id: json['id'],
-            }),
-        ...(json['name'] === undefined
-          ? {}
-          : {
-              name: json['name'],
-            }),
-      } as GetDynamicPropertiesNumbers200Response;
+        id: json['id'],
+        name: json['name'],
+      };
     },
   };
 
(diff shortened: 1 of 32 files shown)


nx-plugin-for-aws:
Pushed a follow-up per feedback: the presence-guard-everything + `as ModelName` cast in `fromJson` weakened type safety for **every** generated model, when it was only needed at the non-discriminated-union merge boundary. Now:

- `fromJson` is back to fully type-checked bodies — required properties convert unconditionally, no casts, so the compiler verifies structural completeness for every model.
- Models that are members of a non-discriminated `oneOf`/`anyOf` additionally emit an internal `$fromJsonPartial` (same precedent as `$fromJsonBase` for discriminator bases) that converts only the properties present on the wire. The union merge composes those partials, so a non-matching member contributes nothing and never fabricates values — while its public `fromJson` stays fully typed.

Still fully deterministic (no branch inference), and the generation-time error for genuinely ambiguous unions is unchanged. Full suite + e2e against a live FastAPI backend re-verified.

cogwirrel (approved):

PR #928, opened 2026-07-11
feat: enable nx migrate for @aws/nx-plugin upgrades

### LINKED ISSUE #703: feat: Define Nx migrations strategy and contributing guidelines
### Description

Define and document a clear strategy for Nx migrations in the `@aws/nx-plugin`. This strategy should guide contributors on what should and shouldn't be handled via automated migrations when breaking changes are introduced between versions.

### Motivation

As we approach v1.0, the plugin will need a reliable migrations story. Nx supports `migrations.json` in plugins which allows `nx migrate` to automatically apply code transformations when users upgrade. Without clear guidelines, contributors will either:
- Skip migrations entirely (leaving users with broken upgrades)
- Over-reach with migrations (overwriting user-customized files like `agent.py` or custom CDK constructs)

### Scope Definition

Migrations **SHOULD** cover:
- Changes to generator metadata in `project.json` (e.g., renamed targets, updated options)
- Updates to shared constructs in `packages/common/` that follow known patterns
- AST-level patches to files that follow a predictable structure (e.g., updating import paths, renaming exported symbols)
- Configuration file updates (`aws-nx-plugin.config.mts`, `nx.json` sync generator entries)
- Dependency version bumps in `package.json` / `pyproject.toml`
- Renaming of generator IDs (updating `project.json` metadata references)

Migrations **SHOULD NOT** cover:
- User-authored business logic (e.g., custom `agent.py`, custom CDK stacks beyond the generated scaffold)
- Files where the generator only provides initial scaffolding and the user is expected to modify heavily
- Stylistic or formatting changes
- Optional enhancements that didn't exist in the prior version

### Dependency Update Strategy

The plugin vends dependencies to users' projects (e.g., TypeScript, Vite, tRPC, Zod, CDK, FastAPI, Strands SDK, etc.). The strategy must cover how these are managed:

**Automated minor/patch updates:**
- The plugin already has an `update-versions` workflow that bumps vended dependency versions
- When a dependency is bumped and no code changes are required in user projects, the version bump itself is sufficient — users get the new version next time they install
- When a dependency bump requires code changes (e.g., new API, deprecated method removed), a migration must accompany the version bump
- The strategy should define how to detect whether a dependency bump needs a migration (e.g., automated CI check that generates a project at the old version, bumps deps, and runs build/lint/tests)

**Manual major version updates (e.g., TypeScript 7, Vite 9, React 20):**
- Major version upgrades often require code changes in generated files
- The strategy should define the process: upgrade locally, identify required changes via `git diff`, write a migration that applies those changes
- Some major upgrades may require changes to user-owned files — these should log warnings rather than auto-patching

**Automating migration generation for dependency bumps:**
- If the `update-versions` workflow bumps a dependency and no code changes are needed, no migration is required (just a version bump in the vended `package.json` / `pyproject.toml`)
- If code changes ARE needed, the workflow should either:
  - Fail CI, signaling that a migration is required before the bump can land
  - Or auto-generate a migration that bumps the version in user projects' dependency files
- The simplest case (most common): a migration that just updates the version string in `package.json` / `pyproject.toml` — this could potentially be automated entirely

### Key Design Problem: Version Targeting

The project uses conventional commits with automatic semver — the next version number isn't known until the release pipeline runs. Nx requires a concrete `"version"` in `migrations.json` to know when to run a migration (it runs migrations where `installed_version < migration_version`).

**Current release model** (important context):
- `package.json` stays at `"version": "0.0.0"` in source — version is never committed
- `nx.json` release config uses `currentVersionResolver: "git-tag"` and `manifestRootsToUpdate: ["dist/{projectRoot}"]`
- The release workflow (`pnpm nx release --yes`) calculates the version, writes it to `dist/`, creates a git tag, and publishes — **no source commit is made**
- Tags point directly at the last merged commit

**This means contributors cannot manually set migration versions.** The solution must automatically determine the correct version at build/release time without requiring CI workflow changes or automated PRs back to source.

**Possible approach: derive versions from git history at compile time**

Source `migrations.json` entries have no version fields. The `compile` step stamps them:
- For migrations that already shipped (file existed in a previous tag): use that tag's version
- For migrations that haven't shipped yet: use `latest_tag + 0.0.1` (or similar bump)

This works because Nx runs migrations where `version > installed_version`, so any version between the user's current and the new version qualifies. No workflow changes, no automated PRs — git history is the source of truth.

This is a suggestion — the actual implementation should be determined when this work is picked up.

### Suggested Implementation

1. **Wire `migrations.json`** into `package.json` under the `nx-migrations` field
2. **Solve version stamping** — contributors should not need to manually assign versions; the build/release process handles it automatically
3. **Document the strategy** in the contributing guide on the docs site
4. **Define a decision tree** for contributors:
   - "Is this a breaking change?" → If yes, a migration is required
   - "Does the file follow a predictable pattern?" → If yes, use AST patching
   - "Is the file user-owned (heavily customized)?" → If yes, emit a warning/log instead of modifying
5. **Document escape hatches**: how users can skip specific migrations, how to handle conflicts
6. **Create a section in CONTRIBUTING.md** pointing to the full guide

### Acceptance Criteria

- [ ] `migrations.json` wired into `package.json` `nx-migrations` field
- [ ] Version stamping solved — contributors don't manually assign versions
- [ ] Contributing guide documents what is/isn't in scope for migrations
- [ ] Decision tree or flowchart for "should this be a migration?"
- [ ] Examples of good migrations vs. things that should NOT be migrations
- [ ] Dependency update strategy documented (minor/patch automation, major version process)
- [ ] Clear guidance on when a dependency bump needs a migration vs. when it doesn't
- [ ] Documentation published on the docs site (English only)

### LINKED ISSUE #704: feat: Scaffold generator for creating migration generators
### Description

Create a standard pattern (and optionally a meta-generator similar to `ts#nx-generator`) for scaffolding new migrations within the plugin. This ensures migrations follow a consistent structure and reduces friction for contributors.

### Motivation

Without a standard pattern, each contributor will implement migrations differently — inconsistent file naming, no tests, no version targeting. A scaffold ensures every migration:
- Has a consistent directory structure
- Includes a test file
- Is automatically registered in `migrations.json`

Ideally the scaffold generator itself should have no parameters — it should infer everything it needs (e.g., migration name from the directory, no version since that's determined at build/release time). Zero-config scaffolding reduces cognitive overhead for contributors.

### Suggested Implementation

One possible approach: a new internal generator (e.g., `ts#nx-migration`) that produces the scaffold and registers it automatically. Alternatively, this could be a documented manual pattern with a template directory to copy.

**Example scaffold structure:**
```
packages/nx-plugin/src/migrations/<name>/
├── migration.ts          # The migration implementation
├── migration.spec.ts     # Unit test
└── schema.json           # (optional) If the migration accepts parameters
```

**Migration implementation pattern:**
```typescript
import { Tree } from '@nx/devkit';

export default function migration(tree: Tree): void {
  // Read/modify files in the tree
  // Use AST utilities for code modifications
  // Log warnings for files that cannot be auto-migrated
}
```

### Acceptance Criteria

- [ ] Consistent, documented pattern for creating new migrations
- [ ] New migrations are registered in `migrations.json` (manually or automatically)
- [ ] Test pattern uses `createTreeUsingTsSolutionSetup()` (consistent with generator tests)
- [ ] At least one example migration included as reference
- [ ] Documentation added to contributing guide

### LINKED ISSUE #808: feat: Single-source migration upgrade guides (docs + MCP migration-guide tool + agentic prompts)
### Description

Define **migration upgrade guides as a single source of truth** that is authored once per version step (`x.x.x -> y.y.y`) and consumed by three surfaces:

1. **The docs site** — a rendered upgrade guide per version step.
2. **The MCP server** — a new tool that accepts `from=<version>` and `to=<version>` and returns the concatenated guidance for every step in that range.
3. **The agentic migration prompt** — the `prompt:` content that Nx 23's agentic `nx migrate` flow feeds to the user's local agent (see #703 and the [Nx 23 release](https://nx.dev/blog/nx-23-release)).

This follows the pattern the repo **already** uses for generator guides: a single set of `.mdx` files under `docs/src/content/docs/en/` is rendered by the Astro docs site *and* served by the MCP `generator-guide` tool, both via the shared `postProcessGuide()` pipeline (`packages/nx-plugin/src/mcp-server/generator-info.ts`, imported by `docs/src/components/markdown-content.astro`). We extend that same single-source mechanism to migrations.

### Motivation

Per the Nx 23 migration strategy (#703), our migrations split into two halves:
- **Deterministic** — handled by `nx migrate` generator-only migrations (version bumps, `project.json`/`nx.json` rewrites, AST patches on files we own). These run unattended, including in CI.
- **Agentic / manual** — changes to user-owned files (`agent.py`, custom CDK/constructs, custom routers) that depend on the user's stack. In Nx 23 these ship as `prompt:` migrations applied by the user's local agent, and as a fallback must be documented for users without an agent.

Today there is **no** place a user upgrading from an arbitrary old version can see "here is everything I need to do to get from where I am to the latest." The same guidance is needed verbatim in three places. Without single-sourcing, the docs, the MCP tool, and the agent prompt drift apart.

### Required structure: one guide per version step

Each version step is authored once, e.g. under a new `docs/src/content/docs/en/guides/migrations/` directory, keyed by the **target** version (the version the step migrates *to*). Each guide has a fixed two-section structure:

```mdx
---
title: Upgrading to y.y.y
migrationStep: y.y.y          # target version this step upgrades to
---

## Run the migration                       <-- "deterministic" section

The deterministic changes are applied automatically. Run:

    npx nx migrate @aws/nx-plugin@y.y.y
    npx nx migrate --run-migrations

<!-- describes what the generator-only migrations do for this step -->

## Suggested changes                        <-- "agentic / manual" section (THE REUSABLE UNIT)

<Snippet name="migrations/y.y.y/suggested-changes" />
```

The **"Suggested changes"** section is the reusable unit. It is authored as its own snippet (e.g. `docs/src/content/docs/en/snippets/migrations/<target-version>/suggested-changes.mdx`) so it can be:

- **Inlined into the docs guide** (as above) — rendered by Astro.
- **Returned by the MCP server** — see below.
- **Used as the `prompt:` content** for that step's agentic migration in `migrations.json` (resolved/copied at compile time, alongside the existing version-stamping in #703).

> Defining the "Suggested changes" as the shared atom (rather than the whole guide) keeps the agent prompt focused on the changes it must make, while the docs guide wraps it with the deterministic-call preamble.

### MCP server tool

Add a new tool (e.g. `migration-guide`) to the MCP server (registered in `packages/nx-plugin-mcp/src/index.ts`, implemented under `packages/nx-plugin/src/mcp-server/tools/`):

- **Input:** `from=<version>`, `to=<version>` (both optional; default `from` = none/earliest, `to` = latest).
- **Behavior:** resolve every migration-step guide whose `migrationStep` target version satisfies `from < step <= to`, ordered ascending, and return for each step its **deterministic call** + **Suggested changes** content (run through the same `postProcessGuide()` pipeline, so `<Snippet>`/option-filtering/component rendering all work identically to `generator-guide`).
- **Output:** one section per step, so an agent (or user) gets the full ordered upgrade path in a single call.

The MCP package already bundles `docs/.../guides` and `docs/.../snippets` into its dist (`packages/nx-plugin-mcp/project.json`); extend the guide/snippet probe paths to include the `migrations/` subtree.

### Single-source flow summary

```
docs/.../snippets/migrations/<target>/suggested-changes.mdx   <-- authored ONCE
        |
        +--> docs guide (guides/migrations/<target>.mdx)  -> Astro renders the upgrade page
        +--> MCP `migration-guide` (from=A, to=B)         -> concatenated steps A..B
        +--> migrations.json `prompt:` for <target>       -> Nx 23 agentic nx migrate
```

### Acceptance Criteria

- [ ] Per-version-step upgrade guides authored under `docs/src/content/docs/en/guides/migrations/`, each with a **deterministic "Run the migration"** section and a **"Suggested changes"** section.
- [ ] The "Suggested changes" content is authored once (as a snippet) and reused — no duplication across docs / MCP / agent prompt.
- [ ] New MCP tool accepts `from=<version>` / `to=<version>` and returns the ordered, concatenated guidance for every step in range, via the shared `postProcessGuide()` pipeline.
- [ ] Each agentic (`prompt:` / hybrid) migration in `migrations.json` sources its prompt from the same "Suggested changes" content for its version step (resolved at compile/release time, consistent with #703's version stamping).
- [ ] MCP server bundling/probe paths updated to resolve the `migrations/` subtree.
- [ ] Docs published (English only), and the contributing guide (from #703) documents the author-once workflow.

### Related

- Part of the Nx Migrations workstream — depends on the strategy in #703 and the scaffold generator in #704; consumed by the e2e harness in #705.
- Tracked under #718.

nx-plugin-for-aws:
### Reason for this change

For v1.0 (#718) users need a reliable upgrade story: running `nx migrate @aws/nx-plugin` should upgrade a generated workspace safely. That needs `migrations.json` wired in, a way to assign migration versions that works with this repo's release model (versions are never hand-written into source), documented guidelines so contributors know what belongs in a migration, a generator to scaffold migrations consistently, and a guide (plus MCP tool) telling users how to actually run an upgrade.

Closes #703
Closes #704
Closes #808

### Description of changes

**Wiring & versioning (#703)**

- **Register `migrations.json`** via the `nx-migrations` field in `packages/nx-plugin/package.json`, and copy it into dist as a compile asset so `nx migrate @aws/nx-plugin` discovers it. The collection starts empty — no released change currently warrants a migration.
- **Versions arrive on their own**, in two steps, so contributors never hand-write a `version` field:
  - The weekly `update-versions` PR runs `scripts/backfill-migration-versions.ts`, recording the version of the release that shipped each migration in the source `migrations.json` and moving it out of `latest/` into that release's `v<x.y.z>/` folder. Only already-released migrations are touched, so `main` converges on the versions of everything published while a net-new migration stays in `latest/`. It lands as part of that PR, so branch protection is respected.
  - At package time `scripts/stamp-migrations.ts` (the `stamp-migrations` target in the `package` chain) stamps any entry still missing a version into the compiled `migrations.json`: one that already shipped gets the version of the earliest release tag registering it, and a net-new one gets `semver.inc(latest, 'prerelease')` — strictly above the latest tag and below any possible next release, including this repo's rc-prereleases which a `patch`/`minor` bump would leapfrog.

  A version already present in source always wins, so backfilled values are stable and never recomputed, and each release only has to resolve entries that are still unversioned.

**`ts#nx-migration` generator (#704)**

A **public** generator that scaffolds a migration in any Nx Plugin project and registers it in that plugin's `migrations.json`, creating the manifest and wiring the `nx-migrations` field into the plugin's `package.json` when absent. Modelled on `ts#nx-generator` (takes a target `--project`, exposed via the SDK, has a docs guide), and reusing the same `readNxPluginProject` / `configureNxPluginPackageJson` / `addNxPluginDependencies` helpers as `ts#nx-plugin`.

`--kind` selects the migration form (default `deterministic`):

```bash
# deterministic: migration.ts codemod (guardrails baked in) + spec
pnpm nx g @aws/nx-plugin:ts#nx-migration --project=@aws/nx-plugin --name=rename-foo-target --description="…"
# agentic: a prompt.md applied by the user's agent
pnpm nx g @aws/nx-plugin:ts#nx-migration --project=@aws/nx-plugin --name=migrate-handlers --description="…" --kind=agentic
# hybrid: migration.ts (returning agentContext) + spec + prompt.md
pnpm nx g @aws/nx-plugin:ts#nx-migration --project=@aws/nx-plugin --name=upgrade-framework --description="…" --kind=hybrid
```

Migrations are grouped by the release that ships them: a new one lands in `src/migrations/latest/<name>/` registered as `latest-<name>`, and the weekly `update-versions` PR moves it into `src/migrations/v<x.y.z>/<name>/` and re-keys it to `v<x.y.z>-<name>` when it records the version, rewriting the entry's paths to match. Keying entries by folder as well as name means reusing a name for a later change adds a new entry alongside the one that already shipped instead of silently overwriting it.

A single template set (`migration.ts`, `migration.spec.ts`, `prompt.md`) covers all three kinds — an `isHybrid` flag drives the parts that deviate (the codemod's `agentContext`, the prompt's phrasing), and the generator prunes the files a kind doesn't use. The skeletons link the Nx migration docs (including `MigrationReturnObject`, so `nextSteps`/`agentContext` are discoverable), finish with `formatFilesInSubtree` so a migration's JSON rewrites don't leave the user's `format` target failing, and leave the body empty. The prompt's audience note is a delete-before-shipping comment so it can't leak into shipped instructions.

Nx ships `@nx/plugin:migration`, but it doesn't fit: it mandates an explicit `--packageVersion` stamped into `migrations.json` (opposite of our model), writes `dist/` implementation paths and adds deps/`files` tuned for a standalone published plugin, and has no notion of our three migration kinds or guardrails.

**Upgrade guide & MCP tool (#808)**

- New <kbd>Get Started → Upgrading your workspace</kbd> guide covering the `nx migrate` cycle, with terminal output captured from a real workspace for the deterministic and agentic paths, and an **Upgrading with an AI agent** section.
- New **`upgrade-workspace`** MCP tool (mirroring `add-to-existing-project`) that returns that guide with commands rendered for the caller's package manager, so an agent can drive an upgrade end to end. Single-sourced from the same `.mdx` the docs site renders, via the existing `fetchGuidePages`/`postProcessGuide` pipeline, so the two can't drift. `InstallCommand`'s package is now optional so the guide can render a plain dependency install.

Per-version-step guides (`guides/migrations/<target>.mdx`) and a `from`/`to` ranged tool are deliberately not included: with an empty migration collection there are no version steps to author yet, and the agentic instructions for a step are already single-sourced as that migration's `prompt.md`, which Nx writes to `tools/ai-migrations/<package>/<version>/` and which the guide tells users (and agents) how to apply. The ranged variant is worth revisiting once there are real steps to concatenate.

**Contributing guide**

- New Migrations section in CONTRIBUTING.md covering the three migration kinds, how to choose between them (prefer deterministic; reach for a prompt only when the edit targets user-owned code), what should and shouldn't be a migration, guardrails, versioning, and testing expectations.

- Unit tests for the versioning helpers (stamping and backfill, including that a version already in source is never recomputed, and that same-named migrations from different releases coexist), the `ts#nx-migration` generator (in-repo and external-workspace modes: manifest creation, `nx-migrations` wiring, all three kinds, not overwriting a released migration of the same name, idempotency) and the `upgrade-workspace` MCP tool.
- Ran the generator via the real nx CLI for all three kinds: correct file sets and `migrations.json` registration; scaffolded output lints clean and its skeleton tests pass; re-running is byte-identical.
- Exercised the backfill and stamping scripts against real and fabricated release-tag histories: a migration first tagged in `v1.0.0` records `1.0.0`, one first tagged in `v1.1.0` records `1.1.0`, an unreleased one is left for the release to stamp, and re-running the backfill is a no-op. With a fake `v9.9.9` tag the backfill `git mv`s the migration into `src/migrations/v9.9.9/<name>/`, re-keys the entry and rewrites both its `implementation` and `prompt` paths; regenerating the same name afterwards lands a fresh `latest-<name>` alongside the untouched released entry.
- Exercised the `upgrade-workspace` tool against the bundled MCP server binary (`tools/list` + `tools/call`).
- Verified the documented upgrade flow in a real workspace against two locally published plugin versions, which is where the guide's output and caveats come from:
  - each `nx migrate <package>` **overwrites** `migrations.json`, so upgrades must complete one package at a time;
  - run from inside an AI agent, Nx defers prompt migrations back to the driving agent with ordered next steps rather than spawning a nested agent, so the whole flow runs non-interactively — confirmed by applying a deferred prompt in-context;
  - multiple migrations can share one version: prompts are de-duplicated by content and disambiguated with `-1`/`-2` suffixes under `tools/ai-migrations/<package>/<version>/`.
- e2e coverage: the shared generator matrix runs a migration of each kind (so the CDK, Terraform and idempotency lanes all exercise the generator), and the `nx-plugin` smoke test asserts the per-kind file sets, `migrations.json` fields, `package.json` wiring, and that the scaffolded codemods type-check.
- Ran the full `package` target and `docs:build`.

### Follow-ups

Upgrading the plugin does not currently move the workspace's `nx`, even though the plugin pins it exactly (`utils/init.ts`) — so an nx pin bump reaches new workspaces only, and none of Nx's own migrations run on a plugin upgrade. Handling that needs a `packageJsonUpdates` block stamped alongside the migration versions, which belongs with the vended-version sync work in #988 rather than here; the upgrading guide deliberately says nothing about Nx upgrades until that lands.


docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,119 @@
...
+```
+
+- **`nextSteps`** — workspace-wide notes surfaced to the user after `nx migrate` runs. Use these to report files you skipped (see the guardrails below) or manual follow-up the migration couldn't perform.
+- **`agentContext`** (hybrid only) — a summary of what the codemod changed, passed to the paired `prompt` when it runs under Nx's agentic flow so the agent can focus on the user-owned parts.
+
+For the same operations you use in generators (`generateFiles`, GritQL transforms, reading and updating JSON), refer to the <Link path="/guides/nx-generator">Nx Generator guide</Link>.
+
+### Guardrails
+
+Migrations run against workspaces you don't control, so follow the same rules the scaffolded skeleton bakes in:
```
cogwirrel: , so it is recommended to adhere to the following:

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+- **Pattern-match before writing.** If a target file has diverged from the shape your generators produce, skip it and report it via `nextSteps` rather than clobbering the user's changes.
+- **Idempotent.** Re-running the migration must be a no-op.
+- **Never make formatting-only changes.**
+
+### Testing Your Migration
+
+The scaffolded `migration.spec.ts` uses `createTreeWithEmptyWorkspace()` to build a virtual workspace. Cover that the migration applies to the shape your generators produce, skips (and reports) customised files, and is idempotent:
```
cogwirrel: “Test” that the migration…

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+describe('my migration', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+
+  it('should update the generated shape', async () => {
+    // Arrange the vended shape, run the migration, assert the changes
```
cogwirrel: Let’s split into 3:

// Set up a tree with the “before” state

// Run the migration generator

// Check that the tree matches the target state

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,119 @@
...
+    const result = await migration(tree);
+    expect(result.nextSteps).toBeDefined();
+  });
+});
+```
+
+## Versioning
+
+Nx runs a migration when the user's installed plugin version is less than the migration's `version`. This generator writes **no** `version` field — assign it (or stamp it from your release process) when you cut the release that ships the migration, so it runs for everyone upgrading from an earlier version.
```
cogwirrel, line 126: Add a snippet of migrations.json below showing an example version, implementation, prompt

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+## Running Migrations
+
+Users apply your plugin's migrations by upgrading it:
+
+<NxCommands commands={['migrate my-plugin@latest', 'migrate --run-migrations']} />
```
cogwirrel, line 159: Do you need to run both or can you upgrade the plugin and run migrations  at the same time? Which commands update nx and which update in the plugin and what is the correct order? Can they be interleaved if bumping multiple versions? Eg we will update our version of nx over time

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+1. **Generate** — `nx migrate <package>@latest` updates the package's versions in `package.json` and writes a `migrations.json` listing the migrations to run. Nothing is installed and no other files are touched yet, so you can review the version changes first.
+
+2. **Install** — install the new versions with your package manager (e.g. `pnpm install --no-frozen-lockfile`).
+
+3. **Run** — `nx migrate --run-migrations` applies each pending migration to your workspace, one at a time.
+
+4. **Commit** — verify your workspace builds, then delete `migrations.json` and commit.
+
+</Steps>
```
cogwirrel: Delete this part.

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,148 @@
...
+- Some prompt migrations were skipped. Review and apply each of the following
+  prompt files to the workspace, in the listed order:
+  - tools/ai-migrations/@aws/nx-plugin/2.0.0/prompt.md
+```
+
+Each skipped prompt is a self-contained markdown instruction file in `tools/ai-migrations/` — paste it into your coding agent, or follow its steps manually.
+
+Finally, verify everything still builds, then delete `migrations.json` and commit:
+
+<NxCommands commands={['run-many --target build --all']} />
```
cogwirrel: package manager short command for build

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+<Aside type="note" title="pnpm build scripts">
+If `nx migrate latest` fails with `ERR_PNPM_IGNORED_BUILDS`, pnpm has blocked the `nx` package's post-install script. Allow it in `pnpm-workspace.yaml` and re-run:
+
+```yaml
+# pnpm-workspace.yaml
+allowBuilds:
+  nx: true
+```
+</Aside>
```
cogwirrel: Delete this.

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+Both Nx and `@aws/nx-plugin` ship migrations that update your workspace automatically when you upgrade them, using [`nx migrate`](https://nx.dev/docs/features/automate-updating-dependencies). This guide walks through upgrading each.
+
+<Aside type="tip">
+You can use your agentic coding tool of choice to upgrade your workspace for you, using the <Link path="get_started/building-with-ai">MCP server</Link>. Try the prompt:
+
+> Upgrade Nx and the Nx Plugin for AWS in this workspace.
+</Aside>
+
+`nx migrate <package>@<version>` targets whichever package you name — so upgrading Nx itself and upgrading `@aws/nx-plugin` are separate upgrades, each following the same cycle.
```
cogwirrel: remove this line

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+<Aside type="caution" title="One package at a time">
+Complete the full cycle for one package before starting the next: each `nx migrate <package>` **overwrites** `migrations.json`, so running `nx migrate @aws/nx-plugin@latest` followed immediately by `nx migrate latest` discards the plugin's pending migrations before they run.
+</Aside>
+
+## Upgrading @aws/nx-plugin
+
+Start the cycle by targeting the plugin:
```
cogwirrel: Change to "First, run the following command to update the plugin version and prepare migrations:"

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+ NX   Next steps:
+
+- Make sure package.json changes make sense and then run 'pnpm install --no-frozen-lockfile',
+- Review and tweak the AI migration prompts in tools/ai-migrations/ as needed.
+- Run 'pnpm exec nx migrate --run-migrations'
+- To learn more go to https://nx.dev/features/automate-updating-dependencies
+```
+
+Review the `package.json` changes, install, then run the migrations:
```
cogwirrel: Review the `package.json` changes then install:

```
<InstallCommand ... />
```

Then run the migrations:

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,160 @@
...
+  await formatFilesInSubtree(tree);
+
+  return { nextSteps };
+}
+```
+
+- **`nextSteps`** — workspace-wide notes surfaced to the user after `nx migrate` runs. Use these to report files you skipped (see the guardrails below) or manual follow-up the migration couldn't perform.
+- **`agentContext`** (hybrid only) — a summary of what the codemod changed, passed to the paired `prompt` when it runs under Nx's agentic flow so the agent can focus on the user-owned parts.
+
+For the same operations you use in generators (`generateFiles`, GritQL transforms, reading and updating JSON), refer to the <Link path="/guides/nx-generator">Nx Generator guide</Link>.
```
cogwirrel: For some example operations you can perform in your migration generator (generating files, modifying existing files using GritQL, etc), refer to the ...

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+## Implementing a Migration
+
+A deterministic (or hybrid) migration is a function which mutates the virtual filesystem (the `Tree`), returning a [`MigrationReturnObject`](https://nx.dev/docs/reference/devkit/MigrationReturnObject):
+
+```ts
+import { type MigrationReturnObject, type Tree } from '@nx/devkit';
+import { formatFilesInSubtree } from '@aws/nx-plugin/sdk/utils/format';
+
+export default async function migration(
```
cogwirrel, line 64: Make this example a hybrid so that the bullet on `agentContext` makes sense.

docs/src/content/docs/en/guides/nx-migration.mdx
```diff
@@ -0,0 +1,160 @@
...
+- **`nextSteps`** — workspace-wide notes surfaced to the user after `nx migrate` runs. Use these to report files you skipped (see the guardrails below) or manual follow-up the migration couldn't perform.
+- **`agentContext`** (hybrid only) — a summary of what the codemod changed, passed to the paired `prompt` when it runs under Nx's agentic flow so the agent can focus on the user-owned parts.
+
+For the same operations you use in generators (`generateFiles`, GritQL transforms, reading and updating JSON), refer to the <Link path="/guides/nx-generator">Nx Generator guide</Link>.
+
+### Guardrails
+
+Migrations run against workspaces you don't control, so it is recommended to adhere to the following (the scaffolded skeleton bakes these in):
+
+- **Pattern-match before writing.** If a target file has diverged from the shape your generators produce, skip it and report it via `nextSteps` rather than clobbering the user's changes.
```
cogwirrel: change to:"..., skip it and report it via `nextSteps`, or consider a a hybrid migration."

docs/astro.config.mjs
```diff
@@ -182,6 +182,20 @@ export default defineConfig({
...
+                fr: 'Mettre à niveau votre espace de travail',
+                it: 'Aggiornare il workspace',
+                es: 'Actualizar tu espacio de trabajo',
+                pt: 'Atualizando seu espaço de trabalho',
+                zh: '升级您的工作区',
+                vi: 'Nâng cấp không gian làm việc',
+              },
+            },
             {
               label: 'Workspaces',
```
cogwirrel: Move this below "Building with AI"

packages/nx-plugin/src/ts/nx-migration/files/migration.ts.template
```diff
@@ -0,0 +1,38 @@
...
+ * - https://nx.dev/docs/kb/migration-generators
+ * - <% if (isHybrid) { %>`nextSteps` and `agentContext`<% } else { %>What `nextSteps` means<% } %>: https://nx.dev/docs/reference/devkit/MigrationReturnObject
+ *
+ * Guardrails:
+ * - Skip files that have diverged from the shape your generators produce and
+ *   report them via `nextSteps` rather than clobbering the user's changes.
+ * - Re-running must be a no-op.
+ * - Never make formatting-only changes.
+ * - Finish with `formatFilesInSubtree`: `updateJson`/`writeJson` re-serialise
+ *   JSON in a shape the vended `format` target rejects.
```
cogwirrel: update this in line with the equivalent bullets in the docs

packages/nx-plugin/src/ts/nx-migration/files/migration.ts.template
```diff
@@ -0,0 +1,38 @@
...
+): Promise<MigrationReturnObject> {
+  const nextSteps: string[] = [];
+<% if (isHybrid) { %>  const agentContext: string[] = [];
+<% } %>
+  // TODO: read the workspace's files and update the ones matching the shape
+  // your generators produce, noting any you skip in `nextSteps`.
+<% if (isHybrid) { %>
+  // `agentContext` reaches the paired prompt under `--agentic` and is dropped
+  // otherwise. TODO: describe what changed, e.g.:
+  // agentContext.push('Renamed the `foo` target to `bar` in every project.json.');
```
cogwirrel, line 50: change to:

// TODO: describe what changed in `agentContext`

packages/nx-plugin/src/ts/nx-plugin/utils.ts
```diff
@@ -32,16 +34,19 @@ export const configureTsProjectAsNxPlugin = (
...
-  // Create a package.json if one doesn't exist, and configure "type", "main"
-  // and "generators"
+/**
+ * Create the plugin project's package.json if absent and point it at the given
+ * manifest (eg `generators`, `nx-migrations`), returning its path.
+ */
+export const configureNxPluginPackageJson = (
+  tree: Tree,
+  project: ProjectConfiguration,
+  manifestField: string,
```
cogwirrel: Type this as `'generators' | 'nx-migrations'`

packages/nx-plugin/src/ts/nx-plugin/utils.ts
```diff
@@ -32,16 +34,19 @@ export const configureTsProjectAsNxPlugin = (
...
+  manifestValue: unknown,
```
cogwirrel: Can we type this based on some types from nx?

packages/nx-plugin/src/utils/migration-versions.ts
```diff
@@ -0,0 +1,170 @@
...
+ * - The weekly `update-versions` PR backfills the version of the release that
+ *   shipped each migration, moving and re-keying it accordingly
+ *   (`scripts/backfill-migration-versions.ts`).
+ * - At package time (`scripts/stamp-migrations.ts`) entries still missing a
+ *   version are stamped into the compiled `migrations.json`: an already shipped
+ *   one gets the earliest release tag registering it, a net-new one gets a
+ *   version strictly between the latest tag and any possible next release.
+ *   `nx migrate` runs a migration when `installed < migration.version`, so any
+ *   version in that interval is correct whatever the next release turns out
+ *   to be.
```
cogwirrel: Reverse the order of these points

docs/src/content/docs/en/get_started/upgrading.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+<Aside type="tip">
+You can use your agentic coding tool of choice to upgrade your workspace for you, using the <Link path="get_started/building-with-ai">MCP server</Link>. Try the prompt:
+
+> Upgrade Nx and the Nx Plugin for AWS in this workspace.
+</Aside>
+
+<Aside type="caution" title="One package at a time">
+Complete the full cycle for one package before starting the next: each `nx migrate <package>` **overwrites** `migrations.json`, so running `nx migrate @aws/nx-plugin@latest` followed immediately by `nx migrate latest` discards the plugin's pending migrations before they run.
+</Aside>
```
cogwirrel: Remove this

packages/nx-plugin/src/ts/nx-migration/files/prompt.md.template
```diff
@@ -0,0 +1,31 @@
+<!-- Authoring guide (delete before shipping): this file is the `prompt` for <% if (isHybrid) { %>the
+agentic half of a hybrid migration. The codemod half (migration.ts) has already
+run its mechanical changes and passes a summary via `agentContext`. The prompt is<% } else { %>an
+agentic migration. It is<% } %>
+applied by the user's AI coding agent, and shown verbatim to users who run
+`nx migrate` without one, so write self-contained, human-actionable steps.
+See https://nx.dev/docs/kb/migration-generators -->
```
cogwirrel: For this and the other conditional parts of this file, prefer duplication rather than if statements chopping up sentences as it's harder to follow, eg
```
<%_ if (isHybrid) { _%>
<!-- Authoring guide ...
<%_ } else { _%>
<!-- Authoring guide ...
<%_ } _%>
```

packages/nx-plugin/src/utils/migration-versions.ts
```diff
@@ -0,0 +1,185 @@
...
+ * reach the published `migrations.json` two ways:
+ *
+ * - The weekly `update-versions` PR backfills the version of the release that
+ *   shipped each migration, moving and re-keying it accordingly
+ *   (`scripts/backfill-migration-versions.ts`).
+ * - At package time (`scripts/stamp-migrations.ts`) entries still missing a
+ *   version are stamped into the compiled `migrations.json`: an already shipped
+ *   one gets the earliest release tag registering it, and a net-new one gets the
+ *   version the release is about to publish, which the release job passes in
+ *   after `nx release version` writes it to the dist manifests.
```
cogwirrel: Swap the order of these two points

packages/nx-plugin/src/utils/migration-versions.ts
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * When no pending version is supplied (packaging locally or on a PR, where no
+ * release follows) a net-new migration falls back to a version strictly between
+ * the latest tag and any possible next release. `nx migrate` runs a migration
+ * when `installed < migration.version`, so that interval is always correct — it
+ * just isn't a version that was really published, which is why the release
+ * itself passes the real one.
```
cogwirrel: Remove fallback - we should always supply a version argument. Remove all relevant fallback code and tests.

packages/nx-plugin/src/utils/migration-versions.ts
```diff
@@ -0,0 +1,185 @@
...
+    );
+    if (movedPath) {
+      const dir = movedPath.slice(0, movedPath.indexOf(latestSegment));
+      moves.push({
+        name,
+        version,
+        from: `${dir}${latestSegment}`.replace(/^\.\/|\/$/g, ''),
+        to: `${dir}${versionSegment}`.replace(/^\.\/|\/$/g, ''),
+      });
+    }
```
cogwirrel: Can we simplify this as a single pass?

scripts/utils/migration-release-tags.ts
```diff
@@ -0,0 +1,54 @@
...
+    // Tag predates migrations.json
+    return undefined;
+  }
+};
+
+/**
+ * Map of migration key -> version of the earliest release tag registering it.
+ * A migration that hasn't been released is absent.
+ */
+export const readShippedMigrationVersions = (
```
cogwirrel: Since tags are processed in ascending order we always look through every single version of the file.

Can we instead:
- Accept the current list of currently unstamped migrations (ie haven't been backfilled yet, missing "version")
- Read migrations.json from each tag in descending order, and for each of the unstamped migrations figure out where it disappears from migrations.json - the one before it disappears is where it's first introduced and therefore its shipped version.

scripts/stamp-migrations.ts
```diff
@@ -0,0 +1,92 @@
...
+const DIST_MIGRATIONS_PATH = 'dist/packages/nx-plugin/migrations.json';
+
+/**
+ * The release the publish is about to make, which the release job reads out of
+ * the dist manifest `nx release version` just wrote.
+ */
+const readPendingVersion = (argv: string[]): string | undefined => {
+  const index = argv.indexOf('--pending-version');
+  if (index === -1) {
+    return undefined;
```
cogwirrel: no fallback, version is required

CONTRIBUTING.md
```diff
@@ -102,6 +102,99 @@ it('should be idempotent when re-run with same options', async () => {
...
+- **agentic** — `prompt.md` (self-contained agent/human instructions).
+- **hybrid** — `migration.ts` (returning `agentContext`) + `migration.spec.ts` + `prompt.md`.
+
+Entries are keyed by folder and name (`latest-<name>`, becoming `v<x.y.z>-<name>` once a release claims them), so reusing a name for a later change can't silently overwrite a migration that has already shipped.
+
+`ts#nx-migration` is a public generator, so it works on any Nx Plugin project (it creates `migrations.json` and wires the `nx-migrations` field into the plugin's `package.json` if absent). See its [guide](./docs/src/content/docs/en/guides/nx-migration.mdx) for the full reference. Within this repo, always pass `--project=@aws/nx-plugin`.
+
+#### What should be a migration
+
+First, **is a migration needed at all?** A migration exists to keep an *already-generated* workspace working across a breaking change. If existing workspaces keep building without any change, there is no migration to write.
```
cogwirrel: Expand this, it's not just breaking changes - new features should be added too. Security hardening improvements, bugfixes, etc that touch common constructs / shared code.

CONTRIBUTING.md
```diff
@@ -102,6 +102,99 @@ it('should be idempotent when re-run with same options', async () => {
...
+
+Not a migration at all:
+
+- Formatting or stylistic changes
+- New optional features — users adopt them by re-running the (idempotent) generator
```
cogwirrel: Remove this one

CONTRIBUTING.md
```diff
@@ -102,6 +102,99 @@ it('should be idempotent when re-run with same options', async () => {
...
+
+#### Choosing the kind: deterministic, agentic, or hybrid
+
+Once a change *is* a migration, pick the kind with two questions, in order:
+
+1. **Do we own the target file?** — i.e. is it a file the generator fully owns and users aren't expected to hand-edit (`project.json`, `nx.json`, `aws-nx-plugin.config.mts`, shared constructs following the vended pattern, generated clients, runtime-config wiring)? Or is it user-authored code built *on top of* our scaffolding (`agent.py`, custom CDK stacks, custom routers and components)?
+2. **Is the edit deterministic?** — can we express it as an exact before/after that holds for every workspace, or does the correct edit depend on what the user has built?
+
+**Always prefer deterministic.** Deterministic migrations are the required unattended upgrade path: they alone must take an uncustomised generated workspace from one version to the next with build, lint and test green, including in CI and non-interactive terminals where no agent runs. Only reach for a prompt when the edit genuinely cannot be expressed mechanically.
```
cogwirrel: Soften this - where a deterministic generator would be really complex we prefer hybrid/agentic

CONTRIBUTING.md
```diff
@@ -102,6 +102,151 @@ it('should be idempotent when re-run with same options', async () => {
...
+
+Unit tests only prove the migration does what you wrote it to do. Also **test the migration manually against a real workspace generated by the latest published version**, since that is the state your users are upgrading from. The goal is to confirm the migrated workspace matches what today's generators would produce.
+
+First build the local packages:
+
+```bash
+pnpm package:all
+```
+
+Then create a workspace on the **published** version, generate the projects your migration touches, and commit it so the migration's changes are easy to inspect. Create it somewhere with no parent `package.json` or `pnpm-workspace.yaml` above it, or pnpm treats that parent as the workspace root and links into the wrong place:
```
cogwirrel: remove "Create it somewhere with no parent `package.json` or `pnpm-workspace.yaml` above it, or pnpm treats that parent as the workspace root and links into the wrong place"

CONTRIBUTING.md
```diff
@@ -102,6 +102,151 @@ it('should be idempotent when re-run with same options', async () => {
...
+
+```bash
+mkdir -p ~/tmp/migtest && cd ~/tmp/migtest
+npx -y create-nx-workspace@latest ws --preset=@aws/nx-plugin --ci=skip --pm=pnpm
```
cogwirrel: `pnpm create @aws/nx-workspace <my-workspace>`

CONTRIBUTING.md
```diff
@@ -102,6 +102,151 @@ it('should be idempotent when re-run with same options', async () => {
...
+First build the local packages:
+
+```bash
+pnpm package:all
+```
+
+Then create a workspace on the **published** version, generate the projects your migration touches, and commit it so the migration's changes are easy to inspect. Create it somewhere with no parent `package.json` or `pnpm-workspace.yaml` above it, or pnpm treats that parent as the workspace root and links into the wrong place:
+
+```bash
+mkdir -p ~/tmp/migtest && cd ~/tmp/migtest
```
cogwirrel: Remove this line

CONTRIBUTING.md
```diff
@@ -102,6 +102,151 @@ it('should be idempotent when re-run with same options', async () => {
...
+npx -y create-nx-workspace@latest ws --preset=@aws/nx-plugin --ci=skip --pm=pnpm
+cd ws
+# Resolve the vended pnpm-workspace.yaml `allowBuilds` placeholder, then install
+pnpm install --no-frozen-lockfile
```
cogwirrel: Remove this, not required via `pnpm create @aws/nx-workspace`

cogwirrel (approved):

PR #937, opened 2026-07-14
fix(open-api#ts-client): send primitive urlencoded request bodies verbatim

BenSMantel:
### Reason for this change

The ts-client generator emitted object marshalling — `this.$urlEncodedForm(input)` — for every `application/x-www-form-urlencoded` request body, without checking whether the body schema is an object or a primitive. `$urlEncodedForm` is typed `(model: { [key: string]: any })` and iterates `Object.entries(model)`, so an operation whose urlencoded body is a raw string produced a client that fails to type-check:

    error TS2345: Argument of type 'string' is not assignable to parameter
    of type '{ [key: string]: any; }'.

### Description of changes

Fix: treat a urlencoded body as form-encoded only when its schema is an object. A new `isUrlEncodedObjectBody(op)` predicate now gates both the `$urlEncodedForm` helper emission (`hasUrlEncodedBody`) and the urlencoded marshalling branch. A primitive urlencoded body falls through to the existing `String(input)` branch, sending the pre-encoded string verbatim. The Content-Type header is set independently of the body branch, so it stays `application/x-www-form-urlencoded` in both cases. Object urlencoded bodies are unchanged (still `$urlEncodedForm`).

Also fixes `open-api#ts-hooks`, which delegates to the ts-client generator and reuses this template.

Adds a `generator.content-type.spec.ts` case covering a primitive-string body (asserts the generated client compiles and sends the raw string, content type preserved) and an object body (asserts `key=value` form encoding).

All existing ts-client tests and snapshots pass unchanged.

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/open-api/ts-client/files/client.gen.ts.template
+++ b/packages/nx-plugin/src/open-api/ts-client/files/client.gen.ts.template
@@ -49,7 +49,10 @@ const bodyMediaTypeOf = (op) => {
   return mediaTypes.find(mt => mt === 'application/json' || mt.endsWith('+json')) || mediaTypes[0];
 };
 const hasMultipartBody = allOperations.some(op => bodyMediaTypeOf(op) === 'multipart/form-data');
-const hasUrlEncodedBody = allOperations.some(op => bodyMediaTypeOf(op) === 'application/x-www-form-urlencoded');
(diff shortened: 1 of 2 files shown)


cogwirrel (approved):
Great catch! Thanks so much for this!

PR #939, opened 2026-07-14
feat(ts#dcr-proxy): add a generator for dynamic client registration enabling coding agents to connect to mcp servers

drskur:
### Reason for this change

Cognito does not support DCR (Dynamic Client Registration), which made it hard to implement authentication via OAuth. So I implemented a generator for a proxy that virtually handles DCR.

### Description of changes

  - Generate the handlers needed for the DCR proxy as a dcr-proxy project
  - Generate the IaC to serve them with AWS Lambda and API Gateway (cdk / terraform)

  - Updated and extended the generator unit tests, including a new terraform describe block covering module generation, the shared core http-api module, all seven routes, token-only secret access, the mcp-proxy
  upstream configuration, the shared terraform build dependency, and idempotency. Snapshots were regenerated.
  - The full @aws/nx-plugin test suite passes (133 files, 2644 tests), along with compile and lint.
  - The docs build succeeds and generates the new guide page.

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist
- [X] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/awslabs/nx-plugin-for-aws/blob/main/DESIGN_GUIDELINES.md)

----


docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+<RunGenerator generator="ts#dcr-proxy" />
+
+### Options
+
+<GeneratorParameters generator="ts#dcr-proxy" />
+
+## Generator Output
+
+The generator creates a standalone TypeScript project containing the Lambda handlers, and infrastructure to deploy them based on your selected `iacProvider`.
```
cogwirrel: Nit: please can we link "TypeScript project" to the ts project guide page? :)

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+<Snippet name="shared-constructs" />
+
+<Infrastructure>
+<Fragment slot="cdk">
+The generator creates a CDK construct which deploys the proxy, residing in the `packages/common/constructs/src/app/dcr-proxies/<dcr-proxy-name>` directory.
+</Fragment>
+<Fragment slot="terraform">
+The generator creates a Terraform module which deploys the proxy, residing in the `packages/common/terraform/src/app/dcr-proxies/<dcr-proxy-name>` directory.
+</Fragment>
+</Infrastructure>
```
cogwirrel, line 74: Nit: please can we use the `<FileTree>` here, eg:

```
For deploying the proxy, the following files are generated:

<Infrastructure>
<Fragment slot="cdk">
<FileTree>
- packages/common/constructs/src
  - app
    - dcr-proxies
      - <dcr-proxy-name>
        - <dcr-proxy-name>.ts
</FileTree>
...
```

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+  asset_bucket_name         = module.asset_bucket.bucket_name
+}
+```
+
+The module exposes the proxy endpoints (`proxy_url`, `mcp_url`, `metadata_url`, `token_endpoint`, `registration_endpoint`) as outputs.
+</Fragment>
+</Infrastructure>
+
+:::note
+The `/mcp` proxy uses API Gateway HTTP API integrations, which have a hard 29 second timeout. Long-running upstream requests are buffered by the proxy rather than streamed.
```
cogwirrel, line 140: Would it be better to use a REST API with response streaming to avoid this limitation? :)

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+| `GET /.well-known/oauth-authorization-server` | Authorization server metadata |
+| `GET /.well-known/openid-configuration` | OpenID configuration (served by the authorization server metadata handler) |
+| `POST /register` | Dynamic Client Registration |
+| `GET /authorize` | Authorization (redirects to the Cognito Hosted UI) |
+| `POST /oauth/token` | Token exchange (injects the App Client secret) |
+| `ANY /mcp` | Proxy to the upstream MCP server |
+
+Only the token handler is granted read access to the Cognito App Client secret in Secrets Manager.
+
+## Using the DCR Proxy
```
cogwirrel: Nit: rename to `## Deploying the DCR Proxy`

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+}
+```
+
+The module exposes the proxy endpoints (`proxy_url`, `mcp_url`, `metadata_url`, `token_endpoint`, `registration_endpoint`) as outputs.
+</Fragment>
+</Infrastructure>
+
+:::note
+The `/mcp` proxy uses API Gateway HTTP API integrations, which have a hard 29 second timeout. Long-running upstream requests are buffered by the proxy rather than streamed.
+:::
```
cogwirrel, line 141: Please could we add a section like `## Consuming a Proxied MCP Server` and include some documentation on how this enables agents to invoke an mcp server with auth? eg how do you set up Claude Code, Kiro CLI, or MCP Inspector to point to your proxied MCP server

packages/nx-plugin/src/ts/dcr-proxy/generator.ts
```diff
@@ -0,0 +1,134 @@
...
+} from '../../utils/nx';
+import { sortObjectKeys } from '../../utils/object';
+import { sharedConstructsGenerator } from '../../utils/shared-constructs';
+import type { TsDcrProxyGeneratorSchema } from './schema';
+
+export const TS_DCR_PROXY_GENERATOR_INFO: NxGeneratorInfo = getGeneratorInfo(
+  import.meta.filename,
+);
+
+export const tsDcrProxyGenerator = async (
```
cogwirrel, line 41: Please could you add an export for this in `src/sdk/ts.ts`?

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+  readonly lambdaEnvironment?: Record<string, string>;
+  readonly logRetention?: logs.RetentionDays;
+  readonly removalPolicy?: RemovalPolicy;
+
+  /**
+   * Timeout applied to the MCP proxy Lambda. Defaults to 29 seconds (the
+   * API Gateway HTTP API integration limit).
+   */
+  readonly mcpProxyTimeout?: Duration;
+  readonly mcpProxyMemorySize?: number;
```
cogwirrel, line 146: These options probably don't need to be exposed, since users can edit the construct anyway and are probably unlikely to change these settings from the defaults.

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+];
+const DEFAULT_RATE_LIMIT = 50;
+const DEFAULT_BURST_LIMIT = 100;
+
+// Resolve a handler's rolldown bundle output directory, relative to this
+// compiled construct file, back to the workspace root.
+const asset = (bundlePathFromRoot: string): lambda.Code =>
+  lambda.Code.fromAsset(
+    url.fileURLToPath(
+      new URL(\`../../../../../../../\${bundlePathFromRoot}\`, import.meta.url),
```
cogwirrel, line 162: Can we use `findWorkspaceRoot` like other constructs? :)

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+  readonly authServerMetadataUrl: string;
+  readonly tokenEndpoint: string;
+  readonly registrationEndpoint: string;
+
+  constructor(scope: Construct, id: string, props: SnapshotProxyProps) {
+    super(scope, id);
+
+    const region = Stack.of(this).region;
+    const logRetentionDefault =
+      props.logRetention ?? logs.RetentionDays.ONE_WEEK;
```
cogwirrel, line 188: Let's make the default a year like the other constructs :)

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+   */
+  readonly userPoolClientId: string;
+
+  /**
+   * ARN of the Secrets Manager secret holding the Cognito App Client secret.
+   * The token handler reads this at runtime to complete the token exchange;
+   * the secret is never exposed to the client. Create the secret in your own
+   * stack (from your App Client's generated secret) and pass its ARN here.
+   */
+  readonly cognitoClientSecretArn: string;
```
cogwirrel, line 116: Would it be cleaner to accept `ISecret` here? :)

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+    ): lambda.Function => {
+      const logGroup = new logs.LogGroup(this, \`\${id_}LogGroup\`, {
+        retention: logRetentionDefault,
+        removalPolicy: removalPolicyDefault,
+      });
+
+      return new lambda.Function(this, id_, {
+        code: asset(bundlePathFromRoot),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_20_X,
```
cogwirrel: Can we use NODEJS_LATEST? :)

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+  params.delete('resource');
+
+  const res = await fetch(\`\${COGNITO_HOSTED_UI_BASE}/oauth2/token\`, {
+    method: 'POST',
+    headers: { 'content-type': 'application/x-www-form-urlencoded' },
+    body: params.toString(),
+  });
+
+  const responseBody = await res.text();
+  console.log('[token]', {
```
cogwirrel, line 613: Nit: should we use powertools logger like other lambdas?

packages/nx-plugin/src/ts/dcr-proxy/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,788 @@
...
+  #checkov:skip=CKV_AWS_115:Concurrent execution limit not required for this use case
+  #checkov:skip=CKV_AWS_173:Lambda environment variables encrypted by managed key
+  for_each = local.standard_handlers
+
+  s3_bucket        = aws_s3_object.handler_zip[each.key].bucket
+  s3_key           = aws_s3_object.handler_zip[each.key].key
+  function_name    = "snapshot-proxy-\${each.key}-\${random_string.suffix.result}"
+  role             = each.key == "token" ? aws_iam_role.token.arn : aws_iam_role.proxy.arn
+  handler          = "index.handler"
+  runtime          = "nodejs20.x"
```
cogwirrel: Same here can we use a newer node? :)

packages/nx-plugin/src/ts/dcr-proxy/generator.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+
+    const construct = tree.read(`${defaultDir}/dcr-proxy.ts`, 'utf-8');
+    expect(construct).toContain('export class DcrProxy extends Construct');
+  });
+
+  it('generates the construct and its 6 lambda handlers', async () => {
+    await tsDcrProxyGenerator(tree, { name: 'my-proxy', iac: 'cdk' });
+
+    expect(tree.exists(`${constructDir}/my-proxy.ts`)).toBe(true);
+    for (const handler of [
```
cogwirrel, line 52: Please could we snapshot each handler? :)

packages/nx-plugin/src/ts/dcr-proxy/generator.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+import {
+  ensureAwsNxPluginConfig,
+  updateAwsNxPluginConfig,
+} from '../../utils/config/utils';
+import { expectHasMetricTags } from '../../utils/metrics.spec';
+import { getNpmScopePrefix } from '../../utils/npm-scope';
+import { createTreeUsingTsSolutionSetup } from '../../utils/test';
+import { TS_DCR_PROXY_GENERATOR_INFO, tsDcrProxyGenerator } from './generator';
+
+describe('ts#dcr-proxy generator', () => {
```
cogwirrel, line 15: Please could you also add `ts#dcr-proxy` to the `generator-matrix.ts` so it participates in the smoke tests?

packages/nx-plugin/src/utils/dcr-proxy-constructs/files/project/src/handlers/mcp-proxy.ts.template
```diff
@@ -0,0 +1,85 @@
+import type {
+  APIGatewayProxyHandlerV2,
+  APIGatewayProxyStructuredResultV2,
+} from 'aws-lambda';
+
+const UPSTREAM_URL = process.env.UPSTREAM_URL!;
+const PROXY_METADATA_URL = process.env.PROXY_METADATA_URL!;
+
+const HOP_BY_HOP = new Set([
```
cogwirrel, line 13: I'm not sure what "hop by hop" means :) Is there a better name and/or could we add a short comment explaining what these headers are used for?

packages/nx-plugin/src/utils/dcr-proxy-constructs/dcr-proxy-constructs.ts
```diff
@@ -0,0 +1,218 @@
...
+  );
+
+  // The token handler imports @aws-sdk/client-secrets-manager (provided by the
+  // Lambda runtime but needed for type-checking and local dev) and the handlers
+  // are typed against @types/aws-lambda.
+  addDependenciesToPackageJson(
+    tree,
+    withVersions(['@aws-sdk/client-secrets-manager']),
+    withVersions(['@types/aws-lambda']),
+  );
```
cogwirrel: I think this should be in `dcr-proxy/generator.ts`? The dependencies are for the handlers and this is currently only called if vending cdk infrastructure.

cogwirrel (commented):
Thanks so much @drskur! This is awesome, just have a few minor comments :)

packages/nx-plugin/src/utils/dcr-proxy-constructs/files/cdk/app/dcr-proxies/__nameKebabCase__/__nameKebabCase__.ts.template
```diff
@@ -0,0 +1,293 @@
...
+    const clientSecret = secretsmanager.Secret.fromSecretCompleteArn(
+      this,
+      'CognitoClientSecret',
+      props.cognitoClientSecretArn,
+    );
+
+    const cognitoIssuer = `https://cognito-idp.${region}.amazonaws.com/${props.userPoolId}`;
+
+    const baseEnv: Record<string, string> = {
+      COGNITO_USER_POOL_ID: props.userPoolId,
```
cogwirrel: nit: looks like this env variable is unused? :)

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+```typescript
+import { DcrProxy } from ':my-scope/common-constructs';
+
+new DcrProxy(this, 'DcrProxy', {
+  userPoolId: userPool.userPoolId,
+  userPoolClientId: userPoolClient.userPoolClientId,
+  cognitoClientSecretArn: clientSecret.secretArn,
+  cognitoHostedUiBase: userPoolDomain.baseUrl(),
+  upstreamUrl: 'https://my-agentcore-runtime-url/mcp',
```
cogwirrel, line 113: Looks like `allowedRedirectPatterns` needs to be set for the proxy to work? We should call out what this should be set to for eg claude code in a `Consuming a Proxied MCP Server`

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -0,0 +1,127 @@
...
+| `POST /register` | Dynamic Client Registration |
+| `GET /authorize` | Authorization (redirects to the Cognito Hosted UI) |
+| `POST /oauth/token` | Token exchange (injects the App Client secret) |
+| `ANY /mcp` | Proxy to the upstream MCP server |
+
+Only the token handler is granted read access to the Cognito App Client secret in Secrets Manager.
+
+## Using the DCR Proxy
+
+The proxy does not create your Cognito resources or your MCP server. Instead, you inject the identifiers of resources managed elsewhere (whether generated by this plugin or provisioned separately), keeping the proxy decoupled from how those resources are provisioned.
```
cogwirrel, line 92: Nice, I like how decoupled this is :) I think we should add a section to the docs for how to use this with the `ts#mcp-server` or `py#mcp-server`, ie I assume you generate it with `--auth cognito`, then pass the same user pool and client it uses. Would be nice to have full examples for both cdk and tf :)

cogwirrel: Also if you rebase, the example can use the new invocation url property from https://github.com/awslabs/nx-plugin-for-aws/pull/942 :)

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
drskur: Good question! To actually get response streaming here we'd need a Lambda Function URL rather than an API Gateway integration — and that has a structural consequence worth calling out.

A Function URL maps one-to-one to a single Lambda, so there's no per-route routing in front of it the way API Gateway gives us today. But the OAuth discovery flow is same-origin: the /mcp response advertises resource_metadata on the same host, and the client then hits /.well-known/..., /register, /authorize, and /oauth/token on that same domain. So if we move /mcp to a Function URL for streaming, everything has to live behind one origin — which means collapsing all six handlers into a single Lambda that routes internally on rawPath. It also means giving up the current least-privilege split where only the token handler can read the App Client secret.

That's a fairly large change, so before going there I'd like to sanity-check whether it's worth it:

- The MCP spec does define streaming, but the AgentCore Gateway implementation we proxy doesn't really appear to be built around streaming either.
- In practice most MCP usage is "do a discrete piece of work and respond quickly" rather than long-lived streamed responses — which is exactly what the buffered 29s path already handles fine.

Given that, do you think streaming is actually a requirement for this proxy's use cases? 
Happy to do the Function URL rework if we think streaming matters here.

cogwirrel: If it’s not too heavy a lift we can use individual lambdas with a REST api? The trpc/fastapi generators configure this (response transfer mode set to STREAM) when a REST api is picked :) https://docs.aws.amazon.com/apigateway/latest/developerguide/response-transfer-mode.html

drskur: The problem: REST APIs force a stage segment into the URL

A REST API is always invoked at https://{id}.execute-api.{region}.amazonaws.com/{stage}/.... The stage name (/prod) is a mandatory path segment — there is no equivalent of the HTTP API $default stage that serves routes at the host root.

That breaks OAuth/MCP metadata discovery. Per RFC 8414 (Authorization Server Metadata) and RFC 9728 (Protected Resource Metadata), clients fetch the well-known documents from the host root:

  - https://{host}/.well-known/oauth-authorization-server
  - https://{host}/.well-known/oauth-protected-resource

With a REST API the actual endpoints live under https://{host}/prod/.well-known/..., so the client's root-relative discovery request https://{host}/.well-known/... returns a 403/404. Rewriting the URLs inside the metadata documents (to include /prod) doesn't help — the client never reaches the metadata document in the first place, because it looks for it at the root.

The only way to make REST API work: a custom domain

cogwirrel: Ahh ok gotcha, in that case let's leave it as-is with an HTTP API :) Thanks!

cogwirrel (dismissed):
This is great, thanks so much @drskur! Awesome to make it nice and easy to expose MCP servers to coding agents 😄 🎉

cogwirrel (approved):

PR #950, opened 2026-07-17
feat(ts#project): vend per-project package.json, npm aliases and dependency catalogs

nx-plugin-for-aws:
### Reason for this change

TypeScript projects previously had no `package.json` (the generator deleted it), imports between projects used a non-standard colon alias (`:scope/name`), and dependency versions were pinned as direct ranges in the root manifest. This deviates from the idiomatic npm/Nx workspace layout, adding friction for users adopting the plugin in standard Nx workspaces and breaking common workflows like `pnpm add <pkg> --filter <project>`.

### Description of changes

**Per-project `package.json`**
- Every generated project (TS projects, websites, agents, MCP servers, tRPC/Smithy APIs, infra, terraform applications, astro-docs, shared constructs/shadcn) now carries a minimal private manifest declaring its package name and an explicit `type` (module format), so package managers link workspace projects and nearest-manifest module-format resolution stays consistent.
- Workspace globs (`pnpm-workspace.yaml` / `workspaces` field) are extended automatically when a project is generated outside the covered directories. Both the array and object (`{"packages": [...]}`) forms of the `workspaces` field are supported (yarn and bun accept either), and negated globs exclude regardless of their position in the list, matching package manager semantics.
- `@nx/js`'s default `./dist`-based entry points are stripped since build output goes to the workspace `dist` and imports resolve via path aliases; user-customised entry points and subpath exports are preserved.
- Running a project generator in a workspace that hasn't been initialised logs a warning pointing at `nx add @aws/nx-plugin` / the `init` generator, since lint/format targets and dependency sync need the workspace-level setup.

**npm-format aliases**
- `tsconfig.base.json` path aliases now use the project's package name (`@scope/name`) instead of the colon form (`:scope/name`). Cross-project imports match npm conventions; legacy colon aliases are cleaned up on re-run.

**Per-project runtime dependencies, enforced**
- Each generated project declares the third-party runtime dependencies its source imports in its own `package.json`; shared build/test tooling stays in the root manifest. Biome's `noUndeclaredDependencies` rule is enforced as an error (config/test/stories files exempt). The docs cover opting back into root-only dependency declarations by disabling the rule.
- Terraform application projects declare their deploy-time AWS SDK dependencies (used by the vended bootstrap/init scripts) in their own manifest, and the astro-docs project declares its runtime dependencies (astro, starlight, translation script imports) in its manifest.
- The `ts#sync` generator derives local workspace dependencies from the Nx project graph (the same import analysis as `@nx/js:typescript-sync`) and declares them automatically using the package manager's local-dependency specifier: `workspace:*` where the workspace protocol is supported (pnpm, bun, yarn berry), `*` on npm and yarn classic, which reject the `workspace:` protocol. Invoking it directly via `nx g` works as well as via `nx sync`.
- A tree-level audit test asserts every vended project's source imports only packages its manifest declares — covering the TS project types plus terraform and astro-docs — so template/dependency drift fails unit tests rather than downstream smoke tests.

**Dependency catalogs (single version policy)**
- Generators route third-party dependency versions through the package manager's dependency catalog where supported: `pnpm-workspace.yaml` `catalog` ([pnpm ≥ 9.5](https://pnpm.io/catalogs)), `.yarnrc.yml` `catalog` ([yarn ≥ 4.10](https://yarnpkg.com/features/catalogs)), root `package.json` `catalog` ([bun ≥ 1.2.14](https://bun.com/docs/install/catalogs)). Root manifest entries become `catalog:` references with the catalog as the single source of truth for versions — including entries other Nx generators write as direct root ranges (`@types/node`, `react`), so only one copy of each package resolves.
- On bun, packages whose versions Nx generators read from the root manifest without resolving catalogs (`vite`, `react`, `react-dom`) keep direct version ranges — Nx ships no bun catalog manager yet, and a `catalog:` reference crashes those generators.
- Existing catalog entries are never downgraded: a generator only updates a catalog version when the incoming version is a strict upgrade, so user-raised versions survive subsequent generator runs; non-semver user customisations (tags, compound ranges like `>=3 <5`) are always kept.
- npm has no catalog feature, so npm workspaces keep direct ranges; the docs recommend [syncpack](https://github.com/JamieMason/syncpack) for aligning versions across manifests on npm.
- Catalogs are optional: `--catalog false` at workspace creation, or `packageManager.catalogs: false` in `aws-nx-plugin.config.mts` at any time. A config file that fails to evaluate logs a warning rather than silently reverting to defaults.

**Docs & guidance**
- Rewrote the dependency/project-structure doctrine in the TypeScript project and workspace guides (per-project manifests, catalogs, syncpack, publishing, the root-only opt-out), updated all generator file trees, and converted all import examples to `@scope/name` (English docs only; translations follow via the translation workflow).
- The "add to an existing workspace" guide documents that the first `ts#project` run renames existing libraries' inferred `build` target to `compile`.
- Updated MCP server general guidance and the vended guidance templates so coding agents get the correct instructions (per-project runtime deps, correct per-package-manager install commands including bun's `--cwd` form).

- Full unit test suite passes (2700 tests, 138 files) with updated snapshots; unit tests cover the manifest vending util, the catalog-aware dependency util (per package manager, version gating, downgrade protection, root-range alignment, bun exclusions, idempotency), the per-package-manager local dependency specifier, array/object workspaces forms and negated globs, terraform application manifests, astro-docs project manifests, and the tree-level undeclared-import audit (verified to fail on an injected undeclared import).
- Built all packages and validated end-to-end against a local registry across all four package managers: created fresh pnpm, npm, yarn 4 and bun workspaces with the locally built `@aws/create-nx-workspace`, ran `ts#project`, `ts#trpc-api`, `ts#react-website`, `ts#infra`, `ts#mcp-server` and `terraform#project`, and verified `nx run-many --target build` is green, every project has a correct manifest, aliases are `@scope/name`, catalog entries are written (pnpm/yarn/bun) / direct ranges written (npm), `pnpm add <pkg> --filter <project>` targets the right project and enforces the catalog (`catalogMode: strict`), the tRPC dev server serves requests, and terraform `bootstrap`/`init` scripts resolve their AWS SDK imports.
- Verified interop with a vanilla `create-nx-workspace --preset=ts` workspace: `ts#project` respects the existing scope, native `@nx/js:library` projects coexist and cross-import in both directions with `ts#sync` maintaining `workspace:*` entries in native manifests, `nx sync` is idempotent with both sync generators registered, and third-party generators (`@nxlv/python`) are unaffected.
- Verified empirically that npm and yarn classic reject `workspace:*` and correctly resolve `*` for workspace-local packages; the smoke harness re-installs after `nx sync` so all package-manager smoke tests exercise the manifests sync writes.
- Catalog support matrix verified empirically for pnpm 11 / yarn 4.17 / bun 1.3 (install with `catalog:` refs in root and project manifests, Nx graph, target execution), including that consecutive vitest-configuring generators succeed on bun.

### Checklist

----


docs/src/content/docs/en/guides/react-website.mdx
```diff
@@ -59,6 +59,7 @@ The generator will create the following project structure in the `<directory>/<n
   - tsconfig.json Base TypeScript configuration for source and tests
   - tsconfig.app.json TypeScript configuration for source code
   - tsconfig.spec.json TypeScript configuration for tests
+  - package.json Minimal project manifest defining the project's package name
```
cogwirrel: add "and dependencies". same for other occurences

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -85,11 +86,11 @@ export * from './algorithms/index.js';
...
 
 ```ts title="packages/my-other-project/src/index.ts"
-import { sayHello } from ':my-scope/my-library';
+import { sayHello } from '@my-scope/my-library';
 ```
 
 :::note[Project Aliases]
-Aliases for your TypeScript projects begin with a `:` rather than the traditional `@`, since this avoids the possibility of name conflicts between local packages in your workspace and remote packages in [NPM](https://www.npmjs.com/).
+Aliases match your projects' package names, so any project in your workspace can be imported by the name declared in its `package.json` — no need to declare local projects as dependencies.
 :::
```
cogwirrel: delete this note

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
 
 ### Dependencies
 
-You will notice that your TypeScript project does not have a `package.json` file, which might be unexpected if you are used to traditional TypeScript monorepos.
+Each TypeScript project declares the third-party **runtime dependencies its source code imports** in its own `package.json`. This keeps every project self-describing — its manifest lists exactly what it needs — and lets Biome's [`noUndeclaredDependencies`](https://biomejs.dev/linter/rules/no-undeclared-dependencies/) lint rule enforce that a project never imports a package it hasn't declared.
 
-To add a dependency for any TypeScript package in your monorepo, simply add the dependency to the `package.json` in the root of your workspace. You can do this via the command line for your package manager:
+**Build and test tooling** (for example `vitest`, `tsx`, `rolldown`, the Nx toolchain (`nx`, `@nx/js`, ...), `typescript`, `@biomejs/biome`) is shared across the workspace, so it lives in the **root** `package.json` `devDependencies` rather than in each project. These tools are only referenced from config files, build scripts and tests, which the lint rule exempts.
```
cogwirrel: Each TypeScript project declares its third-party runtime dependencies in its own `package.json`. Build and test tooling shared across projects is defined in the root `package.json`.

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
 
-<InstallCommand pkg="some-npm-package" />
+To add a runtime dependency to a project, add it to that project's `package.json`. Using your package manager's filter/workspace flag targets the right project, for example with pnpm:
 
-The dependency is then available for any of the TypeScript projects in your workspace to use.
+```bash
+pnpm add some-npm-package --filter @my-scope/my-library
+```
```
cogwirrel: This should show the command for all package managers, not just pnpm example.

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+
+To add a shared dev tool, add it to the workspace root:
+
+<InstallCommand pkg="-D some-dev-tool" />
```
cogwirrel: Make sure this includes the `-w` flag or equivalent if not already

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+
+This follows a [single version policy](https://nx.dev/concepts/decisions/dependency-management#single-version-policy), meaning all projects use the same version of each dependency — avoiding type conflicts and multiple copies of packages in deployment bundles. For package managers which support catalogs ([pnpm](https://pnpm.io/catalogs), [yarn](https://yarnpkg.com/features/catalogs) and [bun](https://bun.com/docs/install/catalogs)), generators record each version in the catalog and reference it with the `catalog:` protocol (eg `"zod": "catalog:"`), so the catalog is the single source of truth for versions regardless of which project declares the dependency. The vended `biome.json` enables Biome's [`experimentalPnpmCatalogs`](https://biomejs.dev/reference/configuration/#javascriptresolverexperimentalpnpmcatalogs) resolver so the lint rule understands `catalog:` references.
```
cogwirrel: Put under header "Catalogs" under Dependencies.

Reword:
``For package managers which support catalogs ([pnpm](https://pnpm.io/catalogs), [yarn](https://yarnpkg.com/features/catalogs) and [bun](https://bun.com/docs/install/catalogs)), generators record each version in the catalog and reference it with the `catalog:` protocol (eg `"zod": "catalog:"`), so the catalog is the single source of truth for versions regardless of which project declares the dependency. This helps you to follow a [single version policy](https://nx.dev/concepts/decisions/dependency-management#single-version-policy) avoid type conflicts and multiple copies of packages in deployment bundles.``

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+
+:::tip[Aligning versions on npm]
+npm (and yarn classic) has no catalog feature, so versions are declared directly in each `package.json`. To keep them consistent across projects, we recommend [syncpack](https://github.com/JamieMason/syncpack) (`npx syncpack lint`).
+:::
+
+:::note[bun]
+On bun, a few packages whose versions Nx generators read from the root manifest (`vite`, `react`, `react-dom`) keep direct version ranges instead of `catalog:` references — Nx does not yet resolve bun catalogs, and a `catalog:` reference would break those generators.
+:::
```
cogwirrel: Delete this

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+
+#### Opting out of catalogs
+
+Catalogs are enabled by default. To turn them off, create your workspace with `--catalog false`:
+
+```bash
+pnpm create @aws/nx-workspace my-project --catalog false
+```
```
cogwirrel: Create workspace command here for all package managers.

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+
+```ts title="aws-nx-plugin.config.mts"
+export default {
+  packageManager: {
+    catalogs: false,
+  },
+} satisfies AwsNxPluginConfig;
+```
+
+With catalogs disabled, generators write dependency versions directly into each project's `package.json`, and keeping versions aligned across projects is your responsibility (for example using [syncpack](https://github.com/JamieMason/syncpack)). When catalogs are enabled on pnpm, the workspace also sets [`catalogMode: strict`](https://pnpm.io/settings#catalogmode) so `pnpm add` records new dependencies in the catalog, keeping the single version policy without manual editing.
```
cogwirrel: Remove syncpack reference here.

Move ``When catalogs are enabled on pnpm, the workspace also sets [`catalogMode: strict`](https://pnpm.io/settings#catalogmode) so `pnpm add` records new dependencies in the catalog, keeping the single version policy without manual editing.`` up to a paragraph below the "For package managers which support catalogs..." paragraph.

Also we should set the equivalent of catalogMode strict for yarn and bun, or failing that consider the commands to run to install dependencies and put their version in the catalog by default.

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+
+When one project imports another project in your workspace, you don't need to declare it by hand: the TypeScript path aliases resolve the import for building, and the `ts#sync` generator (which runs before `compile`, or on demand via `nx sync`) declares the imported project as a workspace dependency in the importing project's `package.json` (`workspace:*` where the package manager supports the workspace protocol, `*` on npm and yarn classic). This keeps each project's manifest an accurate record of what it uses — satisfying the `noUndeclaredDependencies` lint rule — without manual bookkeeping.
```
cogwirrel: This should move up into "Importing your library code in other projects" and remove the part about the biome lint rule. It should be expanded and weaved in with the existing nx sync generator for project references - test in a real workspace and use real log messages etc for the example.

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -136,13 +137,69 @@ If you add custom `compilerOptions.paths` entries in a project's `tsconfig.json`
...
+    "rules": {
+      "correctness": {
+        "noUndeclaredDependencies": "off"
+      }
+    }
+  }
+}
+```
+
+Projects resolve root-installed packages at build time, so everything still builds — you give up per-project manifests as an accurate record of each project's dependencies (which matters if you later publish a project or split the repo).
```
cogwirrel, line 258: Let's make sure that the generators handle when a package.json doesn't exist for a project and add to the root instead (ie when not creating a project, don't add a package.json or fail if a package.json doesn't exist - add to the root if one doesn't exist).

docs/src/content/docs/en/guides/workspace.mdx
```diff
@@ -82,7 +82,9 @@ The default monorepo setup uses a [single version policy](https://nx.dev/concept
 
 This means that all projects within your monorepo use the same version of dependencies by default, reducing issues related to packages in the same monorepo running into version mismatch issues.
 
-From a Node perspective this means a single lockfile at the root with a single `node_modules` containing all dependencies. If you need to add a new dependency, add it in the root `package.json`.
+From a Node perspective this means a single lockfile at the root, with dependencies installed once and linked into each project. Each Node project declares the runtime dependencies its source imports in its own `package.json`, while shared build/test tooling lives in the root `package.json` `devDependencies`. Add a project's runtime dependency with your package manager's filter/workspace flag (eg `pnpm add <pkg> --filter <project>`), and shared dev tooling at the root.
```
cogwirrel: Again use InstallCommand or similar so that this is shown for each package manager.

docs/src/components/install-command.astro
```diff
@@ -1,8 +1,8 @@
 ---
 import PackageManagerCommand from './package-manager-command.astro';
 import { buildInstallCommand } from '../../../packages/nx-plugin/src/utils/commands';
-const { pkg, dev } = Astro.props;
+const { pkg, dev, project } = Astro.props;
 
-const buildCommand = (pm) => buildInstallCommand(pm, pkg, dev);
+const buildCommand = (pm) => buildInstallCommand(pm, pkg, dev, project);
```
cogwirrel, line 6: Can we render the install command in shorthand (scopeless) where supported by package managers, eg you can do `pnpm add foo --filter proj` when the project is `@some-scope/proj`

packages/nx-plugin/src/ts/lib/ts-project-utils.ts
```diff
@@ -42,6 +43,20 @@ export const configureTsProject = async (
...
+  // than leaving them to discover the gaps one failure at a time. Either
+  // init artefact counts as initialised (tests seed biome.json directly).
+  if (
+    !tree.exists(AWS_NX_PLUGIN_CONFIG_FILE_NAME) &&
+    !tree.exists('biome.json')
+  ) {
+    logger.warn(
+      `This workspace has no ${AWS_NX_PLUGIN_CONFIG_FILE_NAME} — run 'nx g @aws/nx-plugin:init' (or 'nx add @aws/nx-plugin') to configure linting, formatting and workspace dependency sync.`,
+    );
+  }
```
cogwirrel: No need to warn if biome is missing, just nx plugin config.

packages/nx-plugin/src/ts/lib/ts-project-utils.ts
```diff
@@ -90,14 +105,13 @@ export const configureTsProject = async (
       baseUrl: undefined,
       rootDir: '.',
       paths: {
-        // Remove any path aliases for this project with the npm scope prefix (eg remove @foo/bar)
+        // Remove any legacy colon-form alias for this project (eg :foo/bar)
```
cogwirrel: No need to remove legacy aliases

packages/nx-plugin/src/ts/sync/generator.ts
```diff
@@ -23,11 +26,20 @@ export const SYNC_GENERATOR_NAME = `${PackageJson.name}:${TS_SYNC_GENERATOR_INFO
 
 export const tsSyncGeneratorGenerator = async (
   tree: Tree,
-): Promise<SyncGeneratorResult> => {
+  // The sync runner invokes with only the tree; `nx g` passes an options
```
cogwirrel: When does this run via nx g?

packages/nx-plugin/src/utils/commands.ts
```diff
@@ -59,9 +59,36 @@ export const buildPackageManagerShortCommand = (
...
+ * bun's `add` has no project filter, so it needs the project's directory
+ * (`projectDir`) instead — falling back to the conventional
+ * `packages/<name>` when not provided.
  */
-export const buildInstallCommand = (pm: string, pkg: string, dev: boolean) => {
+export const buildInstallCommand = (
+  pm: string,
+  pkg: string,
+  dev: boolean,
+  project?: string,
```
cogwirrel, line 70: use shorthand where supported by package managers (ie no scope)

packages/nx-plugin/src/utils/dependencies.ts
```diff
@@ -0,0 +1,473 @@
...
+  tree.exists('pnpm-workspace.yaml') ? 'pnpm' : detectPackageManager(tree.root);
+
+/**
+ * Parse a version or simple range (an optional `^`/`~`/`>=` prefix followed by
+ * x.y.z and an optional prerelease suffix) into numeric parts. Returns
+ * undefined for anything else (tags, protocols, compound ranges) — the whole
+ * string must match, so ranges like `>=3 <5` or `^1 || ^2` are treated as
+ * user-customised and never overwritten.
+ */
+const parseVersionParts = (version: string): number[] | undefined => {
```
cogwirrel: can we use a dependency to simplify any version logic? eg does nx have helpers we can use?

packages/nx-plugin/src/utils/dependencies.ts
```diff
@@ -0,0 +1,473 @@
...
+  // Filesystem / shell build helpers
+  'ncp',
+  'rimraf',
+  'make-dir-cli',
+  'husky',
+  // Inspectors / dev harnesses
+  '@modelcontextprotocol/inspector',
+]);
+
+const isRootDevTooling = (packageName: string): boolean =>
```
cogwirrel: Specific generators should declare these, it shouldn't be in a single long list that pushes them up based on what the dep is. generators should call `addDependenciesToPackageJson` twice, once for project and once for root based on their needs.

packages/nx-plugin/src/utils/format.ts
```diff
@@ -29,6 +29,11 @@ export const DEFAULT_BIOME_CONFIG = {
       quoteStyle: 'single',
       trailingCommas: 'all',
     },
+    // Resolve dependencies declared via the package manager's catalog so
+    // `noUndeclaredDependencies` understands `catalog:` references.
+    resolver: {
+      experimentalPnpmCatalogs: true,
```
cogwirrel: what is behaviour of biome for yarn/bun? do we need this for those pacakge managers?

cogwirrel (dismissed):

cogwirrel (approved):

PR #955, opened 2026-07-17
fix(utils): kill formatter subprocess group on timeout to prevent hangs

nx-plugin-for-aws:
### Reason for this change

Follow-up to #951. That change moved Python/TS formatting to async `execFile` with a `timeout` and set `UV_LOCK_TIMEOUT`, which fixed the common case — but `pnpm build:all` could **still hang indefinitely** (duration counting up forever, only recoverable by killing the run).

Root cause of the residual hang: `execFile`'s `timeout` sends a single `SIGTERM` to the **direct** child. When that child is `uv`/`uvx` — which can spawn its own download/worker grandchildren, or may not exit promptly on `SIGTERM` — the child (or a grandchild holding the stdout pipe) survives, so the promisified call never settles and the event loop stalls. This was verified directly: a subprocess that ignores `SIGTERM` leaves the `execFile` promise pending forever.

### Description of changes

- Replace the `execFile`-with-`timeout` primitive with `spawn` in a dedicated **process group** (`detached: true`) and **`SIGKILL` the whole group** on timeout, so a wedged tool — or any grandchild it spawned — is reaped and the promise always settles.
- `unref()` the timeout timer so a pending deadline never keeps the process alive on its own.
- `execWithInput` now resolves with `{ stdout, stderr, code, timedOut }` instead of throwing on non-zero exit; callers adopt output only from a completed (non-killed) run — `ruff format`/biome require a clean exit, while `ruff check --fix` keeps its fixed stdout on the expected non-zero "unfixable issues" exit.
- Keep `UV_LOCK_TIMEOUT` (default 20s, honouring any user override) so a single ruff attempt rides out transient cache-lock contention and fails fast if the lock is genuinely stuck.

- Happy path: `format.spec.ts` 14/14, ~6s, run repeatedly (5/5) with no flakiness.
- Realistic transient contention (a concurrent `uv` holding the cache lock for 5–15s, cold cache): the single 20s lock wait outlasts the contention, ruff then installs and formats normally — 14/14, wall ~9–19s.
- Adversarial reap test: a subprocess that **ignores `SIGTERM`** and holds stdout open is killed by the process-group `SIGKILL` and the promise settles at the deadline (exit clean, zero leftover processes), whereas the `execFile`-`timeout` approach leaves the promise pending indefinitely.
- Biome format/lint clean (no new warnings) and `tsc --noEmit` passes.

### Issue # (if applicable)

Follow-up to #951.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/utils/format.ts
+++ b/packages/nx-plugin/src/utils/format.ts
@@ -5,58 +5,97 @@
 
 import { Biome } from '@biomejs/js-api/nodejs';
 import { getProjects, type Tree } from '@nx/devkit';
-import { execFile, execFileSync } from 'child_process';
+import { execFileSync, spawn } from 'child_process';
 import { existsSync, readFileSync } from 'fs';
 import { createRequire } from 'module';
 import path from 'path';
-import { promisify } from 'util';
 import { readToml } from './toml';
 
 const require = createRequire(import.meta.url);
-const execFileAsync = promisify(execFile);
 
 /** Hard ceiling on any formatter subprocess so a hung tool cannot stall generation. */
 const EXEC_TIMEOUT_MS = 60_000;
-/** Seconds uv waits for its global cache lock before giving up on a ruff call. */
+/**
+ * Seconds uv waits for its global cache lock before giving up. uv serialises
+ * cache writes behind a global lock (`<uv cache>/.lock`) and by default waits
+ * minutes for it. Capping the wait lets a single attempt ride out transient
+ * contention (another uv process installing ruff finishes in seconds) while
+ * failing fast if the lock is genuinely stuck, rather than stalling generation.
+ */
 const UV_LOCK_TIMEOUT_SECONDS = 20;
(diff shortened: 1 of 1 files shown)


nx-plugin-for-aws:
Superseded by #959, which takes a different approach.

The subprocess-group SIGKILL here (and the async/`UV_LOCK_TIMEOUT` work in the merged #951) hardened the formatter's subprocess handling, but the hang persisted because the real cause is contention on uv's global cache **write-lock**: on a cold cache the first `uvx ruff` call takes an exclusive lock while installing ruff, and ~19 Python-touching specs racing it across one-worker-per-core serialize the whole suite.

#959 warms the uv cache once in a vitest `globalSetup` before workers spawn — so every worker hits a warm cache (shared locks only, no storm) — and reverts the added subprocess complexity back to the simpler synchronous formatter. Verified: cold-cache baseline hangs (400s timeout) vs. 155s with the pre-warm.

PR #961, opened 2026-07-19
fix(rdb-constructs): enable PII-safe database logging by default

### LINKED ISSUE #918: rdb: enable database query logging by default with PII-safe configuration
Currently `ts#rdb` and `py#rdb` both default `enableCloudwatchLogs`/`enable_cloudwatch_logs` to `false` — CloudWatch log export for the Aurora engine logs is opt-in. Raised as a follow-up from PR #902's review (comment: https://github.com/awslabs/nx-plugin-for-aws/pull/902#discussion_r3555761129).

**Goal:** Enable logging by default, with best-practice configuration that gives visibility into e.g. executed DDL, without leaking PII.

**Current gap:** When `enable_cloudwatch_logs` is enabled for PostgreSQL, the Terraform module sets `log_statement = "all"` (`aurora.tf.template:187`), which logs every SQL statement including full bind parameter values — this is not PII-safe as-is. The CDK path exports `postgresql`/`audit,error,general,slowquery` logs but doesn't configure statement-level parameter groups at all.

**Scope:**
- Decide sensible defaults for both engines (Postgres/MySQL) that surface schema-level activity (DDL) without echoing query parameter values.
- Apply consistently across `ts#rdb` and `py#rdb` (shared `aurora.ts.template`/`aurora.tf.template`).
- Consider whether this should be a breaking change to the default, or gated behind a new option with the new behavior as default for new projects only.

AlexTo:
### Reason for this change

`ts#rdb`/`py#rdb` defaulted `enableCloudwatchLogs`/`enable_cloudwatch_logs` to `false` — CloudWatch log export was opt-in. When enabled, PostgreSQL set `log_statement=all`, logging every SQL statement including bind parameter values (a PII leak into CloudWatch Logs). MySQL exported `general` and `slowquery` log streams alongside `audit`, which capture full literal SQL text whenever `general_log`/`slow_query_log` are ever turned on, silently defeating the DDL-only intent.

### Description of changes

- Default `enableCloudwatchLogs`/`enable_cloudwatch_logs` to `true` for both `ts#rdb` and `py#rdb`, in CDK and Terraform.
- PostgreSQL: `log_statement=ddl` — surfaces schema changes (DDL) without logging statement parameter values. One caveat documented: a multi-statement batch mixing DDL with DML (e.g. a single `psql -c "a;b;c"` call) logs the whole batch verbatim, including any DML values in it.
- MySQL: enable Advanced Auditing scoped to `CONNECT,QUERY_DDL`, and trim the CloudWatch export list to `audit`/`error` only (`general`/`slowquery` log full statement text including DML values).
- Updated `ts#rdb`/`py#rdb` guide docs: split the shared logging snippet into engine-specific PostgreSQL/MySQL sections, surfaced via an engine filter dropdown on both guide pages (previously only `ts#rdb` had one, for an unrelated Prisma limitations section).

- Updated generator snapshot tests (`ts#rdb`/`py#rdb`) to reflect the new defaults and template output.
- Ran the full `@aws/nx-plugin` test suite (135 files, 2653 tests, all passing).
- Ran a full `docs` site build to confirm the new snippet structure and engine-filter wiring compile cleanly across all locales, with no broken internal links.
- End-to-end deployment test against real AWS infrastructure: generated both a MySQL and a PostgreSQL Aurora database via the generators, deployed both to a sandbox account with `enableCloudwatchLogs: true`, and ran a workload mixing DDL (`CREATE`/`ALTER`/`DROP TABLE`) with DML containing fake PII (email + SSN) against each.
  - **MySQL**: confirmed via `aws logs filter-log-events` on the `audit` log group that only DDL events were recorded — zero occurrences of the literal PII values anywhere in the log. Holds even when statements are sent as one semicolon-joined batch, since Advanced Auditing classifies per statement.
  - **PostgreSQL**: confirmed `log_statement=ddl` is safe when each statement is sent as its own call (e.g. as Prisma/Alembic-generated migrations always do), but running a DDL+DML batch as a single multi-statement `psql -c "a;b;c"` call does log the entire raw text verbatim, including the literal PII — reproduced this directly and documented it as the caveat above rather than leaving it theoretical.
  - Verified teardown (`cdk destroy`) removed the deployed stack and RDS clusters cleanly.

### Issue # (if applicable)

Closes #918.

### Checklist

----


cogwirrel (approved):
Thanks Alex! 🙌🙌

PR #968, opened 2026-07-21
test(e2e): stop bun and npm smoke lanes silently testing pnpm

nx-plugin-for-aws:
### Reason for this change

The bun smoke test lane has been silently creating **pnpm** workspaces while staying green, so bun support has not actually been exercised by CI. The npm lane was affected the same way.

The leak chain: the e2e suite itself runs under `pnpm nx`, and pnpm exports `npm_config_user_agent=pnpm/...`. The e2e `runCLI` helper spawned commands with the full inherited environment, and both bun and npm pass an inherited user agent through to child processes unchanged. `@aws/create-nx-workspace` detects the package manager from that user agent, so `bun create @aws/nx-workspace` resolved `--pm=pnpm` and produced a pnpm workspace. The yarn lanes were unaffected only because their corepack shims reset the user agent.

Two further e2e bugs compounded this and kept it invisible:
- The e2e-side package manager detection only looked for the legacy `bun.lockb` — bun ≥ 1.2 writes the text-based `bun.lock` — so even a genuine bun workspace would not have been detected as one.
- `runCLI` detected the package manager from the parent temp directory (which has no lockfiles) instead of the workspace under test, so every lane ran subsequent commands via `npx nx` regardless of package manager.

### Description of changes

- Scrub `npm_config_user_agent` from the `runCLI` spawn environment so each spawned package manager sets its own.
- Detect bun workspaces via `bun.lock` as well as legacy `bun.lockb`.
- Detect the package manager from the workspace under test (`opts.cwd`), so post-create commands run via the lane's package manager (`bun nx`, `npx nx`, etc.).
- Add `assertWorkspaceUsesPackageManager`, called after `createTestWorkspace` and in each `create-workspace` variant: the lane now fails loudly if the created workspace's lockfile doesn't match the package manager it claims to test, so a future detection regression can't silently fall back.
- Align vended/docs bun install guidance from `bun install -D <pkg>` to the idiomatic `bun add -D <pkg>` (bun aliases the former to the latter, but the `add` form is the documented one): `buildInstallCommand`, the MCP server general guidance, and the `ts#nx-plugin` vended MCP server templates (+ snapshots).

With the leak fixed, a genuine bun workspace passes the full smoke matrix — no bun-specific generator or build breakage had crept in.

- Verified the leak empirically: with an inherited `pnpm/...` user agent, `bun run`/`bun create` and `npm run` all propagate it unchanged; with it scrubbed, each sets its own.
- Ran the bun smoke test locally end-to-end against a Verdaccio-published local build: the workspace is created with `bun.lock` and bun's isolated `node_modules/.bun` layout, and the full generator matrix + `nx run-many --target build --all` passes (16m40s).
- Ran the npm smoke lane locally to validate the same scrub on the npm side (workspace created with `package-lock.json`).
- Full unit suite passes (2653 tests, including via the pre-commit hook).

### Checklist

----

### DIFF
--- a/e2e/src/utils.ts
+++ b/e2e/src/utils.ts
@@ -39,7 +39,7 @@ export async function runCLI(
   },
 ): Promise<string> {
   try {
-    const pm = getPackageManagerCommand();
+    const pm = getPackageManagerCommand({ path: opts.cwd || tmpProjPath() });
     const commandToRun = `${
       opts.prefixWithPackageManagerCmd !== false ? `${pm.runNxSilent} ` : ''
     }${command} ${opts.verbose ? ' --verbose' : ''}${
@@ -57,6 +57,14 @@ export async function runCLI(
           env: {
             PATH: process.env.PATH,
             ...process.env,
+            // The e2e suite itself runs under `pnpm nx`, which exports
+            // npm_config_user_agent=pnpm/... npm and bun treat an inherited
+            // user agent as config and pass it through to children, so
+            // user-agent-based package manager detection (e.g.
+            // @aws/create-nx-workspace resolving --pm) would see pnpm in
+            // every lane. Drop it so each spawned package manager sets its
+            // own.
+            npm_config_user_agent: undefined,
             ...opts.env,
           },
           shell: true,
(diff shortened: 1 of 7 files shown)


e2e/src/utils.ts
```diff
@@ -107,7 +115,9 @@ export async function runCLI(
 }
 
 function detectPackageManager(dir = ''): PackageManager {
```
cogwirrel, line 131: you just defined LOCKFILES below, use it here

nx-plugin-for-aws: Done in ce3bb094 — hoisted `LOCKFILES` above `detectPackageManager` and both it and the assertion now share a `hasLockfile` helper.

cogwirrel (approved):

nx-plugin-for-aws:
Note: the `rdb` and `terraform` lane failures on this PR are unrelated — new AL2023 glib2/libacl HIGH CVEs landed in the vended migration images' Trivy scans (also failing for the same reason on any branch). Fix: #970.

PR #973, opened 2026-07-21
fix(ts#agent): bundle @ag-ui/aws-strands peer deps into AG-UI runtime

nx-plugin-for-aws:
### Reason for this change

The `cdk-deploy` e2e smoke test on `main` fails: the TypeScript AG-UI agent (`MyTsAguiAgent`) AgentCore runtime returns **HTTP 424** ("An error occurred when starting the runtime"). Its CloudWatch runtime logs show the container crash-looping on startup with:

```
Error: Cannot find module '@ag-ui/a2ui-toolkit'
```

`@ag-ui/aws-strands` declares `@ag-ui/a2ui-toolkit` and `@ag-ui/client` as **peer dependencies** (its own `dependencies` is empty) but imports them **statically** in its compiled output. The `ts#agent` generator declared only `@ag-ui/aws-strands`, `@ag-ui/core`, and `@ag-ui/encoder`, so the two peers were never installed. Since the agent is bundled with rolldown (no externals) and the runtime image only ships the bundled `index.js`, rolldown could not resolve the missing peers at bundle time and left bare `require()` calls in the output — which crash the container at startup.

This surfaced as a nondeterministic green→red flip with no source change because there is no committed lockfile: the `cdk-deploy` and `terraform-deploy` jobs each install fresh, and whether the missing peer happened to be hoisted somewhere rolldown could resolve varied between installs.

### Description of changes

- Add `@ag-ui/a2ui-toolkit` to `TS_VERSIONS` (pinned `0.0.4`).
- Declare `@ag-ui/a2ui-toolkit` and `@ag-ui/client` in the AG-UI agent's dependencies in the `ts#agent` generator, so they are installed and deterministically inlined into the runtime image by rolldown.
- Add generator unit-test assertions for both new dependencies.

- `ts#agent` generator unit tests pass (65/65); full `@aws/nx-plugin` suite green (2646/2646).
- Reproduced the root cause with a controlled rolldown bundle: **without** the fix the bundle leaves bare `require("@ag-ui/a2ui-toolkit")` / `require("@ag-ui/client")` (matching the CloudWatch crash); **with** the fix all `@ag-ui/*` modules are inlined and only Node built-ins remain external.
- Ran the `cdk-deploy` e2e smoke test locally against the compiled packages, deploying to AWS and exercising the website integration test that surfaced the 424. _(result to be appended)_

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/ts/agent/generator.spec.ts
+++ b/packages/nx-plugin/src/ts/agent/generator.spec.ts
@@ -1045,6 +1045,8 @@ describe('ts#agent generator', () => {
 
     const rootPackageJson = JSON.parse(tree.read('package.json', 'utf-8'));
     expect(rootPackageJson.dependencies['@ag-ui/aws-strands']).toBeDefined();
+    expect(rootPackageJson.dependencies['@ag-ui/a2ui-toolkit']).toBeDefined();
+    expect(rootPackageJson.dependencies['@ag-ui/client']).toBeDefined();
     expect(rootPackageJson.dependencies['@ag-ui/core']).toBeDefined();
     expect(rootPackageJson.dependencies['@ag-ui/encoder']).toBeDefined();
     expect(rootPackageJson.dependencies['express']).toBeDefined();
--- a/packages/nx-plugin/src/ts/agent/generator.ts
+++ b/packages/nx-plugin/src/ts/agent/generator.ts
@@ -221,6 +221,11 @@ export const tsAgentGenerator = async (
         : protocol === 'ag-ui'
           ? ([
               '@ag-ui/aws-strands',
+              // @ag-ui/aws-strands declares these as peer dependencies but
+              // statically imports them, so they must be installed for the
+              // bundler to resolve and inline them into the runtime image.
+              '@ag-ui/a2ui-toolkit',
+              '@ag-ui/client',
               '@ag-ui/core',
               '@ag-ui/encoder',
               'express',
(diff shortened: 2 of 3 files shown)


cogwirrel:
## Targeted AG-UI deploy + invoke validation ✅

Since `cdk-deploy` doesn't run in PR CI, I validated the fix directly on AWS (nx-e2e account, us-west-2):

1. Generated a workspace with the locally-compiled plugin (this branch) and confirmed the generated TS AG-UI agent's `package.json` declares `@ag-ui/a2ui-toolkit` + `@ag-ui/client`, and its rolldown bundle inlines all `@ag-ui/*` (zero bare `require()`).
2. Deployed the TS + Python AG-UI agent runtimes (plus the `HostedMcpServer` the TS agent connects to) via CDK.
3. Invoked both over the AG-UI protocol (SigV4, same payload as the e2e `invokeAgentCoreAgUi` helper):

```
MyTsAguiAgent: HTTP 200, streamed RUN_STARTED → TEXT_MESSAGE_CONTENT, no RUN_ERROR  ✅
MyPyAguiAgent: HTTP 200  ✅
```

Runtime logs show `AG-UI server listening on 0.0.0.0:8080` with **no** `Cannot find module '@ag-ui/a2ui-toolkit'` crash (the pre-fix failure). Before this change the same runtime returned HTTP 424 with that module error in a crash loop.

Test stack torn down after validation.

cogwirrel (approved):

cogwirrel (approved):

PR #976, opened 2026-07-22
refactor(ts#dcr-proxy): reuse shared HttpApi construct and document fronting MCP servers/gateways

nx-plugin-for-aws:
### Reason for this change

Follow-up improvements to the `ts#dcr-proxy` generator (#939) covering a few nitpicks and a CDK/Terraform parity gap identified during review.

### Description of changes

- **schema**: set `default: "dcr-proxy"` on the `name` property rather than only describing it, and drop the redundant "(default: dcr-proxy)" hints from the description/prompt.
- **sdk**: export `tsDcrProxyGenerator` / `TsDcrProxyGeneratorSchema` from `src/sdk/ts.ts`.
- **CDK construct**: reuse the shared `HttpApi` core construct (generated into `common/constructs/src/core/api` alongside its `utils`, mirroring how the Terraform flavour already reuses the shared `http-api` module) with a hardcoded operations/integrations map for the proxy's seven routes, instead of hand-rolling a bare `apigwv2.HttpApi`. This brings the CDK flavour to parity with the Terraform flavour and the API generators: the API now has a KMS-encrypted access-log group and the Lambdas use their auto-created log groups, so the construct passes Checkov (CKV_AWS_95, CKV_AWS_158) without requiring the consumer to add stack-level suppressions.
- **docs**: add a **Fronting an MCP Server** section documenting how to front a `ts#mcp-server` / `py#mcp-server` generated with `--auth cognito`; rewrite **Fronting an AgentCore Gateway** to use a gateway generated with the `agentcore-gateway` generator with `--auth cognito` (now on `main`). Both examples use the non-hardcoded invocation/gateway URL exposed by the generated construct/module (`invocationUrl` / `gateway.gatewayUrl`; `invocation_url` / `gateway_url`). Add a tip on reusing the `UserIdentity` construct from `ts#website#auth` for the same user pool, including the required separate confidential App Client and the Managed Login branding needed for it.

- Added a CDK unit test asserting the shared `HttpApi` core construct + `utils` are generated and composed, and that no per-Lambda `LogGroup` is created; regenerated snapshots. Full `@aws/nx-plugin` build + test suite passes.
- Created fresh CDK and Terraform workspaces from the locally built plugin and deployed both to AWS. Verified the refactored CDK construct synthesizes and passes the vended `checkov` target with **zero** stack-level suppressions.
- Exercised the full OAuth Dynamic Client Registration flow end to end on both flavours (metadata discovery, `POST /register`, `/authorize` with the RFC 8707 `resource` parameter stripped, PKCE token exchange with server-side secret injection, `WWW-Authenticate` rewrite), then made an authenticated MCP `tools/call` proxied through to a Cognito-authenticated AgentCore runtime. Confirmed a real MCP client (Claude Code, `claude mcp add --transport http`) discovers the proxy and initiates authentication. Tore down all resources afterwards.

### Issue # (if applicable)

N/A — follow-up to #939.

### Checklist

----


docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -140,96 +140,248 @@ The module exposes the proxy endpoints (`proxy_url`, `mcp_url`, `metadata_url`,
 The `/mcp` proxy uses API Gateway HTTP API integrations, which have a hard 29 second timeout. Long-running upstream requests are buffered by the proxy rather than streamed.
 :::
 
-### Fronting an AgentCore Gateway
+:::tip[Reusing an existing User Pool]
```
cogwirrel, line 143: Shouldn't be a tip because this is long - make it a section in itself and add it at the bottom of of the docs page

docs/src/content/docs/en/guides/ts-dcr-proxy.mdx
```diff
@@ -140,96 +140,248 @@ The module exposes the proxy endpoints (`proxy_url`, `mcp_url`, `metadata_url`,
...
+  userPoolId: identity.userPool.userPoolId,
+  clientId: proxyClient.userPoolClientId,
+  useCognitoProvidedValues: true,
+});
+```
+
+Pass `identity.userPool.userPoolId`, `proxyClient.userPoolClientId` and `identity.userPoolDomain.baseUrl()` to the proxy. The `UserIdentity` construct always creates a domain, so you do not need to add one.
+</Fragment>
+<Fragment slot="terraform">
+```hcl
```
cogwirrel: Include instantiating the user identity module here

packages/nx-plugin/src/utils/dcr-proxy-constructs/files/cdk/app/dcr-proxies/__nameKebabCase__/__nameKebabCase__.ts.template
cogwirrel: Let's get rid of these parameters, they're easy for users to change after generation.

cogwirrel (approved):

PR #978, opened 2026-07-22
fix(ts#website#auth): add custom domain names to cognito callback/logout urls

AlexTo:
### Reason for this change

#948 added `domainNames`/`certificate` support to `StaticWebsite`, letting a CloudFront distribution serve a custom domain instead of only the default `*.cloudfront.net` domain. However, `UserIdentity` (the Cognito construct/module used by `ts#website#auth`) never learned about these custom domains: it only ever added the CloudFront default domain name as a valid OAuth callback/logout URL.

As a result, a user who configures a custom domain per the "Custom Domain & TLS" docs and then signs in via that domain hits Cognito's `redirect_mismatch` error, since the custom domain is never in the User Pool Client's allowed callback/logout URL list. Nothing in the code or docs calls this out as a manual step.

### Description of changes

- **CDK** (`user-identity.ts.template`): `findCloudFrontDistributions` is replaced with `findCloudFrontDomainNames`, which reads each CloudFront distribution's default domain name plus its configured `aliases` (custom domain names) straight off the underlying `CfnDistribution.distributionConfig` (CDK's `Distribution` L2 writes `aliases: props.domainNames` there directly), so no changes to `StaticWebsite` itself were needed.
- **Terraform** (`identity-constructs.ts`): the auto-inserted `add_callback_url` module block is now generated with a `locals.callback_urls` set (CloudFront domain + `for domain in var.custom_domain_names`) and a `for_each` over that set, instead of a single hardcoded `callback_url`.
- Updated the "Remove localhost callback URLs for production" doc note (and translated locales) to mention custom domain names are included automatically.

- Updated/ran the existing `cognito-auth` generator unit tests (CDK + Terraform) and refreshed the snapshots that capture the generated `user-identity.ts` / `static-website.tf` content to reflect the new behavior.
- All tests in `packages/nx-plugin/src/ts/react-website/cognito-auth/generator*.spec.ts` pass.
- Deployed a real distribution with a custom domain and ACM certificate and confirmed the custom domain is added to the User Pool Client's callback and logout URLs.

### Issue # (if applicable)

N/A — found during review, no tracked issue.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/react-website-auth.mdx
+++ b/docs/src/content/docs/en/guides/react-website-auth.mdx
@@ -195,7 +195,7 @@ The user identity module automatically adds the necessary <Link path="guides/rea
 </Infrastructure>
 
 :::caution[Remove localhost callback URLs for production]
-The generated User Pool client allows your CloudFront distribution URL as an OAuth callback/logout URL, plus `http://localhost:4200` and `http://localhost:4300` for local development against the deployed pool.
+The generated User Pool client allows your CloudFront distribution URL (and any custom domain names configured on it) as OAuth callback/logout URLs, plus `http://localhost:4200` and `http://localhost:4300` for local development against the deployed pool.
 
 It is recommended to **remove the `http://localhost` callback/logout URLs for production stages**, keeping the allowlist limited to your real application origins.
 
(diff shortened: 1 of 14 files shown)


cogwirrel (approved):
Great catch, thank you!

PR #997, opened 2026-07-27
feat(smithy#project): support smithy shape libraries

### LINKED ISSUE #304: Support for smithy shape libraries
### Description

Add support for creating Smithy shape libraries, ie model projects without an associated backend.

Please +1 this if it's important to you.

### Motivation

Some users may wish to share models between multiple Smithy APIs

### Suggested Implementation

Generify and expose `smithy#project` 

### Contribute to feature

ArushCreater:
### Reason for this change

Smithy models could only be generated as part of a `ts#api --framework=smithy` project, always
paired with a TypeScript implementation. There was no way to define shapes independently, so users
sharing types across multiple Smithy APIs had to duplicate the definitions in each model and keep
them in sync by hand.

### Description of changes

**Generify and expose `smithy#project`** — adds a `--type` option (`service` | `shapes`), defaulting
to `service` so existing behaviour is unchanged, and unhides the generator. Templates are split into
`files/service/` and `files/shapes/`. A shape library defines no service, so its `smithy-build.json`
configures neither the `openapi` nor the `typescript-ssdk-codegen` plugin — both require a service
shape. Its build validates the model and emits the assembled `model.json`.

**Consuming shape libraries** — adds a `smithy#model-connection` generator, wired into the
`connection` generator as `ts#smithy-api -> smithy#project` and `smithy#project -> smithy#project`.
When the source is an API, the connection resolves to its model project, which is where the shapes
are used.

The Smithy CLI runs inside a Docker build whose context is the project directory, so a sibling
project's sources aren't visible to it. Each dependency's sources are therefore copied into the
consuming project's `.smithy-deps` directory by a `copy-model-deps` target, mirroring the existing
`copy-ssdk` pattern; `compile` depends on it so shapes are in place before the model builds.
Transitive dependencies are resolved into a flat, name-keyed set, so a library reached by more than
one path is copied once and never defines its shapes twice.

Two changes worth flagging for review:

- `isSmithyApi` previously classified *any* `smithy#project` as `ts#smithy-api`. It now also requires
  `backendProject` metadata, so a standalone Smithy project resolves to the new `smithy#project`
  type. The existing test fixture omitted `backendProject`, which `ts#smithy-api` always sets — I
  made the fixture faithful rather than weaken the check.
- `smithy#project` templates moved from `OverwriteStrategy.Overwrite` to `KeepExisting`. A re-run
  previously clobbered hand-edited `.smithy` files, which mattered less while the generator was
  internal-only.

Unit tests for both generators, including the idempotency cases required by CONTRIBUTING: re-running
`smithy#project` preserves user edits; re-running the connection leaves `project.json`, the
Dockerfile, `smithy-build.json` and `.gitignore` byte-identical; a second differently-named library
is additive; and self-connections and cycles are rejected. Transitive and diamond dependency
resolution are covered explicitly.

All 2291 non-Python unit tests pass, along with `tsc --noEmit` and Biome lint/format. (The Python
specs were not run — `uv` and `ruff` are not installed in my environment.)

Since the service-less build was the load-bearing assumption, I verified the design against Smithy
CLI 1.61.0 directly rather than relying on it:

- a service-less build with no codegen plugins succeeds and emits
  `build/smithy/source/model/model.json`, the exact path the shapes Dockerfile copies from
- a consuming model with `.smithy-deps/` in `sources` resolves cross-namespace shapes, and the
  shared shape appears in the generated OpenAPI
- a diamond dependency (`model -> {a, b} -> common`) builds with no duplicate-shape error

**Not validated locally:** the e2e matrix additions and the docs build. No container engine was
available in my environment, so no Smithy build could actually run, and I did not run `docs:build`.
Both rely on CI.

### Issue # (if applicable)

Closes #304.

### Checklist

----


docs/astro.config.mjs
```diff
@@ -393,6 +393,10 @@ export default defineConfig({
                       label: 'Smithy',
                       link: '/guides/ts-smithy-api',
                     },
+                    {
+                      label: 'smithy#project',
+                      link: '/guides/smithy-project',
+                    },
```
cogwirrel: Even though this is a separate public generator, it's very much tied to the Smithy API so I think it's better to remove from the menu and add a section to `ts-smithy-api.mdx` like `Adding a shape library` which introduces what it's for and then links to the smithy project guide.

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Smithy projects build using [Docker](https://www.docker.com/), which runs the [Smithy CLI](https://smithy.io/2.0/guides/smithy-cli/index.html) to validate your model:
+
+<NxCommands commands={['build my-shapes']} />
+
+## Sharing Shapes Between Projects
+
+To use a shape library's shapes in another Smithy project, connect the two with the <Link path="guides/connection/smithy-shapes">`connection` generator</Link>:
+
+<RunGenerator generator="connection" requiredParameters={{ sourceProject: 'my-api', targetProject: 'my-shapes' }} />
```
cogwirrel: I think it's probably overkill to have a connection generator for this, and the intention of the connection generator is more for different services/systems interacting and writing the integration code, rather than language dependencies :)

I think we should just document how to set up the dependencies between Smithy projects here (I think this would be using `"imports"` in `smithy-build.json`)

packages/nx-plugin/src/smithy/model-connection/generator.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  if (modelProjectConfig.name === libraryProjectConfig.name) {
+    throw new Error(
+      `Cannot connect Smithy project ${modelProjectConfig.name} to itself.`,
+    );
+  }
+
+  // A shape library's own dependencies are recorded transitively, so the
+  // library's set plus the library itself is the full set to add
+  const libraryDependencies = readSmithyDependencies(libraryProjectConfig);
```
cogwirrel: (I think we should get rid of the model connection generator, so this comment is more about the process we document for adding dependencies between smithy projects)

Is there a way to build/package a Smithy shape library into a form which includes its dependencies, such that we don't need to add dependencies on its transitive dependencies when consuming it?

I remember you could do it with the Smithy gradle plugin and build a jar, but I'm not sure if it's possible with the Smithy CLI. Adding a dependency on gradle/java etc would definitely be too much, so would need to be a CLI supported method :)

packages/nx-plugin/src/smithy/model-connection/generator.ts
```diff
@@ -0,0 +1,201 @@
...
+  // Projects generated by smithy#project since shape library support was added
+  // need no changes here, but older ones do.
+  ensureSmithyDepsSupport(tree, modelProjectConfig.root);
+
+  const cmd = new FsCommands(tree);
+
+  // Smithy models are resolved by the Smithy CLI inside a Docker build whose
+  // context is the project directory, so each dependency's sources are copied
+  // in before the model is built.
+  modelProjectConfig.targets[COPY_MODEL_DEPS_TARGET] = {
```
cogwirrel: (Again a comment for documenting adding dependencies between smithy projects)

Can we do Smithy docker builds with `--build-context workspace={workspaceRoot}` to avoid having to add a copy target? I think the Dockerfile of the consumer could then reference the built output of the shape library dist from workspace root :)

cogwirrel (commented):
Thanks for the contribution @ArushCreater! Added a few comments, mostly around simplifying to remove the `connection` generator :)

ArushCreater:
Thanks for the review, agreed on dropping the connection generator. i think you're right with that shape library is a compile-time dependency rather than two systems integrating, so it doesn't belong there. I'll remove it.

I looked into the other 2 questions with the Smithy CLI (1.61.0), and the answers change the design quite a bit.

**Packaging a library with its dependencies**

Yes, because smithy build already does it. The build/smithy/source/model/model.json artifact contains the library's own shapes plus anything it depends on, as long as those deps are in sources:

```

# shapes-a depends on shapes-common
a's model.json -> ['com.example.a#Invoice', 'com.example.common#Money']

```
So a consumer only ever declares its direct dependencies, and all the transitive closure logic in this PR disappears.

I'd assumed the diamond case would break this (two libraries each bundling a shared third, defining its shapes twice). but smithy reports Model.IgnoredDuplicateDefinition, "ignoring duplicate but equivalent shape definition", and carries on:


```
api -> {shapes-a, shapes-b}, both bundling common#Money
SUCCESS: Validated 332 shapes (NOTE: 81)
openapi schemas: ['GetResponseContent', 'Invoice', 'Money', 'Refund']
```
Money resolves once in the output. That assumption is what drove the copy based approach in this PR, so it was worth checking.

This showed that those 81 notes are maven trait shapes getting baked into every packaged model. The shape library template I wrote over declares its maven deps. With smithy-model, smithy-aws-traits and smithy-validation-model the packaged model.json carries 41 shapes (aws.api, aws.auth, aws.protocols, smithy.openapi and friends). With just smithy-model it carries exactly 1, the library's own shape. I'll default shape libraries to the minimal set.

`--build-context workspace={workspaceRoot}`

Makes sense, and it composes well with the above. The Dockerfile can do COPY --from=workspace dist/packages/<lib>/build/model/model.json deps/, so the copy-model-deps target goes away entirely. The only thing I want to confirm first is Finch: --build-context is BuildKit specific and resolveContainers supports finch as well as docker. Do you know offhand whether Finch handles it?

another thing is that you suggested imports in smithy-build.json, but imports excludes those shapes from the built projection, so a library consumed via imports would package only its own shapes and consumers would be back to declaring transitive deps by hand. sources is what makes the self contained packaging above work.

Is sources OK with you, or do you specifically want imports semantics (the dev's own projection staying limited to its own shapes) and are happy for consumers to declare transitive deps explicitly?

**Docs**

Will do. Dropping smithy#project from the sidebar and adding an "Adding a shape library" section to ts-smithy-api.mdx that introduces the concept and links through, with the dependency wiring documented there rather than generated.

cogwirrel:
Thanks for digging into this @ArushCreater! Great news on the packaged `model.json` including transitive deps :)

I did a bit of testing with `imports` and it seems like `build/smithy/source/model/model.json` comes out the same as with `sources`, with transitive dependencies included :) So I reckon let's go for `imports` unless you hit something I missed :)

On Finch - `--build-context` works since version 1.6.0 - might be worth updating the prerequisite note for finch to add `(>=1.6.0)`

Appreciate your work on this, thank you!

ArushCreater:
thank you so much for the help! i have pushed the rework. Connection generator and all its wiring are gone (connection/ is byte-identical to main again), dependencies now go through imports, Smithy builds get --build-context workspace={workspaceRoot}, and shape libraries are trimmed to smithy-model only. Docs moved into smithy-project.mdx with the intro section in ts-smithy-api.mdx, and Finch is now noted as >= 1.6.0.

Confirmed your imports finding, you were right that the packaged model carries transitive deps either way.

Also fixed the one CI failure from my first push: unhiding smithy#project added it to the standalone leg, which builds every public generator, and the CodeBuild Windows runner has no Docker. Added it to CODEBUILD_WINDOWS_UNSUPPORTED alongside the other Docker-building generators.

Workflows are showing action_required so CI hasn't run on this yet tho. But let me know what you think :)

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+<NxCommands commands={['build my-shapes']} />
+
+## Depending on a Shape Library
+
+Building a shape library writes an assembled model to `dist/<my-shapes>/build/model/model.json`. This single file contains every shape the library defines, along with any it depends on, so a consumer only ever declares the libraries it references directly.
+
+To depend on a shape library from another Smithy project, make three changes to the consuming project.
+
+### 1. Copy the built model into the Docker build
```
cogwirrel: Instead of headings, can we use `<Steps>`? :)

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -0,0 +1,183 @@
...
+  "version": "1.0",
+  "sources": ["src/"],
+  "imports": ["deps/my-shapes.json"],
+  ...
+}
+```
+
+:::note
+`imports` rather than `sources` keeps the library's shapes out of your project's own source projection, while still resolving them and carrying them through to your build output.
+:::
```
cogwirrel: Nit: let's delete this :)

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -0,0 +1,183 @@
...
+  "targets": {
+    "compile": {
+      "dependsOn": ["@my-scope/my-shapes:build"],
+      ...
+    }
+  }
+}
+```
+
+### Using the shapes
```
cogwirrel: Can remove this heading once the above are `<Steps>` :)

cogwirrel (commented):
Looks great, thank you @ArushCreater! Just 3 minor comments for the docs :)

ArushCreater:
all done @cogwirrel :)

cogwirrel:
> all done @cogwirrel :)

Thank you! Looks great! :) Only thing is it looks like the `—build-context workspace={workspaceRoot}` is causing issues, might need to swap it for `workspace=.`?

ArushCreater:
ye you're right, thanks, i pushed the changes. Root cause was Nx only accepting {workspaceRoot} at the start of an interpolated expression and stripping it elsewhere, so the token never resolved mid-command

cogwirrel (approved):

cogwirrel:
Thanks very much @ArushCreater ! 🚀🎉

PR #1001, opened 2026-07-28
feat(nx-plugin): sync vended versions on nx migrate

### LINKED ISSUE #988: feat: migrate vended dependency versions on nx migrate (evergreen version-sync migration)
### Description

The weekly `update-versions` workflow (`.github/workflows/update-versions.yml`) bumps every vended dependency version the plugin owns — `TS_VERSIONS`, `PY_VERSIONS`, `TERRAFORM_VERSIONS`, `VENDORED_VERSIONS`, `CONTAINER_VERSIONS`/`BASE_IMAGES`, plus refreshed shadcn templates — and raises a `feat: update dependencies` PR. Those bumps reach **newly generated** projects only. An existing workspace stays on whatever versions it was generated with, indefinitely, even after `nx migrate @aws/nx-plugin`.

This is the single largest gap in the v1.0 upgrade story (#718): the overwhelmingly common thing that changes between two releases of this plugin is vended dependency versions, and today none of it migrates.

**Ask:** whenever the version update workflow runs, the corresponding version bumps must also be applied to users' existing workspaces on `nx migrate` — for everything the workflow covers.

### Motivation

- A user who generated a workspace in January and runs `nx migrate @aws/nx-plugin` in July gets new *generator* behaviour but keeps January's `aws-cdk-lib`, `fastapi`, and `hashicorp/aws` provider. They silently miss six months of security patches in dependencies **we chose for them**.
- Several `update-versions` bumps exist specifically to clear CVEs (e.g. the `@opentelemetry/propagator-jaeger` pin above the ADOT transitive version, for CVE-2026-59892 HIGH). Those never reach existing workspaces.
- Manually authoring a migration per weekly bump PR is not viable: the workflow is fully automated, and the bump set differs every week.

### Research: where vended versions actually land

Verified against a generated pnpm workspace containing `ts#trpc-api`, `py#fast-api`, `terraform#project` and shared constructs (plugin built from the #928 branch):

| Location | Example | Notes |
|---|---|---|
| `pnpm-workspace.yaml` `catalog:` | `zod: 4.4.3`, `'@trpc/server': 11.18.0` | **Primary** location for TS deps — 36 entries in a small workspace |
| Root `package.json` | `"@aws-sdk/client-s3": "catalog:"` | Mostly `catalog:` references, not versions |
| Per-project `package.json` | `"zod": "catalog:"` | `catalog:` when catalogs are enabled; **direct pinned ranges** when not (npm, or `packageManager.catalogs: false`) |
| Project `pyproject.toml` | `"fastapi==0.139.2"` | Exact `==` pins in `[project].dependencies` |
| Root `pyproject.toml` | `"ruff==0.15.22"`, `"ty==0.0.61"` | `[dependency-groups].dev` |
| `.tf` files | `version = "6.55.0"` under `source = "hashicorp/aws"` | Both `packages/<app>/src/providers.tf` **and** every shared construct `.tf` (`packages/common/terraform/**`) — ~8 files in a small workspace |
| Vendored/other | `.git-secrets/git-secrets`, Trivy version + base images in build targets, shadcn templates | Covered by the workflow, not by dependency manifests |

Catalog entries carry a further wrinkle: `packages/nx-plugin/src/utils/dependencies.ts` deliberately keeps `vite`/`react`/`react-dom` as direct ranges on bun (Nx ships no bun `CatalogManager`), so a real solution must handle both catalog and direct forms.

### Research: why Nx's declarative `packageJsonUpdates` is not sufficient

Nx's built-in declarative mechanism is the obvious first candidate — and it does not work here. Tested by publishing purpose-built plugin versions to a local verdaccio and running the real `nx migrate` flow:

**Test 1 — plain `packageJsonUpdates`** (`0.902.0`, entries for `zod`, `@trpc/server`, `@aws-sdk/client-s3`):

```
- '@aws-sdk/client-s3': 3.1000.0     →  3.1090.0   ✅ upgraded
  '@trpc/server': 11.10.0            →  unchanged  ❌
  zod: 4.3.0                         →  unchanged  ❌
```

Nx's `shouldApplyPackageUpdate` gates on the dependency being present in the **root** `package.json` (`nx/src/command-line/migrate/migrate.js`). `@aws-sdk/client-s3` was upgraded only because the root manifest happens to carry a `catalog:` reference for it, which Nx's catalog manager then redirects into `pnpm-workspace.yaml`. `zod` and `@trpc/server` are catalog + project-manifest only, so Nx never sees them and skips them **silently**.

**Test 2 — `alwaysAddToPackageJson: true`** (`0.903.0`) makes it actively worse:

```
root package.json:      "zod": "4.4.3"     ← newly injected direct dep
pnpm-workspace.yaml:    zod: 4.3.0         ← stale catalog entry, untouched
```

That is the duplicate-version split that caused the react `Invalid hook call` regression fixed in #986. Also note `packageJsonUpdates` is inherently per-release data — it would have to be generated and stamped for every weekly bump.

Conclusion: `packageJsonUpdates` can stay as a complement for genuinely root-level packages, but it cannot be the mechanism. Python pins and Terraform provider versions are out of its scope entirely.

### Proposed design: one evergreen deterministic migration, re-stamped every release

Add a **single** generic migration to the codebase — not one per bump — registered once in `migrations.json`, whose implementation reads the plugin's *own* `versions.ts` at run time:

```jsonc
// packages/nx-plugin/migrations.json
{
  "generators": {
    "sync-vended-versions": {
      "description": "Sync vended dependency versions to those vended by this release",
      "implementation": "./src/migrations/sync-vended-versions/migration"
    }
  }
}
```

The migration runs inside the **newly installed** plugin, so `TS_VERSIONS` / `PY_VERSIONS` / `TERRAFORM_VERSIONS` are already the target versions. No per-release data, no codegen, nothing for the `update-versions` workflow to author — the weekly PR keeps touching only `versions.ts`, and the migration follows automatically.

**Evergreen stamping.** The one piece of new infrastructure needed on top of #928: this entry must be re-stamped on *every* release rather than pinned to the release that introduced it. #928's `scripts/stamp-migrations.ts` stamps a shipped migration with the earliest tag that registered it — correct for one-shot migrations, wrong here (it would run once and never again). Proposal: mark the entry as evergreen (e.g. an `x-evergreen: true` flag stripped at package time) and have `stamp-migrations.ts` always stamp it with `semver.inc(latestTag, 'prerelease')` — the same "strictly above latest release, strictly below any possible next release" trick the script already uses for unshipped migrations (verified: `0.904.3` → `0.904.4-0`, `1.0.0-rc.45` → `1.0.0-rc.46`). Since `nx migrate` runs a migration when `installed < version <= target`, that guarantees it runs on every upgrade hop regardless of what the next release number turns out to be.

**Scope of the implementation.** Per surface, mechanically:

- **TS** — walk `pnpm-workspace.yaml` `catalog:`/`catalogs:` (and `.yarnrc.yml` / bun's `package.json` `catalog` equivalents), plus direct ranges in every `package.json`, upgrading any entry whose name appears in `TS_VERSIONS`.
- **Python** — rewrite `name==version` / `name[extras]==version` pins in every `pyproject.toml` (project deps *and* root `[dependency-groups]`) against `PY_VERSIONS`.
- **Terraform** — rewrite `version = "…"` inside `required_providers` entries matching `source = "hashicorp/<provider>"` against `TERRAFORM_VERSIONS`, across app and shared-construct `.tf` files.

**Guardrails** (matching the deterministic-migration rules in #928's CONTRIBUTING section):

- **Upgrade-only, exact-only.** Rewrite only when both existing and vended versions parse and vended is strictly greater. A version the user raised above vended, or replaced with a range/tag/protocol specifier, is left alone. This mirrors `isVersionUpgrade` in `utils/dependencies.ts`, so migration and generators agree.
- **Idempotent.** Re-running changes nothing.
- **Never a downgrade**, matching `nx migrate`'s forward-only contract.

**Prototype status.** Implemented on top of #928 and validated end-to-end (branch `experiment/version-sync-migration`, not yet a PR). Published `0.900.0` → `0.901.0` to a local verdaccio and ran the real user-facing flow (`nx migrate @aws/nx-plugin@0.901.0` → `pnpm install` → `nx migrate --run-migrations`) against a workspace with deliberately downgraded versions:

```
  pnpm-workspace.yaml       zod 4.3.0 → 4.4.3, '@trpc/server' 11.10.0 → 11.18.0
  py_api/pyproject.toml     fastapi==0.130.0 → ==0.139.2
  tf-infra/src/providers.tf hashicorp/aws 6.40.0 → 6.55.0
  common/terraform/**.tf     hashicorp/aws 6.40.0 → 6.55.0, random 3.5.0 → 3.9.0
  demo-api/package.json      aws4fetch 1.0.10 → 1.0.20 (direct-range path)
```

Guardrail held: a catalog entry manually raised to `vitest: 5.0.0-beta.7` (above the vended `4.1.10`) was left untouched. Re-running the migration produced a clean `git status` (idempotent). `nx run-many --target build --all` passed afterwards — 4 projects, 18 tasks, checkov 18/0.

`nextSteps` is used to surface the lock-file follow-ups the migration deliberately does not perform (`uv sync`, `terraform init -upgrade`); the JS install is already run by `nx migrate` itself.

### Open questions

1. **Lock files.** The migration edits manifests only. `pnpm-lock.yaml` is refreshed by the install `nx migrate` already runs; `uv.lock` and `.terraform.lock.hcl` are not. Emit as `nextSteps` (prototype's choice), or have the migration shell out via a post-migration callback?
2. **Non-manifest surfaces.** `VENDORED_VERSIONS` (the vendored `git-secrets` script), `CONTAINER_VERSIONS` (Trivy), `BASE_IMAGES` and refreshed shadcn templates are all covered by `update-versions` but aren't dependency pins — they're vended file content. These are file-replacement migrations with a "did the user edit it?" question attached, and are probably better handled as a separate follow-up (possibly hybrid, per #703's taxonomy) rather than folded into this one. Confirm the split.
3. **Major bumps.** This migration only moves version numbers. When a bump needs accompanying code changes, that remains a hand-written migration per #703 — worth stating explicitly in the contributing guide so nobody assumes this covers major upgrades.
4. **Guardrail reporting.** Should skipped entries (user diverged) be reported individually via `nextSteps`, or is silence-plus-summary enough? Verbose reporting on a workspace with many customisations could be noisy.

### Acceptance Criteria

- [ ] A single generic deterministic migration syncs TS (catalog + direct), Python (`pyproject.toml`), and Terraform provider versions from the installed plugin's `versions.ts`
- [ ] Migration is re-stamped every release (evergreen) so it applies on every upgrade hop, with unit coverage on the stamping logic
- [ ] Upgrade-only / exact-only / idempotent guardrails enforced and unit-tested, including the user-raised-version case
- [ ] Non-pnpm catalog forms (yarn `.yarnrc.yml`, bun `package.json`) and the catalogs-disabled direct-range path covered
- [ ] Lock-file follow-ups resolved per open question 1
- [ ] `update-versions` workflow requires no per-bump migration authoring — verified by running the workflow's script and confirming no `migrations.json` change is needed
- [ ] E2E coverage via the #705 harness: workspace generated at version N, migrated to N+1, vended versions match the new release, build + lint + test green
- [ ] Contributing guide states that version-only bumps are covered automatically and code-change-bearing bumps still need their own migration

### Related

- #718 (v1.0 roadmap — Migrations workstream)
- #703 (migrations strategy — this is the "dependency update strategy" section made concrete)
- #928 (migrations foundations PR — provides `migrations.json` wiring and version stamping this builds on)
- #704 (scaffold generator — used to scaffold this migration)
- #705 (e2e rolling-migration harness — provides the acceptance test)

nx-plugin-for-aws:
### Reason for this change

The weekly `update-versions` workflow bumps every vended dependency version the plugin owns — `TS_VERSIONS`, `PY_VERSIONS`, `TERRAFORM_VERSIONS` — but those bumps only reach **newly generated** projects. An existing workspace stays on whatever versions it was generated with, indefinitely, even after `nx migrate @aws/nx-plugin`. Several of those bumps exist specifically to clear CVEs, so users silently miss security patches in dependencies the plugin chose for them.

There is also a latent split around the nx packages. `TS_VERSIONS` carried `@nx/devkit` and `@nx/react`, while `nx`, `@nx/js` and `@nx/workspace` were pinned separately in `utils/init.ts` from `readModulePackageJson('@nx/js').version`, and `commands.ts` derived `NX_VERSION` from `@nx/devkit`. Today those move together by coincidence, so it isn't visible — but any migration syncing `@nx/devkit` while leaving `nx` behind creates exactly the mismatch `init.ts` warns about: *"A mismatched workspace nx (even a patch apart) hoists a second nested nx under @nx/js, and the two instances deadlock `nx sync`."*

Finally, a version sync must not touch dependencies the plugin didn't put there. If a user has their own `zod` and has never run a generator that adds `zod`, that pin is theirs.

### Description of changes

**One migration, run on every upgrade.** `sync-vended-versions` is committed once in `migrations.json` marked `everyMigration: true` (`utils/version-upgrade-migration/`). Rather than generating a fresh migration per release, `stampMigrationVersions` re-stamps this single entry with each pending version, so it runs on every `nx migrate` hop and always syncs to the versions the *newly installed* plugin vends. Nothing is hand-written or generated per release.

It runs **last**. `nx migrate` sorts the run ascending by version and preserves manifest order among equal versions, so `stampMigrationVersions` emits every-migration entries after the release's own code migrations — a code migration that rewrites a construct lands before versions move underneath it. The backfill never pins or moves this entry, which would stop it running again.

Surfaces covered by `syncVendedVersions`:

- **TypeScript** — via the same `addDependenciesToPackageJson` devkit path the generators use, so pnpm/yarn/bun catalogs are handled by nx's own catalog manager rather than reimplemented. Orphaned catalog entries are pruned.
- **Python** — `==` pins in every `pyproject.toml`, including `[extras]` and `[dependency-groups]`
- **Terraform** — `required_providers` versions across app and shared-construct `.tf` files
- **Metrics** — the tracked plugin version in both the CDK `MetricsAspect` and Terraform `metrics.tf`, via the same GritQL patterns `addGeneratorMetricsIfApplicable` uses, leaving metric id and tags alone

**Generators declare what they own, once, and that declaration adds them.** `declareDependencies` (`utils/declared-dependencies.ts`) has each generator export a `DEPENDENCIES` list beside itself — no central registry. Each entry carries an optional `when` predicate over the metadata the generator records, and `addTsDependencies` / `addPyDependencies` add exactly the applicable subset, so the call site names no packages:

```ts
/** The metadata this generator records, which its predicates read. */
export interface TsAgentMetadata {
  readonly port: number;
  readonly auth: string;
  readonly protocol: string;
}

export const DEPENDENCIES = declareDependencies<TsAgentMetadata>()({
  ts: [
    { name: 'zod' },
    { name: '@a2a-js/sdk', when: (m) => m.protocol === 'a2a' },
    { name: '@types/ws', when: (m) => m.protocol === 'http', dev: true },
    { name: 'tsx', dev: true, root: true },
    ...SHARED_CONSTRUCTS_DEPENDENCIES,
  ],
});

addTsDependencies(tree, DEPENDENCIES, { metadata, projectRoot: project.root });
```

Flags cover the rest: `dev`, `root` (workspace manifest), `group` (pyproject dependency group), and `versionOnly` for a package this generator owns but installs elsewhere — `ownedElsewhere(...)` marks a helper's spread constant that way, and `onlyWhen(..., predicate)` first narrows it to the branch that reaches the helper. Both keep the package owned, so the sync still upgrades it wherever it sits; only the install is skipped. Both `addTsDependencies` and `addPyDependencies` take the same `{ metadata, projectRoot }` options, so the two calls a generator makes read alike. `withVersions` / `withPyVersions` still reject an undeclared package as **both a type error and a runtime error**, and nested declarations compose, so a generator delegating to a helper inherits its requirements rather than restating them.

This replaced ~55 lines of nested ternaries in `ts#agent` with four, and removed the second list that could drift from the declaration.

**Ownership is discovered from state that already exists.** No new files are written into user workspaces. `ownedDependencies` reads `metadata.generator` and `metadata.components[].generator` from each `project.json` (plus `aws-nx-plugin.config.mts`, which stands in for `init`), imports the `DEPENDENCIES` of exactly the generators that ran, and unions them. Anything outside that set is left alone.

**The same declaration narrows the upgrade.** Generators pass the very object they record as metadata to the dependency call, so what a generator added and what the migration owns cannot drift. The sync then replays those predicates against each project's recorded metadata and owns only the branch that project took: a `ts#agent` on `protocol: 'a2a'` owns 20 of its 30 declared packages, `http` owns 23, `ag-ui` 26 — where previously every agent owned all 30. The AG-UI react connection narrows the same way now that its callers record the theme (derived from the website's `ux`) and auth its branches read, so a cloudscape website no longer owns the shadcn packages and an http connection owns none of the AG-UI ones. A predicate that reads a field the metadata doesn't carry counts as not applying, so the migration never claims a branch it cannot confirm; a test asserts every field a predicate reads is a member of the generator's metadata interface, so a silently-unowned dependency fails CI.

This required connections to become visible: a connection generator wires one project into another, so nothing else records it. 27 connection generators now call `addComponentGeneratorMetadata`, keyed per target so wiring two tables yields two entries. Measured on a workspace with one instance of every generator, **11 dependencies are owned only because connections record metadata** — `@tanstack/react-query`, `@copilotkit/react-core`, `@smithy/types`, `httpx` and others would otherwise never upgrade.

**The nx packages move via `packageJsonUpdates`, not the sync.** `nx migrate` collects a package's own migrations only for packages it is bumping (`buildPackageJsonUpdates` → `createMigrateJson`), so bumping nx from inside a migration would silently skip Nx's own migrations for that hop. `registerNxPackageUpdates` emits the declarative block whenever an nx package moves, and the sync explicitly skips them.

**`TS_VERSIONS` is now the single source of truth for the nx version.** Added `nx`, `@nx/js`, `@nx/vite`, `@nx/vitest`, `@nx/workspace` alongside the existing entries, exported as `NX_PACKAGES` with `NX_VERSION`. `init.ts` pins the workspace's `nx`, `@nx/js` and `@nx/workspace` through `withVersions`, `commands.ts` reads `NX_VERSION`, and a test guards that every nx package *and* the plugin's own `@nx/*` dependencies hold the same version, so drift fails CI rather than surfacing as an `nx sync` deadlock.

**Guardrails.** Only exact versions the plugin vends are rewritten, only upwards, and only for owned packages. A version the user raised, or replaced with a range, tag or protocol specifier (`workspace:`, `file:`, `npm:`), is left alone. Lock files are left to the user via `nextSteps`: `nx migrate` runs the package manager install itself, but only when the root manifest changes, and it never runs `uv` or `terraform`.

**Unit:** the declaration and version-upgrade-migration suites are green (76 tests), covering per-surface upgrades, every guardrail class, nx exclusion, ownership discovery, predicate evaluation, migration ordering, and idempotency. Six tests enforce the invariants structurally rather than by convention: every generator that adds dependencies must declare them, must record metadata, must cover the declarations of anything it delegates to, must not install what a helper owns, every connection must be identifiable, and every field a predicate reads must be a member of that generator's metadata interface — each verified to fail when the condition it guards is reverted.

Full plugin suite: 203 failed / 2728 passed. The failing set is **byte-identical** to `main` (203 failed / 2726 passed, `diff` empty) — pre-existing `Cannot read properties of null (reading 'version')` crashes from `@nx/js libraryGeneratorInternal`, unrelated to this change.

**End to end**, against a real `nx migrate` flow across **all four package managers** (pnpm, npm, yarn, bun). For each: published an old and a new plugin to a local verdaccio, created a workspace, ran **one instance of every generator in the plugin (39 invocations)**, rolled every vended version back to a genuinely published earlier release, planted an unowned vended dependency, then ran `nx migrate` → install → `--run-migrations`:

| | pnpm | npm | yarn | bun |
|---|---|---|---|---|
| generators | 39/39 | 39/39 | 39/39 | 39/39 |
| versions upgraded | 86 | 156 | 86 | 89 |
| stale owned | 0 | 0 | 0 | 0 |
| unowned drift | 0 | 0 | 0 | 0 |

`sync-vended-versions` ran as **Migration 4 of 4** on all four, after the three code migrations. 38 generators and 12 connections discovered per workspace; the planted unowned `rxjs` was untouched everywhere; re-running reported no changes.

Earlier hops also confirmed **Nx's own migrations get collected and run** (`@nx/js:23-1-0-add-ignore-deprecations-for-ts6`, `@nx/js:23-1-0-set-tsconfig-root-dir-for-ts6`) — the behaviour `packageJsonUpdates` exists for; a migration-only bump skips them — with a single nx instance resolved, no nested copy under `@nx/js`, and `nx sync` completing without deadlocking.

### Issue # (if applicable)

Closes #988.

### Checklist

----


packages/nx-plugin/src/utils/nx-package-updates.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param version version of the plugin release introducing the bump — the key
+ *   the entry is recorded under, and the version `nx migrate` gates it on
+ * @param nxVersion nx version to bump to; defaults to the vended one
+ */
+export const nxPackageJsonUpdates = (
+  version: string,
+  nxVersion: string = NX_VERSION,
+): PackageJsonUpdates => ({
+  [version]: {
```
cogwirrel: We shouldn't key by the version. if for whatever reason we have some packageJsonUpdates defined by another PR and the backfill hasn't happened yet, this will overwrite those as the key will be `latest` for both. Key can be anything per https://nx.dev/docs/kb/migration-generators - let's make this the same as the migration generator key, ie <version>-<name>

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+const syncPyProjects = (tree: Tree): string[] => {
+  const updated: string[] = [];
+
+  visitNotIgnoredFiles(tree, '.', (path) => {
+    if (!path.endsWith('pyproject.toml')) {
+      return;
+    }
+
+    const original = tree.read(path, 'utf-8') ?? '';
+    const synced = original.replace(
```
cogwirrel: No regex. we have toml helpers.

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+      updated.push(path);
+    }
+  });
+
+  return updated;
+};
+
+// A pinned pip requirement as written in pyproject.toml, eg
+// `"fastapi==0.139.2"` or `"aws-lambda-powertools[tracer]==3.31.1"`.
+const PY_PIN = /"([A-Za-z0-9._-]+(?:\[[A-Za-z0-9._,-]+\])?)==([^"\s]+)"/g;
```
cogwirrel: gross, we have a pip version parser - look at py.ts

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+ * Only versions the plugin vends are rewritten, and only upwards, so a version
+ * the user raised or replaced with a range or protocol specifier is left alone.
+ *
+ * The `nx` and `@nx/*` versions are deliberately NOT synced here — see
+ * `nxPackageJsonUpdates` in `nx-package-updates.ts`. `nx migrate` only collects
+ * a package's own migrations for packages it is bumping, so bumping nx from
+ * inside a migration would silently skip Nx's migrations for that hop. Those
+ * are declared in `migrations.json`'s `packageJsonUpdates` instead, which is
+ * also what keeps every nx pin moving in lockstep.
+ */
```
cogwirrel: Too long, didn't read. Remove this

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+  'devDependencies',
+  'catalog',
+];
+
+/**
+ * The generators vend exact versions, so anything else — a range, a tag, a
+ * protocol specifier — is a deliberate change by the user that this migration
+ * must not undo by pinning it back to the vended version.
+ */
+const isExactVersion = (version: string): boolean => valid(version) !== null;
```
cogwirrel: This is too brittle. Can we check the actual installed version of a package that a user has in their workspace, and then apply if it'd be an upgrade?

eg we write zod as 4.0.1, user changes to `^4`, they upgrade their lockfile so they're actually on 4.1.0. Then a migration tries to upgrade it to 4.0.2 but skips because 4.1.0 is installed, but the next migration to 4.1.2 should apply because 4.1.2 > 4.1.0, and we rewrite the ^4 to our 4.1.2 pin.

Same for python too, perhaps we need to run an uv command...

Check what nx does, we might just be able to use their helper that generators use to add dependencies for typescript deps (see comment below).

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+ * The vended version for a TypeScript dependency, or `undefined` if the plugin
+ * doesn't vend it. nx packages are excluded: `packageJsonUpdates` owns them so
+ * that `nx migrate` collects Nx's own migrations for the hop.
+ */
+const vendedTsVersion = (name: string): string | undefined =>
+  isNxPackage(name) ? undefined : TS_VERSIONS[name as keyof typeof TS_VERSIONS];
+
+/** PY_VERSIONS carries the `==` operator; the pin holds the bare version. */
+const vendedPyVersion = (name: string): string | undefined =>
+  PY_VERSIONS[name as keyof typeof PY_VERSIONS]?.replace(/^==/, '');
```
cogwirrel: shouldn't be splitting here, parse properly with library - see py.ts

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    const original = tree.read(path, 'utf-8') ?? '';
+    let synced = original;
+
+    for (const [provider, vended] of Object.entries(TERRAFORM_VERSIONS)) {
+      const requiredProvider = new RegExp(
+        `(source\\s*=\\s*"hashicorp/${provider}"\\s*\\n\\s*version\\s*=\\s*")([^"]+)(")`,
+        'g',
+      );
```
cogwirrel: We should use gritql, not brittle regex

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+ * export default (tree: Tree) => syncVendedVersions(tree);
+ * ```
+ */
+export const syncVendedVersions = async (
+  tree: Tree,
+): Promise<MigrationReturnObject> => {
+  const nextSteps: string[] = [];
+
+  const catalogs = syncYamlCatalogs(tree);
+  const packageJsons = syncPackageJsons(tree);
```
cogwirrel: Can we reuse the addDependenciesToPackageJson helper here which uses nx's utility and then sets the catalog references?

packages/nx-plugin/src/utils/sync-vended-versions.ts
```diff
@@ -0,0 +1,301 @@
...
+      `\`const version = $old\` => \`const version = '${version}'\`` +
+        ' where { $old <: within `class MetricsAspect implements $_ { $_ }` }',
+    );
+  }
+
+  if (tree.exists(TERRAFORM_METRICS_FILE_PATH)) {
+    await applyGritQL(
+      tree,
+      TERRAFORM_METRICS_FILE_PATH,
+      `\`metric_version = $old\` => \`metric_version = "${version}"\``,
```
cogwirrel: Add a within check

packages/nx-plugin/src/utils/register-sync-versions-migration.ts
```diff
@@ -0,0 +1,136 @@
...
+    'src',
+    'migrations',
+    LATEST_MIGRATIONS_DIR,
+    NAME,
+  );
+  const migrationPath = joinPathFragments(dir, 'migration.ts');
+  const specPath = joinPathFragments(dir, 'migration.spec.ts');
+
+  tree.write(migrationPath, migrationSource());
+  tree.write(specPath, specSource());
```
cogwirrel: Can we just write 1 migration generator which calls the helper and has the test, and then we just point at the same entry for the new migration entry.

packages/nx-plugin/src/utils/register-sync-versions-migration.ts
```diff
@@ -0,0 +1,136 @@
+/**
```
cogwirrel: The files you add specific to the version upgrade migration should be in a dedicated subfolder in utils, ie utils/version-upgrade-migration/xxx

nx-plugin-for-aws:
Thanks — all eleven addressed. Summary of the substantive ones:

**GritQL for HCL.** You were right that there was a way. The blocker was that a `.tf` pattern needs an explicit `language hcl` header — without it the pattern is parsed as the default language and silently matches nothing, which is what made me think the grammar couldn't reach into `required_providers`. With the header it works, and scoping works too, so provider versions are now rewritten with GritQL scoped by the provider's `source`:

```
language hcl
`version = $version` => `version = "6.55.0"` where {
  $version <: within `{ source = "hashicorp/aws" $... }` }
```

Scoping by `source` rather than the attribute name means an aliased provider (`aws_primary = { source = "hashicorp/aws" ... }`) is still matched, which the previous approach handled only by coincidence. Tests cover a file with two vended providers, a non-vended one (`DataDog/datadog`), a `required_version`, and an unrelated `default = "0.0.1"` — only the vended providers move.

**`within` on the metrics `.tf`.** GritQL rejects a `within locals { … }` clause outright ("Could not find variable $_"), so that one is narrowed with `$old <: string()` instead, verified to ignore a `metric_version` mentioned inside another expression. Same constraint the existing `metrics.ts` patterns live with.

**Installed-version comparison.** Implemented, and it found a real bug that only reproduced under `nx migrate`. Note nx's own `addDependenciesToPackageJson` doesn't solve this case — `isIncomingVersionGreater` coerces `^4` to `4.0.0`, so it would happily write a downgrade. The subtle part is pnpm's virtual store: resolving from the workspace root reaches `node_modules/.pnpm/zod@4.4.3/…`, which contains every version *any* project uses, so a catalog entry declaring `4.3.0` looked up to date merely because a project pinned `^4`. Resolution is now confined to a manifest's own `node_modules`.

**`packageJsonUpdates` key** is now `<dir>-<name>` (`latest-nx-packages`), re-keyed to `v<version>-nx-packages` at stamp/backfill time, with a test asserting two releases' entries coexist.

**Also:** one committed `migration.ts` that every release's entry points at (nothing generated); everything moved under `utils/version-upgrade-migration/`; `pyproject.toml` through `readToml`/`updateToml` + `parsePipRequirementsLine`; long docblock removed.

Verified end to end against a real `nx migrate` (old plugin on nx 23.0.2 → new on 23.1.0, local verdaccio), asserting 12 outcomes: nx in lockstep, Nx's own `@nx/js` migrations collected, catalog/py/tf/metrics all synced, aliased provider matched, `~> 3.5` and widened `^4` preserved, decoys untouched, single nx instance, `nx sync` no deadlock, build + lint green, re-run idempotent. Full suite 2866 tests.

packages/nx-plugin/src/py/agent/generator.ts
```diff
@@ -44,14 +52,128 @@ import {
...
+}
+
+/**
+ * The TypeScript dependencies this generator adds itself: the chat CLI, which
+ * runs standalone via tsx for every protocol and resolves the deployed agent
+ * from AppConfig. `agent-chat-cli` transitively bundles the protocol clients
+ * (@a2a-js/sdk, @ag-ui/client). The spread helper constants below are added by
+ * the helpers that own them, so only these are passed to `addTsDependencies`.
+ */
+const OWN_TS_DEPENDENCIES = [
```
cogwirrel: Doesn't look like it needs to be its own variable, inline please.

packages/nx-plugin/src/py/project/generator.ts
```diff
@@ -275,12 +288,8 @@ export const pyProjectGenerator = async (
...
   );
 
-  // Pin ruff to the version generation-time formatting uses (PY_VERSIONS), so
-  // generated files stay `ruff format --check`-clean across ruff releases.
-  addDependenciesToDependencyGroupInPyProjectToml(tree, '.', 'dev', [
-    'ruff',
-    'ty',
-  ]);
+  // The linting and type checking tools live in the workspace root pyproject.
+  addPyDependencies(tree, DEPENDENCIES, '.');
```
cogwirrel: We should use the same pattern `root: true`

packages/nx-plugin/src/py/rdb/generator.ts
```diff
@@ -30,18 +37,74 @@ import {
...
+export interface PyRdbMetadata {
+  readonly engine: PyRdbGeneratorSchema['engine'];
+}
+
+/**
+ * The TypeScript dependencies this generator adds itself, used by the local dev
+ * script wait-for-*-db.ts. The spread helper constants below are added by the
+ * helpers that own them, so only these are passed to `addTsDependencies`.
+ */
+const OWN_TS_DEPENDENCIES = [
```
cogwirrel: Same here, inline (and other occurances)

packages/nx-plugin/src/ts/react-website/agui/generator.ts
```diff
@@ -15,14 +15,42 @@ import {
...
+    { name: 'aws4fetch' },
+    { name: '@aws-sdk/credential-provider-cognito-identity' },
+    { name: 'react-oidc-context' },
+    { name: '@smithy/types' },
+    ...SHADCN_DEPENDENCIES,
+  ],
+});
+
+/** Narrows a conditional list to this generator's declared dependencies. */
+const declared = (
```
cogwirrel: This should follow the same pattern where conditions are defined with `when` - record what is needed in metadata to achieve this.

packages/nx-plugin/src/utils/version-upgrade-migration/nx-package-updates.ts
```diff
@@ -0,0 +1,55 @@
...
+  version: string;
+  packages: Record<string, { version: string; alwaysAddToPackageJson: false }>;
+}
+
+export type PackageJsonUpdates = Record<string, PackageJsonUpdate>;
+
+/**
+ * `alwaysAddToPackageJson: false` so only packages already present are updated.
+ * Keyed `<dir>-<name>` rather than by version, which would collide with another
+ * PR's entry while both are unversioned in `latest`.
```
cogwirrel: name is constant so this is misleading. I guess this is because this is always written as the latest version and we don't keep history in migrations.json as we run once at the end of every migration?

packages/nx-plugin/src/utils/version-upgrade-migration/owned-dependencies.spec.ts
```diff
@@ -0,0 +1,434 @@
...
+
+/**
+ * Generators discovered through something they delegate to instead of their own
+ * metadata: `preset` marks the workspace via `aws-nx-plugin.config.mts`, which
+ * stands in for `init` (whose declaration therefore carries `husky`), and
+ * `ts#dcr-proxy` creates its project through `ts#project`.
+ */
+const DISCOVERED_INDIRECTLY = new Set([
+  'preset',
+  'ts#dcr-proxy',
```
cogwirrel: For this we should write the dcr proxy as the id in metadata

packages/nx-plugin/src/ts/agent/a2a-connection/generator.ts
```diff
@@ -184,19 +205,16 @@ export const tsAgentA2aConnectionGenerator = async (
...
-      '@aws-lambda-powertools/parameters',
-      '@aws-sdk/client-appconfigdata',
-      'aws4fetch',
-      '@aws-sdk/credential-providers',
-      'zod',
-    ]),
-    {},
+    sourceProject.name,
+    TS_AGENT_A2A_CONNECTION_GENERATOR_INFO,
+    toProjectRelativePath(sourceProject, agentFilePath),
```
cogwirrel, line 219: For component -> component connections we may also need to know what the source component is in future. Likely simplest to do by "path", ie "path" is target path as it currently stands, but we add sourcePath too?

packages/nx-plugin/src/utils/version-upgrade-migration/owned-dependencies.spec.ts
```diff
@@ -0,0 +1,432 @@
...
+  // installs into its own project. Those are owned here even though this
+  // generator installs none of them, or the sync would leave them behind.
+  it('should own the dependencies its helpers install elsewhere', async () => {
+    addProject(tree, 'api', { generator: 'ts#trpc-api' } as never);
+
+    const owned = await ownedDependencies(tree);
+
+    // `sharedConstructsGenerator` puts these in the shared constructs project.
+    expect(owned.ts.has('aws-cdk-lib')).toBe(true);
+    expect(owned.ts.has('constructs')).toBe(true);
```
cogwirrel, line 217: Only when iac is cdk though?

packages/nx-plugin/src/utils/version-upgrade-migration/owned-dependencies.spec.ts
```diff
@@ -0,0 +1,432 @@
...
+const RECORDS_METADATA =
+  /addGeneratorMetadata\(|addComponentGeneratorMetadata\(|metadata: \{/;
+
+/**
+ * The explicit helpers. A connection wires into a project it did not create, so
+ * an inline `metadata: {` in its source belongs to something else and doesn't
+ * count.
+ */
+const CALLS_METADATA_HELPER =
+  /addGeneratorMetadata\(|addComponentGeneratorMetadata\(/;
```
cogwirrel: We should use gritql matches rather than regex

packages/nx-plugin/src/utils/version-upgrade-migration/owned-dependencies.spec.ts
```diff
@@ -0,0 +1,432 @@
...
+        }
+      }
+      if (read.size === 0) {
+        continue;
+      }
+      const source = readFileSync(`${info.resolvedFactoryPath}.ts`, 'utf-8');
+      // The metadata interface the generator declares and hands to both the
+      // dependency call and the recording helper. A field a predicate reads must
+      // be one of its members, or no project will carry it.
+      const declared = /interface \w*Metadata \{([^}]*)\}/.exec(source)?.[1];
```
cogwirrel: Again no brittle regexes - same elsewhere in this test file

packages/nx-plugin/src/utils/version-upgrade-migration/register.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param nxChanged whether this update bumped any nx package
+ * @returns paths written, for the update report
+ */
+export const registerNxPackageUpdates = (
+  tree: Tree,
+  nxChanged: boolean,
+): string[] => {
+  if (!nxChanged) {
+    return [];
+  }
```
cogwirrel: the boolean flag to then return [] within the function is ugly, please remove the flag and just skip calling this method if nx didn't change.

packages/nx-plugin/src/utils/version-upgrade-migration/register.ts
```diff
@@ -0,0 +1,47 @@
...
+    tree.read(MIGRATIONS_JSON_PATH, 'utf-8') ?? '{}',
+  );
+
+  // Recorded under `latest` until stamping resolves the version it ships with.
+  migrations.packageJsonUpdates = {
+    ...migrations.packageJsonUpdates,
+    ...nxPackageJsonUpdates(LATEST_MIGRATIONS_DIR, LATEST_MIGRATIONS_DIR),
+  };
+
+  tree.write(MIGRATIONS_JSON_PATH, `${JSON.stringify(migrations, null, 2)}\n`);
```
cogwirrel: use updateJson from @nx/devkit

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -0,0 +1,461 @@
...
+    }
+  });
+  return referenced;
+};
+
+/**
+ * Upgrade a pip requirement to the vended version, or return it unchanged.
+ *
+ * Parsed rather than pattern-matched so extras and operators are read the way
+ * `uv` reads them. Only a single `==` pin is rewritten; anything looser is the
```
cogwirrel: Behaviour should be the same as ts, where if the range doesn't satisfy our upgrade we upgrade

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -0,0 +1,461 @@
...
+  const catalogs = syncOrphanCatalogEntries(tree, owned);
+  const pyProjects = syncPyProjects(tree, owned);
+  const terraformFiles = await syncTerraformProviders(tree);
+  await syncMetricsVersion(tree);
+
+  // Lock files are left to the user: `nx migrate` only installs when the root
+  // manifest changes, and never runs `uv` or `terraform`.
+  if (catalogs.length > 0 || packageJsons.length > 0) {
+    nextSteps.push(
+      'TypeScript dependency versions were updated. Run your package manager install to update the lock file.',
```
cogwirrel: Let's make this package manager specific

packages/nx-plugin/src/utils/add-dependencies.ts
```diff
@@ -0,0 +1,122 @@
...
+    string,
+    { runtime: Record<string, string>; dev: Record<string, string> }
+  >();
+  for (const entry of applicable) {
+    // `root: true` is for shared tooling; everything else belongs to the project
+    // so `noUndeclaredDependencies` passes.
+    const manifestPath =
+      entry.root || !projectRoot
+        ? 'package.json'
+        : joinPathFragments(projectRoot, 'package.json');
```
cogwirrel: We should throw if we have an entry that doesn't declare root and projectRoot is not passed rather than silently adding to root.

packages/nx-plugin/src/utils/add-dependencies.ts
```diff
@@ -0,0 +1,122 @@
...
+ * belongs, mirroring the TypeScript side.
+ */
+export const addPyDependencies = (
+  tree: Tree,
+  declaration: DependencyDeclaration,
+  { metadata = {}, projectRoot }: AddDependenciesOptions = {},
+): void => {
+  const applicable = applicableDependencies(declaration.py, metadata);
+  const targetRoot = (entry: { root?: boolean }) =>
+    entry.root || !projectRoot ? '.' : projectRoot;
```
cogwirrel: Same here, throw if projectRoot not specified and dep not explicitly marked root.

packages/nx-plugin/src/utils/declared-dependencies.ts
```diff
@@ -0,0 +1,225 @@
...
+   * Whether this dependency applies. Reads the metadata the generator records,
+   * so the migration can evaluate the same predicate later.
+   */
+  readonly when?: (metadata: M) => boolean;
+  /** Added as a dev dependency rather than a runtime one. */
+  readonly dev?: boolean;
+  /** Added to the workspace root manifest rather than the project's. */
+  readonly root?: boolean;
+  /** Added to this pyproject dependency group rather than the main list. */
+  readonly group?: string;
```
cogwirrel: can we split types for ts and py so that dev is only assignable for ts and group for py?

packages/nx-plugin/src/utils/declared-dependencies.ts
```diff
@@ -0,0 +1,225 @@
...
+>(
+  entries: readonly Entry[],
+  metadata: M,
+): Entry[] =>
+  entries.filter((entry) => {
+    if (!entry.when) {
+      return true;
+    }
+    try {
+      return entry.when(metadata) === true;
```
cogwirrel: I think better to allow when to return something truthy rather than only `true`?

CONTRIBUTING.md
```diff
@@ -192,6 +192,80 @@ Do not add a `version` field to `migrations.json` entries — versions arrive on
...
+
+Two cases devkit can't reach are handled in `syncOrphanCatalogEntries`, comparing with `isVendedUpgrade` (`vended-upgrade.ts`): a catalog entry **no manifest references**, which devkit only reaches through those references; and **bun catalogs**, since `getCatalogManager` covers pnpm and yarn only, and with no manager devkit rewrites a `catalog:` reference to a literal, severing the catalog. Python pins and Terraform providers use `isVendedUpgrade` too.
+
+**The nx packages are the exception**, listed in `NX_PACKAGES`:
+
+- `TS_VERSIONS.nx` (`NX_VERSION`) is the single source of truth. All of them must hold that value and match the plugin's own `@nx/*` dependencies (guarded in `versions.spec.ts`) — a workspace nx even a patch apart hoists a second nested nx and the two deadlock `nx sync`.
+- They are bumped via `packageJsonUpdates`, **not** the sync migration: `nx migrate` collects a package's migrations only for packages it is bumping, so bumping nx from a migration would skip Nx's own migrations for that hop. `update-versions` writes the block whenever an nx package moved.
```
cogwirrel: Let's trim this part, making sure it's just the relevant parts for contributors writing generators rather than explaining in detail how the upgrade works.

packages/nx-plugin/src/utils/version-upgrade-migration/owned-dependencies.ts
```diff
@@ -118,6 +137,14 @@ export const ownedDependencies = async (
     }
   }
 
+  // The shared constructs CDK packages only reach a CDK workspace, so a
+  // Terraform one never owns them however many generators declare them.
+  if (workspaceIac(tree) === 'terraform') {
```
cogwirrel: No, this is a terrible hack. Own them properly and make sure generators write their iac to metadata based on the time of generation (if it was generated with infra)

cogwirrel (dismissed):

cogwirrel (approved):

PR #1002, opened 2026-07-28
test(e2e): migrate a released workspace to the local build

### LINKED ISSUE #705: feat: E2E test harness for rolling migrations between versions
### Description

Create an end-to-end test harness that validates migrations work correctly across the previous N versions of the plugin. This ensures that users upgrading from any recent version get a working workspace after running `nx migrate`.

### Motivation

Unit tests validate individual migrations in isolation, but don't catch:
- Migration ordering issues (migration A depends on migration B running first)
- Interactions between multiple migrations in a single upgrade path
- Real-world workspace states that differ from test fixtures
- Regressions where a new migration breaks a previously-working upgrade path

### Suggested Approaches

**Option A: Git-tag based** — Check out previous tagged releases, generate a workspace, then migrate forward
```
v0.115.0 workspace → migrate to v0.116.0 → migrate to v0.117.0 → ... → current
```

**Option B: Snapshot based** — Store committed workspace snapshots for key versions
```
e2e/migrations/fixtures/v0.115.0/  (generated workspace files)
e2e/migrations/fixtures/v0.118.0/  (generated workspace files)
```

Each has tradeoffs — Option A avoids fixture maintenance but is slower; Option B is faster but requires upkeep.

**CI considerations:**
- Should run on PRs that touch migration source files
- Possibly a nightly run against all supported versions
- Configurable `N` (number of previous versions to test)

### Acceptance Criteria

- [ ] Harness can generate a workspace at a given tagged version
- [ ] Harness applies migrations sequentially from source to target version
- [ ] Harness validates the migrated workspace builds and lints cleanly
- [ ] At least 3 previous versions tested
- [ ] CI workflow triggers on migration-related changes
- [ ] Clear failure messages indicating which migration broke and between which versions

nx-plugin-for-aws:
### Reason for this change

#928 wired up `nx migrate @aws/nx-plugin`, but nothing exercises it end to end. Unit tests prove a migration does what it was written to do against a synthetic `Tree`; they can't catch ordering problems between migrations, interactions across a whole upgrade, a migration that leaves the workspace unbuildable, or drift between what a migration writes and what today's generators produce.

The contract worth defending is narrow and testable: **deterministic migrations alone must keep a vanilla generated workspace green across an upgrade, non-interactively.** This adds the harness that holds it.

Closes #705

### Description of changes

**New `migrate` smoke test**

Creates a workspace on a **released** version of the plugin, scaffolds a recipe with that version's generators, then upgrades it to the local build with the real `nx migrate` cycle (`migrate` → install → `migrate --run-migrations`) and asserts the result. One `it` per start version, so a failure names the hop that broke.

The "before" workspace is created for real from the registry rather than from a committed fixture, so each hop starts from exactly what a user on that version has — no fixture upkeep, and no building old tags.

Per hop it asserts:

- **The upgrade did something** — `nx migrate` wrote the new version and generated `migrations.json` (a silent "No updates were applied" fails).
- **The installed plugin is the target** — so the migrations that run are the ones the target version ships, not the ones already installed.
- **Codemods ran** — every queued entry appears in the run output; a throwing codemod fails the command.
- **Prompts degrade correctly** — prompt and hybrid migrations are materialised under `tools/ai-migrations/` and reported in next steps rather than silently disappearing.
- **User code is safe** — a deliberately customised user-owned file must survive verbatim (guardrail: report what you don't recognise, don't rewrite it).
- **Idempotency** — re-running `--run-migrations` leaves an empty `git status`.
- **The core contract** — the migrated workspace still `sync`s and builds.

An empty collection is a valid outcome (that's the state on `main` today): the test says so explicitly and still asserts the upgraded workspace is green.

**Registry setup (`global-setup.ts`)**

Two things the migrate test needs that the other smoke tests don't:

- **The local build is republished at a version above every release.** `nx migrate` only runs a migration whose version is greater than the installed one, so publishing the in-repo `0.0.0` would put every migration out of range and silently assert nothing — verified: it reports "No updates were applied". It's republished under a `prerelease` bump of the latest tag, with `migrations.json` stamped by `scripts/stamp-migrations.ts` exactly as the release job does, under a dedicated dist-tag so `latest` keeps pointing at the `0.0.0` build every other smoke test resolves.
- **Each start version is mirrored into verdaccio.** `.verdaccio/config.yml` deliberately doesn't proxy the three local packages to npmjs (our `0.0.0` collides with an early public release), so a released version isn't reachable through the registry the workspaces are pinned to. Mirroring the tarballs in means one `@aws:registry` pin serves both ends of the upgrade.

A failure in this setup is recorded and surfaced by the migrate test rather than thrown, so a registry hiccup fetching an old release doesn't take down every other smoke test sharing the same setup.

**Supporting changes**

- `createTestWorkspace` takes an optional `version`, pinning `@aws/create-nx-workspace` (and hence the preset) to a released version. It already threaded a `tag` argument through to `buildCreateNxWorkspaceCommand` and dropped it.
- `pinAwsScopeToLocalRegistry` extracted from `smoke-test.ts` (unchanged behaviour) and reused, with an `extra` parameter for additional npmrc lines. The migrate test needs `legacy-peer-deps` — the preset is pinned to an exact version, whose peer ranges npm's strict resolver rejects during `create` (matching what `runInstall` already does for npm).
- `scripts/stamp-migrations.ts` takes `--out` so the same stamping logic writes into the staging copy, rather than duplicating it in the e2e setup.
- Runs in the existing `smoke_tests` matrix on both PR and CI — it's no slower than the other workspace-building smoke tests, and migration regressions are exactly what PR CI should catch.

**Start-version matrix**

Read from `v*` git tags, newest first, defaulting to the latest release — the version a user upgrading today actually starts from. `NX_E2E_MIGRATE_VERSIONS=<n>` widens it to the last _n_ releases.

**Recipe scope**

A fixed representative slice of the dungeon-adventure shape (infra, website + auth, tRPC API, FastAPI, a lambda function, an agent) rather than the full generator matrix.

This is a coupling and cost tradeoff rather than a compatibility one — checking the matrix's generators and options against past releases, it's close to compatible over the realistic window (only `ts#dcr-proxy` is absent at rc.40, plus `py#rdb` at rc.30, and `py#agent --framework=langchain` back at rc.20; every flag name it passes still exists in all of them). The problem is that the matrix grows with every new generator, and one that postdates a start version makes `nx g` exit non-zero — failing the hop during scaffolding as though a migration broke. Guarding that means version-gating each of its ~60 invocations, most of which add no migration coverage (the full matrix takes ~31 minutes on the `idempotency` lane, and that cost multiplies per start version). The fixed recipe avoids both, and adding a project type a migration touches is one more line.

Ran the new test locally against real migrations, using three throwaway example migrations (one of each kind, scaffolded with `ts#nx-migration` and given real bodies) published to verdaccio. These are **deliberately not in this PR** — they exist only to prove the harness detects what it claims to:

- **All three kinds queued and handled**: `nx migrate` resolved the stamped local build, wrote `migrations.json` with all three entries, and materialised two prompts to `tools/ai-migrations/@aws/nx-plugin/<version>/` (de-duplicated as `prompt.md` / `prompt-1.md`). The deterministic and hybrid codemods applied (`UPDATE nx.json`, `UPDATE package.json`); both agentic halves were skipped with "agentic flow disabled" and listed in next steps. Idempotency re-run produced an empty `git status`, the customised user file survived, and `run-many --target build --all` passed.
- **With the empty collection this PR ships**: `nx migrate` reports "There are no migrations to run", the test logs that no migrations were queued, and still asserts the upgraded workspace builds. Passes.
- **The `0.0.0` failure mode is real**: publishing the local build at its in-repo version makes `nx migrate` report "No updates were applied" and run nothing — which is why the republish-at-a-higher-version step exists.
- **The harness catches a broken migration**: an early version of the example hybrid migration wrote an invalid key into `biome.json`; the test failed on the workspace build (`format` target rejected the config) rather than passing silently.
- Stamping verified against real tag history (`scripts/stamp-migrations.ts --pending-version 1.0.0-rc.49` stamps all unversioned entries; `--out` writes to the staging path), plus the existing `migration-versions.spec.ts` unit tests (19) still pass.
- `pnpm nx run-many --target lint --all` passes.

### Issue # (if applicable)

Closes #705.

### Checklist

----


e2e/src/smoke-tests/migrate.ts
```diff
@@ -0,0 +1,330 @@
...
+ * strictly above every released one (see `resolveMigrateTargetVersion`), which
+ * is the window `nx migrate` executes migrations in.
+ *
+ * Only the deterministic path is asserted. Prompt migrations are deferred by
+ * Nx in a non-interactive run, and the contract this test holds is that
+ * deterministic migrations alone keep a generated workspace green.
+ */
+
+/** Package manager the migrate hops run under. */
+export const MIGRATE_PKG_MGR = 'npm';
```
cogwirrel: use pnpm

e2e/src/smoke-tests/migrate.ts
```diff
@@ -0,0 +1,330 @@
...
+
+  // 2. Scaffold the recipe with the START version's generators — the workspace
+  // state the user is upgrading from.
+  await runMigrateRecipe(opts);
+
+  // A deliberately customised user-owned file, to assert deterministic
+  // migrations report rather than rewrite what they don't recognise.
+  const userOwnedPath = join(projectRoot, 'packages/website/src/main.tsx');
+  expect(existsSync(userOwnedPath)).toBe(true);
+  appendFileSync(userOwnedPath, `\n${USER_MARKER}\n`, 'utf-8');
```
cogwirrel: no need for this, this test should stay generic

e2e/src/smoke-tests/migrate.ts
```diff
@@ -0,0 +1,330 @@
...
+ *
+ * Reusing the full matrix would couple this test to it: the matrix grows with
+ * every new generator, and one that postdates a start version makes `nx g` exit
+ * non-zero, failing the hop during scaffolding as though a migration broke.
+ * Guarding that means version-gating each of its ~60 invocations, most of which
+ * add no migration coverage. A fixed recipe of generators present in every
+ * supported start version covers every hop instead, and adding a project type a
+ * migration touches is one more line here.
+ */
+const runMigrateRecipe = async (opts: {
```
cogwirrel: This should use the internal#test-matrix generator

e2e/src/global-setup.ts
```diff
@@ -38,6 +52,119 @@ const restoreBackup = (path: string) => {
...
+ *   an early public release), so a released version isn't reachable through the
+ *   registry the workspaces are pinned to. Mirroring the tarballs in means one
+ *   `@aws:registry` pin serves both ends of the upgrade.
+ *
+ * A failure here is recorded rather than thrown: every other smoke test shares
+ * this setup and doesn't need these publishes, so a registry hiccup fetching an
+ * old release shouldn't take the whole run down. The migrate test reads the
+ * recorded reason and fails with it.
+ */
+const publishForMigrateSmokeTest = (localRegistry: string) => {
```
cogwirrel: This should be in migrate-versions if possible.

e2e/src/smoke-tests/migrate-versions.ts
```diff
@@ -0,0 +1,87 @@
...
+  '@aws/nx-plugin-mcp',
+  '@aws/create-nx-workspace',
+] as const;
+
+/**
+ * Number of previous releases to migrate from. Currently one — the latest
+ * release, which is what a user upgrading today starts from. Raise via
+ * `NX_E2E_MIGRATE_VERSIONS` to cover more of the supported range.
+ */
+const DEFAULT_START_VERSION_COUNT = 1;
```
cogwirrel: Set to 5

e2e/src/smoke-tests/migrate.ts
nx-plugin-for-aws: Done — hops now run under pnpm. The npmrc extra had to change with it: `legacy-peer-deps` is npm-only, so it is now pnpm's `strict-peer-dependencies=false`.

This also surfaced a real bug in the test. The old assertion read the `@aws/nx-plugin` range straight out of `package.json`, which on a pnpm workspace is `catalog:` — the version lives in `pnpm-workspace.yaml`. It now asserts against the installed `node_modules/@aws/nx-plugin/package.json`, which is package-manager agnostic and is what the migrations actually run from.

e2e/src/smoke-tests/migrate-versions.ts
nx-plugin-for-aws: Done — 5.

e2e/src/smoke-tests/migrate.ts
nx-plugin-for-aws: Removed — it hardcoded `packages/website/src/main.tsx` from the recipe, so the driver now stays generic.

e2e/src/smoke-tests/migrate.ts
nx-plugin-for-aws: Done, with one caveat worth flagging. Each hop scaffolds with the **start** version's generators, and no release carries `internal#test-matrix` yet — the newest tag is `v1.0.0-rc.50`, which predates #1004. So a plain swap would fail every hop during scaffolding.

It now calls `internal#test-matrix` when the start version ships it, and falls back to the fixed recipe for older ones (`hasTestMatrixGenerator`, gated on `1.0.0-rc.51`). Once the supported range no longer reaches back past that release, the fallback and the constant both go — noted in the comment.

Verified locally against `rc.50` (so via the fallback path): the hop migrates, migrations are idempotent, and the workspace builds — 8 projects, 44 tasks. The matrix path will take effect from the next release without any further change.

e2e/src/global-setup.ts
nx-plugin-for-aws: Moved to `migrate-versions`, next to the version resolution it depends on, so `global-setup` just calls `publishForMigrateSmokeTest(localRegistry)`. The two migrate dist-tag constants moved with it.

While moving it I also gated it: `global-setup` runs for every smoke lane, so with 5 start versions it would mirror 15 tarballs on all 28 lanes for the benefit of one. It now returns early unless the run includes the migrate test (read off the `-t` filter; a bare local `vitest` still publishes).

e2e/src/smoke-tests/migrate.ts
nx-plugin-for-aws: Reworked per Jack: the hardcoded version gate is gone. The hop now asks the workspace it just created whether the generator exists — `nx list @aws/nx-plugin`, checked for `internal#test-matrix` — and skips with a warning if not. No fallback recipe, and no release boundary encoded anywhere.

Verified against a local verdaccio with two plugin builds published, one carrying `internal#test-matrix` and one with it stripped: the predicate returns true and false respectively. Also verified against a real `1.0.0-rc.50` workspace (built from the published tarball, which genuinely lacks the generator) — correctly false. Worth noting `nx list` does surface hidden generators, which is what makes this work.

That exercise turned up a **pre-existing hole in the harness**, unrelated to these changes: the "before" workspace is not actually landing on the start version. The create log shows `preset=@aws/nx-plugin@1.0.0-rc.50`, but the resulting catalog pins `0.0.0` and `node_modules/@aws/nx-plugin` is `0.0.0` — the local build, which verdaccio serves as `latest`. Since `nx migrate` only runs migrations above the installed version, every hop has effectively been `0.0.0 -> target` and asserting nothing about a real upgrade.

I have added an assertion that the installed plugin equals the start version before scaffolding, so this fails loudly instead of passing vacuously. **It fails right now** — deliberately left that way rather than papering over it. Reproducing the same create in isolation against a registry with identical dist-tags resolves `rc.50` correctly, so it is something about the e2e environment specifically; I have not pinned down which part yet. Happy to dig further or split it into its own issue — let me know which you would prefer.

nx-plugin-for-aws: Root-caused and fixed — the harness now always starts from the correct version.

`@nx/workspace`'s new generator resolves the preset dependency as:

```js
[preset]: process.env['NX_E2E_PRESET_VERSION'] ?? getNpmPackageVersion(preset, presetVersion)
```

The env var takes priority over the version parsed from `--preset`. `global-setup` sets it to the local build's version so the other lanes install that, which silently overrode the migrate test's pin — `create` reported `preset=@aws/nx-plugin@1.0.0-rc.50` while the workspace was scaffolded and installed at `0.0.0`. `pnpm create @aws/nx-workspace@<version>` was doing its job; the override happened two layers deeper.

`createTestWorkspace` now points `NX_E2E_PRESET_VERSION` at the version it was asked for, so the variable cannot contradict the pin. Only the migrate test passes a version, so no other lane changes. The version assertion I added proves it: it failed before the fix and passes after, with the workspace on `1.0.0-rc.50`.

One consequence worth being explicit about: `rc.50` genuinely does not vend `internal#test-matrix`, so with the harness now honest, every hop in the supported range skips and the lane asserts nothing until a release ships that generator. The run logs that in as many words rather than leaving a green tick to imply coverage — `no hop actually migrated (N skipped)`. It starts doing real work automatically from the next release, no code change needed.

cogwirrel (approved):

PR #1004, opened 2026-07-28
feat(nx-plugin): internal#test-matrix generator composing every generator

nx-plugin-for-aws:
### Reason for this change

`e2e/src/smoke-tests/generator-matrix.ts` lives in the test repo, so it can only ever describe **today's** generators. That's fine for the existing smoke tests, which always generate with the local build — but not for the migrate harness (#705), which has to scaffold a workspace using a **released** version's generators. A generator that postdates the start version makes `nx g` exit non-zero, failing the hop during scaffolding as though a migration broke, and guarding that means version-gating ~60 invocations.

A matrix that ships **with the plugin** dissolves the problem: each released version carries the matrix of the generators it had, so it is self-consistent by construction and no version gating is ever needed. That doesn't make the CLI matrix redundant — it still tests the path users take — so this adds the second matrix rather than replacing the first.

### Description of changes

Adds a hidden `internal#test-matrix` generator that scaffolds every generator, component and connection permutation by composing the generators **in-process on a single tree** — the pattern `connection/generator.ts` already uses to dispatch to ~30 sub-generators. It imports them through the plugin's own `sdk/*` entrypoints (two generators missing from the SDK, `py#dynamodb` and `ts#astro-docs`, are now exported there).

**Both matrices are kept**, because they test different things:

- `e2e/src/smoke-tests/generator-matrix.ts` (unchanged) runs each generator as its own `nx g` invocation — the path users actually take. This remains what the package manager and IaC provider smoke tests scaffold with.
- `internal#test-matrix` covers the composed path, with its own `test-matrix` smoke lane that installs, syncs and **builds** the workspace.

A new generator goes in both; `CONTRIBUTING.md` and the contribute-generator tutorial now say so.

Beyond version-independence, composing in-process is stronger in two ways:

- **Options are typed against each generator's schema.** In-process calls bypass the CLI's schema-default application, so TypeScript forces every required option to be named explicitly (`iac`, `auth`, `framework`, `infra`, …). A new required option or a changed enum becomes a **compile error**, rather than a dropped flag that silently reduces coverage. This caught real gaps: `py#rdb` types `iac` as optional (it has a schema default) so composed calls needed it passed explicitly, and `directory` defaults to `packages` — without it every project landed at the workspace root.
- **One workspace covers both IaC providers.** `iac` is per-invocation (`inherit|cdk|terraform`).

The infrastructure project every other generator deploys into *is* provider-specific, so `--infra` picks the generator that owns it (`ts#infra` for CDK, `terraform#project` for Terraform), preserving the existing split where `ts#infra` was deliberately CDK-lane-only.

No intermediate installs are needed: nothing composed here reads the project graph (only `ts#sync` does, which isn't part of the matrix), so the whole matrix runs on one tree and the returned `GeneratorCallback` installs once.

**Also fixes a latent non-idempotency this surfaced.** TanStack Router's vite plugin rewrites `routeTree.gen.ts` in its own unformatted shape whenever `vite.config.mts` is loaded — including when Nx computes the project graph. The old matrix hid this: each generator was a separate `nx g`, so the graph reloaded after the website existed and the plugin rewrote the file, meaning the *unformatted* version reached the baseline. Composing on one tree computes the graph once, before the website exists, so our formatter's output survived to the baseline and the next run reverted it.

`formatFilesInSubtree` therefore takes an optional `{ ignore }` list of explicit workspace-relative paths, and `ts#react-website` — the generator that owns the file — passes its own route tree path. Safe to leave unformatted because the vended biome config excludes `**/*.gen.*`, verified by running the `format` target against a generated workspace.

The ignore list is retained **per tree** rather than scoped to the call that passes it. `formatFilesInSubtree` formats every change pending in the tree, not only the ones its caller made, so on a shared tree the *next* generator's plain `formatFilesInSubtree(tree)` would reformat the route tree and reintroduce the non-idempotency. Once the owning generator declares a path ignored it stays ignored for the rest of the run; there's a regression test for exactly that.

- **`npm` smoke test, end to end, locally:** the composed matrix generated **798 files** and `nx run-many --target build` succeeded for **34 projects / 207 tasks**, zero failures (14m43s). This is the real check — the matrix is exercised through the CLI exactly as CI drives it, and the workspace is built rather than just asserted to exist.
- **Terraform path verified directly.** CI caught a genuine bug in the first revision: pulling `ts#infra` into the shared matrix made the Terraform lane generate a CDK app it never had, failing with `Cannot set properties of undefined (setting 'outputs')`. After adding `--infra`, running the matrix against a real Terraform workspace produced **737 files** with no error, `packages/infra` is a Terraform project (`main.tf`, `outputs.tf`), only `packages/common/terraform` shared constructs are present, and the CDK-only projects are correctly absent.
- **The `idempotency` smoke lane passes** — the lane that originally surfaced the route tree problem, so it's the direct check that the per-tree `ignore` mechanism holds on a shared tree.
- Added the `ts#nx-migration` coverage (all three kinds) that the CLI matrix had and the first revision of this one was missing — confirmed generated in the `npm` run.
- `tsc --noEmit` clean over the plugin and the e2e specs; `biome check` clean.

Note: `packages/nx-plugin` unit tests fail in my environment on unmodified `main` (`TypeError: Cannot read properties of null (reading 'version')` from `@nx/js`'s catalog manager), confirmed in a clean worktree at `origin/main` — pre-existing and unrelated, which is why the real-CLI runs above are the substantive local validation. CI's 10 unit shards are the authority for the snapshots and the `format.spec.ts` regression tests.

### Issue # (if applicable)

Groundwork for #705.

### Checklist

----


e2e/src/smoke-tests/test-matrix.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // Both IaC providers land in one workspace, so a migration touching either
+    // provider's vended files is covered by a single hop.
+    expect(
+      existsSync(join(projectRoot, 'packages/common/constructs/src/index.ts')),
+    ).toBe(true);
+    expect(existsSync(join(projectRoot, 'packages/common/terraform'))).toBe(
+      true,
+    );
```
cogwirrel: check it builds instead

packages/nx-plugin/src/internal/test-matrix/generator.ts
```diff
@@ -0,0 +1,595 @@
...
+ * Options are typed against each generator's schema, so adding a required option
+ * or changing an enum breaks the build here rather than silently reducing
+ * coverage.
+ *
+ * Projects are generated with `preferInstallDependencies: false` by default and
+ * the returned callback installs once. Nothing composed here reads the project
+ * graph (only `ts#sync` does, which is not part of the matrix), so no
+ * intermediate install is needed.
+ */
+export const internalTestMatrixGenerator = async (
```
cogwirrel, line 72: I think this should replace generator-matrix.ts in that case (and update contributing guide to update this instead)

packages/nx-plugin/generators.json
```diff
@@ -495,6 +495,13 @@
       "description": "Connect a py#mcp-server to a py#rdb project",
       "metric": "g59",
       "hidden": true
+    },
+    "internal#test-matrix": {
+      "factory": "./src/internal/test-matrix/generator",
+      "schema": "./src/internal/test-matrix/schema.json",
+      "description": "Scaffold every generator permutation for end to end tests",
+      "metric": "g69",
```
cogwirrel: Make this t1

docs/src/content/docs/en/get_started/tutorials/contribute-generator.mdx
```diff
@@ -374,7 +374,12 @@ Write some unit tests for the generator. These are quite straightforward to impl
...
+We have a suite of "smoke tests" which run generators in a fresh workspace and make sure that everything builds. At a minimum, your new generator should be added to **both** generator matrices so that it's exercised by the smoke tests:
+
+- `e2e/src/smoke-tests/generator-matrix.ts` — runs each generator through the CLI, one invocation at a time, exactly as a user would.
+- `packages/nx-plugin/src/internal/test-matrix/generator.ts` — a hidden generator which composes all the others in-process on a single tree.
+
+Both are worth keeping. The CLI matrix tests the path users actually take. The composed matrix ships with the plugin, so each released version carries the matrix of the generators _it_ had — which lets tests that upgrade an older workspace scaffold it with that version's own matrix. Because it calls generators directly rather than through the CLI, its options are type-checked against each generator's schema, so a new required option or a changed enum fails the build there rather than silently reducing coverage.
```
cogwirrel: Remove this

docs/src/content/docs/en/get_started/tutorials/contribute-generator.mdx
```diff
@@ -374,7 +374,12 @@ Write some unit tests for the generator. These are quite straightforward to impl
 
 #### 4. End to End Tests
 
-We have a suite of "smoke tests" which run generators in a fresh workspace and make sure that everything builds. At a minimum, your new generator should be added to the generator matrix in `e2e/src/smoke-tests/generator-matrix.ts` so that it's exercised by the smoke tests.
+We have a suite of "smoke tests" which run generators in a fresh workspace and make sure that everything builds. At a minimum, your new generator should be added to **both** generator matrices so that it's exercised by the smoke tests:
+
+- `e2e/src/smoke-tests/generator-matrix.ts` — runs each generator through the CLI, one invocation at a time, exactly as a user would.
+- `packages/nx-plugin/src/internal/test-matrix/generator.ts` — a hidden generator which composes all the others in-process on a single tree.
```
cogwirrel: change to "a hidden generator which composes all the others for testing migrations between versions"

packages/nx-plugin/src/internal/test-matrix/generator.ts
```diff
@@ -0,0 +1,635 @@
...
+import type { TsNxMigrationGeneratorSchema } from '../../ts/nx-migration/schema';
+import { tsNxPluginGenerator } from '../../ts/nx-plugin/generator';
+import { tsRdbGenerator } from '../../ts/rdb/generator';
+import type { TsRdbGeneratorSchema } from '../../ts/rdb/schema';
+import { tsWebsiteGenerator } from '../../ts/website/app/generator';
+import { tsWebsiteAuthGenerator } from '../../ts/website/auth/generator';
+import { installDependencies } from '../../utils/install';
+import { toSnakeCase } from '../../utils/names';
+import { getNpmScope, getNpmScopePrefix } from '../../utils/npm-scope';
+import type { InternalTestMatrixGeneratorSchema } from './schema';
```
cogwirrel, line 49: Let's import other generators from the sdk instead

packages/nx-plugin/src/internal/test-matrix/generator.ts
```diff
@@ -0,0 +1,635 @@
...
+  tree: Tree,
+  options: InternalTestMatrixGeneratorSchema = {},
+): Promise<GeneratorCallback> => {
+  const preferInstallDependencies = options.preferInstallDependencies ?? false;
+  const defaults = { preferInstallDependencies };
+  // Composing in-process bypasses the CLI's schema-default application, so any
+  // default a user would get has to be passed explicitly. Project generators all
+  // default `directory` to `packages` — without it projects land at the
+  // workspace root, unlike a real workspace. Component generators take no
+  // directory, and `iac` is passed per call.
```
cogwirrel: drop this comment

packages/nx-plugin/src/internal/test-matrix/generator.ts
```diff
@@ -0,0 +1,635 @@
...
+      stageConfig: true,
+      ...projectDefaults,
+    });
+    // A Terraform project alongside the CDK ones, so a CDK workspace still
+    // covers the Terraform generators' output.
+    await terraformProjectGenerator(tree, {
+      name: 'tf-infra',
+      type: 'application',
+      ...projectDefaults,
+    });
```
cogwirrel: Remove since this is no longer used for the other smoke tests

packages/nx-plugin/src/utils/format.ts
```diff
@@ -116,6 +116,22 @@ const BIOME_FORMATTABLE_EXTENSIONS = new Set([
...
+ * generation non-idempotent: the file reverts on the next run and the workspace
+ * flips between the two forms.
+ *
+ * Safe to leave: the vended biome config excludes `**\/*.gen.*`, so the
+ * workspace's own `format` target does not check it either.
+ *
+ * Narrow by design — the OpenAPI clients we generate are also `.gen.ts`, but
+ * they are ours to format and nothing else rewrites them.
+ */
+const isRouteTree = (filePath: string): boolean =>
```
cogwirrel: We should instead allow ignore patterns to be passed as extra optional options to formatFilesInSubtree and the generator which owns the route tree should pass this in.

CONTRIBUTING.md
```diff
@@ -256,6 +248,13 @@ Worth repeating with a *customised* workspace: hand-edit the files your migratio
 
 The end to end tests run our generators and check that generated projects function correctly (usually by performing a build).
 
+A new generator must be added to **both** generator matrices:
+
+- `e2e/src/smoke-tests/generator-matrix.ts` — runs each generator through the CLI, one invocation at a time, as a user would. This is what the package manager and IaC provider smoke tests scaffold with.
+- `packages/nx-plugin/src/internal/test-matrix/generator.ts` — a hidden generator composing all the others in-process on a single tree, covered by the `test-matrix` smoke test.
+
+Both earn their place. The CLI matrix exercises the path users actually take. The composed matrix ships with the plugin, so each released version carries the matrix of the generators it had — which is what lets a test upgrade an older workspace and scaffold it with that version's own matrix. It also type-checks options against each generator's schema, so a new required option or changed enum breaks the build rather than silently reducing coverage.
```
cogwirrel: Make same changes as my comment on contribute-generator.mdx

packages/nx-plugin/generators.json
nx-plugin-for-aws: Done — metric is now `t1`.

e2e/src/smoke-tests/test-matrix.spec.ts
nx-plugin-for-aws: Done — the lane now installs, syncs and runs `run-many --target build --all`, asserting `Successfully ran target build`, rather than checking files exist. That immediately caught a missing `aws-nx-plugin.config.mts` (the smoke tests don't run in a git repo, so `license-check` fails without it).

packages/nx-plugin/src/internal/test-matrix/generator.ts
nx-plugin-for-aws: Partly — I did replace it, then Jack asked to keep `generator-matrix.ts` too, since driving one CLI call at a time is how users actually invoke generators and that's worth testing directly. So it's back to identical-to-main, and both `CONTRIBUTING.md` and the contribute-generator tutorial now say to add a new generator to **both** matrices, with the reasoning for each. Happy to collapse to one if you'd rather.

packages/nx-plugin/src/internal/test-matrix/generator.ts
nx-plugin-for-aws: Done — the matrix now imports from `../../sdk/{agentcore-gateway,connection,license,py,terraform,ts}`. Two generators it uses were missing from the SDK, so I added them: `pyDynamoDBGenerator` and `tsAstroDocsGenerator` (with their schema types).

packages/nx-plugin/src/internal/test-matrix/generator.ts
nx-plugin-for-aws: Dropped.

packages/nx-plugin/src/internal/test-matrix/generator.ts
nx-plugin-for-aws: Removed — the CDK path no longer generates `tf-infra`.

docs/src/content/docs/en/get_started/tutorials/contribute-generator.mdx
nx-plugin-for-aws: Done, in the English source and all eight translations.

docs/src/content/docs/en/get_started/tutorials/contribute-generator.mdx
nx-plugin-for-aws: Removed.

CONTRIBUTING.md
nx-plugin-for-aws: Done — same bullet wording, and the rationale paragraph dropped.

packages/nx-plugin/src/utils/format.ts
nx-plugin-for-aws: Done — `formatFilesInSubtree` now takes an optional third argument, `{ ignore }`, and `ts#react-website` passes its own route tree path (`joinPathFragments(libraryRoot, 'src', 'routeTree.gen.ts')`, only when `tanstackRouter` is on). Per Jack these are explicit workspace-relative paths rather than globs. Also exported the option type from `sdk/utils/format`.

One wrinkle worth flagging: the paths have to be retained per tree rather than scoped to the call that passes them. `formatFilesInSubtree` formats every change pending in the tree, not only the ones its caller made — so on the composed matrix's shared tree, the *next* generator's plain `formatFilesInSubtree(tree)` reformats the route tree and the non-idempotency is back. I confirmed that empirically before settling on it. So the list is held in a `WeakMap<Tree, Set<string>>` and a path stays ignored for the rest of the run once its owner declares it. There's a regression test for exactly that ("should keep ignoring a file on later calls for the same tree").

cogwirrel (approved):

PR #1007, opened 2026-07-29
fix(api-constructs): include CloudFront custom domain aliases in restrictCorsTo

AlexTo:
### Reason for this change

When a website's CloudFront distribution has a custom domain configured (via `ts#website`'s `domainNames`/`certificate` props), `restrictCorsTo` only ever allow-listed the distribution's default `*.cloudfront.net` domain. A browser loading the site from the custom domain sends that domain as the `Origin` header, which never matches the allow-listed `*.cloudfront.net` origin — so the CORS preflight fails and the browser blocks all subsequent API requests.

### Description of changes

- Added a shared `findCloudFrontDomainNames` helper (`common/constructs/src/core/cloudfront.ts`) that reads a CloudFront distribution's aliases directly off its `CfnDistribution`, returning the default domain plus any custom domains.
- Updated `restrictCorsTo` in both the REST and HTTP API app constructs to use this helper, so custom domains are now included automatically alongside the default CloudFront domain.
- Refactored `UserIdentity`'s Cognito callback/logout URL logic to reuse the same shared helper instead of duplicating the alias-extraction logic.
- Updated the `dungeon-game` tutorial docs (all 9 locales) to keep the generated code excerpt in sync with the fix.

- Rendered the updated templates through the `trpc`, `fastapi`, `smithy`, and `cognito-auth` generator test suites and confirmed the output matches expectations; updated snapshots accordingly.
- Deployed a workspace with a custom domain configured on the CloudFront distribution and confirmed the API's CORS preflight (`OPTIONS`) response returns the correct `Access-Control-Allow-Origin`, both when accessing via the custom domain and via the default CloudFront domain.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/migration.ts
@@ -1,0 +1,169 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+  type MigrationReturnObject,
+  type Tree,
+  visitNotIgnoredFiles,
+} from '@nx/devkit';
+import {
+  addDestructuredImport,
+  addStarExport,
+  applyGritQL,
+} from '../../../utils/ast';
+import { formatFilesInSubtree } from '../../../utils/format';
+import { isEsmWorkspace } from '../../../utils/module-format';
+import {
+  PACKAGES_DIR,
+  SHARED_CONSTRUCTS_DIR,
+} from '../../../utils/shared-constructs-constants';
+
+/**
+ * Include CloudFront custom domain aliases in restrictCorsTo and UserIdentity callback URLs
+ *
+ * How to write a migration:
+ * - https://nx.dev/docs/kb/migration-generators
+ * - What `nextSteps` means: https://nx.dev/docs/reference/devkit/MigrationReturnObject
+ *
+ * Guardrails:
+ * - Pattern-match before writing: skip files that have diverged from the shape
+ *   your generators produce and report them via `nextSteps`, rather than
+ *   clobbering the user's changes.
+ * - Idempotent: re-running must be a no-op.
+ * - Format what you write: finish with `formatFilesInSubtree` so the files your
+ *   migration wrote are formatted correctly.
+ */
+
+const CORE_DIR = `${PACKAGES_DIR}/${SHARED_CONSTRUCTS_DIR}/src/core`;
+const APIS_APP_DIR = `${PACKAGES_DIR}/${SHARED_CONSTRUCTS_DIR}/src/app/apis`;
+const CLOUDFRONT_CORE_FILE = `${CORE_DIR}/cloudfront.ts`;
+const CORE_INDEX_FILE = `${CORE_DIR}/index.ts`;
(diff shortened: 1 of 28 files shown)


cogwirrel (dismissed):
Looks great, thank you! I wonder if you'd be able to write a migration for the fix? :) I'll ping you offline :)

packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/migration.ts
```diff
@@ -0,0 +1,191 @@
...
+    if (!contents.includes(OLD_ALLOWED_ORIGINS_BLOCK)) {
+      agentContext.push(
+        `${filePath}: restrictCorsTo has diverged from the generated shape - could not confirm/apply the CloudFront custom domain fix.`,
+      );
+      return;
+    }
+    let updated = contents.replace(
+      OLD_ALLOWED_ORIGINS_BLOCK,
+      NEW_ALLOWED_ORIGINS_BLOCK,
+    );
```
cogwirrel: Can we use gritql instead of regex? :) Same for the other code changes :)

packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/migration.ts
```diff
@@ -0,0 +1,191 @@
...
+    if (!updated.includes(`from '${apiCloudFrontImportSpecifier}'`)) {
+      updated = updated.replace(
+        CORE_API_UTILS_IMPORT_RE,
+        `$1import { findCloudFrontDomainNames } from '${apiCloudFrontImportSpecifier}';\n`,
+      );
+    }
```
cogwirrel: We can use the `addDestructuredImport` helper here instead I think :) Same below

packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/prompt.md
```diff
@@ -0,0 +1,58 @@
+# Include CloudFront custom domain aliases in restrictCorsTo and UserIdentity callback URLs
```
cogwirrel: I'm wondering whether it might be overkill to have this as hybrid after all, as after some research it looks like there's no way to skip the agentic portion - I reckon let's get rid of the prompt (and the pointer in migrations.json) and keep just the deterministic portion :)

I'll update the guidance as I'd assumed the migration generator could tell Nx that no agentic part is needed :)

packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/migration.ts
AlexTo: done, GritQL works now after the fix for `FunctionProps` landed

packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/migration.ts
AlexTo: done

packages/nx-plugin/src/migrations/latest/restrict-cors-to-custom-domains/prompt.md
AlexTo: `prompt.md` and the entry in `migrations.json` removed

cogwirrel (approved):
Great stuff, thanks @AlexTo !

PR #1008, opened 2026-07-29
chore(dev): add @astral-sh/ruff-wasm-nodejs to root devDependencies for pnpm link workflow

AlexTo:
### Reason for this change

When developing locally using `pnpm link dist/packages/nx-plugin`, any generator that formats Python fails with:

```
Error: Cannot find package '@astral-sh/ruff-wasm-nodejs' imported from .../dist/packages/nx-plugin/src/utils/ruff.js
```

This happens because the linked dist package resolves its runtime dependencies from the workspace root `node_modules`, not from inside the dist directory. Since `@astral-sh/ruff-wasm-nodejs` (added in #1006 for in-process Python formatting) was only declared as a dependency in `packages/nx-plugin/package.json`, it is absent from the root workspace `node_modules` and therefore unavailable during the `pnpm link` dev workflow. Same root cause as #782's `jiti` fix.

### Description of changes

Add `@astral-sh/ruff-wasm-nodejs` (pinned to the same version already used in `packages/nx-plugin/package.json`) to the root `package.json` devDependencies, and update `pnpm-lock.yaml` accordingly.

Confirmed the version matches the existing declaration in `packages/nx-plugin/package.json`, and that `pnpm install` resolves and installs the package into the root `node_modules`, mirroring the fix applied for the identical `jiti` issue in #782.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "devDependencies": {
     "@a2a-js/sdk": "0.3.14",
     "@apidevtools/swagger-parser": "12.1.0",
+    "@astral-sh/ruff-wasm-nodejs": "0.15.22",
     "@astrojs/react": "6.0.1",
     "@astrojs/starlight": "0.41.3",
     "@astrojs/starlight-tailwind": "5.0.0",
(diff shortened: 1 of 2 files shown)


cogwirrel (approved):
Oops, thank you!

cogwirrel:
Added a unit test that fails the build when we miss these! :) https://github.com/awslabs/nx-plugin-for-aws/pull/1010

PR #1009, opened 2026-07-29
fix(api-constructs): replace legacy FunctionProps angle-bracket cast with as syntax

AlexTo:
### Reason for this change

While working on #1007, a GritQL-based rewrite surfaced a blocker: the generated API constructs' `defaultIntegrations` uses the legacy angle-bracket type assertion `<FunctionProps>{ ... }`. This is functionally identical to the modern `{ ... } as FunctionProps` form (verified both compile to identical JS output), but GritQL's TypeScript parser cannot handle the legacy angle-bracket form. It silently fails to parse the *entire containing file*, which blocks any GritQL-based pattern matching anywhere else in that file, including in unrelated methods like `restrictCorsTo`. Since #1007 and future GritQL-based migrations need to operate on these generated API construct files, this legacy syntax needs to go first.

### Description of changes

- Replaced `<FunctionProps>{ ... }` with `{ ... } as FunctionProps` in both the REST and HTTP API constructs templates (`api-constructs/files/cdk/app/apis/{rest,http}`).
- Added a `modernize-function-props-cast` migration so existing workspaces pick up the same change: it locates generated API app construct files, rewrites the cast if it still matches the generated shape, and reports (without touching) any file that has diverged.

- Rendered the updated templates through the `trpc`, `fastapi`, and `smithy` generator test suites and confirmed the output matches expectations; updated snapshots accordingly.
- Unit tested the migration (apply / skip-customised / idempotent).
- Ran the migration against a real project generated on a previous version and confirmed it still builds and works correctly afterwards.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
+++ b/docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
@@ -241,7 +241,7 @@ export class GameApi<
   public static defaultIntegrations = (scope: Construct) => {
     return IntegrationBuilder.rest({
       operations: routerToOperations(appRouter),
-      defaultIntegrationOptions: <FunctionProps>{
+      defaultIntegrationOptions: {
         runtime: Runtime.NODEJS_LATEST,
         handler: 'index.handler',
         code: Code.fromAsset(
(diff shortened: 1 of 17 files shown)


cogwirrel (approved):
Great find, thank you!

PR #1012, opened 2026-07-30
chore(terraform#project): migrate existing workspaces to adopt an existing state bucket

nx-plugin-for-aws:
### Reason for this change

#1011 fixed the vended `bootstrap.ts` but shipped no migration. Per CONTRIBUTING, any change to a generator's output needs one — otherwise a workspace generated before the fix silently diverges from one generated today, and in this case keeps the defect: if `bootstrap.tfstate` is lost while the state bucket survives, `bootstrap` reads the 404 as a first run and asks terraform to create a bucket that already exists, failing with `BucketAlreadyOwnedByYou` on every run with no retry that clears it.

### Description of changes

Deterministic, not hybrid. CONTRIBUTING prefers hybrid by default, but here neither half would be shared: the edit has a single recognisable shape, and anything the codemod can't match needs no agent judgement — just the one-line `terraform import` recovery, which `nextSteps` states directly.

**Every edit is a GritQL rewrite, matched on the AST rather than on literal text.** Users own this file, so formatting drift and nearby edits are expected; literal anchors would skip a workspace that plainly still needs the fix. Matching structurally makes the migration strictly more capable — a customised script now *gets the fix with the customisation preserved*, where a string-based version could only report it and walk away.

- **Discovers projects via generator metadata** (`terraform#project`) through the Nx project graph, not a path glob — terraform *libraries* carry the same metadata but never vend `bootstrap.ts`, so the `tree.exists` filter drops them.
- **All-or-nothing site check**, now keyed on the GritQL patterns: every edit site must match before anything is written, so a script genuinely missing one (e.g. state fetched without `GetObjectCommand`) is reported via `nextSteps` rather than part-rewritten.
- **`haveState` is anchored on `try_statement()` containing `GetObjectCommand`**, which places the declaration exactly where the generator puts it. Note `catch ($p)` does *not* match the vended `catch (err: any)` — the typed parameter needs `try_statement()` or `catch ($p: any)`.
- **The `HeadBucketCommand` import is rewritten into the generator's position** rather than appended, so a migrated workspace matches a freshly generated one.
- **The header comment uses `comment() where { $c <: includes ... }`** — GritQL has no pattern that matches JSDoc text, so the comment node is matched and replaced wholesale.

9 unit tests, all passing: applies to the vended shape, no-op with no terraform project, already-migrated is silent, **customised script migrated with the customisation intact**, **reformatted script still migrated**, script missing an edit site skipped + reported, terraform library untouched, idempotent. The central assertion reads today's template **from disk** and requires byte-identical output, so the suite fails if the template later moves on without the migration following.

I mutation-tested the assertions that matter, since a migration test that passes against a broken migration is worse than none:

- Removing the import rewrite → byte-identical test fails.
- Removing the all-sites guard → the missing-edit-site test fails.

I also ran the migration against the **real** pre-fix template extracted from `990ffb82^` (not just a derived fixture) and diffed the result: byte-identical to today's generator output.

Manually, against a real workspace with the local build (`pnpm package:all`, `nx migrate --run-migrations` via a hand-written `migrations.json`):

1. **Diff is exactly the fix** — doc comment, import, `bucketExists`, `haveState`, import step. Nothing else.
2. **Byte-identical to fresh generation** — migrated `bootstrap.ts` `diff`s clean against the current template.
3. **Builds** — `nx run-many --target build --all` passes.
4. **Idempotent** — second run reports "No changes were made".
5. **Customised workspace** — added a `terraform workspace select` call in the targeted region; the migration applied the fix, preserved the customisation, and the workspace still builds.

One gap worth stating: CONTRIBUTING asks for a baseline on the **published** version, which I could not build. `create-nx-workspace` with `--preset=@aws/nx-plugin@1.0.0-rc.49` fails resolving `@biomejs/wasm-nodejs@2.5.4`, which is not on npm (latest is 2.4.8) — an unrelated packaging issue in the published rc. I used the local build with `bootstrap.ts` reverted to the exact pre-fix content from git, which reproduces the same starting state. Worth a look separately, since it likely blocks real `create-nx-workspace` runs on rc.49.

### Issue # (if applicable)

Follow-up to #1011.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/migrations.json
+++ b/packages/nx-plugin/migrations.json
@@ -9,6 +9,10 @@
     "latest-restrict-cors-to-custom-domains": {
       "description": "Include CloudFront custom domain aliases in restrictCorsTo and UserIdentity callback URLs",
       "implementation": "./src/migrations/latest/restrict-cors-to-custom-domains/migration"
+    },
+    "latest-terraform-bootstrap-adopt-existing-bucket": {
+      "description": "Adopt an existing Terraform state bucket in the vended bootstrap script when its state object is missing",
+      "implementation": "./src/migrations/latest/terraform-bootstrap-adopt-existing-bucket/migration"
     }
   }
 }
--- a/packages/nx-plugin/src/migrations/latest/terraform-bootstrap-adopt-existing-bucket/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/terraform-bootstrap-adopt-existing-bucket/migration.spec.ts
@@ -1,0 +1,222 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import type { Tree } from '@nx/devkit';
+import { terraformProjectGenerator } from '../../../terraform/project/generator';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test';
+import migration from './migration';
+
+const BOOTSTRAP_FILE = 'packages/infra/scripts/bootstrap.ts';
+
+// Today's vended script, read from the template the generator uses — so this
+// suite fails if the template moves on without the migration following it.
+const CURRENT_TEMPLATE = readFileSync(
+  join(
+    import.meta.dirname,
+    '../../../terraform/project/files/application/scripts/bootstrap.ts.template',
+  ),
+  'utf-8',
+);
+
+// The script as generated before the fix — verbatim, so the "before" state is
+// exactly what users are upgrading from rather than something derived.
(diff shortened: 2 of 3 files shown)


cogwirrel (approved):

nx-plugin-for-aws:
Correction to the validation notes above: the claim that `@biomejs/wasm-nodejs@2.5.4` "is not on npm (latest is 2.4.8)" is **wrong**, and there is no packaging issue with rc.49.

That lookup was hitting a stale local npm mirror on my dev host. Querying `registry.npmjs.org` directly shows 139 versions with `2.5.4` present (latest `2.5.6`), and `npm install @aws/nx-plugin@1.0.0-rc.49` succeeds once the registry is set correctly.

So the "could not build a baseline on the published version" caveat does not apply either — a published-version baseline is achievable, and I'm using one to validate the released migration. Nothing about the migration itself changes; only that note was incorrect.

PR #1016, opened 2026-08-01
fix(ts#website#auth): count the SSRF WAF rule blocking localhost sign-in

nx-plugin-for-aws:
### Reason for this change

Signing in against a local dev server fails with an opaque `403 Forbidden` before the Cognito login page renders.

The `UserIdentity` Web ACL applies `AWSManagedRulesCommonRuleSet`, whose `EC2MetaDataSSRF_QUERYARGUMENTS` rule inspects query arguments for what looks like an SSRF attempt against the EC2 instance metadata endpoint — and it treats a loopback address as one. Running the website locally sends `redirect_uri=http://localhost:4200` to the Hosted UI, so the rule matches and WAF blocks the request.

The generator already whitelists `http://localhost:4200` in the user pool client's callback URLs, so the intent is clearly that local sign-in should work; the WAF silently prevents it.

Reproduced against a deployed workspace — only the loopback host is affected:

```
redirect_uri=https://d2cs8rjqegjl78.cloudfront.net → 302 ✓
redirect_uri=http://localhost:4200                 → 403 ✗
redirect_uri=https://localhost:4200                → 403 ✗
redirect_uri=https://127.0.0.1:4200                → 403 ✗
redirect_uri=https://example.com                   → 302 ✓
```

WAF logs name the rule directly:

```
action: BLOCK
terminatingRule: CRSRule
  AWSManagedRulesCommonRuleSet → EC2MetaDataSSRF_QUERYARGUMENTS
```

### Description of changes

The local dev server origins are lifted into a named `LOCAL_CALLBACK_URLS` constant (CDK) / `local.local_callback_urls` (Terraform), and `EC2MetaDataSSRF_QUERYARGUMENTS` is overridden to [Count](https://docs.aws.amazon.com/waf/latest/developerguide/web-acl-rule-action.html#web-acl-rule-action-count) **only while that list is non-empty**. Removing the local callback URLs restores the rule to Block, so the relaxation can't outlive the reason for it.

Matches are still recorded in the Web ACL metrics and WAF logs while counted, every other rule in the group — including the rest of the SSRF protections — still blocks, and the client's callback URL allowlist continues to constrain where users can actually be redirected.

A deterministic migration (`latest-user-identity-waf-allow-localhost-callback`) applies the same change to existing workspaces. Every edit site is expressed as a GritQL rewrite — TypeScript for the CDK construct, HCL for the Terraform module — so the Web ACL is matched on its AST rather than its formatting, and all sites are confirmed present before any are written so a partly-matching file is left whole. A workspace whose Web ACL has diverged from the generated shape is skipped and reported via `nextSteps`.

The `react-website-auth` guide gains a short caution block noting the rule is counted, why, and that removing the local callback URLs restores it to Block.

**Unit tests** — 14 cases parameterised over both IaC providers, including the idempotency cases required by CONTRIBUTING. Rather than a hardcoded fixture, each test generates a website with auth and reverses each of the migration's edits, so the "before" state can't drift away from what the generators actually vend (the rewind throws if it stops matching). Covered: the migrated file is byte-identical to a freshly generated one; a re-run is a no-op that reports nothing; the override is derived from the callback URLs rather than restating the condition; the `KnownBadInputs` rule group is untouched; a Web ACL with a custom managed rule group is skipped and reported; a Web ACL with an extra user-added rule is migrated without disturbing it. Full suite passes (2882 tests), along with `tsc`, Biome lint/format and checkov.

**The conditional actually toggles** — synthesised a CDK workspace both ways:

```
with local callback URLs    → RuleActionOverrides: [{Count, EC2MetaDataSSRF_QUERYARGUMENTS}]
with the list emptied       → RuleActionOverrides: null   (rule reverts to Block)
```

For Terraform, `terraform validate` passes on the generated module both with the local callback URLs present and with the list emptied, confirming the `dynamic` block is valid HCL either way.

**Deployed e2e** — a fresh CDK workspace (website + auth + tRPC API + infra) deployed to AWS. Confirmed the override is live on the deployed Web ACL, then verified `http://localhost:4200` returns 302 where it previously returned 403.

**Local sign-in via `serve`, not `dev`** — deliberately tested the `serve` target, which serves the real deployed `runtime-config.json` and calls the deployed API, rather than `dev` which redirects the API to localhost. Completed the full Hosted UI flow in a browser (username → password → TOTP MFA → callback), landed back on the app authenticated, and confirmed the deployed API returned `200` from `http://localhost:4200` with the real Cognito token (CORS preflight `204`, then `200`).

This confirms the WAF override is the **only** change required — no HTTPS dev server, no extra callback URLs and no CORS changes were needed.

**Migration e2e** — rewound the same deployed workspace's construct to the pre-fix shape, ran the migration from the built package, and confirmed it produced a file byte-identical to the template, that a second run changed nothing, that the workspace still built, and that redeploying put the override back on the live Web ACL with `http://localhost:4200` returning 302 again.

All provisioned resources were torn down afterwards.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/user-identity-waf-allow-localhost-callback/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/user-identity-waf-allow-localhost-callback/migration.spec.ts
@@ -1,0 +1,165 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Tree } from '@nx/devkit';
+import { tsReactWebsiteGenerator } from '../../../ts/react-website/app/generator';
+import { tsWebsiteAuthGenerator } from '../../../ts/website/auth/generator';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test';
+import migration from './migration';
+
+const CDK_FILE = 'packages/common/constructs/src/core/user-identity.ts';
+const TERRAFORM_FILE =
+  'packages/common/terraform/src/core/user-identity/identity/identity.tf';
+
+const SSRF_RULE = 'EC2MetaDataSSRF_QUERYARGUMENTS';
+
+/**
+ * Generate a website with auth, then rewind the Web ACL to the shape the
+ * generators vended before this fix — so the "before" state is derived from
+ * today's output rather than a fixture that can drift away from it.
+ */
+const generateAndRewind = async (tree: Tree, iac: 'cdk' | 'terraform') => {
+  await tsReactWebsiteGenerator(tree, {
+    name: 'test-website',
+    iac,
+    ux: 'cloudscape',
+  });
+  await tsWebsiteAuthGenerator(tree, {
+    project: 'test-website',
+    allowSignup: false,
+    iac,
+  });
+
+  const filePath = iac === 'cdk' ? CDK_FILE : TERRAFORM_FILE;
+  const current = tree.read(filePath, 'utf-8') ?? '';
+
+  // Cut from the end of the vendor name line to the statement's closing brace,
+  // matched by indentation so nothing after the override is left orphaned.
+  const [anchor, closer, replacement] =
+    iac === 'cdk'
+      ? ["              vendorName: 'AWS',\n", '\n            },\n', '']
+      : ['        vendor_name = "AWS"\n', '\n      }\n', ''];
+
+  const start = current.indexOf(anchor) + anchor.length;
+  const end = current.indexOf(closer, start);
+  const rewound =
+    current.slice(0, start) + replacement + current.slice(end + 1);
+
+  tree.write(filePath, rewound);
+  return { filePath, current, rewound };
+};
+
+describe('user-identity-waf-allow-localhost-callback migration', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeUsingTsSolutionSetup();
+  });
+
+  it('should be a no-op when the workspace has no user identity', async () => {
+    const result = await migration(tree);
+    expect(result.nextSteps).toEqual([]);
+  });
+
+  describe.each(['cdk', 'terraform'] as const)('%s', (iac) => {
+    it('should start from a fixture that lacks the override', async () => {
+      // Guards the rewind: if it left the override in place, every assertion
+      // below would pass without the migration doing anything.
+      const { current, rewound } = await generateAndRewind(tree, iac);
+      expect(current).toContain(SSRF_RULE);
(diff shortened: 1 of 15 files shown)


packages/nx-plugin/src/migrations/latest/user-identity-waf-allow-localhost-callback/migration.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+  'packages/common/terraform/src/core/user-identity/identity/identity.tf';
+
+const SSRF_RULE = 'EC2MetaDataSSRF_QUERYARGUMENTS';
+
+/**
+ * Each entry reverses one of the migration's edits, so the "before" state is
+ * derived from today's generator output rather than a fixture which can drift
+ * away from it.
+ */
+const REWINDS: Record<'cdk' | 'terraform', Array<[string, string]>> = {
```
cogwirrel: No - this will cause drift/problems in future. Just store a string with the pre-change state of the file

nx-plugin-for-aws: Agreed, thanks - swapped both over to verbatim pre-fix strings. I captured them by generating from the templates at the pre-fix commit rather than hand-writing them, so they're exactly what users are upgrading from. The `should produce exactly what the current generator vends` test still asserts the migrated output is byte-identical to freshly generated output, so if a template moves on without the migration following it that test fails rather than the fixture silently drifting.

packages/nx-plugin/src/migrations/latest/user-identity-waf-allow-localhost-callback/migration.spec.ts
```diff
@@ -0,0 +1,1011 @@
...
+      const { filePath, current } = await generateWithPreFixConstruct(
+        tree,
+        iac,
+      );
+
+      const result = await migration(tree);
+
+      // The point of a migration: the workspace ends up byte-identical to one
+      // generated from today's generators.
+      expect(tree.read(filePath, 'utf-8')).toEqual(current);
```
cogwirrel: Today's may change in future though and this migration shouldn't ever change, as it's job is to bring it to the next version we will release with this change. Save the expected output as a variable too.

nx-plugin-for-aws: Good point - pinned the expected output too, so both ends are now fixed strings and the migration is decoupled from whatever the templates vend later. Also added a fixture guard asserting the pre-fix state lacks the override and the expected state contains it, so neither can silently go stale.

cogwirrel (approved):

PR #1017, opened 2026-08-02
feat(ts#agent): add session management support

AlexTo:
### Reason for this change

`ts#agent`-generated agents had no built-in way to persist conversation state (message history, tool state) across invocations. Since Bedrock AgentCore Runtime can scale or restart between requests, this meant conversations couldn't survive beyond a single process lifetime.

### Description of changes

- Add a `session` option to `ts#agent`: `s3` (default) or `in-memory`.
- Generate a `session.ts` exporting `getSessionManager()`, wired into `agent.ts` for HTTP/A2A agents (directly on the `Agent` constructor) and into `index.ts` for AG-UI agents (via `sessionManagerProvider`, since AG-UI clones a template agent per thread).
- When `session: s3`, CDK/Terraform provision a dedicated S3 bucket (encrypted with a dedicated KMS key, all public access blocked, server access logs delivered to CloudWatch) and grant the agent's IAM role read/write access; the bucket name is registered in AppConfig alongside the agent's runtime ARN.
- Local development always uses `LocalFileStorage` under `tmp/agents/strands/<agent-name>` at the workspace root, regardless of the configured `session` option.
- Add `@aws-sdk/client-s3` as an explicit dependency when `session: s3`, so the bundler can resolve the Strands SDK's lazy `import('@aws-sdk/client-s3')` — without it, the deployed container (which ships no `node_modules`) fails at runtime.
- Add a `ts-agent-session-management-support` Nx migration to bring existing (pre-feature) workspaces up to the new shape: reshapes `agentRuntimes` entries to `{ arn, session }`, backfills `session.ts` and wires it into existing agents, and migrates the error-logging hooks (both AG-UI and HTTP/A2A) to `StrandsAgent`/`Agent` constructor plugins.
- `py#agent` also gains the `session` option (schema only, always `in-memory`) for API parity — Python session manager support isn't implemented yet.
- Docs: new "Session Management" section in the `ts#agent` guide; fixed stale `FileTree`/code samples that predated this change.

- Unit tests.
- Generated a baseline test project with the pre-session-feature version of `ts#agent`/`py#agent`, including MCP server, A2A, and React website connections, in [this commit](https://github.com/AlexTo/test-ts-agent-session/commit/a9f2422170ab6fa9a7b16cb4fb9a9e3747bbcf72).
- Applied the migration in [this commit](https://github.com/AlexTo/test-ts-agent-session/commit/de3eaf5e32d76eb452b976c3d8d91cc023bb5f57) to verify existing HTTP/A2A/AG-UI agents (TypeScript and Python), along with their MCP/A2A/React connections, were correctly upgraded — `session.ts` backfilled and wired in, error-logging hooks migrated to `StrandsAgent` plugins for the AG-UI agent, `agentRuntimes` reshaped, and connection client/provider code updated to read the new `{ arn, session }` shape.
- Generated new agents with `session: s3`, an A2A connection between them, and wired them to CDK infra in [this commit](https://github.com/AlexTo/test-ts-agent-session/commit/3d8deb03da4f874836d111e71e1deef3acd253c1), deployed to AWS, and verified the sessionId flows through end-to-end — each connected agent persists its session under the same sessionId, in its own dedicated S3 bucket. Also chatted with the agents locally to verify each agent's local session directory (`tmp/agents/strands/<agent-name>`) is created and populated correctly.

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/ts-agent.mdx
+++ b/docs/src/content/docs/en/guides/ts-agent.mdx
@@ -60,6 +60,7 @@ The generator will add the following files to your existing TypeScript project. 
         - init.ts tRPC initialization
         - router.ts tRPC router with agent procedures
         - agent.ts Main agent definition with sample tools
+        - session.ts Resolves the SessionManager used to persist conversation state
         - client.ts Vended client for invoking your agent
         - agent-core-trpc-client.ts Client factory for connecting to agents on AgentCore Runtime
         - Dockerfile Entry point for hosting your agent (excluded when `infra` is set to `None`)
@@ -79,6 +80,7 @@ The entry point uses the [Strands A2A Express Server](https://strandsagents.com/
       - agent/ (or custom name if specified)
         - index.ts A2A Express server entry point
         - agent.ts Main agent definition with sample tools
+        - session.ts Resolves the SessionManager used to persist conversation state
         - Dockerfile Entry point for hosting your agent (excluded when `infra` is set to `None`)
     - package.json Updated with Strands and Express dependencies
     - project.json Updated with agent serve targets
@@ -96,6 +98,7 @@ The entry point uses [`@ag-ui/aws-strands`](https://www.npmjs.com/package/@ag-ui
       - agent/ (or custom name if specified)
         - index.ts AG-UI server entry point (Express + SSE)
         - agent.ts Main agent definition with sample tools
+        - session.ts Resolves the SessionManager used to persist conversation state
         - Dockerfile Entry point for hosting your agent (excluded when `infra` is set to `None`)
     - package.json Updated with Strands and AG-UI dependencies
     - project.json Updated with agent serve targets
@@ -185,7 +188,7 @@ const letterCounter = tool({
 });
 
 // Add tools to your agent
-export const getAgent = async (sessionId: string) => {
+export const getAgent = async () => {
   return new Agent({
     systemPrompt: 'You are a helpful assistant with access to various tools.',
     tools: [letterCounter],
@@ -347,6 +350,37 @@ Your agent is automatically configured with observability using the [AWS Distro 
 You can find traces in the CloudWatch AWS Console, by selecting "GenAI Observability" in the menu. Note that for traces to be populated you will need to enable [Transaction Search](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Transaction-Search.html).
 
 For more details, refer to the [AgentCore documentation on observability](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability-configure.html).
+
+### Session Management
+
+The `session` option controls how your agent persists conversation state (message history, tool state, etc.) across invocations, using the Strands SDK's [SessionManager](https://strandsagents.com/docs/user-guide/concepts/agents/session-management/):
+
+- **`s3`** (default): The CDK/Terraform infrastructure provisions a dedicated S3 bucket for session data, encrypted with SSE-S3 and with all public access blocked. The agent's IAM role is granted read/write/list/delete access to it, and the bucket name is registered alongside the agent's ARN in AppConfig runtime configuration.
+- **`in-memory`**: No bucket is provisioned. Conversation state is kept in memory only for the lifetime of the running process and does not survive restarts or scale-in.
+
+This is implemented in the generated `session.ts`, which exports a `getSessionManager()` function resolving a `SessionManager` for the current session:
+
+```typescript
+export const getSessionManager = async (): Promise<SessionManager> => {
+  const sessionId = getCurrentSessionId();
+  // local development always uses local file storage...
+  // ...otherwise S3Storage (session: 's3') or InMemoryStorage (session: 'in-memory')
+};
+```
+
+<OptionFilter when={{ protocol: 'ag-ui' }} description="AG-UI session wiring">
+Since AG-UI clones a template agent per conversation thread, `getSessionManager` is wired in as a `sessionManagerProvider` on the `StrandsAgent` config in `index.ts`, so a fresh `SessionManager` is resolved for each thread rather than baked into the shared template agent.
+</OptionFilter>
+
+<OptionFilter when={{ protocol: ['http', 'a2a'] }} description="HTTP/A2A session wiring">
+Since `withSessionId` already caches one `Agent` instance per session, `getSessionManager` is called directly inside `getAgent`'s `new Agent({ sessionManager: await getSessionManager() })` call in `agent.ts`.
+</OptionFilter>
+
+The session ID itself comes from the AgentCore Runtime session (propagated via the `x-amzn-bedrock-agentcore-runtime-session-id` header for A2A/AG-UI, or the WebSocket connection context for HTTP/tRPC) and is bound to an [`AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage)-based context so `getCurrentSessionId()` can resolve it anywhere in the request — including in any downstream MCP or A2A clients wired up via the <Link path="/guides/connection">`connection` generator</Link>, so the whole call chain shares a consistent session.
+
+:::note[Local Development]
+When running locally (`LOCAL_DEV=true`, set automatically by the `-dev` target), session data is always stored on disk under `tmp/agents/strands/<agent-name>` at the workspace root, regardless of the configured `session` option, for convenience.
+:::
 </OptionFilter>
 
 ## Invoking your Agent
--- a/docs/src/content/docs/en/guides/ts-agent.mdx
+++ b/docs/src/content/docs/en/guides/ts-agent.mdx
@@ -355,18 +355,10 @@ For more details, refer to the [AgentCore documentation on observability](https:
 
 The `session` option controls how your agent persists conversation state (message history, tool state, etc.) across invocations, using the Strands SDK's [SessionManager](https://strandsagents.com/docs/user-guide/concepts/agents/session-management/):
 
-- **`s3`** (default): The CDK/Terraform infrastructure provisions a dedicated S3 bucket for session data, encrypted with SSE-S3 and with all public access blocked. The agent's IAM role is granted read/write/list/delete access to it, and the bucket name is registered alongside the agent's ARN in AppConfig runtime configuration.
+- **`s3`** (default): The CDK/Terraform infrastructure provisions a dedicated S3 bucket for session data, encrypted with a dedicated KMS key and with all public access blocked; server access logs are delivered to a CloudWatch Logs log group via the same key. The agent's IAM role is granted read/write/list/delete access to the bucket and decrypt/generate-data-key access to the key, and the bucket name is registered alongside the agent's ARN in AppConfig runtime configuration.
 - **`in-memory`**: No bucket is provisioned. Conversation state is kept in memory only for the lifetime of the running process and does not survive restarts or scale-in.
 
-This is implemented in the generated `session.ts`, which exports a `getSessionManager()` function resolving a `SessionManager` for the current session:
-
-```typescript
-export const getSessionManager = async (): Promise<SessionManager> => {
-  const sessionId = getCurrentSessionId();
-  // local development always uses local file storage...
-  // ...otherwise S3Storage (session: 's3') or InMemoryStorage (session: 'in-memory')
-};
-```
+This is implemented in the generated `session.ts`, which exports a `getSessionManager()` function resolving a `SessionManager` for the current session.
 
 <OptionFilter when={{ protocol: 'ag-ui' }} description="AG-UI session wiring">
 Since AG-UI clones a template agent per conversation thread, `getSessionManager` is wired in as a `sessionManagerProvider` on the `StrandsAgent` config in `index.ts`, so a fresh `SessionManager` is resolved for each thread rather than baked into the shared template agent.
--- a/packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
@@ -13,6 +13,7 @@ import { applyGritQL, captureGritQL, matchGritQL } from '../../../utils/ast';
 import { formatFilesInSubtree } from '../../../utils/format';
 import { isEsmWorkspace } from '../../../utils/module-format';
 import { kebabCase } from '../../../utils/names';
+import { getRelativePathToRootByDirectory } from '../../../utils/paths';
 
 /**
  * Add session management support to ts#agent.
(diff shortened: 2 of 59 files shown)


docs/src/content/docs/en/guides/ts-agent.mdx
```diff
@@ -347,6 +350,37 @@ Your agent is automatically configured with observability using the [AWS Distro
...
+
+This is implemented in the generated `session.ts`, which exports a `getSessionManager()` function resolving a `SessionManager` for the current session:
+
+```typescript
+export const getSessionManager = async (): Promise<SessionManager> => {
+  const sessionId = getCurrentSessionId();
+  // local development always uses local file storage...
+  // ...otherwise S3Storage (session: 's3') or InMemoryStorage (session: 'in-memory')
+};
+```
```
cogwirrel: Nit: probably don't need this example :)

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
```diff
@@ -647,6 +653,44 @@ export class SnapshotBedrockAgent
       platform: Platform.LINUX_ARM64,
     });
 
+    const sessionBucket = new Bucket(this, 'SessionBucket', {
+      enforceSSL: true,
+      autoDeleteObjects: true,
+      removalPolicy: RemovalPolicy.DESTROY,
```
cogwirrel: While useful for testing, I think it's safer not to delete data on destroy, like we do for databases :)

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
```diff
@@ -647,6 +653,44 @@ export class SnapshotBedrockAgent
...
+    suppressRules(
+      sessionBucket,
+      ['CKV_AWS_21'],
+      'Session data does not need versioning enabled',
+    );
+    suppressRules(
+      sessionBucket,
+      ['CKV_AWS_145'],
+      'AES256 (S3-managed) encryption is sufficient for session data',
+    );
```
cogwirrel: Probably better to use a CMK :)

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
```diff
@@ -647,6 +653,44 @@ export class SnapshotBedrockAgent
...
+    suppressRules(
+      sessionBucket,
+      ['CKV2_AWS_62'],
+      'Event notifications not required for session data',
+    );
+    suppressRules(
+      sessionBucket,
+      ['CKV_AWS_18'],
+      'Access logging not required for session data',
+    );
```
cogwirrel, line 806: I think we should follow the best practice and store access logs in cloudwatch like for other buckets :)

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
```diff
@@ -805,6 +857,36 @@ variable "appconfig_application_arn" {
...
 
+# Bucket storing the agent's session data
+resource "aws_s3_bucket" "session" {
+  #checkov:skip=CKV2_AWS_61:Lifecycle configuration not required for session data
+  #checkov:skip=CKV_AWS_144:Cross-region replication not required for session data
+  #checkov:skip=CKV2_AWS_62:Event notifications not required for session data
+  #checkov:skip=CKV_AWS_18:Access logging not required for session data
+  #checkov:skip=CKV_AWS_21:Session data does not need versioning enabled
+  #checkov:skip=CKV_AWS_145:AES256 (S3-managed) encryption is sufficient for session data
+  force_destroy = true
```
cogwirrel: Same comments as above for s3, best to retain data, use a cmk and store s3 access logs in cloudwatch :)

packages/nx-plugin/src/ts/agent/generator.ts
```diff
@@ -84,6 +84,26 @@ export const tsAgentGenerator = async (
...
+  // (`tmp/agents/strands/<agent-name>`), not inside the project, so each agent
+  // gets its own storage directory. The `-dev`/`-serve` targets run with
+  // cwd={projectRoot}, so compute that directory relative to the project root
+  // here rather than resolving it at runtime (e.g. via import.meta.url),
+  // which would need an extra runtime helper.
+  const projectDepth = project.root.split('/').filter(Boolean).length;
+  const localSessionsDir = joinPathFragments(
+    Array(projectDepth).fill('..').join('/'),
+    `tmp/agents/strands/${name}`,
+  );
```
cogwirrel, line 187: Nit: you can use `getRelativePathToRootByDirectory` :) https://github.com/awslabs/nx-plugin-for-aws/blob/5d440b48adfd62582575f3c8549cb8724dd3fa30/packages/nx-plugin/src/utils/paths.ts#L18

packages/nx-plugin/src/utils/agent-connection/files/core-strands/base/model-errors-strands.ts.template
```diff
@@ -20,17 +24,28 @@ const hasErrorNamed = (error: unknown, name: string): boolean => {
   return false;
 };
 
+/**
+ * Plugin that logs model invocation errors, calling out missing credentials or
+ * denied permissions.
+ */
+export class ModelErrorLoggingPlugin implements Plugin {
```
cogwirrel, line 31: Nice one! :)

packages/nx-plugin/src/utils/agent-core-constructs/files/cdk/app/agent-core/__nameKebabCase__/__nameKebabCase__.ts.template
```diff
@@ -114,12 +160,24 @@ export class <%- nameClassName %> extends Construct implements IGrantable, IConn
...
 <%_ } _%>
     rc.grantReadAppConfig(this.agentCoreRuntime);
 
     rc.set('agentcore', 'agentRuntimes', {
       ...rc.get('agentcore').agentRuntimes,
-      <%- nameClassName %>: this.agentCoreRuntime.agentRuntimeArn,
+      <%- nameClassName %>: {
+        arn: this.agentCoreRuntime.agentRuntimeArn,
+        session: {
+          storage: '<%- session %>',
```
cogwirrel: Do we need to store `storage`? I might have missed it but it doesn't look like anything actually checks this, as the `getSessionManager` is generated to look for the bucketName regardless :)

packages/nx-plugin/src/ts/react-website/agui/files/common/src/hooks/useAgui__agentNameClassName__.tsx.template
```diff
@@ -107,9 +107,11 @@ export const useAgui<%= agentNameClassName %> = (): Record<string, AbstractAgent
...
   const url = useMemo(
     () =>
-      agentRuntimeValue.startsWith('arn:')
-        ? buildAgentCoreUrl(agentRuntimeValue)
-        : agentRuntimeValue,
+      // A string value is a local-dev override URL; otherwise it's the
+      // deployed runtime entry, so build the invocation URL from its ARN.
+      typeof agentRuntimeValue === 'string'
+        ? agentRuntimeValue
+        : buildAgentCoreUrl(agentRuntimeValue.arn),
```
cogwirrel: Do we get a type when the connection generator sets the value which is meant to be an object as a string? :)

cogwirrel (commented):
Looks great, thanks @AlexTo! :) Few minor comments :)

packages/nx-plugin/src/ts/react-website/agui/files/common/src/hooks/useAgui__agentNameClassName__.tsx.template
AlexTo: I think no, because IRuntimeConfig is `any` 
image

packages/nx-plugin/src/utils/agent-core-constructs/files/cdk/app/agent-core/__nameKebabCase__/__nameKebabCase__.ts.template
AlexTo: ah actually no, it is just metadata. removed to simplify the config

packages/nx-plugin/src/ts/react-website/agui/files/common/src/hooks/useAgui__agentNameClassName__.tsx.template
AlexTo: I also realised that the `shape` of the client side `connection` doesn't have to follow the shape of the server side `agentcore` config so it can remain as `agentRuntimes: string`. But maybe it's a good idea to just refactor it into an object `agentRuntimes: { arn: string }` to be consistent and in case we have to include other settings in the future?

docs/src/content/docs/en/guides/ts-agent.mdx
AlexTo: removed

packages/nx-plugin/src/ts/react-website/agui/files/common/src/hooks/useAgui__agentNameClassName__.tsx.template
cogwirrel: Ooh yes true- I think it might be best to keep as a string to keep it simple, we can always refactor later if needed :) probably better to have a bit of friction to do so too as we ideally don’t want to send the bucket name to the front end :)

packages/nx-plugin/src/utils/agent-connection/files/core-strands/base/model-errors-strands.ts.template
AlexTo: actually HTTP/A2A should also use the plugin form instead of calling `log*Errors`, let me update it

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
AlexTo: updated

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
AlexTo: done

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
AlexTo: done

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
AlexTo: done

packages/nx-plugin/src/ts/agent/generator.ts
AlexTo: done

packages/nx-plugin/src/ts/react-website/agui/files/common/src/hooks/useAgui__agentNameClassName__.tsx.template
AlexTo: ah , same shape as in having object type `{ arn: str }` for `connection` without the `session` attr. But I reverted it back to the simple string as done previously.

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
```diff
@@ -0,0 +1,673 @@
...
+// idempotent (nothing to match once removed). The import itself is removed
+// via removeLogErrorsImport below (see its comment for why a plain GritQL
+// rewrite can't express this).
+const AGENT_TS_LOG_MODEL_CALL_PATTERN = '`logModelErrors(agent);` => .';
+const AGENT_TS_LOG_TOOL_CALL_PATTERN = '`logToolErrors(agent);` => .';
+
+// AG-UI is the only protocol whose index.ts imports from '@ag-ui/aws-strands'
+// (see files/ag-ui/index.ts.template), so that's a reliable signal that the
+// sibling agent.ts belongs to an AG-UI agent.
+const isAgUiAgentDir = (tree: Tree, agentTsPath: string): boolean => {
```
cogwirrel: Nit: can we infer this from the component metadata instead? :)

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
```diff
@@ -0,0 +1,673 @@
...
+        anyChanges = true;
+      }
+    }
+  }
+
+  if (anyChanges) {
+    nextSteps.push(
+      'If you have custom code reading agentRuntimes entries as plain ARN strings ' +
+        '(e.g. custom Lambda handlers or agent code), update it to read the ARN from `.arn` ' +
+        'on the new `{ arn }` shape.',
```
cogwirrel: Nit: let's call out that it's runtime config `agentcore.agentRuntimes` entries as users might not immediately know what `agentRuntimes entries` are :)

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
```diff
@@ -0,0 +1,673 @@
...
+      );
+      const rewroteConstructor = await applyGritQL(
+        tree,
+        filePath,
+        AGENT_TS_SESSION_MANAGER_CONSTRUCTOR_PATTERN,
+      );
+      if (rewrotePlugins || rewroteConstructor) {
+        nextSteps.push(
+          `${filePath}: wired sessionManager and the error-logging plugins into the Agent constructor (see ${sessionPath}).`,
+        );
```
cogwirrel: `nextSteps` should be actions for the user where we couldn't make changes, rather than a log of what's been done :) eg we might want to report in the opposite case if we couldn't add the plugins.

Same for a few other `nextSteps` pushes below :)

Also as a general rule it's good to make sure we're fairly "all or nothing" with changes that depend on one another - eg if we can't add the plugins we probably shouldn't add the imports or remove the logModelErrors calls

packages/nx-plugin/src/py/agent/react-connection/__snapshots__/generator.spec.ts.snap
```diff
@@ -312,6 +318,8 @@ export const useAguiTestAgent = (): Record<string, AbstractAgent> => {
 
   const url = useMemo(
     () =>
+      // A local-dev override is a plain URL; otherwise it's the agent's
+      // runtime ARN, so build the invocation URL from it.
```
cogwirrel: Nit: can probably get rid of this comment since it's unchanged :)

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
```diff
@@ -647,6 +663,107 @@ export class SnapshotBedrockAgent
...
+      });
+    const sessionAccessLogsDelivery = new CfnDelivery(
+      this,
+      'SessionAccessLogsDelivery',
+      {
+        deliverySourceName: sessionAccessLogsSource.name,
+        deliveryDestinationArn: sessionAccessLogsDestination.attrArn,
+      },
+    );
+    sessionAccessLogsDelivery.addDependency(sessionAccessLogsSource);
```
cogwirrel, line 844: We should also make sure we order this after the bucket policy too :) see https://github.com/awslabs/nx-plugin-for-aws/pull/1018

cogwirrel (commented):
Looking great @AlexTo! Sorry just a couple more comments :)

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
AlexTo: done, mention reading from runtime config to be updated for custom code if any rewrites above happens

packages/nx-plugin/src/py/agent/react-connection/__snapshots__/generator.spec.ts.snap
AlexTo: removed

packages/nx-plugin/src/ts/agent/__snapshots__/generator.spec.ts.snap
AlexTo: applied the same `addDependency`

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
AlexTo: added a method to infer from component metadata

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
AlexTo: done

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
```diff
@@ -0,0 +1,741 @@
...
+  );
+  const toolPluginReady = await ensurePluginClass(
+    tree,
+    nextSteps,
+    toolErrorsPath,
+    'ToolErrorLoggingPlugin',
+    TOOL_ERRORS_IMPORT_PATTERN,
+    TOOL_ERRORS_CLASS_PATTERN,
+  );
+  const pluginsReady = modelPluginReady && toolPluginReady;
```
cogwirrel: Here we're adding the plugins independently, but then only modify agent below if both succeed, so we could end up in a situation where we don't compile if 1 plugin can't be updated :)

Perhaps we test if the patterns match for both, and if so update both and proceed, otherwise leave both untouched?

cogwirrel (dismissed):
Looks great @AlexTo ! Just 1 comment for the migration :)

packages/nx-plugin/src/migrations/latest/ts-agent-session-management-support/migration.ts
AlexTo: ah right, the migration updates each plugin file individually. Replaced `ensurePluginClass` with `checkPluginClass` to check both are not diverged first to update and declare `pluginsReady` for modifying agent class, otherwise leave untouched.

cogwirrel (approved):

PR #1029, opened 2026-08-03
feat(nx-plugin): sync vended versions baked into generated file bodies

nx-plugin-for-aws:
### Reason for this change

> Supersedes #1034 (now closed), whose commit is the base of this branch. That PR bumped `rolldown` and `rolldown-plugin-dts` in the smithy SSDK bundle image; those pins only *stay* current because of the sync landing here, so the two are one change.

The version sync migration added in #1001 visits `package.json`, `pyproject.toml` and the `required_providers` block of `.tf` files. Versions a generator bakes into the **body** of a generated file were therefore never upgraded.

That set is almost entirely security pins. `ts#agent` and `ts#mcp-server` vend a Dockerfile whose own comment says the overrides "pin transitive dependencies with known HIGH/CRITICAL vulnerabilities to fixed versions" — so the pins most in need of an upgrade were the only ones guaranteed not to get one. A workspace generated today and migrated for a year would still build images against the CVE-affected versions it was scaffolded with.

Verified unreachable by running the real migration over planted content — every one was left untouched:

- `npm pkg set "overrides.@opentelemetry/propagator-jaeger=<version>"` and `overrides.minimatch=<version>`, plus `npm install @aws/aws-distro-opentelemetry-node-autoinstrumentation@<version>`
- `prisma@<version>` and `npm@<version>` in the `ts#rdb` Dockerfile
- `boto3` / `httpx` / `mcp` pins in the `uv run --with` inline scripts of five `.tf` templates
- `BASE_IMAGES` (node, python) and `CONTAINER_VERSIONS.trivy`, the latter landing in a `project.json` target command

### Description of changes

New `sync-embedded-versions.ts`, called from `syncVendedVersions`, covering all of the above.

**Which files are visited.** A Dockerfile is matched on its basename case-insensitively, allowing a per-variant prefix (`build.Dockerfile`, which the smithy templates vend) as well as a suffix (`Dockerfile.migration`). `.tf` and `project.json` are matched by extension.

**Locating the pins.** These sit inside shell commands, and GritQL has no Dockerfile or shell grammar (`language dockerfile` / `sh` / `bash` all error). Rather than fall back to string replace, each pin is matched by an anchored GritQL regex whose single capture group is the version, and **only that capture is rewritten**:

```
r"[\s\S]*overrides\.minimatch=([^\s\\\"]+)[\s\S]*"($version) where { $version => `10.2.5` }
```

The file's own text around the pin is never reconstructed, so a workspace whose Dockerfile was reformatted or hand-edited syncs exactly as a freshly generated one does — the reason for using an AST rewrite rather than a literal anchor. For `.tf`, `template_literal()` reaches inside `<<-EOT` heredocs *and* quoted strings, which is what makes those pins addressable.

The install shape accepts `install`, `i` and `add`, and finds the package anywhere in the command — `npm i -g pnpm@x rolldown@y` is the smithy build image's form. The whitespace before the name is matched rather than asserted, since GritQL's regex engine rejects lookbehind; that also stops `some-npm` matching `npm` while still reaching a `@scope/name`, which a `\b` boundary would not.

**Generic over the owned set, not a snapshot.** Both Dockerfile pin shapes and the `--with` shape iterate the packages the workspace's generators own, not a list of the ones pinned today. A CVE pin added to a template in a later release is covered without touching this module — otherwise this PR would fix four holes and leave the category open. A plain regex prefilters before GritQL is invoked, so cost scales with pins *present* rather than searched (~0.3ms per Dockerfile across the full vended set).

**Ownership.** 19 vended packages were declared by no generator; these surfaces accounted for several. Image pins are declared via `NODE_IMAGE_DEPENDENCIES` / `ADOT_IMAGE_DEPENDENCIES` in `docker.ts`, beside the `nodeImageVersions()` template data that emits them. The Terraform script pins are declared via a `*_PY_DEPENDENCIES` constant on each construct helper, gated on a new `generatedTerraform` predicate (`iac === 'terraform'`) since the CDK branch writes no such script — `when` predicates over recorded metadata, not a filter on the result set. All are spread through `ownedElsewhere`, so nothing is installed; the pin is owned only so its version stays current. A package the user pinned themselves is never matched.

**Comparison.** A base image tag has no ordering, so it is replaced whenever it differs from what this release vends. Every other pin only moves forward, so a version the user raised above the vended one is left alone.

**No `nextSteps`.** Following #1025: each of these is read the next time the thing holding it runs, so there is no lock file to reconcile and no follow-up left for the user.

**28 unit tests** across two new specs, all passing.

`real-templates.spec.ts` is the load-bearing one — it renders the **actual** vended templates through devkit's `generateFiles`, rolls the versions back to stand in for an older release, runs the migration, and asserts the result is **byte-identical** to what this release vends. That covers the agent, MCP server, rdb and Python agent Dockerfiles, the gateway `.tf` script and the trivy target.

It also adds **coverage guards**. One scans every real `.template` for vended-version substitutions and fails on one whose shape the sync cannot reach, plus a companion that fails on a new `*Version` var it doesn't recognise. That is what stops this class of bug recurring silently rather than relying on an audit. A third asserts the *filename* each pin-bearing template emits is one the sync actually visits — matching a pin's shape is only half of reachability. All three were confirmed to fire: adding a `RUN pip install boto3<%= boto3Version %>` line fails the shape check and names the file; adding a `someToolVersion` var fails the var check; and reverting the Dockerfile match to a `Dockerfile`-prefix test fails the filename check, naming `build.Dockerfile`.

Every assertion was **mutation-tested** — disabling the sync fails 15 tests; removing the TypeScript ownership gate fails 1; removing the Python gate fails 1; collapsing the tag/version distinction fails 3; relaxing the package-name boundary (so `some-npm` could be hit when syncing `npm`) fails 1.

Tool images come from a `CONTAINER_REPOSITORIES` map beside `CONTAINER_VERSIONS`, so a second pinned tool is one entry rather than a reference built in `docker.ts` plus a pattern in the sync — the same symmetry the Dockerfile and Terraform paths get from the owned set.

`tsc --noEmit`, lint and format are clean. Full `@aws/nx-plugin:test` diffed by test **name** against a clean worktree at `d17dd616`: identical failure sets (215 both sides, all pre-existing on this host), **zero new, zero fixed**.

### Issue # (if applicable)

N/A — follow-up deferred from #1001.

### Checklist

----


nx-plugin-for-aws:
## Review: end-to-end tested + extensibility findings

Tested this end-to-end by generating real CDK / Terraform / bun workspaces, downgrading every embedded pin to stand in for an old workspace, and running the **compiled** migration through the same internal path `nx migrate --run-migrations` uses (`FsTree` → migration → `flushChanges`, with `backfill-generator-metadata` ordered first). All pins restore to vended, idempotent, and user-raised / unowned pins are correctly left alone across all three package managers and both the heredoc and single-line `uv run --with` forms. Mutation-disabling the embedded sync fails exactly 15 tests, and both coverage guards fire when provoked. Nicely done — the owned-set-driven design means a new Dockerfile in a new location and a new `.tf` module with owned pins are both picked up with zero changes here (verified).

Two concrete extensibility gaps worth addressing, since they undercut the "add new Dockerfiles/Terraform and still have them included" goal:

### 1. Dockerfile detection is a `startsWith('Dockerfile')` prefix match

`syncDockerfiles` collects files with:

```ts
if (path.split('/').pop()?.startsWith('Dockerfile')) {
```

So a Dockerfile whose name does **not** start with `Dockerfile` is silently skipped:

- `build.Dockerfile` — which **already exists**, rendered from `smithy/project/files/{service,shapes}/build.Dockerfile.template`
- `service.dockerfile` (lowercase)
- any `*.Dockerfile`

Verified: a `build.Dockerfile` and a `service.dockerfile` carrying a downgraded `npm@` pin were both left untouched by the migration. No live bug today — the smithy `build.Dockerfile` carries only hardcoded literals (`pnpm@11.1.1`, `node:24`, smithy `1.61.0`), none of which are owned/vended — but it's a latent trap for exactly the case a future contributor hits when they add a pin to that template or vend a `*.Dockerfile`.

**Suggested fix:** match on the basename case-insensitively and allow the suffix form, e.g.

```ts
const name = path.split('/').pop() ?? '';
if (/(^|\.)dockerfile$/i.test(name) || /^dockerfile/i.test(name)) {
```

### 2. The coverage test doesn't check filename *reachability*

The `template pin coverage` test validates pin **shape** + var name, but not whether the emitted filename is one `syncDockerfiles` will actually visit. Verified: adding a fully-reachable `FROM <%= nodeBaseImage %>` to `build.Dockerfile.template` **passes both coverage tests**, yet the rendered `build.Dockerfile` would be skipped by the sync. So gap #1 wouldn't be caught if someone added a real pin to that template.

**Suggested fix:** extend the coverage test so that for each pin-bearing template it asserts the rendered filename is reachable by the corresponding sync (`Dockerfile*` → `syncDockerfiles`, `*.tf` → `syncTerraformScriptPins`, `project.json` → `syncTargetToolImages`). That closes the loop the same way the shape check already does.

### Minor: trivy pin is hardcoded, unlike the owned-set-driven pins

`syncTargetToolImages` hardcodes the single `aquasecurity/trivy` image rather than iterating an owned set, so a second tool-image pin added to a `project.json` command later wouldn't be covered without editing the sync. Fine today (trivy is the only one) — just less future-proof than the Dockerfile/Terraform paths. Non-blocking; flagging for symmetry.

nx-plugin-for-aws:
## Re-review of the two fix commits (`7f16013a`, `33ed8e45`)

Thanks for the quick turnaround — all three of my earlier points are addressed well: `isDockerfile` now matches `Dockerfile`/`*.Dockerfile` exactly (and correctly excludes `Dockerfile.bak`/`.md`), the new "should visit the file every pin-bearing template emits" test closes the filename-reachability half, `CONTAINER_REPOSITORIES` gives the tool-image path the same owned-set symmetry, and per-file `ownedForFile` fixes the cross-project leak (nice catch on the `prisma`-in-an-agent-Dockerfile case — I'd seen that in my own e2e and wrongly waved it off).

Re-tested end-to-end against freshly generated workspaces. Full suite is green (3126). **But I found one regression that the two commits create together**, and it happens to hit the exact pin `7f16013a` set out to fix.

### `rolldown` in the smithy `build.Dockerfile` is no longer synced

`7f16013a` broadened the npm-install shape specifically so the `rolldown@…` pin in `smithy/project/files/service/build.Dockerfile.template` becomes reachable — the commit message calls it *"live-unreachable, not just latent."* Then `33ed8e45` scoped ownership per file via `ownedForFile`. The problem:

- `build.Dockerfile` lands at `packages/<api>/model/build.Dockerfile`, under the **`smithy#project`** root.
- `smithy#project`'s `DEPENDENCIES` declares only `[...FS_DEPENDENCIES]` — it does **not** own `rolldown`. (`rolldown` is owned by `BUNDLE_DEPENDENCIES`, i.e. `ts#rdb` / `ts#dcr-proxy` / `ts#smithy-api` / `mcp-server` — none of which own this file.)
- So `ownedForFile(owned, 'packages/api/model/build.Dockerfile')` returns a set **without** `rolldown`, and the pin is skipped.

Verified on a fresh `ts#api --framework=smithy` workspace with the compiled PR head:

```
TS_VERSIONS.rolldown (vended)                         = 1.2.0
template pins                                          rolldown@1.0.0-beta.38
workspace-union owns rolldown?                         true
ownedForFile('packages/api/model/build.Dockerfile')?   false   <-- skipped
```

Under `7f16013a` alone the sync used the workspace union (`dockerfilePins(owned)`), which owns `rolldown`, so it **would** have synced. `33ed8e45`'s switch to `dockerfilePins(ownedForFile(owned, path))` is what regressed it. Tightening ownership to kill the `prisma` false-positive reintroduced a false-negative for a pin whose owning generator under-declares.

### Why the tests stay green

`build.Dockerfile.template` has **zero EJS vars** — `rolldown@1.0.0-beta.38`, `node:24`, `pnpm@11.1.1` are all hardcoded literals. The coverage tests key off `VENDED_TEMPLATE_VARS` (EJS var names), so `hasPin` is `false` for this template and both the shape and the new filename-reachability checks skip it entirely. Nothing asserts that a pin *present in a rendered template* is owned by the project that file lands in.

### Suggested fix

Root cause is that `smithy#project` vends a `build.Dockerfile` carrying a `rolldown` pin but doesn't declare ownership of it. Spread the relevant deps into its declaration the way the agent/mcp generators spread `NODE_IMAGE_DEPENDENCIES`, e.g.

```ts
export const DEPENDENCIES = declareDependencies()({
  ts: [...FS_DEPENDENCIES, ...ownedElsewhere([{ name: 'rolldown' }])],
});
```

I confirmed that flips `ownedForFile(build.Dockerfile)` to include `rolldown` and the pin then syncs.

Two things worth hardening so this class doesn't recur silently:
1. **The coverage guard is blind to hardcoded literal pins.** Consider also scanning rendered templates for literal `<pkg>@<version>` / `<repo>:<tag>` matches of vended packages, not just EJS vars — otherwise `node:24` and `pnpm@11.1.1` in this same file are also drifting unnoticed (node:24 isn't `BASE_IMAGES.node` = `node:lts-slim`; pnpm isn't vended at all — may be intentional, but nothing enforces it either way).
2. **Add an ownership-reachability assertion**: for each real template, every vended pin it emits should be owned by the project the emitted file belongs to. That's the check that would have caught this.

Happy to open a small PR with the `smithy#project` declaration fix + a regression test if useful.

nx-plugin-for-aws:
## Re-review of `34fce1a3` / `a210f87e` / `3f82a266` + an adversarial robustness pass

The three follow-ups all look right, and I'm glad the e2e caught more than my static read did:

- **Build-stage base image** (`34fce1a3`) — matching only a whole-instruction `FROM` is the correct fix; I verified on a fresh smithy workspace that `FROM node:24 AS builder` is now left alone (previously it'd rewrite to `lts-slim`, and the builder's `curl`/`unzip` step dies with exit 127 — genuinely build-breaking, good catch).
- **Shared-project ownership** (`a210f87e`) — right call; `common/terraform` *is* a registered project (resolves to `terraform#project`, which owns none of the `uv --with` pins), so the union fallback for shared projects is correct. I'd wrongly assumed it was project-less in my last comment — the union-vs-`ownedForFile` reasoning here is sound.
- **Holding the rolldown pin** (`3f82a266`) — agreed this is the correct resolution over my "declare ownership" suggestion: bumping beta.38→1.2.0 moves the SSDK `.d.ts` to a hashed name the handler import can't resolve. Holding it and listing it in the guard's held-pin set is the right shape.

Verified per-project scoping on a real multi-generator CDK workspace (agent + mcp + rdb + trpc): a `prisma` pin planted in the **agent** Dockerfile stays put (agent doesn't own prisma) while the **rdb** Dockerfile's prisma syncs, and `npm`/`minimatch`/`jaeger`/base-image/trivy all move. Idempotent on re-run. 

### Adversarial robustness pass — the migration can be made to crash the whole `nx migrate`

I threw a large suite of malformed / pathological inputs at `syncVendedVersions` asserting only that **it never throws** (a thrown migration aborts the entire upgrade — "leave it untouched" is fine, "crash" strands the user). 11 inputs throw or hang. Triaged each against the base branch to separate pre-existing from PR-introduced:

**Pre-existing (already true of the #1001 manifest sync — unguarded parses):**
- malformed `package.json`, `project.json`, `pyproject.toml`, `pnpm-workspace.yaml` → the underlying `readJson` / `readToml` / `yaml.load` throw
- `pyproject.toml` whose `dependencies` array holds non-string entries → `TomlError`

**New with this PR:**
1. **Non-string dependency version crashes the run.** A `package.json` with `"dependencies": { "minimatch": 123 }` (or `null`/`true`) throws `TypeError: declared.startsWith is not a function` at `sync-vended-versions.ts:81` (`isSyncableSpecifier`). That function is pre-existing, but it was never reached for these packages before — the PR added `minimatch`/`npm` to `ts#agent`'s owned set via `NODE_IMAGE_DEPENDENCIES`, so a non-string version now flows into it. Guarding with `typeof declared !== 'string'` fixes it.
2. **`project.json` with `targets` as a non-object** → `TypeError: Cannot use 'in' operator`.
3. **Quadratic blow-up on many-pin Dockerfiles.** Timing the embedded sync against a Dockerfile with N identical `RUN npm install -g npm@x` lines:

   | pins | time |
   |---|---|
   | 100 | 0.2s |
   | 400 | 1.4s |
   | 800 | 4.6s |
   | ~5000 | >120s (times out) |

   `applyPins` re-reads the file and runs a whole-file-anchored (`[\s\S]*…[\s\S]*`) GritQL rewrite once per pin, and each rewrite rescans the entire file — so cost grows ~O(pins x filesize). A hand-authored monorepo Dockerfile with a few hundred `RUN npm install` lines is not far-fetched, and this runs unattended.

**Realism caveat:** the parse-crash inputs (1, 2, and the pre-existing ones) require a genuinely malformed file, which a freshly generated workspace won't have — so low likelihood. But users hand-edit these files, and a single malformed file *anywhere* in the workspace takes down the whole `nx migrate`, not just its own sync. Since the migration's contract is already "only touch what I own, leave everything else exactly as-is", the natural fix is to treat a file that won't parse the same as one that isn't owned: **wrap each per-file parse/rewrite in try/catch and skip on failure** (`tryReadToml` already models this for TOML). The perf cliff (3) is the one I'd treat as more than defensive — worth an early `declared.every(...)`-style short-circuit or a cheaper single-pass rewrite.

Happy to share the adversarial spec (kept out of the PR) or open a follow-up with the guards + a perf fix if useful. Everything else here is in great shape.

nx-plugin-for-aws:
## Stacked on #1034 — rolldown is now synced here too

Rebased onto #1034 and retargeted this PR's base at `feat/smithy-rolldown-bump`, so the diff above shows only this PR's own changes. **#1034 should merge first**, after which this retargets to `main` on its own.

### Why

The last round held the smithy `build.Dockerfile`'s `rolldown` pin back, because moving it alone made the SSDK bundle emit `index-<hash>.d.ts` and broke `compile` with TS7016. That was a workaround for a real defect, and #1034 fixes it at the source: `rolldown-plugin-dts` is vended at a release that keeps the declaration on `index.d.ts`, and all three bundle pins are declared on `smithy#project` and rendered from the template.

So there is nothing left to hold. The commit dropped from this branch is `3f82a266` ("hold the smithy SSDK rolldown pin where the template put it") — its entire purpose was the workaround — and with it the coverage guard's `DELIBERATELY_HELD` exemption. `rolldown` is now synced like any other embedded pin, which was the original intent of the earlier `7f16013a`.

### Changes in this push

- Dropped the hold commit and the held-pin exemption set.
- Coverage guard learns the three new template vars (`rolldownVersion`, `rolldownDtsVersion`, `esmShimVersion`) — it caught all three unregistered on its own, which is the guard working as designed.
- The roundtrip winds all three pins back so it actually exercises them. `rolldown-plugin-dts` is wound back **before** `rolldown`, since `rolldown@` is a prefix of it.
- The smithy `build.Dockerfile` test now runs the migration over the file the **real generator renders** rather than writing the raw template verbatim — the template no longer stands in for a rendered file now that its pins are EJS.

### Verification

Confirmed the pins actually move on a real generated project: `rolldown@1.2.0`, `rolldown-plugin-dts@0.28.0`, `@rollup/plugin-esm-shim@0.1.8` all synced from wound-back versions.

Mutation-tested the new coverage: dropping `rolldown-plugin-dts` from the declaration fails the roundtrip on **both** cdk and terraform with `DRIFT .../build.Dockerfile`. So the roundtrip genuinely guards it rather than passing vacuously.

143 migration tests pass; `tsc` and lint clean. Full `@aws/nx-plugin:test` diffed against a pristine `main` worktree by failing test **name**: identical 215 failures (pre-existing `libraryGeneratorInternal` crash on this host), **zero new** — 2928 passing here vs 2864 on `main`.

nx-plugin-for-aws:
## Added: checkov in infra projects was uncovered

Good catch — `checkov` was **not** covered, and it is the same class of gap this PR set out to close. Confirmed empirically before fixing: planted an older `checkov==3.2.0` pin, ran the real migration, and it came back untouched for both `terraform#project` and `infra#app`.

`checkov` is pinned into a `project.json` target command by `uvxCommand`:

```
uvx --from checkov==3.3.8 checkov --directory . -o cli -o json ...          # terraform#project
uvx --from checkov==3.3.8 checkov --config-file {projectRoot}/checkov.yml ... # infra#app
```

The `project.json` sync only matched container repositories, so this shape fell through — meaning the security scan every infra project runs stayed on whatever version generated the workspace.

Fixed in `85009a4e` by extending that sync to `uvx --from <tool>==<version>`, driven off `PY_VERSIONS` rather than a list of the tools pinned today, so another tool invoked from a target command later is covered with no change here. It is deliberately **unscoped**, unlike the file-body pins: `uvx --from` installs nothing into the project, so there is no dependency for a generator to own — only a tool this release vends at a higher version moves.

### Swept the category, not just this case

`uvxCommand` has three call sites. `checkov` is the only one whose pin lands in a workspace file — `pip-licenses` (license dependency-check) runs in-process via `execSync` and `ruff` runs in-process via wasm, so neither is written anywhere to drift. So this closes the whole `uvx` class.

### Verification

- Real-generator check: generated a `terraform#project` infra project, wound its pin back to `3.2.0`, synced → restored to `3.3.8` and **byte-identical** to fresh. The CDK path can't run its generator on this host (the pre-existing `libraryGeneratorInternal` crash), so I verified it emits the identical `uvx --from checkov==` shape and covered it directly in the spec.
- 4 new tests (both target-command shapes, plus a user-raised `checkov==99.0.0` left alone). Mutation-verified: dropping the uvx pins fails exactly those.
- 146 migration tests pass (was 143). `tsc` and lint clean. Full suite vs current `main`: **zero new failures** (216 both sides), 2966 passing.

nx-plugin-for-aws:
## Consolidated: #1034 folded in and closed, `--with` gap fixed, roundtrip spec removed

**#1034 is closed and its work lives here now** — no stack, one PR. Both commits carried over intact, including the three review fixes cogwirrel raised there (hardcoded once-off migration versions, scoping by generator id + `smithyType`, trimmed `nextSteps`):

- `40f774bc` feat(smithy#project): bundle the SSDK with the rolldown this release vends
- `41e0afde` fix(smithy#project): scope and pin the SSDK bundle migration

### `--with` was indeed uncovered

Right again. `uvxCommand` pins **two** things and I'd only handled one:

```
uvx --from checkov==3.2.0 --with boto3==1.40.0 --with httpx==0.27.0 checkov -d .
       ^^^^ synced             ^^^^ not synced      ^^^^ not synced
```

Confirmed by planting that exact command and running the real migration: `checkov` moved, `boto3` and `httpx` did not. Fixed in `df499e4d` — both shapes now come from one `UVX_PIN_SHAPES` list, so a target command using either form is covered. (The `.tf` path already matched `--with` for its inline `uv run` scripts; this was the `project.json` path only.)

No generator passes `withDeps` today, so there's no live bug — but it is exactly the latent trap this PR exists to close, and the shape is already in `uvxCommand`'s signature waiting to be used.

### Removed `ws-roundtrip.spec.ts`

Dropped as requested. Nothing referenced it; the coverage it provided over the embedded surfaces is in `sync-embedded-versions.spec.ts` and `template-pin-ownership.spec.ts`.

### Verification

189 tests pass across the migration, smithy project and SSDK-pin specs. `tsc` and lint clean. Full `@aws/nx-plugin:test` diffed against current `main` (`c88001dc`) by failing test **name**: **zero new failures** (216 both sides).

One heads-up unrelated to this PR: `ts#agent > should not create session bucket infrastructure when session is in-memory` is failing on `main` itself, arriving with #1017.

cogwirrel (approved):

PR #1034, opened 2026-08-04
feat(smithy#project): bundle the SSDK with the rolldown this release vends

nx-plugin-for-aws:
### Reason for this change

`smithy/project/files/service/build.Dockerfile.template` pinned `rolldown` at `1.0.0-beta.38` and `rolldown-plugin-dts` at `0.16.5`, both as hardcoded literals in the file body, so neither moved with the versions this plugin vends (`TS_VERSIONS.rolldown` is already `1.2.0`).

Bumping `rolldown` alone breaks the build, which is why the pin had been left behind. Before 0.20.0 the dts plugin named the bundled declaration after its content hash, so on rolldown 1.2.0 the SSDK bundle emits `index-<hash>.d.ts`. The generated handler imports `./generated/ssdk/index.js`, finds no `index.d.ts` beside it, and `compile` fails with TS7016.

The two pins have to move together — which is what this PR does.

### Description of changes

- Vend `rolldown-plugin-dts` (`0.28.0`) and `@rollup/plugin-esm-shim` (`0.1.8`) in `utils/versions.ts`, with a comment recording *why* the plugin is held at or above 0.20.0.
- Render all three pins from the template rather than hardcoding them, following the `ts#agent` Dockerfile pattern (`<%- nodeBaseImage %>`, `<%- npmVersion %>`, …).
- Declare them on `smithy#project` as `SSDK_BUNDLE_DEPENDENCIES`, spread through `ownedElsewhere` — never installed into the workspace, but from here on the version sync keeps the pins current instead of this recurring.
- Add `latest-smithy-ssdk-bundle-pins`, a deterministic migration that moves an existing project's pins forward. Each pin is matched **by name** so the `rolldown-plugin-dts` and `@rollup/plugin-esm-shim` versions sharing that `pnpm add` command keep their own (`rolldown` is a prefix of the plugin's name). A workspace already on the vended versions is left untouched; one whose pins have diverged is skipped and reported via `nextSteps`. GritQL for the transform, per CONTRIBUTING.

**Reproduced the root cause on the real SSDK, both directions.** Generated a real service project, then built `build.Dockerfile` with Docker:

| rolldown | rolldown-plugin-dts | declaration emitted |
|---|---|---|
| 1.2.0 | 0.16.5 (old) | `index-B2P1qQoJ.d.ts` ❌ |
| 1.2.0 | 0.28.0 (this PR) | `index.d.ts` ✅ |

So the plugin bump is precisely what makes the rolldown bump safe. Also bisected the plugin: 0.16.5 hashes, ≥ 0.20.0 does not.

**End-to-end on real generated output, not fixtures.** Generated a service project with the actual generator, wound both pins back to an older release's versions, ran the migration, and asserted the result is **byte-identical to a fresh workspace** — it is, with empty `nextSteps`. Then built that migrated `build.Dockerfile` for real: it completes and lands `index.d.ts` alongside `index.js` in `/out/ssdk`.

**Tests.** 6 new migration tests (applies, leaves neighbouring pins in the same command alone, no-op when already current, idempotent, ignores the shapes build image which bundles no SSDK, no-op with no smithy project) — all rendered from the real template via the real generator, so they can't pass while the template's shape drifts.

`compile` and lint clean. Full `@aws/nx-plugin:test` diffed against a pristine `main` worktree by failing test **name**: identical 215 failures on both (a pre-existing `libraryGeneratorInternal` crash on this host), **zero new** — and 6 additional passing tests.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/smithy-ssdk-bundle-pins/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/smithy-ssdk-bundle-pins/migration.ts
@@ -1,0 +1,113 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+  type MigrationReturnObject,
+  type Tree,
+  visitNotIgnoredFiles,
+} from '@nx/devkit';
+import { applyGritQL } from '../../../utils/ast';
+import { formatFilesInSubtree } from '../../../utils/format';
+import { TS_VERSIONS } from '../../../utils/versions';
+
+/**
+ * Move a Smithy `build.Dockerfile` onto the SSDK bundle pins it now vends.
+ *
+ * `rolldown-plugin-dts` named the bundled declaration after its content hash
+ * before 0.20.0, so the generated handler's `./generated/ssdk/index.js` import
+ * found no `index.d.ts` beside it and `compile` failed with TS7016. The two pins
+ * move together: the plugin release that fixed the naming needs a rolldown past
+ * the pin this shipped with.
+ *
+ * There is no Dockerfile grammar, so each pin is matched as an anchored rewrite
+ * over the whole file with the version captured — `${...}` in these patterns is
+ * GritQL metavariable syntax, not interpolation.
+ */
+
+/** A pin in the file body, and the version this release vends for it. */
+interface Pin {
+  /** The package as the install command names it. */
+  readonly name: string;
+  readonly vended: string;
+}
+
+const PINS: readonly Pin[] = [
+  { name: 'rolldown', vended: TS_VERSIONS.rolldown },
+  {
+    name: 'rolldown-plugin-dts',
+    vended: TS_VERSIONS['rolldown-plugin-dts'],
+  },
+  {
+    name: '@rollup/plugin-esm-shim',
+    vended: TS_VERSIONS['@rollup/plugin-esm-shim'],
+  },
+];
+
+/** A version ends at whitespace, a line break or the end of the command. */
+const VERSION = String.raw`[^\s\\]+`;
+
+const escapeRegExp = (value: string): string =>
+  value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
+/**
+ * Rewrite every occurrence of one pin to the vended version.
+ *
+ * The pin is matched by name, so a `rolldown-plugin-dts@x` in the same command
+ * as `rolldown@y` keeps its own version — the longer name is not a prefix of the
+ * shorter one here, since the `@` before the version terminates the match.
+ */
+const rewritePin = ({ name, vended }: Pin): string =>
+  `r"[\\s\\S]*${escapeRegExp(name)}@(${VERSION})[\\s\\S]*"($version) where { $version => \`${vended}\` }`;
+
+/** Whether the file pins this package at a version other than the vended one. */
+const isOutOfDate = (contents: string, { name, vended }: Pin): boolean => {
+  const matches = contents.matchAll(
+    new RegExp(`${escapeRegExp(name)}@(${VERSION})`, 'g'),
+  );
+  const declared = [...matches].flatMap((match) =>
+    match[1] ? [match[1]] : [],
+  );
+  // A rewrite replaces every occurrence at once, so this has to hold for all of
+  // them — a copy already on the vended version is left alone.
+  return declared.length > 0 && declared.every((version) => version !== vended);
+};
+
+const divergedNextStep = (path: string): string =>
+  `${path}: the Smithy build image's bundle pins have diverged from the generated shape - left untouched. Update \`rolldown\` to ${TS_VERSIONS.rolldown} and \`rolldown-plugin-dts\` to ${TS_VERSIONS['rolldown-plugin-dts']} by hand: before 0.20.0 the plugin named the bundled SSDK declaration after its content hash, so the generated handler's \`./generated/ssdk/index.js\` import fails to resolve its types with TS7016.`;
+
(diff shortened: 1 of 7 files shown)


packages/nx-plugin/src/migrations/latest/smithy-ssdk-bundle-pins/migration.ts
```diff
@@ -0,0 +1,113 @@
...
+const PINS: readonly Pin[] = [
+  { name: 'rolldown', vended: TS_VERSIONS.rolldown },
+  {
+    name: 'rolldown-plugin-dts',
+    vended: TS_VERSIONS['rolldown-plugin-dts'],
+  },
+  {
+    name: '@rollup/plugin-esm-shim',
+    vended: TS_VERSIONS['@rollup/plugin-esm-shim'],
+  },
```
cogwirrel: we should hardcode the versions for the once-off migration

packages/nx-plugin/src/migrations/latest/smithy-ssdk-bundle-pins/migration.ts
```diff
@@ -0,0 +1,113 @@
...
+  tree: Tree,
+): Promise<MigrationReturnObject> {
+  const nextSteps: string[] = [];
+  const buildDockerfiles: string[] = [];
+
+  visitNotIgnoredFiles(tree, '.', (path) => {
+    if (path.split('/').pop() === 'build.Dockerfile') {
+      buildDockerfiles.push(path);
+    }
+  });
```
cogwirrel: We should filter to within smithy projects, marked by the generator id.

packages/nx-plugin/src/migrations/latest/smithy-ssdk-bundle-pins/migration.ts
```diff
@@ -0,0 +1,113 @@
...
+  const declared = [...matches].flatMap((match) =>
+    match[1] ? [match[1]] : [],
+  );
+  // A rewrite replaces every occurrence at once, so this has to hold for all of
+  // them — a copy already on the vended version is left alone.
+  return declared.length > 0 && declared.every((version) => version !== vended);
+};
+
+const divergedNextStep = (path: string): string =>
+  `${path}: the Smithy build image's bundle pins have diverged from the generated shape - left untouched. Update \`rolldown\` to ${TS_VERSIONS.rolldown} and \`rolldown-plugin-dts\` to ${TS_VERSIONS['rolldown-plugin-dts']} by hand: before 0.20.0 the plugin named the bundled SSDK declaration after its content hash, so the generated handler's \`./generated/ssdk/index.js\` import fails to resolve its types with TS7016.`;
```
cogwirrel: Can remove " by hand: before 0.20.0 the plugin named the bundled SSDK declaration after its content hash, so the generated handler's \`./generated/ssdk/index.js\` import fails to resolve its types with TS7016."

nx-plugin-for-aws:
All three addressed in `f177e2d1`.

**1. Hardcode the versions for the once-off migration** — done. `PINS` now carries literal targets (`rolldown@1.2.0`, `rolldown-plugin-dts@0.28.0`, `@rollup/plugin-esm-shim@0.1.8`) and the migration no longer imports `TS_VERSIONS` at all. The spec hardcodes them too.

This surfaced a real bug that reading `TS_VERSIONS` had been masking: with the targets fixed, a workspace a **later** release had already moved past them would be wound **back**. I caught it by bumping vended `rolldown` to `9.9.9` locally and watching the migration downgrade the fresh pin. Fixed by only rewriting a pin that sits *behind* the target, via the existing `isVendedUpgrade` — with a test that fails if it regresses. Re-ran the suite with vended at `9.9.9` to confirm it holds.

**2. Filter to within smithy projects by generator id** — done. It now selects from `getProjects` where the recorded metadata has `generator === SMITHY_PROJECT_GENERATOR_INFO.id` **and** `smithyType === 'service'`, following `terraform-bootstrap-adopt-existing-bucket`'s shape. That also replaces the content sniff I had for excluding the shapes build image, which was the weaker signal.

New test: a `build.Dockerfile` under a non-smithy project with those pins is left exactly as the user wrote it, while the smithy one still migrates. Mutation-verified — removing the scoping fails precisely that test.

**3. Trim the `nextSteps` message** — done, it now ends at "Update `rolldown` to 1.2.0 and `rolldown-plugin-dts` to 0.28.0."

### Verification

8 migration tests pass (up from 6). Migrated output still byte-identical to a fresh workspace with empty `nextSteps`. `tsc` and lint clean.

Full `@aws/nx-plugin:test` diffed against current `main` (`c88001dc`) by failing test **name**: **zero new failures** (216 both sides). Worth noting one of those 216 — `ts#agent > should not create session bucket infrastructure when session is in-memory` — is failing on `main` itself, arriving with #1017; it is not from this PR.

nx-plugin-for-aws:
Superseded by #1029, which carries this commit unchanged (`feat(smithy#project): bundle the SSDK with the rolldown this release vends`) as its base along with the rest of the version-sync work — the SSDK bundle pins are only kept current going forward by the sync landing in that PR, so the two don't make sense to review or merge apart.

Closing in favour of #1029. No changes are lost: `git merge-base --is-ancestor` confirms this branch's tip is an ancestor of `feat/version-sync-unreachable-surfaces`.

nx-plugin-for-aws:
Closing this — the smithy rolldown work has been folded into #1029, so it lands as one PR rather than a stack. Both commits are on that branch:

- `40f774bc` feat(smithy#project): bundle the SSDK with the rolldown this release vends
- `41e0afde` fix(smithy#project): scope and pin the SSDK bundle migration (this review's three fixes)

No content lost; the review feedback here is all carried over. Further discussion on #1029.

PR #1037, opened 2026-08-04
docs: add a graph builder for sketching a workspace visually

nx-plugin-for-aws:
### Reason for this change

Getting started with the plugin means knowing which generators to run, in which order, and which connections are supported between which project types. That knowledge is spread across a dozen guide pages, and the ordering constraints aren't obvious until a command fails — a component needs its host project to exist first, a connection needs both of its ends, and several connections quietly require particular options on their endpoints.

### Description of changes

A drag-and-drop canvas at **Tools → Graph Builder**: drag components from the palette onto the canvas, drag between them to connect, edit each component's properties, and copy the commands that scaffold exactly what you drew.

**The builder stays current as the plugin evolves.** Nothing in the UI hand-lists generators:

- The palette and the edges you may draw come from `SUPPORTED_CONNECTIONS`
- Every property field comes from that generator's own `schema.json`
- A new `scaffold-catalog.ts` **infers** how to run each endpoint rather than declaring it:
  - `kind` — a generator whose schema `required`s a `project` adds a component; anything else creates a project
  - `host` — a component lands in its language's base project generator, with that generator's own required options
  - `generator` + `options` — a `hidden` generator is reached through the public wrapper whose enum value reconstructs its name (`ts#trpc-api` ← `ts#api --framework=trpc`)

So adding a generator plus a connection for it needs no edit to the builder. Two things stay declared, because the metadata genuinely doesn't carry them: display labels (schema `title`/`description` are CLI prompt prose — several are empty strings), and the connection constraints (each is an imperative `throw` inside a connection generator, reachable only by running it). Both degrade gracefully: an unlabelled type falls back to its generator id, and one with no artwork still appears in the palette.

Validation mirrors those generator guards, so a graph that validates is a graph that scaffolds. An `infra` project (`ts#infra` or `terraform#project`) is appended to match the workspace's IaC choice, and its name is reserved.

Also included:
- `docs:test` added to the CI and PR workflows — the docs project had no test coverage before
- Generator-guide sidebar subsections collapsed by default, and a graph icon in the header

**Integration tests against the real generators (33 cases).** For every supported connection, both endpoints are scaffolded with exactly the options the builder emits, then the connection generator is run between them. This is the strongest available check that the emitted commands actually work.

**Unit tests.** 70 docs tests and 17 catalogue tests. I verified the derivation reproduces all 13 previously hand-written recipes byte-for-byte, then pinned that output so a generator restructured such that the rules infer something different shows as a diff rather than a subtly wrong command.

**Drift detection proven, not assumed.** I temporarily added a connection using a fake endpoint type and confirmed the suite fails naming exactly what's missing.

**Browser verification against the production build.** Add, drag-to-connect, invalid-drop refusal, constraint errors, canvas panning, shift-click multi-select and group drag, light and dark mode, and preset layout at four viewport widths (asserting no node lands off-canvas and none overlap).

Full plugin suite (3187 tests), lint, and the docs build with link validation all pass.

Two things worth a reviewer's attention:

1. **`scaffold-catalog.ts` lives in `packages/nx-plugin`, not in `docs`.** It has to: the docs site imports it and it reads `generators.json`. It's plugin-internal and not exported from the SDK, but it does mean a docs feature added a file to the plugin package.
2. **Deduplicating the schema lookup fixed a latent bug.** The docs catalogue had its own lookup using suffix matching, which resolved `ts#api` to `smithy/ts/api/schema.json` — the wrong file. It only worked by luck because the option it needed existed in both. Both sides now share one exact-path resolver.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/lib/graph-builder/commands.ts
```diff
@@ -0,0 +1,311 @@
...
+    const args = [
+      kebabCase(node.name),
+      ...Object.entries(type.variantOptions).map(([k, v]) => flag(k, v)),
+      ...type.properties
+        .filter((p) => shouldEmit(node, type, p.name))
+        .map((p) => flag(p.name, node.options[p.name])),
+    ];
+    commands.push({
+      command: generatorCommand(type.generator, args),
+      comment: `Create ${node.name} — a ${type.label} project`,
```
cogwirrel: change to `Create ${node.name} (${type.label})`

packages/nx-plugin/src/connection/scaffold-catalog.ts
```diff
@@ -0,0 +1,408 @@
...
+  return undefined;
+};
+
+/**
+ * Display labels. The one thing the metadata cannot supply: schema `title` and
+ * `description` are prose aimed at CLI prompts ("Create a relational database
+ * project"), not the short noun a palette entry needs. A type with no entry
+ * falls back to its generator id, so a new endpoint type still appears.
+ */
+const LABELS: Readonly<Record<string, string>> = {
```
cogwirrel: Add an "x-label" to schema.json for the relevant generators and then derive from there

packages/nx-plugin/src/connection/scaffold-catalog.ts
```diff
@@ -0,0 +1,408 @@
...
+  },
+];
+
+/**
+ * The option-level requirements each connection places on its endpoints.
+ *
+ * Only connections carrying a requirement appear. Keyed by `ConnectionKey`, so a
+ * key that no longer names a supported connection is a compile error.
+ */
+export const CONNECTION_CONSTRAINTS = {
```
cogwirrel, line 359: Add to contributing guide a very short section on connection generators and the need to add constraints here, as well as the location we map generator ids to logos etc. ie when building a new generator it should point at the extra work required to ensure it works properly in the graph builder

docs/src/lib/graph-builder/commands.ts
nx-plugin-for-aws: Done — comments now read `Create <name> (<Label>)`, and the component ones `Add <name> (<Label>) to <project>`.

packages/nx-plugin/src/connection/scaffold-catalog.ts
nx-plugin-for-aws: Done. Every generator that can take part in a connection now has an `x-label` in its `schema.json`, and the label is read from there — the hand-written `LABELS` map is gone.

A test asserts each connectable generator has one (`should give every connectable generator an x-label`), so a new generator is told rather than silently falling back to its raw id. I left `title`/`description` out of it deliberately: those are prose written for CLI prompts ("Create a relational database project"), and several are empty.

One knock-on worth flagging: this module is imported by the docs site, so it can no longer use `node:fs`. Reading schemas from disk moved to `schema-resolver.ts` and the resolver is now injected — the plugin passes the disk one, the docs site passes one over its own bundle. The derivation itself is shared. Without that split the builder failed to hydrate in the browser, which I caught in the browser rather than in tests (the docs tests run in Node, so they passed).

packages/nx-plugin/src/connection/scaffold-catalog.ts
nx-plugin-for-aws: Added — `CONTRIBUTING.md` now has a "Connection Generators and the Graph Builder" section just above End to End Tests. It covers the three things a new connection generator needs (declare the connection, add an `x-label`, record any endpoint option constraints), plus where to map the logo and palette grouping, and notes that everything else is inferred and that the spec fails with a specific message if it cannot work out how to run a generator.

cogwirrel (approved):

PR #1041, opened 2026-08-05
feat(ts#nx-migration): assemble migrations.json to end merge conflicts

nx-plugin-for-aws:
### Reason for this change

Recent PRs frequently add migrations, and every one of them edits the shared `generators` object in `packages/nx-plugin/migrations.json`. Because the generator re-sorts and rewrites that single object, two migration PRs conflict on that one file even though the migrations themselves are entirely independent. This has become a constant source of merge conflicts.

### Description of changes

Make `migrations.json` a build artifact assembled from the plugin's source, so two PRs adding migrations touch **disjoint files** and never conflict:

- Each migration folder carries a `metadata.json` (its description) beside the code. The migration's **kind** is inferred from the files present (`migration.ts` → deterministic, `prompt.md` → agentic, both → hybrid) and its **version** from the folder it lives in (`latest` vs `v<version>`).
- The nx `packageJsonUpdates` move to their own committed `packageJsonUpdates.json`, so the weekly version-update bot never shares an edit with a migration PR.
- The `sync-vended-versions` `everyMigration` entry becomes a fixed code constant (never changes, never conflicts).
- `migrations.json` is gitignored and generated — into both the source tree (so nx can load the plugin locally) and `dist` (what gets published) — by a new `generate-migrations` target that `compile`/`package` depend on.

Release tooling is repointed at the assembled manifest, with the subtle pure versioning functions (`stampMigrationVersions`, `backfillMigrationVersions`, `readShippedMigrationVersions`) left **unchanged**:

- `stamp-migrations.ts` and `backfill-migration-versions.ts` assemble the source manifest from the folders instead of reading the committed file. Backfill now records a shipped release purely by moving the migration's folder into `v<version>/`.
- The tag-walk that dates migrations (`readReleasedMigrations`) derives each release's registered keys from that tag's migration folders, unioning with the committed `migrations.json` for releases predating the split — so the entire release history stays resolvable across the transition.

In the `ts#nx-migration` generator, the folder-based registration applies **only to the plugin's own repo**. Other workspaces still get a written `migrations.json` exactly as before.

- Verified the assembled `migrations.json` is byte-identical (content) to the previously committed one (only key ordering differs — alphabetical, which stamping re-sorts anyway).
- Proved the rewritten tag-walk returns **identical key sets** to the old committed-file reader across all 10 release tags that carried a `migrations.json`, so version resolution is provably unchanged.
- Ran `stamp-migrations.ts` and `backfill-migration-versions.ts` against the real release tags and confirmed behavior matches the old tooling (e.g. a migration shipped in `v1.0.0-rc.58` resolves to that version; net-new migrations get the pending version).
- Full `package:all` build, `@aws/nx-plugin` test suite, and lint all green. Confirmed a fresh-clone build (no source `migrations.json`) produces both manifests. Added unit tests for the new assembler and updated the generator/register specs.

### Issue # (if applicable)

N/A

### Checklist

----


packages/nx-plugin/project.json
```diff
@@ -87,10 +105,15 @@
...
+      "dependsOn": ["generate-migrations", "generate-3p-license"]
     },
     "build": {
-      "dependsOn": ["compile", "lint", "test", "generate-3p-license"]
+      "dependsOn": [
+        "generate-migrations",
+        "lint",
+        "test",
+        "generate-3p-license"
+      ]
```
cogwirrel, line 117: We should still keep the explicit compile dependency - nx will resolve the duplicate and it's safer if we were to remove generate migrations.

cogwirrel (dismissed):

packages/nx-plugin/project.json
nx-plugin-for-aws: Good call — restored the explicit `compile` dependency on both `package` and `build`. nx dedupes it against the one `generate-migrations` already pulls in, and it keeps them compiling if `generate-migrations` is ever removed. Pushed in 39f5ce57.

cogwirrel (approved):

PR #1042, opened 2026-08-05
feat(smithy#project): build with the Smithy CLI rather than a container

nx-plugin-for-aws:
### Reason for this change

Smithy projects built inside a container image. The `build.Dockerfile` installed the Smithy CLI, built the model, and — for a service — installed and bundled the generated TypeScript Server SDK. Two costs followed from that:

- **Docker became a prerequisite** for any workspace holding a Smithy project, which is why `smithy#project` and `ts#smithy-api` were excluded from the CodeBuild Windows smoke tests.
- **Consuming a shape library took three manual edits**, one of them a `COPY --from=workspace` into the image, because the build could only see files the build context carried in.

### Description of changes

Smithy projects now build with the Smithy CLI directly.

**The build.** `compile` runs `mise exec smithy@<version> -- smithy build`, so nothing is installed beyond the workspace's own dev dependencies — `mise` downloads and caches the pinned CLI on first use. A service then publishes its OpenAPI spec and bundles the generated Server SDK with the workspace's own rolldown (the config is vended as `ssdk.rolldown.config.mjs`). Everything the CLI writes stays under `dist`, so a build no longer writes into the project directory.

**Shape libraries.** Smithy resolves `imports` relative to `smithy-build.json`, so depending on a library is now a single relative path — the `COPY` step and the Dockerfile both go, taking the docs from three steps to two.

**Windows.** `mise` publishes no Windows package to npm: the `mise` package is a stub whose `preinstall` fetches `@jdxcode/mise-<platform>-<arch>`, and `scripts/release-npm.sh` upstream publishes only linux and darwin. A root dependency on it would fail *every* install on Windows, not just Smithy builds. So on Windows the Smithy CLI is a documented prerequisite, the target invokes `smithy` directly, `mise` is not added, and the generator warns when the CLI isn't on `PATH`. CI installs it on the Windows runners, which lets `smithy#project` and `ts#smithy-api` come off the exclusion lists.

**Keeping versions current.** Two new surfaces hold versions no manifest carries — the CLI pin in the `project.json` command, and the Maven coordinates in `smithy-build.json`. Both are now reached by the version sync migration. `update-versions` asks `mise` for the CLI (so it never vends a version mise cannot install) and Maven Central's `maven-metadata.xml` for the artifacts, moving the `software.amazon.smithy:*` set only once every artifact has published the CLI's version — the CLI cannot load a model built against a newer `smithy-model`.

`smithy-aws-typescript-codegen` is held at **0.50.0** via a new documented `SMITHY_VERSION_CEILINGS`. 0.51.0 generates a server importing `ServerInterceptor` from `@aws-smithy/server-common`, which its latest release (`1.0.0-alpha.10`) does not export, so the SSDK fails to bundle. The ceiling keeps the hold visible in one place rather than buried in a reject list; lifting it is deleting an entry.

**Migration.** `latest-smithy-build-without-docker` moves existing service and shape projects across: it rewrites the `compile` target from the same helper the generator uses (so a migrated project cannot drift from a fresh one), deletes the Dockerfile, writes the bundle config a service needs, and carries any `COPY --from=workspace` over to `imports`. A project whose `compile` commands the user has customised is left alone and reported rather than clobbered.

**Unit tests.** Zero new failures against the pre-change baseline on this machine (25 failing files in both; 3017 passing vs 2993 — this environment has a pre-existing set of failures unrelated to the change). 37 tests on `smithy#project`, 15 on the new migration, 5 on the Windows CLI branches, and the version-sync pin coverage rewritten from the deleted Dockerfile onto the two new surfaces.

**Deployed to AWS and invoked it.** A workspace created through the preset, with `ts#api --framework=smithy`, `smithy#project --type=shapes` and `ts#infra`, deployed to the `nx-e2e` account and the live endpoint invoked with SigV4:

```
HTTP 200
{"message":"hello-from-smithy"}
```

Stack destroyed afterwards and confirmed deleted.

**Also validated locally:**
- Shape library builds in ~1.6s, a service in ~10s, neither touching Docker.
- The migration run against a real workspace wound back to the container build — the projects it migrated then built successfully.
- A shape library consumed by a service through the documented `imports` entry: 830 shapes validated vs 825 without it, and `org#ExampleShape` present in the consumer's assembled model.
- CDK synth clean with 74 checkov checks passing, and `docs:build` clean.
- `update-versions`' Smithy logic exercised against the live registries: CLI resolved via mise (1.72.1), all four `software.amazon.smithy:*` artifacts confirmed published at that version, and codegen correctly held at the 0.50.0 ceiling.

**Two pre-existing issues surfaced that are out of scope here**, both reproducing identically on `main`:
- `@middy/core` 7 is ESM-only, so the Smithy `handler.ts` fails to compile in a CommonJS workspace. Workspaces created by the preset are ESM, which is why CI is green; I hit it only by hand-rolling a CJS workspace.
- The `smithy-ssdk-bundle-pins` migration shipped in rc.58 and still targets a `build.Dockerfile`, so it has to stay for anyone upgrading from rc.57. Its spec now runs against a snapshot of that release's rendered template, since the generator no longer vends one.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -110,39 +109,43 @@ Building a service project generates an OpenAPI specification and a TypeScript S
...
 
-Building a shape library writes an assembled model to `dist/<my-shapes>/build/model/model.json`. This single file contains every shape the library defines, along with any it depends on, so a consumer only ever declares the libraries it references directly.
+:::caution[Windows]
+`mise` publishes no Windows package to npm, so install the Smithy CLI yourself before building:
 
-To depend on a shape library from another Smithy project, make three changes to the consuming project:
+```sh
+scoop bucket add smithy-lang https://github.com/smithy-lang/scoop-bucket
+scoop install smithy-lang/smithy-cli
+```
```
cogwirrel: no scoop commands, just link to the install instructions

docs/src/content/docs/en/guides/ts-smithy-api.mdx
```diff
@@ -576,10 +576,23 @@ export const Echo: EchoOperation<ServiceContext> = async (input, ctx) => {
...
 
+The CLI is resolved by [mise](https://mise.jdx.dev/), added to your workspace as a dev dependency, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+
+:::caution[Windows]
+`mise` publishes no Windows package to npm, so install the Smithy CLI yourself before building:
+
+```sh
+scoop bucket add smithy-lang https://github.com/smithy-lang/scoop-bucket
+scoop install smithy-lang/smithy-cli
+```
```
cogwirrel: same

packages/nx-plugin/src/smithy/project/generator.spec.ts
```diff
@@ -50,8 +53,12 @@ describe('smithyProjectGenerator', () => {
     expect(projectConfig.targets.compile.executor).toBe('nx:run-commands');
     expect(projectConfig.targets.compile.options.commands).toEqual([
       'rimraf dist/{projectRoot}/build',
+      'rimraf dist/{projectRoot}/smithy',
       'make-dir dist/{projectRoot}/build',
-      'docker build -f {projectRoot}/build.Dockerfile --build-context workspace=. --target export --output type=local,dest=dist/{projectRoot}/build {projectRoot}',
+      'mise exec smithy@1.72.1 -- smithy build -c {projectRoot}/smithy-build.json --output dist/{projectRoot}/smithy',
```
cogwirrel: don't fix version in test or it'll fail on upgrade

packages/nx-plugin/src/utils/init.ts
```diff
@@ -61,7 +61,18 @@ export const INIT_DEPENDENCIES = [
 // pnpm 11 reads `allowBuilds` instead. Any dep NOT in this allowlist will
 // have its install scripts skipped with a warning — matching pnpm 10's
 // default behaviour. The user can opt-in later via `pnpm approve-builds`.
-export const PNPM_BUILT_DEPENDENCIES = ['@swc/core', 'esbuild', 'nx', 'sharp'];
+//
+// `mise` is here for the Smithy generators: it fetches its own binary in a
+// `preinstall`, and pnpm fails the install outright (ERR_PNPM_IGNORED_BUILDS)
+// for a script it skipped. Listed unconditionally so adding a Smithy project to
+// an existing workspace doesn't have to repair the allowlist first.
```
cogwirrel: adding a smithy project should update the allow list, not part of init.

packages/nx-plugin/src/utils/smithy.ts
```diff
@@ -0,0 +1,76 @@
...
+  onPath: () => boolean = isSmithyOnPath,
+): void => {
+  if (!isWindows() || onPath()) {
+    return;
+  }
+  logger.warn(
+    `The Smithy CLI was not found on your PATH. Smithy projects build with it directly on Windows, so install it before building:
+
+  scoop bucket add smithy-lang https://github.com/smithy-lang/scoop-bucket
+  scoop install smithy-lang/smithy-cli
```
cogwirrel: no scoop

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -271,6 +273,81 @@ export const VENDORED_VERSIONS = {
...
+ * and lifting it is deleting an entry once the blocking release lands.
+ */
+export const SMITHY_VERSION_CEILINGS: Partial<Record<ISmithyVersion, string>> =
+  {
+    // 0.51.0 generates a server importing `ServerInterceptor` from
+    // `@aws-smithy/server-common`, which its latest release (1.0.0-alpha.10) does
+    // not export — the generated SSDK then fails to bundle. Lift once a
+    // `server-common` exporting it ships.
+    'software.amazon.smithy.typescript:smithy-aws-typescript-codegen': '0.50.0',
+  };
```
cogwirrel: Let's leave this out, version update can resolve all deps and see if it fails - e2e tests should fail.

scripts/update-versions.ts
```diff
@@ -352,6 +356,134 @@ const updateGitSecrets = async (
...
+  const versionsFilePath = 'packages/nx-plugin/src/utils/versions.ts';
+
+  const rewrite = async (key: ISmithyVersion, from: string, to: string) => {
+    const ceiling = SMITHY_VERSION_CEILINGS[key];
+    if (ceiling && compareSemver(to, ceiling) > 0) {
+      console.log(
+        `Holding ${key} at ${ceiling}: ${to} is above its ceiling (see SMITHY_VERSION_CEILINGS)`,
+      );
+      return;
+    }
```
cogwirrel: no ceilings please

docs/src/content/docs/en/guides/smithy-project.mdx
nx-plugin-for-aws: Done — now just links to the install guide.

docs/src/content/docs/en/guides/ts-smithy-api.mdx
nx-plugin-for-aws: Done.

packages/nx-plugin/src/smithy/project/generator.spec.ts
nx-plugin-for-aws: Good catch — reads `SMITHY_VERSIONS.cli` now, like the other two assertions in this file already did.

packages/nx-plugin/src/utils/init.ts
nx-plugin-for-aws: Moved — `smithy#project` now calls `registerPnpmBuiltDependencies({ mise: true })` itself, and so does the migration (a migrated workspace newly gains `mise`). Off the workspace-wide list, so a workspace with no Smithy project never carries the entry. Covered by a test in each spec.

packages/nx-plugin/src/utils/smithy.ts
nx-plugin-for-aws: Done — the warning is now one line pointing at the install guide.

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Ceilings gone. The codegen artifact stays at 0.50.0 with the reason recorded in the doc comment, and the version update proposes the latest so e2e catches a release that doesn't build. FWIW the specific breakage I hit: 0.51.0 generates a server importing `ServerInterceptor` from `@aws-smithy/server-common`, which no published release of that package exports, so the SSDK fails to bundle.

scripts/update-versions.ts
nx-plugin-for-aws: Removed.

nx-plugin-for-aws:
Thanks — all seven addressed in `76cb31c0`, rebased onto latest main:

- **No ceilings.** `SMITHY_VERSION_CEILINGS` is gone from both `versions.ts` and `update-versions.ts`. The version update now proposes whatever Maven reports as latest and lets e2e catch a release that doesn't build. The codegen artifact stays at 0.50.0 with the reason recorded in the doc comment beside it — 0.51.0 generates a server importing `ServerInterceptor` from `@aws-smithy/server-common`, which no published release of that package exports, so the SSDK fails to bundle.
- **No scoop.** Both guides and the generator's warning now just link to the Smithy CLI install guide.
- **No fixed version in the test.** Reads `SMITHY_VERSIONS.cli`, matching the two assertions in that file that already did.
- **Allowlist moved out of init.** `smithy#project` calls `registerPnpmBuiltDependencies({ mise: true })` itself, and so does the migration — a migrated workspace newly gains `mise`. A workspace with no Smithy project never carries the entry. Covered by a test in each spec.

One thing worth flagging that came out of this: the translation job's output for this PR was corrupt in four languages — ko/pt/vi `smithy-project.mdx` each had a mangled duplicate of the document tail appended (an unclosed code fence and a second `</Steps>`), which is what broke the docs build. Fixed by truncating them at their proper end. The `jp/ts-smithy-api.mdx` replacement characters pre-exist on main and don't affect the build, so I left those alone.

packages/nx-plugin/src/smithy/project/generator.ts
```diff
@@ -28,35 +28,155 @@ import {
...
+    return [
+      ...commands,
+      cmd.cp(
+        `${SOURCE_PROJECTION}/model/model.json`,
+        `${BUILD_DIR}/model.json`,
+      ),
+    ];
+  }
+
+  return [
```
cogwirrel: split out smithy compile to be the smithy model build. and add a generate ssdk target which does the ssdk generation. generate-ssdk depend on compile, and build depend on compile and generate-ssdk

scripts/update-versions.ts
```diff
@@ -352,6 +355,131 @@ const updateGitSecrets = async (
...
+  }
+
+  // The TypeScript codegen artifact tracks its own releases.
+  const codegen =
+    'software.amazon.smithy.typescript:smithy-aws-typescript-codegen';
+  const latestCodegen = await getLatestMavenVersion(codegen);
+  if (latestCodegen) {
+    await rewrite(codegen, SMITHY_VERSIONS[codegen], latestCodegen);
+  }
```
cogwirrel: feels wrong to have a special case here. let's have a JAVA_VERSIONS in versions.ts and these are all just fetched and checked with maven. also MISE_VERSIONS for smithy cli which are all checked with mise.

packages/nx-plugin/src/smithy/project/generator.ts
cogwirrel: Done in cc653676. `compile` is now the model build alone, and a new `generate-ssdk` target does the install + tsc + rolldown, with `dependsOn: ['compile']`; `build` depends on both. A shape library has no SDK so it keeps the single target.

One thing worth flagging: the two targets both write into `dist/{projectRoot}/build`, so I listed outputs per artifact (`build/openapi` and `smithy/` for compile, `build/ssdk` for generate-ssdk) rather than the whole directory — otherwise restoring one from cache discards the other's output.

The migration moves existing projects onto the same structure from the same shared helpers, with a test asserting a migrated project's targets deep-equal a freshly generated one's (mutation-tested that it actually fails if the migration skips the new target).

scripts/update-versions.ts
cogwirrel: Done in a834aba5 — `SMITHY_VERSIONS` is gone, replaced by `JAVA_VERSIONS` (keyed by Maven coordinate, all fetched from Maven Central) and `MISE_VERSIONS` (keyed by mise tool name, all checked with `mise latest`).

Both now go through the same `applyUpdatedVersions` the TS/Python/Terraform sets use, so the bespoke `updateSmithyVersions` pass is gone along with the CLI/artifact lockstep gate it carried — each pin just moves to its own latest and e2e catches a combination that doesn't build, per your earlier comment about ceilings.

Verified against the real registries: all five Maven coordinates resolve (codegen correctly reports 0.51.0 as available, so the weekly run will propose it), and `mise latest smithy` returns 1.72.1.

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -110,39 +109,36 @@ Building a service project generates an OpenAPI specification and a TypeScript S
...
 
 <NxCommands commands={['build my-shapes']} />
 
-## Depending on a Shape Library
+The CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
 
-Building a shape library writes an assembled model to `dist/<my-shapes>/build/model/model.json`. This single file contains every shape the library defines, along with any it depends on, so a consumer only ever declares the libraries it references directly.
+:::tip[Building in CI]
+`mise` reads the CLI's release metadata from the GitHub API, whose unauthenticated rate limit is shared per IP address and easily exhausted on a CI runner — the download then fails with `403 Forbidden`. Set `GITHUB_TOKEN` in your build environment to lift it. Subsequent builds hit the local cache and make no request at all.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/ts-smithy-api.mdx
```diff
@@ -576,10 +576,16 @@ export const Echo: EchoOperation<ServiceContext> = async (input, ctx) => {
...
-The Smithy model project uses [Docker](https://www.docker.com/) to build the Smithy artifacts and generate the TypeScript Server SDK:
+The Smithy model project uses the [Smithy CLI](https://smithy.io/2.0/guides/smithy-cli/index.html) to build the Smithy artifacts and generate the TypeScript Server SDK:
 
 <NxCommands commands={['build <model-project>']} />
 
+The CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+
+:::tip[Building in CI]
+`mise` reads the CLI's release metadata from the GitHub API, whose unauthenticated rate limit is shared per IP address and easily exhausted on a CI runner — the download then fails with `403 Forbidden`. Set `GITHUB_TOKEN` in your build environment to lift it. Subsequent builds hit the local cache and make no request at all.
+:::
```
cogwirrel: remove

e2e/src/smoke-tests/standalone.spec.ts
```diff
@@ -82,10 +82,13 @@ const categorizeGenerators = () => {
...
-// projects build a Docker image) and can't run checkov (ts#infra and
-// terraform#project). These stay covered on the Linux standalone leg and the
-// windows-latest dungeon-adventure test.
+// The CodeBuild Windows runner lacks Docker (agents, MCP servers and rdb build a
+// Docker image) and can't run checkov (ts#infra and terraform#project). These stay
+// covered on the Linux standalone leg and the windows-latest dungeon-adventure
+// test.
+//
+// Smithy is not among them: it builds with the Smithy CLI, which mise resolves on
+// demand, rather than in a container.
```
cogwirrel: delete this part of the comment

packages/nx-plugin/src/migrations/latest/smithy-build-without-docker/migration.spec.ts
```diff
@@ -0,0 +1,363 @@
...
+import {
+  SMITHY_PROJECT_GENERATOR_INFO,
+  smithyProjectGenerator,
+} from '../../../smithy/project/generator';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test';
+import { MISE_VERSIONS, TS_VERSIONS } from '../../../utils/versions';
+import migration from './migration';
+
+/**
+ * The `build.Dockerfile` a v1.0.0-rc.58 service project holds, and the
```
cogwirrel: no need to name everything after specific versions. just call this legacy smithy build dockerfile or similar. same for filename of dockerfile etc.

packages/nx-plugin/src/utils/version-upgrade-migration/sync-embedded-versions.ts
```diff
@@ -391,10 +405,45 @@ const syncTargetToolPins = async (tree: Tree): Promise<string[]> => {
...
+ * there is no dependency for a generator to own.
+ */
+const syncSmithyMavenPins = async (tree: Tree): Promise<string[]> => {
+  const smithyBuilds: string[] = [];
+
+  visitNotIgnoredFiles(tree, '.', (path) => {
+    if (path.endsWith('smithy-build.json')) {
+      smithyBuilds.push(path);
+    }
+  });
```
cogwirrel: these should be within smithy projects we've generated only, ie metadata says smithy#project

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -271,6 +273,71 @@ export const VENDORED_VERSIONS = {
   'git-secrets': '1.3.0',
 } as const;
 
+/**
+ * Versions of Java dependencies added by generators, keyed by Maven coordinate.
+ *
+ * Every entry is resolved from Maven Central by the version update, and lands in a
+ * generated `smithy-build.json` as `<group>:<artifact>:<version>`.
```
cogwirrel: not necessarily true that it'd go in a smithy-build.json - write this as non smithy specific

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -271,6 +273,71 @@ export const VENDORED_VERSIONS = {
...
+ *
+ * The TypeScript codegen artifact sits behind its latest release: 0.51.0 generates
+ * a server importing `ServerInterceptor` from `@aws-smithy/server-common`, which no
+ * published release of that package exports, so the generated SDK fails to bundle.
+ * Nothing pins it back — the version update proposes the latest and the e2e tests
+ * catch a release that doesn't build.
```
cogwirrel: remove

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -271,6 +273,71 @@ export const VENDORED_VERSIONS = {
...
+  smithyValidationModelDependency: javaMavenDependency(
+    'software.amazon.smithy:smithy-validation-model',
+  ),
+  smithyOpenApiDependency: javaMavenDependency(
+    'software.amazon.smithy:smithy-openapi',
+  ),
+  smithyTypeScriptCodegenDependency: javaMavenDependency(
+    'software.amazon.smithy.typescript:smithy-aws-typescript-codegen',
+  ),
+});
```
cogwirrel: this belongs in the smithy generator or smithy utils, not versions.ts

scripts/update-versions.ts
```diff
@@ -352,6 +356,83 @@ const updateGitSecrets = async (
...
+ * Asked of mise rather than of the tool's own releases, so the version this vends
+ * is one the tool that resolves it can actually install — a release mise's registry
+ * hasn't picked up yet would leave every build failing to resolve it.
+ *
+ * Run through `pnpm dlx` rather than added to this repo's dependencies: the `mise`
+ * npm package fetches its own binary in a `preinstall` and publishes nothing for
+ * Windows, so depending on it would fail `pnpm i` on every Windows CI job. This
+ * script only ever runs on Linux.
+ */
+const getLatestMiseVersion = (tool: string): string | undefined => {
```
cogwirrel: type tool as IMiseVersion

scripts/update-versions.ts
```diff
@@ -352,6 +356,83 @@ const updateGitSecrets = async (
...
+/**
+ * The latest release of a Maven artifact, from its `maven-metadata.xml` on Maven
+ * Central.
+ *
+ * The metadata document rather than the search API: the latter's index lags well
+ * behind (it reported smithy-model 1.58.0 while 1.72.1 was current), which would
+ * silently hold these pins back.
+ */
+const getLatestMavenVersion = async (
+  coordinate: string,
```
cogwirrel: type properly as IJavaVersion (or what is actually meant to be passed)

docs/src/content/docs/en/guides/smithy-project.mdx
nx-plugin-for-aws: Removed — the section now just links to mise, no CI tip block. (66db8c5)

docs/src/content/docs/en/guides/ts-smithy-api.mdx
nx-plugin-for-aws: Removed. (66db8c5)

e2e/src/smoke-tests/standalone.spec.ts
nx-plugin-for-aws: Deleted that part of the comment. (66db8c5)

packages/nx-plugin/src/migrations/latest/smithy-build-without-docker/migration.spec.ts
nx-plugin-for-aws: Renamed — the constants are LEGACY_* now and the fixture is legacy-service.Dockerfile.fixture local to this migration, so nothing is named after a specific release. (66db8c5)

packages/nx-plugin/src/utils/version-upgrade-migration/sync-embedded-versions.ts
nx-plugin-for-aws: Scoped — syncSmithyMavenPins now walks projects whose recorded generator is smithy#project rather than every smithy-build.json, so a hand-written one keeps the versions the user chose. Added a test covering that. (66db8c5)

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Reworded to be non-Smithy-specific: it just says the coordinate is named <group>:<artifact>:<version> where a generator writes it. (66db8c5)

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Removed. (66db8c5)

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Moved to utils/smithy.ts as smithyMavenVersions, imported by the generator from there. (66db8c5)

scripts/update-versions.ts
nx-plugin-for-aws: Typed as IMiseVersion. (66db8c5)

scripts/update-versions.ts
nx-plugin-for-aws: Typed as IJavaVersion. (66db8c5)

cogwirrel (dismissed):

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -113,7 +113,40 @@ Smithy projects build with the [Smithy CLI](https://smithy.io/2.0/guides/smithy-
...
+On macOS and Linux the CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+
+### Building on Windows
+
+`mise` publishes no Windows package to npm, so on Windows the Smithy CLI is a prerequisite you install yourself. Install it once following the [Smithy CLI installation guide](https://smithy.io/2.0/guides/smithy-cli/cli_installation.html) (for example `winget install smithy` or `scoop install smithy`), and make sure `smithy` is on your `PATH`. A Smithy project generated on Windows runs `smithy` directly rather than through `mise`.
+
+:::caution[Upgrading the CLI]
+Because the CLI is installed globally rather than pinned in your workspace, upgrading the workspace with `nx migrate` does **not** upgrade it. When a migration moves the Smithy dependencies forward, upgrade your global Smithy CLI to a matching version yourself — the CLI cannot load a model built against a newer `smithy-model`.
```
cogwirrel: remove `` — the CLI cannot load a model built against a newer `smithy-model```

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -113,7 +113,40 @@ Smithy projects build with the [Smithy CLI](https://smithy.io/2.0/guides/smithy-
...
+    }
+  }
+}
+```
+
+To go back to `mise` resolving the CLI, restore the `npx -y mise@<version> exec smithy@<version> --` prefix.
+
+:::caution
+When you invoke the CLI directly, its version is no longer pinned in the workspace, so `nx migrate` cannot move it forward — upgrade your global Smithy CLI independently when upgrading the workspace.
+:::
```
cogwirrel: Remove

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -113,7 +113,40 @@ Smithy projects build with the [Smithy CLI](https://smithy.io/2.0/guides/smithy-
 
 <NxCommands commands={['build my-shapes']} />
 
-The CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+On macOS and Linux the CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+
+### Building on Windows
+
+`mise` publishes no Windows package to npm, so on Windows the Smithy CLI is a prerequisite you install yourself. Install it once following the [Smithy CLI installation guide](https://smithy.io/2.0/guides/smithy-cli/cli_installation.html) (for example `winget install smithy` or `scoop install smithy`), and make sure `smithy` is on your `PATH`. A Smithy project generated on Windows runs `smithy` directly rather than through `mise`.
```
cogwirrel: Add another short paragraph afterwards covering:

- using wsl instead to allow use of mise option
- other developers working on the project on linux/mac will need to have smithy cli

docs/src/content/docs/en/guides/smithy-project.mdx
```diff
@@ -113,7 +113,40 @@ Smithy projects build with the [Smithy CLI](https://smithy.io/2.0/guides/smithy-
 
 <NxCommands commands={['build my-shapes']} />
 
-The CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+On macOS and Linux the CLI is resolved by [mise](https://mise.jdx.dev/), which the build fetches on demand, so there is nothing to install — it downloads and caches the pinned version the first time you build.
+
+### Building on Windows
```
cogwirrel: Swap this section to the ts-smithy-api guide, and put the short version linking back to that guide on this page instead.

packages/nx-plugin/src/smithy/project/files/ssdk-bundle/ssdk.rolldown.config.mjs.template
```diff
@@ -45,24 +45,40 @@ const re2WasmPlugin = () => ({
...
+    const self = path.basename(id).replace(/\.[cm]?ts$/, '');
+    const withoutSelfImport = code.replace(
+      new RegExp(`^\\s*import\\s+[^;]*?from\\s+["']\\./${self}["'];?\\s*$`, 'gm'),
+      '',
+    );
+    return withoutSelfImport === code
+      ? null
+      : { code: withoutSelfImport, map: null };
+  },
+});
```
cogwirrel: this seems very odd, please investigate more deeply, this fix is a hack. we should also not vend long comments like this into the code.

packages/nx-plugin/src/smithy/project/files/ssdk-bundle/ssdk.rolldown.config.mjs.template
```diff
@@ -45,24 +45,40 @@ const re2WasmPlugin = () => ({
...
 /**
- * The Server SDK's own TypeScript compiles it before this runs, so both passes
- * bundle emitted JavaScript and declarations rather than the generated sources —
- * see the compile target's commands for why.
+ * Bundle the Server SDK straight from the generated TypeScript sources, emitting
+ * the module and its declarations in one pass. The sources rather than a `tsc`
+ * pre-pass: the SDK merges an `interface` with a `namespace` and a `class` across
+ * an operation↔service import cycle, which tsc mis-resolved on a case-insensitive
+ * filesystem, emitting a module that imported a name from itself.
  */
```
cogwirrel: too much historic context, delete this comment

packages/nx-plugin/src/smithy/project/files/ssdk-bundle/ssdk.rolldown.config.mjs.template
nx-plugin-for-aws: Investigated: it's a path-separator bug in the pinned Smithy Java codegen (0.50.0) — on Windows its import writer emits a module that imports its own exports (`import { Echo, ... } from "./Echo"` inside Echo.ts). We can't patch the Maven artifact and can't move off 0.50.0 (0.51.0 imports a ServerInterceptor that @aws-smithy/server-common doesn't export). A module importing itself is invalid on every platform, so rather than match a Windows filename I now drop any import whose specifier resolves to the importing module itself — a platform-agnostic no-op elsewhere. Reproduced locally by injecting the self-import into a Linux-generated project: bundle fails without it, succeeds with it, same module + types emitted. (7f3a499)

packages/nx-plugin/src/smithy/project/files/ssdk-bundle/ssdk.rolldown.config.mjs.template
nx-plugin-for-aws: Trimmed to two lines describing what the drop does and why. (7f3a499)

docs/src/content/docs/en/guides/smithy-project.mdx
nx-plugin-for-aws: Moved — the detailed section now lives in the ts#smithy-api guide (Building on Windows), and this page keeps a one-line pointer to it. (7f3a499)

docs/src/content/docs/en/guides/smithy-project.mdx
nx-plugin-for-aws: Added: a note on developing in WSL to keep the mise path, and that a project generated on Windows commits a bare `smithy` compile command, so macOS/Linux collaborators need the CLI on PATH too (with a pointer to switch it back to mise). (7f3a499)

docs/src/content/docs/en/guides/smithy-project.mdx
nx-plugin-for-aws: The switch-to/from instructions are kept (as 'Choosing how the CLI is resolved' in the ts#smithy-api guide) since they're referenced by the collaborators note, but the standalone section is gone from this page. (7f3a499)

docs/src/content/docs/en/guides/smithy-project.mdx
nx-plugin-for-aws: Removed that clause. (7f3a499)

cogwirrel (dismissed):

cogwirrel (approved):

PR #1045, opened 2026-08-05
fix(agentcore-gateway): destroy WAF log group on delete and fix multi-agent VPC docs

nx-plugin-for-aws:
### Reason for this change

While testing the HTTP AgentCore Gateway fronting VPC-deployed agents end to end (CDK + Terraform, deploy + invoke), two issues surfaced:

1. **WAF log group orphaned on rollback (CDK).** The vended `AgentCoreGateway` construct's WAF log group uses a deterministic name (`aws-waf-logs-<prefix>-<addr>`) and, without an explicit removal policy, inherited CloudFormation's default **RETAIN** for log groups. When a first deploy rolled back, the log group was left behind, so the retry failed with `… already exists` and required a manual `aws logs delete-log-group` before redeploying.

2. **CDK docs snippet breaks with more than one agent.** The VPC tip in the gateway→agent connection guide showed `RuntimeNetworkConfiguration.usingVpc(this, { … })`, passing the stack as the scope. The CDK L2 creates `new ec2.SecurityGroup(scope, 'SecurityGroup')` under that scope, so fronting two agents (the whole point of a gateway) throws `There is already a Construct with name 'SecurityGroup'`.

### Description of changes

- Set `removalPolicy: cdk.RemovalPolicy.DESTROY` on the WAF log group in the vended CDK `agentcore-gateway` construct template so it is torn down with the stack instead of orphaned.
- Add a new migration (`latest-gateway-waf-log-group-removal`) that adds the same removal policy to the WAF log group in **existing** workspaces' vended constructs (idempotent, skips diverged/http-only constructs, reports a manual follow-up if it can't apply cleanly).
- Fix the CDK VPC snippet in `guides/connection/agentcore-gateway-agent.mdx` to give each agent its own scope for `usingVpc`, avoiding the `SecurityGroup` construct-id collision, and note the AgentCore Runtime unsupported-AZ constraint.

- New unit tests for the migration (apply, idempotency, user-added-policy left alone, no-web-acl no-op, no-construct no-op) — all pass, with a snapshot of the patched construct.
- Ran the full `@aws/nx-plugin` test suite with snapshot updates: **3283 tests pass**.
- Compiled + linked the plugin and ran the compiled migration against a real workspace generated on an older release: it patched the WAF log group cleanly (`nextSteps: []`) and the workspace's `common-constructs` still compiles.
- Confirmed the migration is assembled into `migrations.json` as `latest-gateway-waf-log-group-removal`.

### Issue # (if applicable)

N/A

### Checklist

----


nx-plugin-for-aws:
**Note on the `Smoke Tests - trivy` failure:** this is unrelated to the changes in this PR. Trivy reports 2 HIGH CVEs (`HIGH: 2, CRITICAL: 0`) in the **Python agent base container images** (`e2e_test.py_project` / `e2e_test.py_langchain_project` agent/mcp-server images). This PR's diff touches only a CDK TypeScript construct template, migration files, and a docs `.mdx` — no Python, container, or dependency files (see `git diff`), so it cannot affect those images. The finding is from a newly-published CVE in the upstream base image that trivy flags on any build after disclosure; addressing it is separate base-image work, out of scope for this change. All other 62 checks pass.

cogwirrel (approved):

cogwirrel (approved):

nx-plugin-for-aws:
Pushed a second commit extending this PR to also fix the **rest-api** construct's WAF log group (same latent RETAIN-on-deterministic-name issue as the gateway) and covering both constructs in the migration (renamed `waf-log-group-removal`). The migration's rewrite is scoped to the `WebAclLogs` log group so the rest api's access-log group is untouched — verified end-to-end against a real workspace and via unit tests.

CI is green except `Smoke Tests - trivy`, which fails identically to before: the failing tasks are exclusively the **Python agent / mcp-server container images** (`e2e_test.py_project`, `e2e_test.py_langchain_project`), a pre-existing HIGH CVE in the upstream Python base image. Neither commit in this PR touches Python, containers, or dependencies (the diff is a CDK TypeScript construct, migration, and docs), so it cannot affect those images. All 62 other checks pass, including all 5 `migrate` smoke tests.

PR #1047, opened 2026-08-06
fix(py#agent): construct agents in a FastAPI lifespan instead of at import time

AlexTo:
### Reason for this change

py#agent built the Strands agent / LangChain graph eagerly at module import
time. Concretely, this broke the build: generating the OpenAPI spec failed
when a website connects to an agent that is itself connected to an MCP
server. The MCP client only skips `get_agentcore_runtime_config()` (which
needs `RUNTIME_CONFIG_APP_ID`) when `LOCAL_DEV=true`, and the `build-openapi`
target doesn't set `LOCAL_DEV` (only `dev` does) - so importing the module to
generate the OpenAPI spec hit that code path and failed with no
`RUNTIME_CONFIG_APP_ID` set. More generally, eager construction breaks any
codepath that imports the module before AgentCore env vars are set (OpenAPI
spec generation, tests), and for AgentCore, construction belongs at
container/session startup rather than import time, since each session gets
its own dedicated container.

### Description of changes

- HTTP and AG-UI templates for both Strands and LangChain now construct the
  agent/graph inside a FastAPI `lifespan` handler, storing it on `app.state`
  instead of a module-level global.
- Templates reorganized from flat per-protocol dirs (`http`, `http-langchain`,
  `common`, `common-langchain`, ...) into `files/<framework>/<protocol>`, so
  Strands and LangChain each own their own `init.py` - needed because the
  lifespan body now differs per framework, whereas the old `http/init.py` was
  shared between them.
- Added the `py-agent-lifespan-construction` Nx migration to move existing
  generated projects onto the new shape. It pattern-matches each file via
  GritQL before rewriting; anything diverged from the generated shape is left
  completely untouched, with a `nextSteps` entry pointing at the relevant
  template as a manual-migration reference.

- Updated `generator.frameworks.spec.ts` / snapshot tests to assert the
  lifespan handler shape (`app.state.agui_agent`, `app.state.graph`, etc.)
  instead of the old eager-construction shape.
- Manually validated the migration end-to-end against a generated workspace
  with locally linked build output (`py#agent` -> `latest-py-agent-lifespan-construction`).
- Manually validated LangChain and Strands end to end (HTTP and AG-UI), confirming
  the session ID is still propagated correctly now that the agent/graph is built
  inside the `lifespan` handler rather than at import time.

### Checklist

----


cogwirrel (dismissed):
Thanks @AlexTo!

cogwirrel (approved):

PR #1057, opened 2026-08-07
ci: single release script with publish retry and rc/stable toggle

nx-plugin-for-aws:
### Reason for this change

The release job's publish logic was spread across inline workflow bash and several scripts, and had no resilience to a transient publish failure (the class of failure that lost `@aws/nx-plugin-mcp@1.0.0-rc.61` to a Sigstore transparency-log 409). There was also no clean way to move the project off release candidates onto stable releases.

### Description of changes

Consolidated into one `scripts/release.ts` that the release job runs:

1. `nx release version` (writes the pending version into the dist manifests)
2. stamp `migrations.json` with the shipping version — it's assembled unversioned; now via the imported `stampMigrationsFile`, no subprocess
3. `nx release --skip-publish` (tag + changelog)
4. `nx release publish`, **retried on a transient failure** — nx skips a package already on the registry and tolerates an already-present 409, so a retry only re-attempts what didn't land, and is correct through npm's publish-time scanning window.

**rc/stable toggle — a committed constant, `RELEASE_TRAIN: 'rc' | 'stable'`** (not an env var):

- `'rc'` (default): `--preid rc`, bumping `1.0.0-rc.x → 1.0.0-rc.x+1`.
- `'stable'` while the current tag is still an rc: `--specifier <STABLE_PROMOTION_BUMP> --preid rc`, promoting the rc series to its stable version (e.g. `1.0.0-rc.x → 1.0.0`). The `--preid` is what lets nx's git-tag resolver match the rc tag; the explicit specifier drops the `-rc` suffix. (No flags errors here — nx can't resolve an rc tag as the base for a stable bump.)
- `'stable'` once on a stable tag: no flags, following conventional-commits semver (`1.0.1`, `1.1.0`, …).

The rc-vs-stable transition is auto-detected from the current git tag.

`stamp-migrations.ts` keeps its CLI (the migrate smoke test uses it with `--out`) but its core is now the importable `stampMigrationsFile`.

**No backfill.** Per discussion, a package a past release failed to publish (`mcp@rc.61`) is left as-is; the publish retry prevents the transient failures that caused it, which is the robustness that matters going forward.

- **Release e2e smoke test** against the e2e Verdaccio (real `npm publish`/`view` + real `stamp-migrations.ts`): confirms stamping versions **every** entry in the shipped `migrations.json`, and that re-publishing an already-present version is a benign conflict — the property that makes the publish retry safe. Wired into the CI smoke matrix (green in CI).
- **Full rc/stable switchover, end-to-end against Verdaccio** (real `nx release version` + `npm publish` + dist-tags, asserting the published version and `latest` at each step):
  - `rc` train, `v1.0.0-rc.61` → `1.0.0-rc.62` (rc continues)
  - `stable` train, `v1.0.0-rc.62` → `1.0.0` (promotion)
  - `stable` train, `v1.0.0` + `fix:` → `1.0.1`; `+ feat:` → `1.1.0` (conventional semver)
  - `rc` train, `v1.1.0` → `1.1.1-rc.0` (start a new rc series)
- Verified `nx release publish` behaviour empirically: a re-run skips already-published packages and exits 0.
- `stamp-migrations.ts` CLI still works (migrate lane depends on it); full migration unit suite (408 tests), `compile`, and lint pass.

### Issue # (if applicable)

Follow-up to the reverted #1054.

### Checklist

----


nx-plugin-for-aws:
Thanks — actioned all points from the latest review:

- **🟠 `--tag latest` silently dropped by `npm exec`** (real bug): fixed. `nx release publish` now runs via `pnpm` (`run('pnpm', ['nx','release','publish','--tag','latest'])`), which forwards the flag. Verified `npm exec` swallows `--tag` while `pnpm` passes it through.
- **🟡 e2e used `localeCompare` not the production comparator**: the release e2e now imports and uses `compareVersions` (re-exported for exactly this), so the latest-advance rule under test matches production for prereleases.
- **Coverage gap on the stamping logic**: extracted `stampManifest` as a pure, directly unit-tested helper (version stamp, `@aws/*` sibling-pin rewrite, wrong-package guard, no-mutation) — the most dangerous logic now has direct coverage, not only the e2e mirror. The script uses it and still re-reads the file to confirm the stamp landed on disk.
- **Vacuous-lane risk**: the release lane now throws if it's the targeted test but `NX_E2E_LOCAL_REGISTRY` is unset, so a missing registry is loud rather than a green no-op.
- **`MIGRATE_SKIP_START_VERSIONS` staleness**: left in place for now, as agreed — to be removed once the backfill has repaired rc.61.

424 unit tests + 6 real-Verdaccio e2e tests pass locally; lint/compile clean.

nx-plugin-for-aws:
Substantial rework in response to the latest review — thank you, all three findings were correct and are addressed by **delegating to `nx release publish` instead of reimplementing it**:

**#3 (registry-verification miscalibrated for npm's scanning window) — the core fix.** I investigated `nx release publish`'s executor and verified its behaviour empirically against Verdaccio:
- Re-running **skips** a package already on the registry (`Skipped … already exists`, exit 0), and in a partial release publishes **only the missing** package.
- It tolerates an already-present **409** as success (`isAlreadyPublishedPublishError`).

So during the ~5–15 min a fresh publish is invisible to `npm view`, a re-run's view misses it → nx re-publishes → 409 → **treated as success**. The delay can't turn a real publish into a failed release. The custom `publishAndVerify`/`publishRelease` retry-and-verify layer (which had the 15s-vs-300s budget problem and re-published into the window) is **removed**. The pending release is now just `nx release publish --tag latest`.

**#1 (mutation gap — e2e paraphrased the latest-advance rule).** `shouldAdvanceLatest` is now a pure exported helper; the release e2e **imports** it (and `stampManifest`) and drives them against real Verdaccio, so breaking the production rule fails the test. Confirmed the old gap is closed.

**#2 (scope-prefix stamping landmine).** `stampManifest` re-pins siblings by **exact name against the release set** (from `nx.json`), not `@aws/` prefix — so a third-party `@aws/`-scoped dep (`@aws/pdk`) is never rewritten. There's a unit test asserting `@aws/pdk`/`@aws-sdk/*` are left alone.

**Backfill** now: build-from-tag → `stampManifest` (still needed — tags carry `0.0.0`) → `nx release publish --projects <p> --tag <holding>` (nx skips/tolerates/verifies) → advance `latest` only if newer → `npm dist-tag rm <holding>` so no stray tag remains (per the "no-tag end state" ask).

Net **−229 lines**. 18 unit tests + 4 real-Verdaccio e2e tests pass; compile/lint clean. Also fixed the earlier `npm exec` → `pnpm` `--tag` bug (now moot for the release path since nx is invoked via pnpm).

nx-plugin-for-aws:
Rebased onto main after the `revert(ci)` of #1054, and squashed to a single self-contained commit. The revert removed the migrate-test fixes too, so this PR now reintroduces those (the sound half of #1054) alongside the reworked release-publish.

Also, following the "does nx do the sibling stamping?" question: **it does** — `nx release version <version>` writes the version into the dist manifests *and* re-pins sibling packages (`updateDependents: 'always'`), which is what the original release flow used. So the backfill now delegates versioning to nx as well:

- Backfill builds all release packages from the git tag, runs `nx release version <shipped-version>` (nx stamps version + sibling pins — the tag's manifest is `0.0.0`, so this is what prevents a `0.0.0` publish), then `nx release publish --projects <missing> --tag <transient>`. The hand-rolled `stampManifest` is gone.
- (The fixed release group can't be versioned a subset at a time, so all three are built/versioned together; only the missing one is published.)
- A dist-version assertion (`built.version === version`) guards against publishing anything unexpected.

Net effect unchanged from the previous review round otherwise: nx owns publish (skip/409-tolerance/subset) and now versioning; this script only adds backfill discovery, the transient-tag/advance-latest dance, and the ci.yml wiring. 14 unit tests + 4 real-Verdaccio e2e tests green.

nx-plugin-for-aws:
You're right, and thank you — that was my error: I'd verified the fixed-group `--projects` constraint earlier (for `nx release version`) and applied build-all/version-all there, but inconsistently left `--projects` on the *publish*, which hits the identical constraint and failed silently. Fixed:

- **Backfill now publishes the built dist directly with `npm publish --tag <transient>`** (reviewer option 1), sidestepping the fixed-group constraint. The version is already stamped by `nx release version`, and there's a `dist.version === version` guard before publishing.
- **409 tolerance kept** for belt-and-braces: a new pure `isAlreadyPublished` helper accepts `EPUBLISHCONFLICT` / "previously published" / `409 … already present`, so a re-run or a re-publish racing npm's scanning window is tolerated. (The main release still goes through `nx release publish`, which has its own tolerance.)
- **The coverage gap is closed.** The release e2e now runs the **real publish command** against Verdaccio, including two new tests: re-publishing an already-present version is tolerated, and a **real npm 409 is asserted to classify as already-published**. This is the path that failed every prior round — it now executes in the test.

Verified locally against real Verdaccio (6 e2e tests) + 17 unit tests; compile/lint clean. The `nx release version` versioning rework you validated is unchanged.

cogwirrel (approved):

PR #1075, opened 2026-08-11
fix(py#agent): construct A2A agents in a FastAPI lifespan instead of at import time

AlexTo:
### Reason for this change

Adding a `SessionManager` to an agent (for persisted session state) requires a real session id at construction time. The Strands A2A template didn't build its agent eagerly by itself — `_agent_ctx.__enter__()` just returns a lazy session-routing proxy that defers real `Agent` construction until first use. But `strands.multiagent.a2a.executor.StrandsA2AExecutor.__init__`, in its deprecated single-`agent=` mode, does `getattr(agent, "_session_manager", None)` — confirmed directly in the installed SDK source — which the proxy's `__getattr__` happily serves by building the real underlying `Agent` right then, under a bogus `"default"` session, since no request has bound a real session id yet. That call happens inside `A2AServer.__init__`, which the old template constructed at plain module import time — so the proxy's laziness gets forced open anyway, with no legitimate session id to construct against, breaking any `get_agent()` that requires one (e.g. via a `SessionManager`).

The earlier `py-agent-lifespan-construction` migration moved HTTP and AG-UI (both frameworks) to lifespan-based construction, but deliberately left A2A alone, reasoning that `with_session_id` only touches its agent factory lazily on first use, so the eager `_agent_ctx.__enter__()` was never actually eager in practice — true as far as it went, but it didn't account for the executor's own probe forcing that laziness open regardless.

### Description of changes

- **Strands A2A template**: agent now built inside a FastAPI `lifespan` handler and read off `app.state.agent`. `A2AServer` uses `agent_factory=` (not the deprecated `agent=`), sidestepping the `_session_manager` probe entirely — that mode takes a different code branch with no such check. The one synchronous call `A2AServer.__init__` makes before `lifespan` has run gets a `SimpleNamespace` placeholder (cast to `Agent` so `ty check` passes) carrying just the real `name`/`description` needed for the agent card.
- **LangChain A2A template**: moves `_graph = get_agent()` into `lifespan` too, storing on `app.state.agent` — for the same reason HTTP/AG-UI already do this: it can require env vars or network setup not guaranteed available at import time (test collection, OpenAPI spec generation tooling).
- **`with_session_id_strands.py`/`session_context.py`**: return type annotations changed from `Iterator[T]` to `Generator[T]`, resolving a Pylance deprecation warning on `@contextmanager` functions.
- **New migration** `py-agent-a2a-lifespan-construction` brings existing generated A2A agents (both frameworks) up to the new shape via GritQL rewrites, leaving diverged files untouched with `nextSteps` guidance.
- Updated `generator.protocols.spec.ts` assertions to match the new template output.

- `migration.spec.ts` (14 tests) and `generator.protocols.spec.ts` (17 tests) pass on an uncached run.
- Verified the rendered Strands template compiles (`python -m py_compile`) and passes `ty check` (Astral's type checker) cleanly, both before/after reproducing and fixing a real type error found there.
- End-to-end: generated a real workspace (`test-a2a-agents`) with both Strands and LangChain A2A agents using the locally-linked updated generator/migration. Confirmed the generated `main.py` files match the new template shape, and `ty check` passes cleanly against both packages' real installed dependencies.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/py-agent-a2a-lifespan-construction/metadata.json
+++ b/packages/nx-plugin/src/migrations/latest/py-agent-a2a-lifespan-construction/metadata.json
@@ -1,0 +1,3 @@
+{
+  "description": "Move py#agent A2A agent construction (Strands and LangChain) out of module import time and into a FastAPI lifespan handler, stored on app.state; also updates with_session_id/session_id_context to use Generator instead of the deprecated Iterator return type"
+}
--- a/packages/nx-plugin/src/migrations/latest/py-agent-a2a-lifespan-construction/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/py-agent-a2a-lifespan-construction/migration.spec.ts
@@ -1,0 +1,363 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { addProjectConfiguration, type Tree } from '@nx/devkit';
+import { PY_AGENT_GENERATOR_INFO } from '../../../py/agent/generator';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test';
+import migration from './migration';
+
+const PROJECT_ROOT = 'apps/test-project';
+const AGENT_DIR = `${PROJECT_ROOT}/proj_test_project/agent`;
+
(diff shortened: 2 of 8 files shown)


cogwirrel (approved):
Thanks for the fix @AlexTo !

AlexTo:
> AlexTo !

you're welcome

PR #1076, opened 2026-08-12
fix(smithy#project): pre-install the Smithy CLI in CI so a GitHub API blip cannot fail a build

nx-plugin-for-aws:
### Reason for this change

The `Smoke Tests - migrate (4/5)` lane failed on an unrelated PR ([run 31496110706](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/31496110706/job/93924591051?pr=1075)). The failing task was `@migrate-test/my-smithy-api-model:compile`:

```
mise ERROR Failed to install aqua:smithy-lang/smithy@1.72.1: HTTP status server error
(500 Internal Server Error) for url (https://api.github.com/repos/smithy-lang/smithy/releases/tags/1.72.1)
```

A generated Smithy project's `compile` target resolves the CLI through `mise` on first use, and mise reads the release's digest from the GitHub API to verify the download. That call happens **mid-build**, so a transient upstream 5xx has no outcome other than a failed compile — which fails the hop, the shard, and the run. The migrate lane is the most exposed to it: a hop scaffolds the full test matrix and builds it, so it is the first build of a Smithy project on a fresh machine with a cold mise cache.

Nothing about the migration under test was wrong. Worth recording what was ruled out, since the symptom points elsewhere:

- **Start versions were correct.** Shard 4/5 correctly selected `1.0.0-rc.66`, published the local build as `1.0.0-rc.70`, and the test's own assertion that `node_modules/@aws/nx-plugin` is exactly the start version passed. 4 migrations were queued and ran, so the hop asserted real work rather than silently skipping.
- **pnpm's release-age gate was not the cause.** It does fire (`Added 1 entry to minimumReleaseAgeExclude ... @aws/create-nx-workspace@1.0.0-rc.66`), but self-resolves: pnpm auto-excludes exact pins, and both ends of the upgrade are pinned exactly. Confirmed by installing a same-day release under pnpm 11.20.0 — it succeeds with an auto-exclude entry.
- **`GITHUB_TOKEN` is already set** workflow-wide for exactly this reason (rate limiting), so this was not a 403.

A second, latent problem in the same area: the action carried its own `smithy-version: '1.72.1'` default, a hand-maintained copy of `MISE_VERSIONS.smithy` whose only guard was a comment saying to keep them in step. Nothing failed if they drifted — the Windows lane would just prepare a different CLI than the one generated projects build with.

### Description of changes

- **Pre-install the pinned Smithy CLI in `init-monorepo`, before any build runs.** The GitHub API call moves into a step that owns nothing else, so a blip retries there (3 attempts, backoff) instead of failing a compile. Once the tool is installed, every build in the job resolves it from the mise cache with no network call at all — verified: a warm `mise exec` makes zero network requests and succeeds with `MISE_OFFLINE=1`.
- **Cache the install across runs**, keyed on the pinned version, so the common case makes no API call either. A cache hit makes the pre-warm step a no-op success.
- **Non-fatal on exhaustion.** Only some lanes build a Smithy project; if the pre-warm can't complete it emits a `::warning::` and those builds resolve the CLI on demand exactly as a user's machine would. This keeps a hard upstream outage from failing lanes that never touch Smithy.
- **Unconditional rather than gated on `setup-smithy`** — seven lanes build a Smithy project (standalone, smithy-api, migrate, test-matrix, local-dev, rdb, terraform-deploy), a warm cache costs seconds, and a lane that grows a Smithy project shouldn't have to remember to opt in.
- **Both pins now come from `versions.ts`** via `.github/scripts/tool-pins.mjs`, so the runner cannot prepare a different CLI than the one a generated project's build command carries. Removes the `smithy-version` input (no call site passed it). The script parses by regex rather than importing, because it runs before `pnpm i`, so there is no tsx to load the module with — and it errors if a pin isn't found exactly once, rather than silently resolving nothing.

Reproduced and verified the mechanism locally against the pinned mise/Smithy versions:

- **Reproduced the failure mode**: a cold `mise install smithy@1.72.1` contacts `api.github.com` (confirmed via `MISE_VERBOSE=1`: `using GitHub API digest for checksum verification`), and fails the install when that call fails.
- **Verified the fix removes the call from the build path**: after the pre-warm, `mise exec smithy@1.72.1 -- smithy --version` makes **zero** network requests (verbose trace empty) and succeeds under `MISE_OFFLINE=1`.
- **Verified the cache-restore path**: restoring only `installs/smithy` makes the pre-warm step a no-op success, and a subsequent build resolves the CLI offline.
- **Verified the retry loop** at 0, 2 and 3+ consecutive failures — it succeeds on recovery, does not sleep after the final attempt, and warns without failing the step on exhaustion.
- **Verified the drift guard fails**: reshaping the `mise:` declaration in `versions.ts` makes both the script and the new unit test fail loudly (``Expected exactly one `mise` pin ... found 0``) rather than silently preparing nothing.
- Added a unit test in `smithy.spec.ts` asserting the script reports both pins, alongside the existing assertions that the same pins stay greppable for the version sync.
- `pnpm nx run @aws/nx-plugin:test` — 3459 tests / 184 files pass. `pnpm nx run-many --target build --all` and `lint --all --fix` both pass.

### Note on a coverage gap found while investigating (not addressed here)

The migrate lane starts from the newest 5 releases (`DEFAULT_START_VERSION_COUNT`), but migration folders in the tree go back to `v1.0.0-rc.50` — 20 releases. At the current ~1-2 releases/day cadence, 5 releases is about 3 days, so migrations older than that are never exercised end-to-end. Left as-is deliberately: widening multiplies the lane's wall-clock, and that's a separate call from fixing this flake.

### Issue # (if applicable)

N/A — found via CI flake on #1075.

### Checklist

----

### DIFF
--- a/.github/actions/init-monorepo/action.yml
+++ b/.github/actions/init-monorepo/action.yml
@@ -29,10 +29,6 @@ inputs:
     description: 'Install the Smithy CLI on Windows, where it is a prerequisite rather than resolved by mise. Off by default — only the jobs that build a Smithy project need it. No effect on other runners.'
     required: false
     default: 'false'
-  smithy-version:
-    description: 'Smithy CLI version to install on Windows. Keep in step with MISE_VERSIONS.smithy in packages/nx-plugin/src/utils/versions.ts.'
-    required: false
-    default: '1.72.1'
 
 runs:
   using: 'composite'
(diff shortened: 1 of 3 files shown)


nx-plugin-for-aws:
Closing — over-engineered for what turned out to be transient upstream flakiness (a GitHub API 500 while mise resolved the Smithy CLI). Not worth a CI pre-warm step, a cache and a pin-extraction script.

PR #1078, opened 2026-08-13
fix(py#agent): rebuild the A2A httpx client per call instead of reusing one built at factory time

AlexTo:
### Reason for this change

Every LangChain and Strands A2A agent that delegates to another agent over A2A hits a reproducible `RuntimeError: Event loop is closed` on the second call through the same client. `AgentCoreA2aClientConfig` builds one `httpx.AsyncClient` a single time, at factory call time, and that client's connection pool binds to whichever event loop first touches it. Every A2A delegate call runs through a fresh `asyncio.run()` loop (`_run_sync`, and — for Strands — `A2AAgent.__call__` does the same internally), so call 1 binds the pool to its loop, that loop closes when the call returns, and call 2's new loop then touches the same pool and blows up.

Screenshot From 2026-08-13 17-08-22

### Description of changes

- `agentcore_a2a_client_langchain.py.template` / `agentcore_a2a_client_strands.py.template` — both clients now open a fresh `httpx.AsyncClient` per call (`async with`, reusing only the `auth`/`timeout` off the once-built shared config) instead of reusing the one built at factory time. For Strands this also means the factory no longer hands out a raw `A2AAgent` — it now returns a private `_A2AClient` that builds a fresh `A2AAgent` per call via `invoke_async`.
- `__targetAgentSnakeCase___client_strands.py.template` — the per-target `<Agent>ClientStrands.create()` wrapper's now-stale `-> A2AAgent` return annotation and unused `A2AAgent` import are dropped, since the factory no longer returns that type.
- New migration `py-agent-a2a-httpx-client-per-call` carries existing workspaces to the same shape: it finds `agentcore_a2a_client_langchain.py`, `agentcore_a2a_client_strands.py`, and any `<target>_client_strands.py` wrapper by content shape, rewrites them via GritQL, and reports (without touching) any file that has diverged from the generated shape.

- Added `migration.spec.ts` (11 tests): applies cleanly to the generated shape for all three files, detects and reports divergence without partial rewrites, is idempotent, and leaves an unrelated strands wrapper (e.g. an MCP client, which shares the `_client_strands.py` suffix) untouched.
- Dry-ran the migration's exact GritQL rewrites against the real pre-fix template content and diffed the result against the fixed templates — byte-identical for both core clients.
- `pnpm nx run @aws/nx-plugin:test` (migrations, py/agent, agent-connection suites): 493 tests pass. Snapshot updated for the Strands core client's new shape.
- `pnpm nx run @aws/nx-plugin:typecheck` and `:lint` pass.
- Manually reproduced the bug for both frameworks in a real generated workspace and confirmed the fix resolves it: [reproduction](https://github.com/AlexTo/py-agent-session-manager/commit/1012a40ca721d7de5c8445fe3aca333e54e6b7df), [migration + fix applied](https://github.com/AlexTo/py-agent-session-manager/commit/66f5a3f239d9c1922e08eb18dadd74820fd08327) — the second commit is `nx migrate` actually running this PR's migration against a real workspace, landing byte-for-byte the same rewrite the unit tests assert.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/py-agent-a2a-httpx-client-per-call/metadata.json
+++ b/packages/nx-plugin/src/migrations/latest/py-agent-a2a-httpx-client-per-call/metadata.json
@@ -1,0 +1,3 @@
+{
+  "description": "Rebuild the httpx.AsyncClient (and Strands A2AAgent) per A2A delegate call instead of reusing one built at factory time, fixing a RuntimeError: Event loop is closed on the second call through a shared client"
+}
--- a/packages/nx-plugin/src/migrations/latest/py-agent-a2a-httpx-client-per-call/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/py-agent-a2a-httpx-client-per-call/migration.spec.ts
@@ -1,0 +1,387 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Tree } from '@nx/devkit';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test';
+import migration from './migration';
+
+const AC_DIR = 'packages/common/agent_connection/proj_agent_connection/core';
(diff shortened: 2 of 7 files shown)


cogwirrel (approved):
Thanks @AlexTo !

PR #1081, opened 2026-08-14
fix(py#fast-api): replace bare multi-exception except with split clauses

nx-plugin-for-aws:
## Summary

- The `py#fast-api` generator emits `except (json.JSONDecodeError, KeyError):` in the `__init__.py` template, but `formatFilesInSubtree` runs ruff with `target-version=py314` (because `py#project` defaults to `requires-python = ">=3.14"`), which converts it to the bare PEP 758 form `except json.JSONDecodeError, KeyError:` — a Python 3.14+ only syntax.
- Fix: split into two separate `except` clauses, which is valid on all Python versions and is not altered by ruff's py314 formatter.
- Snapshot updated to reflect the corrected output.

Fixes: STE-14

## Test plan

- [ ] `pnpm nx run @aws/nx-plugin:test` passes with no snapshot failures
- [ ] Generated `__init__.py` no longer contains bare multi-exception `except` clause
- [ ] Snapshot at `src/py/fast-api/__snapshots__/generator.spec.ts.snap` shows two separate `except` clauses

### DIFF
--- a/packages/nx-plugin/src/py/fast-api/__snapshots__/generator.spec.ts.snap
+++ b/packages/nx-plugin/src/py/fast-api/__snapshots__/generator.spec.ts.snap
@@ -147,7 +147,9 @@ async def add_correlation_id(request: Request, call_next):
             try:
                 lambda_context = json.loads(lambda_context_header)
                 corr_id = lambda_context.get("request_id")
-            except json.JSONDecodeError, KeyError:
+            except json.JSONDecodeError:
+                pass
+            except KeyError:
                 pass
     if not corr_id:
         # If still empty, use uuid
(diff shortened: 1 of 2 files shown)


nx-plugin-for-aws:
Closing per founder decision on STE-14: the org targets Python 3.14 exclusively (generator defaults requires-python >=3.14 and all generated Lambda runtimes are python3.14), so the bare multi-exception except emitted by ruff's py314 formatting is valid everywhere the generated code runs. The fix only mattered for consumers lowering requires-python below 3.14 after generating. Commit 9fae05de remains recoverable from this PR if that case ever surfaces.

PR #1082, opened 2026-08-14
feat(py#agent): add session management support

AlexTo:
### Reason for this change

`py#agent` had no way to persist conversation state across invocations: every restart, redeploy, or scale-in event silently dropped a user's conversation history. `ts#agent` already got this in #1017; this brings the same capability to `py#agent` for both supported frameworks.

### Description of changes

- **Strands framework**: adds a `session` option (`s3` default, or `in-memory`) that resolves a `strands.session.SessionManager` via a generated `session.py`. `s3` provisions a dedicated, KMS-encrypted S3 bucket (CDK/Terraform) with access logs delivered to CloudWatch Logs, and registers the bucket name in AppConfig runtime config alongside the agent's ARN. Wired into the `Agent` constructor directly for HTTP/A2A, and via a `session_manager_provider` on `StrandsAgentConfig` for AG-UI (since its adapter clones one template agent per conversation thread).
- **LangChain framework**: adds the same `session` option, but with `s3` / `dynamodb-s3` / `in-memory` choices (defaulting to `s3`), resolved via a generated `session.py`'s `get_checkpointer()`:
  - `s3` uses a vendored, async-capable `S3CheckpointSaver` (adapted from the `langgraph-checkpoint-s3` package, MIT licensed; vendored rather than installed since its botocore pin conflicts with this project's boto3 pin. Async methods offload the sync boto3 calls via `run_in_executor`, mirroring `langgraph-checkpoint-aws`'s own `DynamoDBSaver`).
  - `dynamodb-s3` provisions a DynamoDB table (unified `PK`/`SK` schema, `PAY_PER_REQUEST`, point-in-time recovery, TTL) plus an S3 bucket for offloading checkpoints over 350KB, per AWS's own guidance for LangGraph-on-DynamoDB.
  - Local dev (`LOCAL_DEV=true`) always uses `AsyncSqliteSaver` (Strands: `FileSessionManager`) under `tmp/agents/<framework>/<name>` at the workspace root, regardless of the configured `session` option.
- **Generator plumbing**: per-framework validity table (`s3`/`in-memory` for Strands; `s3`/`dynamodb-s3`/`in-memory` for LangChain) throws a clear error on an unsupported combination; `log_tool_errors` hook added alongside the existing `log_model_errors` for every protocol.
- **Migration**: `py-agent-session-management-support` retrofits existing workspaces: wires `session_manager`/`get_checkpointer()` into `agent.py` (Strands) or `create_agent(...)` (LangChain), backfills `session.py` (or the LangChain equivalent) defaulting to in-memory (no prior storage to preserve), and retrofits `log_tool_errors`. Reports `nextSteps` for anything that's diverged from the generated shape rather than guessing.
- **Docs**: session management section for both frameworks (English + regenerated translations for all 9 locales), explaining the mapping to Strands' own session-management concept and LangGraph's checkpointer concept respectively, with links to both and to the AWS DynamoDB guide.

- Full generator test suite (unit + snapshot tests across `generator.core/frameworks/protocols/constructs.spec.ts`) covering every framework × session × protocol combination, including the throw-on-unsupported-combination paths.
- Migration test suite covering idempotency and divergence-reporting for Strands HTTP/A2A/AG-UI and LangChain shapes.
- **Live-tested the vendored `S3CheckpointSaver` against a real S3 bucket** (temporary bucket, torn down after): verified `put`/`get_tuple` round-trip (content + metadata), "latest checkpoint" resolution, `put_writes`/pending writes, `list` ordering/`limit`/`filter`, the full async surface (`aput`/`aget_tuple`/`aput_writes`/`alist`/`adelete_thread`), and `delete_thread` actually removing objects. 20/20 checks passed.
- Full `pnpm nx run @aws/nx-plugin:test` suite green.
- **End-to-end validation in a real generated workspace**, linked against this branch's locally-built plugin, per [CONTRIBUTING.md](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md)'s testing guidance ([AlexTo/py-agent-session-manager](https://github.com/AlexTo/py-agent-session-manager)):
  - [`8853887`](https://github.com/AlexTo/py-agent-session-manager/commit/88538879f3816b37bca5970fcccfff620cb0535c): linked the locally-built `@aws/nx-plugin` into a fresh workspace and added the new `py-agent-session-management-support` migration.
  - [`2ca9efc`](https://github.com/AlexTo/py-agent-session-manager/commit/2ca9efccccb52579af669347f0703b8e2b78ed68): ran the migration against existing Strands and LangChain agents (HTTP, AG-UI, A2A), confirming it backfills `session.py`, `session_context.py`, `with_session_id_strands.py` and `tool_errors_strands.py` correctly.
  - [`330fd4b`](https://github.com/AlexTo/py-agent-session-manager/commit/330fd4b097c1260fb3bab9eeb8072a6d57d38442): generated a fresh Strands agent set with `session: s3` end to end (agents, MCP server, website, CDK infra), deployed it to AWS, and confirmed the `S3SessionManager`-backed session works both under `LOCAL_DEV` and against the real deployment.
  - [`78fdf2b`](https://github.com/AlexTo/py-agent-session-manager/commit/78fdf2b5ee37a9590dc14578ffb4016eae6554b9): same exercise for LangChain with `session: s3`, including the vendored `s3_checkpoint_saver_langchain.py` in the generated `agent_connection` package; confirmed working both locally and deployed to AWS.
  - [`1012a40`](https://github.com/AlexTo/py-agent-session-manager/commit/1012a40ca721d7de5c8445fe3aca333e54e6b7df): same exercise for LangChain with `session: dynamodb-s3`, including the generated DynamoDB table + S3 offload bucket CDK stack; confirmed working both locally and deployed to AWS.
  - [`66f5a3f`](https://github.com/AlexTo/py-agent-session-manager/commit/66f5a3f239d9c1922e08eb18dadd74820fd08327): applied the unrelated `py-agent-a2a-httpx-client-per-call` migration on top, confirming migrations continue to compose cleanly against the new session-management generator output.

### Checklist

----


nx-plugin-for-aws:
Hi @AlexTo — I gave this a thorough end-to-end run (both CDK and Terraform, local + deployed to a sandbox account) because session persistence is one of those features where the interesting paths only light up once it's really running. The short version: **it works**, and the design holds up well — the local-dev/deployed split, the per-framework validity table, and the S3/DynamoDB wiring all behaved exactly as documented.

Three things I found that I think are worth a look before merge, plus one optional follow-up. Full validation summary at the bottom for context.

---

### 1. The migration derives a slightly different local session directory than the generator

`migration.ts` reconstructs the agent's name by kebab-casing its class name (`rc`), and that round trip loses hyphens around single-letter segments:

```
kebabCase("OldSHttp") = "old-shttp"    # the agent is actually 'old-s-http'
kebabCase("SHttpS3")  = "shttp-s3"
```

Generator vs. migration for the *same* agent name:

```
GENERATOR  name 'old-s-http' -> storage_dir=".../tmp/agents/strands/old-s-http"
MIGRATION  name 'old-s-http' -> storage_dir=".../tmp/agents/strands/old-shttp"
```

Two effects: a migrated workspace points at a different local-dev directory than a freshly generated one (so existing local history is orphaned), and — the one I'd care more about — two distinct agents can land on the *same* store and share conversation history in local dev:

```
old-s-http -> rc=OldSHttp -> old-shttp   ┐ same directory
old-shttp  -> rc=OldShttp -> old-shttp   ┘
my-x-api   -> rc=MyXApi   -> my-xapi     ┐ same directory
my-xapi    -> rc=MyXapi   -> my-xapi     ┘
```

Affects all three call sites (the langchain one plus the two strands ones). Deployed behaviour is unaffected, since those paths resolve storage from runtime config rather than this string.

Reading the agent's real name from component metadata instead of round-tripping the class name would fix it. Very easy to miss — the existing `migration.spec.ts` fixtures (`agent`, `my-agent`) all round-trip cleanly, so a case with a name like `my-s-agent` would lock it down.

### 2. `terraform destroy` can't complete, and there's no way to opt into deletion

In the Terraform template, `force_destroy = false` on the session bucket and `deletion_protection_enabled = true` on the checkpoint table are both hardcoded with no variable to override them. Destroying a 7-agent stack gave six hard errors and left it half-deleted:

```
Error: deleting S3 Bucket (terraform-198778584bae98212d2c724b7d): BucketNotEmpty
Error: deleting AWS DynamoDB Table (tf-l-ddb-checkpoints-3f1302c2): ValidationException:
       Resource cannot be deleted as it is currently protected against deletion.
Error: deleting Cognito user pool (...): deletion protection is activated
```

It only completed after I manually emptied 5 buckets and disabled protection on 2 tables.

To be clear, I think the **defaults are right** — session data shouldn't evaporate on a `destroy`. The gap is just that callers can't override. There's already a nice precedent in the repo: `app/dynamodb/<table>` sets the *same* attribute as a variable with the *same* safe default —

```hcl
variable "deletion_protection_enabled" {
  description = "Whether to enable deletion protection on the table."
  type        = bool
  default     = true
}
```

— and the repo's own e2e Terraform test relies on that override (`deletion_protection_enabled = false` for `my_table`). Mirroring it here (e.g. `session_force_destroy` / `session_deletion_protection_enabled`, plus matching CDK props threading to `removalPolicy`/`deletionProtection`) would keep production safe while making sandbox and CI teardown a one-liner.

Worth noting the asymmetry between providers: CDK's `RemovalPolicy.RETAIN` orphans the resources silently, whereas Terraform blocks the teardown outright. Same intent, quite different operator experience — and neither cleans up unaided. Also relevant if a Terraform deploy smoke test ever covers `session=s3`/`dynamodb-s3`: it wouldn't be able to tear down, which is exactly why `my_table` needed the override.

### 3. Terraform session buckets get auto-generated names with no agent identifier

`aws_s3_bucket.session` doesn't set a `bucket` argument, so AWS generates the name. Across 5 agents I got:

```
terraform-198778584bae98212d2c724b7d   (tf-s-s3)
terraform-663cb664c77c0326e6d339550a   (tf-l-s3)
terraform-75048e180f083371b8b29a5289   (tf-l-ddb)
terraform-123d952878e6c05d3bcca99eef   (tf-s-cog)
terraform-b4d37af7639918d3e838d29f5f   (tf-l-cog)
```

Nothing says which agent owns which. The DynamoDB table in the same template already does the nice thing:

```hcl
name = "<%= nameKebabCase %>-checkpoints-${random_id.session_unique_suffix.hex}"
```

so this looks like an oversight rather than a decision. CDK isn't affected — CloudFormation derives a descriptive name from the construct path. This compounds #2: when the missing override forces manual cleanup, doing it across five indistinguishable buckets is uncomfortably error-prone (ask me how I know 🙂).

### 4. Optional follow-up: e2e coverage for session persistence

Not a blocker, and happy to pick this up separately. #1080 just landed the pattern in `local-dev.spec.ts` (repeated invocation on a reused `x-amzn-bedrock-agentcore-runtime-session-id`), so extending it for sessions should be cheap. Two traps I'd suggest encoding, both of which caught me out:

- `LOCAL_DEV=true` always uses `FileSessionManager`/`AsyncSqliteSaver` regardless of `session`, so locally every backend behaves identically and even `in-memory` persists to disk. A local test can prove persistence works but **can't** distinguish backends — the `s3`/`dynamodb-s3` branches only execute when deployed.
- A **stable** session id across calls is essential; a fresh id per call builds a new agent each time and hides the behaviour entirely.

---

### Validation summary

Everything else passed, including the parts I most expected to be fragile.

**Build/tests:** full build, 3514 unit tests, typecheck and lint all green.

**Permutation matrix:** 15 core combinations (framework × protocol × session). All 3 unsupported combinations (`strands` + `dynamodb-s3`) throw the clear message from the validity table. The `infra=none` warning fires on exactly the 9 non-`in-memory` permutations and stays quiet for the 6 `in-memory` ones.

**Local (CDK workspace): 15/15 pass.** Turn counts grow 1→2→3 on a reused session, the secret is retained, and a different session id sees nothing. History also survives a full dev-server restart on both frameworks, and MCP + gateway tools keep working with a session store active across all four on/off corners.

**Deployed CDK: 7/7 pass** — verified against the real backends rather than just the replies: 10 S3 objects for the Strands manager, 36 DynamoDB items with `CHUNK_`/`WRITES_` keys for the LangChain saver, and no session resources at all for `in-memory`. Cognito agents correctly reject SigV4 (`403 Authorization method mismatch`) and pass with a real JWT.

**Deployed Terraform: 7/7 pass** — `validate`/`fmt` clean, Checkov 197+198 passed / 0 failed, and the deployed AppConfig content matches CDK's shape exactly (`tableName`+`bucketName` for `dynamodb-s3`, `bucketName` for `s3`, no session block for `in-memory`), so the same `session.py` works under either provider.

**Durability, on both providers:** wrote state on a session, forced container replacement via redeploy, then read back — both S3 and DynamoDB returned the value. That's the core claim of the feature and it holds.

All sandbox resources torn down and confirmed gone. Nice work on this — the vendored `S3CheckpointSaver` and the AG-UI `session_manager_provider` seam in particular were a pleasure to read.

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -486,6 +489,47 @@ Your agent is automatically configured with observability using the [AWS Distro
...
+The `session` option controls how your agent persists conversation state (message history, tool state, etc.) across invocations, using the Strands SDK's [SessionManager](https://strandsagents.com/docs/user-guide/concepts/agents/session-management/):
+
+- **`s3`** (default): The CDK/Terraform infrastructure provisions a dedicated S3 bucket for session data, encrypted with a dedicated KMS key and with all public access blocked; server access logs are delivered to a CloudWatch Logs log group via the same key. The agent's IAM role is granted read/write/list/delete access to the bucket and decrypt/generate-data-key access to the key, and the bucket name is registered alongside the agent's ARN in AppConfig runtime configuration.
+- **`in-memory`**: No bucket is provisioned. Conversation state is kept in memory only for the lifetime of the running process and does not survive restarts or scale-in.
+
+This is implemented in the generated `session.py`, which exports a `get_session_manager()` function resolving a `SessionManager` for the current session.
+
+<OptionFilter when={{ protocol: 'ag-ui' }} description="AG-UI session wiring">
```
cogwirrel: Nit: probably too much detail to have this section (and the one below for http/a2a). Also means we won’t have nested option filters which may get confusing :)

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -486,6 +489,47 @@ Your agent is automatically configured with observability using the [AWS Distro
...
+
+:::note[Local Development]
+When running locally (`LOCAL_DEV=true`, set automatically by the `-dev` target), session data is always stored on disk under `tmp/agents/strands/<agent-name>` at the workspace root, regardless of the configured `session` option, for convenience.
+:::
+</OptionFilter>
+
+<OptionFilter when={{ framework: 'langchain' }} description="LangChain session management (session.py)">
+The `session` option controls how your agent's LangGraph checkpointer persists conversation state:
+
+- **`s3`** (default): The deployed agent uses an `S3CheckpointSaver` with the provisioned session bucket, storing checkpoints and pending writes under the `checkpoints/` prefix. This class (adapted from [`langgraph-checkpoint-s3`](https://github.com/Isa-rentacs/langgraph-checkpoint-s3), MIT licensed) lives in `s3_checkpoint_saver_langchain.py` in the shared agent-connection project.
```
cogwirrel: Nit: no need to link to originating codebase given its MIT :)

cogwirrel (commented):
Looks great Alex, thank you! Couple of minor comments from me and a couple from AI testing :)

docs/src/content/docs/en/guides/py-agent.mdx
AlexTo: yup, removed all details about protocols

docs/src/content/docs/en/guides/py-agent.mdx
AlexTo: done, we have also modified the original class quite a bit

AlexTo:
the dungeon adventure tutorial has been updated as well to remove the manual provision of the S3 session manager

nx-plugin-for-aws:
Thanks for turning these around so quickly @AlexTo — and for picking up the optional e2e one too. I re-ran the relevant checks against `f88da02` and **all four are fixed**. Details below in case it's useful for the record.

### 1. Migration local session directory ✅

Switching to ComponentMetadata's `name` (the target-prefix) instead of `kebabCase(rc)` is exactly right, and the `'agent'`-sentinel fallback is a nicer catch than anything I'd flagged — I hadn't considered that an unnamed agent records a literal `agent` and is indistinguishable from one explicitly named that.

Re-tested by generating agents with the published rc.70 plugin and migrating, then generating the same names with this branch:

```
                MIGRATION                          GENERATOR
old-s-http      .../strands/old-s-http             .../strands/old-s-http      ✅
old-l-http      .../langchain/old-l-http           .../langchain/old-l-http    ✅
(unnamed)       .../strands/agents-agent           .../strands/agents-agent    ✅
```

Byte-identical across all three, including the fallback path. The previous run produced `old-shttp` for the first case, so the collision scenario is gone.

### 2. `terraform destroy` overrides ✅

`session_force_destroy` and `session_deletion_protection_enabled` both present with the safe defaults, and I confirmed they reach the resources rather than just existing as variables — a plan with both set to the teardown-friendly values:

```
module.ddb_agent.session: deletion_protection=False
module.ddb_agent.session: force_destroy=True
module.s3_agent.session:  force_destroy=True
```

The CDK-side `sessionBucketRemovalPolicy` / `sessionTableDeletionProtection` / `sessionTableRemovalPolicy` props give the same escape hatch there, which closes the provider asymmetry nicely.

### 3. Terraform session bucket naming ✅

```hcl
bucket = "t2-s-s3-sessions-${random_id.session_unique_suffix.hex}"
bucket = "t2-l-ddb-sessions-${random_id.session_unique_suffix.hex}"
```

Now matches the DynamoDB table's convention, so multi-agent stacks are legible.

### 4. e2e session-persistence coverage ✅

The restart-based test is the stronger version of what I suggested — I'd only proposed repeated invocation, whereas gating the recall reply on the establish turn still being in the message history means it can *only* pass if the session was genuinely reloaded from disk. A live Python object can't fake that. Good call also adding the `waitForPortFree(ports.pyA2a, ...)` alongside `pyAgent`, since that dev chain starts both.

### Re-verification on `f88da02`

- Build, typecheck, lint: green
- Unit tests: **3515 passed** (187 files)
- Terraform: `validate` clean, Checkov 68 passed / 0 failed, `plan` succeeds against the real AWS provider (plan-only, nothing applied — verified no resources were created, scratch root removed)

I didn't re-run the full deploy matrix, since the changed surface is the migration string, two Terraform attributes plus a bucket name, the CDK props, and test/docs — none of which alters the runtime session paths I exercised on `c257ba2` (15/15 local, 7/7 CDK deployed, 7/7 Terraform deployed, durability across container replacement on both providers). Happy to redo the deployed runs if you'd like belt-and-braces before merge.

One note unrelated to your changes: CI on the new commits is sitting at `action_required`, so it needs a maintainer to approve the workflow run before the checks will execute.

cogwirrel (approved):
Thanks so much @AlexTo ! Awesome to see the tutorial getting even simpler! :)

PR #1090, opened 2026-08-17
feat(ts#mcp-server): upgrade MCP Inspector to v2 and drop its license exceptions

nx-plugin-for-aws:
### Reason for this change

Two related things prompted this:

1. **The MCP Inspector was four majors behind.** The workspace pinned `@modelcontextprotocol/inspector` at `0.22.0`; the current major is `2.2.0`.
2. **Investigating whether the MCP "v2" protocol (spec revision `2026-07-28`) could be adopted.** The conclusion is that it cannot yet — see below — but the Inspector upgrade is independently safe and unblocks the license-exception cleanup.

### Description of changes

**MCP Inspector 0.22.0 → 2.2.0.** The Inspector is standalone dev tooling invoked by the generated `<server>-inspect` / `<server>-inspect-stdio` targets, not a library the generated code links against, so it upgrades independently of the MCP SDKs. Both command shapes the generators emit are unchanged in v2.

**Removed the MCP Inspector license exceptions entirely.** Inspector v1 declared `SEE LICENSE IN LICENSE`, which resolves to no SPDX identifier, so the license check reported `UNKNOWN` and needed an exception per published package. v2 ships as a single package declaring `MIT`, which `DEFAULT_LICENSE_ALLOWLIST` already covers.

Worth calling out: keeping the exceptions would have been actively wrong, not merely redundant — `spdx: 'Apache-2.0'` **overrides** the now-correctly-declared `MIT`, misreporting the license in the dependency report. The exceptions' stated reason ("Tarball omits the LICENSE file") was never the actual mechanism either: `runCheck` reads declared package metadata and never looks at the LICENSE file.

Knock-on cleanups this enabled:
- Dropped two circular imports (`license/generator.ts` ↔ `{ts,py}/mcp-server/generator.ts`), which only existed to read the MCP generator IDs.
- Re-enabled Inspector coverage in the e2e `license-check` smoke test, which previously skipped every `@modelcontextprotocol/inspector*` package.

**Added `removeLicenseExceptions` + a migration** so existing workspaces have the four stale entries cleared rather than silently carrying a wrong `Apache-2.0` override.

**Registered the inspector as a rejected pnpm build.** v2 adds a `postinstall` script that v1 did not have, and under pnpm 11's default `strictDepBuilds=true` an unreviewed install script fails the whole install with `ERR_PNPM_IGNORED_BUILDS`. The script cascades installs into the inspector's own client packages for a from-source checkout and exits early when the package lives under `node_modules`, so it is a no-op for a consuming workspace — it is registered `false` (reviewed, deliberately not run) rather than `true`, via the existing `registerPnpmBuiltDependencies` helper, matching the `@scarf/scarf` precedent.

Both generators do this for new MCP servers, and the migration does it too: an existing workspace receives the v2 bump from the version sync without passing through a generator, so it needs the entry as well (guarded on the workspace actually depending on the inspector).

#### The MCP SDKs are deliberately left on v1

`mcp==2.0.0` (Python) and `@modelcontextprotocol/{core,server,client}@2.0.0` (TS) implement spec revision `2026-07-28`. They are **not** adopted here because the agent frameworks are not ready:

| Dependency | Constraint | Under forced `mcp` 2.0 |
|---|---|---|
| `strands-agents` 1.52.0 | `mcp>=1.23.0,<2.0.0` | `ImportError: cannot import name 'streamablehttp_client'` |
| `langchain-mcp-adapters` 0.3.2 | `mcp>=1.24.0,<2.0.0` | `ImportError: cannot import name 'RequestContext'` |

`uv add` fails outright with an unsatisfiable resolution on both. Upstream work is in flight but unmerged: strands-agents/harness-sdk#1659 (tracker) with its compat PR still open and explicitly keeping the `<2` pin, and langchain-ai/langchain-mcp-adapters#578 unanswered.

Separately, `mcp.server.fastmcp` — which the `py#mcp-server` template imports — no longer exists in `mcp` 2.0, so that adoption will need a template change, not just a version bump.

**AgentCore does support the new revision** (Gateway serves `2025-03-26`, `2025-06-18`, `2025-11-25` and `2026-07-28`, selected per request via `supportedVersions`), so it is not the blocker. AWS's own readiness guidance is to confirm client SDK support first, which is exactly what is missing. Note also that `aws-cdk-lib` 2.265.0's `MCPProtocolVersion` only exposes up to `MCP_2025_06_18`. The existing `2025-03-26` pins in the gateway constructs are left unchanged.

**Unit tests:** 3539 passing across 188 files (`pnpm nx run @aws/nx-plugin:test -u`), no snapshot churn from the version bump. Added coverage for `removeLicenseExceptions` (removal, no-match no-op, missing-config no-op) and 4 migration tests including the empty-array and no-config edge cases.

**Build and lint:** `pnpm nx run-many --target build --all` passes. Lint warnings went 1036 → 1032 (the 4 removed exceptions); no new warnings.

**End-to-end in a real workspace**, using a locally compiled + `pnpm link`ed plugin, for both `ts#mcp-server` and `py#mcp-server`:

- Verified the Inspector v1 → v2 root cause empirically: v1 reports `Custom: LICENSE`, v2 reports `MIT`.
- `license-check` passes with **442 dependencies inspected and zero exceptions**, including the new transitive `@modelcontextprotocol/{core,client,server,server-legacy}` packages.
- Drove both generated servers with the vended Inspector v2 CLI: `tools/list`, `tools/call` (TS `divide` → `2.5`; Python `add` → `12` with `structuredContent`), and `resources/list` all correct against the v1-SDK servers.
- Ran the generated `-inspect` and `-inspect-stdio` targets; both start via `dependsOn` and the inspector API confirms it is wired to the right server (`streamable-http` at the dev port, and `stdio` with the right command/cwd).
- Applied the **compiled** migration to a workspace seeded with the four old exceptions: all removed, no array hole, config still parses, and `license-check` still passes afterwards.
- Reproduced the pnpm 11 `ERR_PNPM_IGNORED_BUILDS` install failure in a clean workspace on pnpm 11.20.0, confirmed the `allowBuilds` entry lands as `false`, and confirmed `pnpm install` then completes and the inspector still drives a server correctly with its postinstall skipped.
- Separately reproduced the **upgrade** path on pnpm 11: seeded a workspace pinning inspector 0.22.0 with the old exceptions and no `allowBuilds`, ran the compiled migration, then bumped to 2.2.0 and installed — succeeds. Removing the entry again reproduces the failure, confirming the migration is what fixes it.
- Confirmed the docs' `localhost:6274` and "Connect" button guidance still hold in the v2 UI, so no doc changes were needed.

### Issue # (if applicable)

N/A

### Checklist

----


cogwirrel (dismissed):

cogwirrel (approved):

nx-plugin-for-aws:
### CI status: green except a pre-existing `trivy` failure

**57 / 58 jobs pass.** The one failure is `Smoke Tests - trivy`, and it is not caused by this PR.

**What it is:** 9 HIGH CVEs in the Debian `util-linux` family (`bsdutils`, `mount`, `libblkid1`, `libmount1`, `libsmartcols1`, `libuuid1`) in the vended Python base image — headline `CVE-2026-53615`. Trivy reports `Total: 9 (HIGH: 9, CRITICAL: 0)` for `e2e_test.py_project` and `e2e_test.py_langchain_project`.

**Why it isn't from this change:**

- `main`'s own recent CI runs are failing, and `main`'s run for `b5839c3` fails on the same job with the **identical** `CVE-2026-53615` and identical `Total: 9 (HIGH: 9, CRITICAL: 0)`.
- This PR touches no Dockerfile, no `BASE_IMAGES` entry, and no trivy pin (`git diff origin/main` is empty for all three).
- The failing job's log contains no `ERR_PNPM_IGNORED_BUILDS` and no PyPI errors — it is purely the image scan.

**Why it can't be fixed here:** Debian has published a fix (`2.41.5-0+deb13u1`), but the upstream image has not picked it up yet — pulling `public.ecr.aws/docker/library/python:3.14-slim` right now still reports `bsdutils 1:2.41-5`. It needs a rebuilt upstream image (or a separate base-image bump on `main`), so it is out of scope for this PR rather than something to mask with a `.trivyignore`.

**On the other failures:** the earlier red legs were transient PyPI 502s (`Failed to download ...`, "Request failed after 8 retries" — the flakiness #1089 targets), which cleared on re-run. Notably the 5 `migrate` shards that failed on the first push were a **real** bug in this PR — an upgrading workspace never passed through a generator, so it never got the pnpm `allowBuilds` entry for the inspector's new v2 postinstall. That is fixed in 91676e8 and `ERR_PNPM_IGNORED_BUILDS` no longer appears in any job log.

PR #1093, opened 2026-08-18
chore(ci): share one uv cache across all jobs, runs and workflows

nx-plugin-for-aws:
### Reason for this change

pypi/warehouse#1089 got the uv cache working, but keyed it **per job**. That was a workaround for `setup-uv` restoring by exact key only, not a claim that the contents differ — and it left real waste, as raised in review:

> can't the cache be across runs though? ie a package is the same regardless of which workflow downloaded it

Correct, and measurably so. On `main` there were **46 uv cache entries totalling 4.6GB** (mean 103MB) against a 10GB repo limit, most of it the same wheels — `boto3`, `botocore`, `pydantic` — stored dozens of times. Worse than the storage: no job benefited from what its siblings had already fetched, so every leg paid its own PyPI round trip for the common set, and every one of those round trips was a chance to hit the 502s.

### Description of changes

One cache shared by every job, accumulating across runs, workflows and PRs.

`setup-uv` cannot express this. It restores by exact key only, so a shared key degenerates: the first job to finish saves, every later job gets an exact hit and **skips saving** (it skips when `matchedKey === cacheKey`), and the entry freezes at whatever that one job happened to download. So this manages the cache directly and keeps `setup-uv` only for installing uv and exporting `UV_CACHE_DIR`:

- `enable-cache: 'false'` on `setup-uv`, retaining `cache-local-path` — that step still calls `setCacheDir` unconditionally, which is what makes this work without a version bump.
- `actions/cache/restore` with a **run-unique primary key** and a `restore-keys` prefix ladder. Each run saves a fresh entry; the prefix match restores the most recent one. That is what lets the cache accumulate instead of fragmenting or freezing.
- A small `save-uv-cache` action as each job's last step, since `enable-cache: false` also removes `setup-uv`'s post-job save (and a composite action cannot declare post steps).

Details worth noting:

- **Cross-PR/CI sharing comes for free.** Branches inherit the default branch's caches, so PR jobs restore what CI on `main` saved. This already demonstrably happens — a PR job on `refs/pull/1092/merge` restored an entry created on `refs/heads/main`.
- **The prefix ladder degrades gracefully:** `uv-<os>-<versions.ts hash>-` then `uv-<os>-`. A dependency bump warms from the previous manifest instead of starting cold, which previously reset every entry for a full run.
- **Saves under `always()`**, so a job that fails part way still banks what it downloaded. This is the one useful idea from the closed pypi/warehouse#1092, reduced to a two-line condition rather than a second key space and an env-var protocol.
- **Every uv job now contributes**, including `package`/`build`/`unit_tests`. `release` deliberately has no save — it publishes and installs no Python deps.
- Drops the now-dead `prune-uv-cache` input, since `setup-uv` no longer manages the cache. The prune hazard that originally disabled caching (8952c20c) is gone with it.

No unit tests — CI configuration with no code surface.

- Confirmed uv's cache is content-addressed internally (`archive-v0/` holds hash-named entries) and that syncing two disjoint projects into one cache dir **merges** rather than conflicts — so one shared entry is sound.
- **Corrected a wrong assumption of mine before building this.** I initially proposed upgrading `setup-uv` on the belief that newer versions support `restore-keys`. They do not: I checked v10.0.1's `restore-cache.ts` and it still calls `cache.restoreCache([path], cacheKey)` with no restore-keys argument. Upgrading would not have helped, hence managing the cache directly.
- Verified `cache-primary-key` is a real output of `actions/cache/restore@v4` at the pinned SHA, so the save key cannot silently resolve empty.
- Verified `setup-uv` exports `UV_CACHE_DIR` even with `enable-cache: false`.
- Programmatically checked every job in both workflows: each uv-using job has a save step, it is the job's last step, and it runs under `always()`.
- Validated all four YAML files parse.

Cache behaviour across successive runs is observable in this PR's runs — see the follow-up comment.

### Issue # (if applicable)

Follows up pypi/warehouse#1089 (review feedback). Supersedes the closed pypi/warehouse#1092 with a much smaller change.

### Checklist

----


nx-plugin-for-aws:
### Observed across multiple runs

Iterated on this across several runs on this branch. Two keying mistakes surfaced and were fixed; the final shape is verified working.

**Storage — the problem this set out to fix**

| | entries | total | mean |
| --- | --- | --- | --- |
| Per-job keys (`main` today) | ~40 | 3.93 GB | 101 MB |
| Shared cache (this branch) | 8 per run | 724 MB | 91 MB |

Currently 14 entries / 1.20 GB across two attempts, against 3.93 GB for a single run before. Roughly **5x less storage** on a repo that was sitting at 9.96 GB against GitHub's 10 GB limit.

**Sharing works — jobs restore each other's caches**

The point of `restore-keys`. Jobs log restores from entries *other jobs* created, which exact-key-only matching could never do:

```
Cache restored from key: uv-Windows-<hash>-32138771951-1     # created by a different run
Cache Size: ~634 MB                                          # saved back, compounded
```

**It compounds across runs.** A Windows leg restored a 50 MB entry and saved 634 MB. Entries grow run over run rather than fragmenting, because every job restores before it runs, so each save is a superset of what it restored.

**Cross-PR / cross-CI sharing is real**, not theoretical: a job on `refs/pull/1092/merge` restored an entry created on `refs/heads/main`. Branches inherit the default branch's caches, so PR runs start warm from CI on `main`.

### Two bugs found and fixed while iterating

1. **Empty `with:` blocks.** Removing the now-unused `uv-cache-suffix` left `with:` holding nothing on the Windows smoke jobs. That is valid YAML but Actions rejects it, and **both workflows failed at startup before any job ran**. Fixed in 6335c778. I added a structural lint check for this locally, since plain YAML parsing does not catch it.
2. **Keying per run was wrong.** With one key per run, the first job to finish wins the save — and that is typically a Python-light job. `package` saved a **22 MB** cache and every heavier leg's 395 MB save lost the race, leaving the run's entry nearly empty. Fixed in 311a39f1 by keying on `github.job`: matrix siblings share a job id so each matrix still collapses to one entry (~30 legs → 1), but heavy legs no longer compete with light ones.

That second one is worth noting as a design constraint: a shared cache needs the key to be coarse enough to dedupe but fine enough that dissimilar workloads do not clobber each other. Per job id is that line.

### Remaining failures

All PyPI 502s (pypi/support#11908), unrelated to this change — verified per job. Failures went 10 → 4 across the two attempts as the cache warmed, and every failing job logged a successful cache restore first. One `migrate` shard failure is the pre-existing `no hop actually migrated` assertion, also unrelated and present on runs predating this branch.

nx-plugin-for-aws:
Tracked in #1096 alongside the upstream PyPI incident (pypi/support#11908) that motivated this change.

PR #1105, opened 2026-08-21
feat(ts#agent): scope session storage per user for cognito agents

nx-plugin-for-aws:
### Reason for this change

Agents generated with `auth=cognito` are reached by end users who each sign in as themselves, but their conversation storage was keyed only on identifiers the client supplies — the AgentCore Runtime session id, or the AG-UI thread id. Offering more than one user a private conversation therefore meant every project adding its own namespacing first.

These agents already receive the caller's validated JWT, so the generator can derive a per-user namespace and give projects sensible session storage by default.

### Description of changes

**Per-user session storage (`auth=cognito`)**

The shared session-id middleware reads the `sub` claim from the caller's JWT — validated by the AgentCore Runtime authorizer and allowlisted through `requestHeaderConfiguration` — and binds it for the request. Storage keys become `u-<sub>-<conversationId>`.

Each segment is escaped into the character set Strands accepts for session ids (`^[a-z0-9_-]+$`) using fixed-width UTF-8 byte escapes, so the encoding is injective, and the `u-` / `s-` prefixes keep namespaced and unnamespaced keys disjoint.

Agents using `auth=iam` are unchanged: SigV4 identifies the calling principal to AgentCore Runtime rather than to the agent code, so those keys stay unnamespaced and `bedrock-agentcore:InvokeAgentRuntime` remains the boundary.

**AG-UI thread id**

The AG-UI adapters cache one agent — and its session manager — per thread id, and call `sessionManagerProvider` once per thread, so the namespace is applied to the thread id itself. The TypeScript middleware folds the protocol's `thread_id` alias into `threadId` first, since the adapter maps it first-key-wins when parsing the body.

**Local development**

`nx dev` serves agents without an authorizer, so unauthenticated requests fall back to a single local user, gated on `LOCAL_DEV` (set only by the `-dev` targets). Health checks and CORS preflights bypass the identity check, so `/ping` stays healthy and browser preflights are not blocked.

**MCP servers**

`createServer` now receives an optional `McpRequestContext` carrying the calling agent's session id, so tools can scope stored data per caller rather than on a model-supplied argument. The parameter is optional, so STDIO servers and existing projects are unaffected.

**Documented alternative**

The `ts#agent` and `py#agent` guides also describe a stronger option for projects that need it: issue session IDs from an API, record the owning user against each, and have the agent validate that the caller owns the session it was given. Namespacing partitions conversations, whereas an ownership registry also decides which sessions a user may resume, and avoids two callers sharing a microVM's filesystem by sending the same ID. This is the approach taken in #1106, which is documentation only.

**Dungeon adventure tutorial**

Updated to match the new defaults: games are partitioned on the signed-in user's `sub`, inventory is keyed on the calling session, and the Game API resolves the caller from its IAM authorizer context via a shared `identity.ts`. Two players can now also choose the same hero name.

- New behavioural tests for the vended TypeScript and Python session-context helpers — claim extraction, key namespacing, encoding injectivity, path traversal and empty-user handling — plus generator tests asserting the wiring per protocol and auth mode.
- Cross-checked that the TypeScript and Python key encoders produce identical output across 1,952 shared inputs, and fuzzed 400,000 random `(user, session)` pairs with no collisions.
- Deployed the dungeon adventure tutorial to AWS with two Cognito users and confirmed each sees only their own games and conversations, that an unauthenticated caller is refused, and that persisted DynamoDB rows carry the owner in the partition key. Resources torn down afterwards.
- Ran the generated TypeScript and Python agents locally to confirm unauthenticated requests succeed under `nx dev` and are rejected without `LOCAL_DEV`, and that `/ping` and CORS preflights are unaffected.
- Full plugin test suite, lint and docs build pass.

### Issue # (if applicable)

N/A

### Checklist

----


nx-plugin-for-aws:
Closing in favour of #1106, which documents the session ownership approach rather than changing the generated defaults.

PR #1106, opened 2026-08-21
docs(ts#agent): describe restricting agent sessions to their owner

nx-plugin-for-aws:
### Reason for this change

The session ID an agent receives names a conversation, but on its own it does not say who that conversation belongs to. AgentCore Runtime authorizes an invocation against the agent runtime resource ARN rather than against an individual session, which leaves it to the agent to decide what a session means to the application.

Projects that want each signed-in user to reach only their own conversations currently have to work that pattern out for themselves. This documents it.

This is an alternative to #1105, which instead changes the generated defaults. This PR is documentation only — no generated code changes, so existing projects are unaffected and readers can adopt the pattern at whatever granularity suits them. #1105 also documents this pattern alongside its own change, so the two are not mutually exclusive.

### Description of changes

Adds a "Restricting sessions to their owner" section to the <Link path="guides/ts-agent">`ts#agent`</Link> and <Link path="guides/py-agent">`py#agent`</Link> guides, describing a session ownership registry:

1. **Create sessions through your API.** A procedure mints an opaque session ID, resolves the caller from its own verified request context, and records the pair — for example in a DynamoDB table partitioned on the session ID, with a secondary index on the user for listing their own conversations. The caller is never taken from an input field.

2. **Validate ownership in the agent.** The agent looks up the owner for the session ID it is given, compares it against the caller's `sub` — available with `auth=cognito`, since the AgentCore Runtime authorizer validates the JWT and the generated infrastructure allowlists the `Authorization` header — and refuses the request when they do not match.

Both guides include example code in their respective idioms, and explain why validating in both places is worthwhile: the API decides ownership, the agent enforces it, so a session ID that leaks is still refused on the next invocation and any future client inherits the same rule.

The section also notes that a session ID selects the microVM serving the request, so its filesystem, `/tmp` and in-process caches are shared by every request resolving to that session. Generating opaque IDs in the API avoids two callers colliding there; deriving them from user-supplied values such as a conversation name does not.

Separately, documents how the `connection` generator derives the AgentCore session ID from the AG-UI thread ID — including the 33-character padding — since that page previously did not mention either identifier, and points to the new guidance from it.

Documentation only. Verified the docs site builds and all internal links resolve. Reviewed the described pattern against the generated code paths: the session ID header and JWT claim are both available to the agent as described for `auth=cognito`, and the caller-resolution approach for the API matches the identity middleware already documented in the <Link path="guides/trpc">tRPC guide</Link>.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/connection/react-agui.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-agui.mdx
@@ -92,6 +92,25 @@ The generated code handles authentication depending on your agent's configuratio
 - **IAM** (default): uses AWS SigV4-signed HTTP requests. Credentials are obtained from the Cognito Identity Pool configured with your website's auth.
 - **Cognito**: embeds the JWT access token in the `Authorization` header as a Bearer token.
 
+### Sessions and Threads
+
+AG-UI and AgentCore Runtime each identify a conversation differently, and the generated hook ties them together:
+
+- **`threadId`** — the AG-UI conversation identifier, sent in the request body. CopilotKit generates a random UUID per chat unless you pass an explicit `threadId`.
+- **Session ID** — the AgentCore Runtime session, sent in the `X-Amzn-Bedrock-AgentCore-Runtime-Session-Id` header. It selects the [microVM](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-how-it-works.html) serving the request, and is what your agent's `session.ts` / `session.py` keys conversation state on.
+
+The hook derives the session ID from the thread ID, right-padding it to the 33 characters AgentCore Runtime requires:
+
+```ts
+function agentCoreSessionId(input: RunAgentInput): string {
+  return (input.threadId ?? '').padEnd(33, '0');
+}
+```
+
+Leaving `threadId` unset is simplest — CopilotKit's generated UUID is already 36 characters. If you pass one explicitly, make it at least 33 characters, since padding maps thread IDs that differ only in trailing characters onto the same session.
+
+Both Session ID and Thread ID are provided by the browser. To restrict each user to their own conversations, refer to the <Link path="guides/py-agent">`py#agent`</Link> or <Link path="guides/ts-agent">`ts#agent`</Link> guide.
+
 ## Infrastructure
 
 <Snippet name="connection/react-agent-infrastructure" parentHeading="Infrastructure" />
--- a/docs/src/content/docs/en/guides/py-agent.mdx
+++ b/docs/src/content/docs/en/guides/py-agent.mdx
@@ -4,7 +4,7 @@ description: Generate a Python Agent for building AI agents with tools and deplo
 generator: py#agent
 ---
 
-import { FileTree, Tabs, TabItem, CardGrid } from '@astrojs/starlight/components';
+import { FileTree, Tabs, TabItem, CardGrid, Steps } from '@astrojs/starlight/components';
 import Astro from '@astrojs/react';
 import ConnectionCard from '@components/connection-card.astro';
 import RunGenerator from '@components/run-generator.astro';
@@ -516,6 +516,23 @@ The session ID itself comes from the AgentCore Runtime session (propagated via t
 :::note[Local Development]
 When running locally (`LOCAL_DEV=true`, set automatically by the `-dev` target), session data is always stored on disk under `tmp/agents/strands/<agent-name>` at the workspace root, regardless of the configured `session` option, for convenience.
 :::
+
+#### Restricting sessions to their owner
+
+The session ID arrives from the caller, so on its own it identifies a conversation but not who the conversation belongs to. AgentCore Runtime authorizes an invocation against the agent runtime resource ARN rather than against an individual session, which leaves the agent free to decide what a session means to your application.
+
+:::caution[The session also selects the microVM]
+A session ID selects the [microVM](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-how-it-works.html) that serves the request, so its filesystem, `/tmp` and in-process caches are shared by every request resolving to that session. Two callers who send the same session ID share those resources even when their stored conversations are separate.
+:::
+
+To restrict each user to their own conversations:
+
+<Steps>
+  1. Add an API to create a session, using <Link path="guides/trpc">tRPC</Link>, <Link path="guides/fastapi">FastAPI</Link> or <Link path="guides/ts-smithy-api">Smithy</Link>. Generate an opaque session ID (at least 33 characters) and store it alongside the calling user's ID — for example in a table created with the <Link path="guides/py-dynamodb">`py#dynamodb`</Link> generator. Each API guide shows how to retrieve the calling user's ID.
+  2. In your agent, look up the stored user ID for the session ID it was given, and reject the request when it does not match the caller. With `auth=cognito` the caller's JWT reaches your agent code, so its `sub` claim identifies them.
+</Steps>
+
+Generate the session ID rather than deriving it from user-supplied values such as a conversation name — anything a caller can predict, a caller can send.
 </OptionFilter>
 
 <OptionFilter when={{ framework: 'langchain' }} description="LangChain session management (session.py)">
--- a/docs/src/content/docs/en/guides/ts-agent.mdx
+++ b/docs/src/content/docs/en/guides/ts-agent.mdx
@@ -4,7 +4,7 @@ description: Generate a TypeScript Agent for building AI agents with tools and d
 generator: ts#agent
 ---
 
-import { FileTree, Tabs, TabItem, CardGrid } from '@astrojs/starlight/components';
+import { FileTree, Tabs, TabItem, CardGrid, Steps } from '@astrojs/starlight/components';
 import Astro from '@astrojs/react';
 import ConnectionCard from '@components/connection-card.astro';
 import RunGenerator from '@components/run-generator.astro';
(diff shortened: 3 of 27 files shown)


docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -513,6 +513,78 @@ This is implemented in the generated `session.py`, which exports a `get_session_
...
+      .go();
+    return { sessionId };
+  });
+```
+
+The API can be in any language; the example above uses the <Link path="guides/trpc">`ts#trpc-api`</Link> generator.
+
+Resolve the caller from the verified request context rather than from an input field, so it cannot be supplied by the browser. <Link path="guides/trpc">Implementing Custom Middleware</Link> shows how to do this for both IAM and Cognito authenticated APIs.
+
+A second procedure querying that index lets the UI offer a user their own conversations without ever seeing anyone else's.
```
cogwirrel: Too verbose, reduce. Remove "A session ID is then something...." to this end line and replace with simple instructions and links to other docs. No need for sample code or tables here. Link to each api generator and mention the guide has examples for how to retrieve the calling user id, and that the user id and session id should be stored together.

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -513,6 +513,78 @@ This is implemented in the generated `session.py`, which exports a `get_session_
...
+
+        with session_id_context(session_id):
+            return await call_next(request)
+```
+
+Decoding the claim without verifying its signature is sufficient here, since AgentCore Runtime validated the token during inbound authorization.
+
+:::tip[Why validate in the agent as well as the API]
+The API decides who owns a session; the agent enforces it. Checking in both places means a session ID that leaks — copied from a browser, or reused from an old conversation — is still refused on the next invocation, and any future client of the agent inherits the same rule.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -513,6 +513,78 @@ This is implemented in the generated `session.py`, which exports a `get_session_
...
+
+:::caution[The session also selects the microVM]
+A session ID selects the [microVM](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-how-it-works.html) that serves the request, so its filesystem, `/tmp` and in-process caches are shared by every request resolving to that session. This is worth attention if your agent or its tools write to local disk: two callers who send the same session ID share those resources even when their stored conversations are separate.
+
+Generating opaque session IDs in your API avoids this, because a caller cannot land on a session it was not given. Deriving one from user-supplied values — a conversation name, an email, a game name — is predictable and lets two callers collide.
+:::
+
 :::note[Local Development]
 When running locally (`LOCAL_DEV=true`, set automatically by the `-dev` target), session data is always stored on disk under `tmp/agents/strands/<agent-name>` at the workspace root, regardless of the configured `session` option, for convenience.
 :::
```
cogwirrel: move this just before the restricting sessions to their owner section.

docs/src/content/docs/en/guides/ts-agent.mdx
```diff
@@ -362,6 +362,76 @@ This is implemented in the generated `session.ts`, which exports a `getSessionMa
...
+    await sessionEntity
+      .put({ sessionId, userId: callerSub(ctx.event) })
+      .go();
+    return { sessionId };
+  });
+```
+
+Resolve the caller from the verified request context rather than from an input field, so it cannot be supplied by the browser. <Link path="guides/trpc">Implementing Custom Middleware</Link> shows how to do this for both IAM and Cognito authenticated APIs.
+
+A second procedure querying that index lets the UI offer a user their own conversations without ever seeing anyone else's.
```
cogwirrel: same comment applies as I wrote for the equivalent part of the py#agent guide

docs/src/content/docs/en/guides/ts-agent.mdx
```diff
@@ -362,6 +362,76 @@ This is implemented in the generated `session.ts`, which exports a `getSessionMa
...
+
+  runWithSessionId(sessionId, () => next());
+};
+```
+
+Decoding the claim without verifying its signature is sufficient here, since AgentCore Runtime validated the token during inbound authorization.
+
+:::tip[Why validate in the agent as well as the API]
+The API decides who owns a session; the agent enforces it. Checking in both places means a session ID that leaks — copied from a browser, or reused from an old conversation — is still refused on the next invocation, and any future client of the agent inherits the same rule.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/guides/ts-agent.mdx
```diff
@@ -362,6 +362,76 @@ This is implemented in the generated `session.ts`, which exports a `getSessionMa
...
+
+:::caution[The session also selects the microVM]
+A session ID selects the [microVM](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-how-it-works.html) that serves the request, so its filesystem, `/tmp` and in-process caches are shared by every request resolving to that session. This is worth attention if your agent or its tools write to local disk: two callers who send the same session ID share those resources even when their stored conversations are separate.
+
+Generating opaque session IDs in your API avoids this, because a caller cannot land on a session it was not given. Deriving one from user-supplied values — a conversation name, an email, a game name — is predictable and lets two callers collide.
+:::
+
 :::note[Local Development]
 When running locally (`LOCAL_DEV=true`, set automatically by the `-dev` target), session data is always stored on disk under `tmp/agents/strands/<agent-name>` at the workspace root, regardless of the configured `session` option, for convenience.
 :::
```
cogwirrel: move up as per py#agent guide

docs/src/content/docs/en/guides/ts-agent.mdx
```diff
@@ -362,6 +362,76 @@ This is implemented in the generated `session.ts`, which exports a `getSessionMa
...
+
+:::tip[Why validate in the agent as well as the API]
+The API decides who owns a session; the agent enforces it. Checking in both places means a session ID that leaks — copied from a browser, or reused from an old conversation — is still refused on the next invocation, and any future client of the agent inherits the same rule.
+:::
+
+:::caution[The session also selects the microVM]
+A session ID selects the [microVM](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-how-it-works.html) that serves the request, so its filesystem, `/tmp` and in-process caches are shared by every request resolving to that session. This is worth attention if your agent or its tools write to local disk: two callers who send the same session ID share those resources even when their stored conversations are separate.
+
+Generating opaque session IDs in your API avoids this, because a caller cannot land on a session it was not given. Deriving one from user-supplied values — a conversation name, an email, a game name — is predictable and lets two callers collide.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -513,6 +513,78 @@ This is implemented in the generated `session.py`, which exports a `get_session_
...
+
+:::tip[Why validate in the agent as well as the API]
+The API decides who owns a session; the agent enforces it. Checking in both places means a session ID that leaks — copied from a browser, or reused from an old conversation — is still refused on the next invocation, and any future client of the agent inherits the same rule.
+:::
+
+:::caution[The session also selects the microVM]
+A session ID selects the [microVM](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-how-it-works.html) that serves the request, so its filesystem, `/tmp` and in-process caches are shared by every request resolving to that session. This is worth attention if your agent or its tools write to local disk: two callers who send the same session ID share those resources even when their stored conversations are separate.
+
+Generating opaque session IDs in your API avoids this, because a caller cannot land on a session it was not given. Deriving one from user-supplied values — a conversation name, an email, a game name — is predictable and lets two callers collide.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/guides/connection/react-agui.mdx
```diff
@@ -92,6 +92,25 @@ The generated code handles authentication depending on your agent's configuratio
...
+
+```ts
+function agentCoreSessionId(input: RunAgentInput): string {
+  return (input.threadId ?? '').padEnd(33, '0');
+}
+```
+
+Leaving `threadId` unset is simplest — CopilotKit's generated UUID is already 36 characters. If you pass one explicitly, make it at least 33 characters, since padding maps thread IDs that differ only in trailing characters onto the same session.
+
+Both values come from the browser, so they name a conversation without saying whose it is. To restrict each user to their own conversations, see <Link path="guides/ts-agent">Restricting sessions to their owner</Link>.
```
cogwirrel: Change to: `Both Session ID and Thread ID are provided by the browser. To restrict each user to their own conversations, refer to the <link...>py#agent</link> or <link>ts#agent</link> guide.`

docs/src/content/docs/en/guides/py-agent.mdx
nx-plugin-for-aws: Reduced to two numbered steps. Dropped the sample code and the table, and now link tRPC / FastAPI / Smithy for creating the session API, noting each guide shows how to retrieve the calling user's ID and that the user ID and session ID should be stored together.

docs/src/content/docs/en/guides/py-agent.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/py-agent.mdx
nx-plugin-for-aws: Moved — the microVM caution now sits directly under the section heading, before the numbered steps.

docs/src/content/docs/en/guides/ts-agent.mdx
nx-plugin-for-aws: Applied the same rewrite here.

docs/src/content/docs/en/guides/ts-agent.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/ts-agent.mdx
nx-plugin-for-aws: Moved up to match.

docs/src/content/docs/en/guides/ts-agent.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/py-agent.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/connection/react-agui.mdx
nx-plugin-for-aws: Updated to your suggested wording, linking both the py#agent and ts#agent guides.

cogwirrel (approved):

PR #1107, opened 2026-08-22
feat(ts#website): make WAF, encryption and KMS key rotation configurable on the static website construct/module

AlexTo:
### Reason for this change

The `StaticWebsite` CDK construct and its Terraform equivalent always attached a WAF Web ACL and always created a new customer-managed KMS key, with no way to opt out or bring your own. Every consumer paid for and managed a KMS key and Web ACL even when their use case didn't need one, and there was no way to point at an existing key or a different encryption mode.

This configuration was also baked into the shared base class/module instead of exposed on the per-website construct. A workspace with more than one website had no way to configure them differently; every website was forced into the same WAF and encryption behavior.

### Description of changes

- CDK `StaticWebsite` gains `enableWaf`, `encryption`, `encryptionKey` and `enableKeyRotation` props. The WAF Web ACL is now optional. The KMS key is only created when `encryption` is `BucketEncryption.KMS` and no `encryptionKey` is supplied. `enableKeyRotation` (default `true`) only applies to that auto-created key.
- The generated per-website CDK app construct (`MyWebsite`) now takes an optional `props` parameter that forwards through to `StaticWebsite`, so these props (and `domainNames`/`certificate`) can be set per website where the construct is instantiated in the application stack, instead of editing the generated file.
- Terraform `core/static-website` gains the equivalent `enable_waf`, `encryption`, `kms_key_arn`, `enable_key_rotation` variables, with the KMS key/alias and WAF ACL made conditional via `count`.
- Extended this to the Terraform app-level module too (`app/static-websites/<name>/<name>.tf`), which previously hardcoded everything except `website_name`/`website_file_path`. It now forwards all of the above plus the pre-existing `custom_domain_names`/`acm_certificate_arn`, matching the pass-through convention every other app-wraps-core Terraform module in this plugin already follows (`rdb`, `agent-core`, `dcr-proxies`, the REST/HTTP API). Each website's module can now be configured independently.
- Added an Nx migration (`static-website-configurable-waf-encryption`) to bring existing workspaces up to this shape across all three surfaces: the CDK core construct, CDK app constructs, and Terraform core plus app modules. It pattern-matches the vended shape before rewriting anything and reports diverged files via `nextSteps` instead of touching what it can't confirm.
- Updated the `react-website` guide with new WAF and Bucket Encryption sections, and rewrote the Custom Domain & TLS example. All three, plus the existing Security Headers section, moved from under Generator Output → Infrastructure to their own sections under Deploying Your Website, matching how `trpc.mdx`/`fastapi.mdx` place WAF and Access logging under their own "Deploying your X" heading rather than under Generator Output. The CDK examples now show configuring `domainNames`/`certificate`/`enableWaf`/`encryption` at the application-stack level (`new MyWebsite(this, 'MyWebsite', { ... })`), matching the equivalent `enableWaf` example already documented for the tRPC API generator.
- Added a note with working code showing that disabling WAF on CDK requires suppressing the checkov `CKV_AWS_68` check. Terraform doesn't need the same suppression, since checkov doesn't statically evaluate its runtime conditional the same way.
- Fixed a stale code example in the dungeon-game tutorial: the literal `GameUI` construct snippet no longer matched what `ts#website` generates, since it predated this change. Updated it to show the new `Props` type and `props` constructor parameter.

Unit tests for the migration cover 17 cases: applying to the vended shape, idempotency on re-run, and workspaces that already customized the pre-existing `domainNames`/`certificate` (CDK) or `custom_domain_names`/`acm_certificate_arn` (Terraform) properties. That last case caught two real bugs while writing the migration:

- A GritQL pattern that anchored on `websiteName` being followed by exactly one more property, which only matches a single trailing node, not an arbitrary list. A workspace with more than one existing customization (say `domainNames` and `certificate` both) failed the match and was safely skipped, though not migrated.
- A missing divergence check on the Terraform side that would have silently produced a duplicate `custom_domain_names` argument, which is invalid HCL, with no warning to the user.

Both are fixed and now have regression tests.

Also validated end to end against real generated workspaces, not just unit tests:

- Built the plugin locally, generated fresh CDK and Terraform workspaces on the published version, and ran the actual `nx migrate` cycle against the local build.
- `terraform validate` passes on the migrated module, including with `enable_waf = false` and `encryption = "S3_MANAGED"` set at the root module call.
- The Terraform `checkov` target reports 0 failed checks for the default configuration. Setting `encryption = "S3_MANAGED"` or `enable_key_rotation = false` does fail checkov (`CKV_AWS_145`/`CKV_AWS_7`); this needs a `.checkov.yaml` skip, now documented in the Bucket Encryption section.
- The CDK `checkov` target genuinely fails with `CKV_AWS_68` when `enableWaf: false`, confirming the new doc note is necessary, and passes again once suppressed per the documented example.
- Re-running the migration reports "No changes were made" against the real project.
- Checked the one change that looked riskiest for anyone with an already-deployed Terraform stack: introducing `count` on the KMS key, KMS alias, and WAF ACL changes their resource addresses. Reproduced the exact same pattern locally (a `local` expression combining two variables, same as the real template) with `random_id` resources so I could actually apply and plan without AWS credentials. Terraform's plan showed 0 to add, 0 to change, 0 to destroy, just a "has moved to" note. This has been automatic since Terraform 1.1, so it's not a break for anyone on a reasonably current CLI.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/react-website.mdx
+++ b/docs/src/content/docs/en/guides/react-website.mdx
@@ -334,6 +334,22 @@ export class ApplicationStack extends Stack {
 }
 ```
 
+:::caution[Suppress CKV_AWS_158 when using S3_MANAGED]
+Choosing `BucketEncryption.S3_MANAGED` means the access log group is no longer KMS encrypted, which fails the `checkov` security scan with `CKV_AWS_158`. Suppress it:
+
+```ts title="packages/infra/src/stacks/application-stack.ts"
+import { LogGroup } from 'aws-cdk-lib/aws-logs';
+import { suppressRules } from '@my-scope/common-constructs';
+
+suppressRules(
+  website,
+  ['CKV_AWS_158'],
+  'Access log group is not KMS encrypted when encryption is S3_MANAGED',
+  (c) => c instanceof LogGroup,
+);
+```
+:::
+
 To use your own KMS key instead of one created automatically, pass `encryptionKey`:
 
 ```ts {2}
@@ -342,10 +358,36 @@ new MyWebsite(this, 'MyWebsite', {
 });
 ```
 
-`enableKeyRotation` (default `true`) only applies to the automatically created key, i.e. when `encryption` is `BucketEncryption.KMS` (the default) and no `encryptionKey` is supplied.
+:::caution[Imported keys need their own permissions]
+A key imported via `Key.fromKeyArn` must already grant the CloudWatch Logs, S3 and CloudFront service principals the permissions they need in its own key policy. `addToResourcePolicy` is a no-op on an imported key, so this construct can't grant them for you, and `cdk synth` won't warn you - the failure only shows up at deploy time. A key created in the same app (`new Key(this, 'WebsiteKey')`) doesn't have this problem, since CDK can update its policy directly.
+:::
+
+`enableKeyRotation` (default `true`) only applies to the automatically created key, i.e. when `encryption` is `BucketEncryption.KMS` (the default) and no `encryptionKey` is supplied:
+
+```ts {2}
+new MyWebsite(this, 'MyWebsite', {
+  enableKeyRotation: false,
+});
+```
+
+:::caution[Suppress CKV_AWS_7 when disabling key rotation]
+Disabling key rotation on the automatically created key fails the `checkov` security scan with `CKV_AWS_7`. Suppress it:
+
+```ts title="packages/infra/src/stacks/application-stack.ts"
+import { Key } from 'aws-cdk-lib/aws-kms';
+import { suppressRules } from '@my-scope/common-constructs';
+
+suppressRules(
+  website,
+  ['CKV_AWS_7'],
+  'Key rotation is managed externally',
+  (c) => c instanceof Key,
+);
+```
+:::
 </Fragment>
 <Fragment slot="terraform">
-If you want to use a different encryption configuration, set the `encryption`, `kms_key_arn` and `enable_key_rotation` variables:
+If you want to use a different encryption configuration, set the `encryption`, `kms_key_arn`, `create_kms_key` and `enable_key_rotation` variables:
 
 ```hcl title="packages/infra/src/main.tf" {7}
 module "my_website" {
@@ -358,9 +400,52 @@ module "my_website" {
 }
 ```
 
-To use your own KMS key instead of one created automatically, pass `kms_key_arn`. A customer-supplied key must already grant the CloudWatch Logs, S3 and CloudFront service principals the permissions they need in its own key policy.
+:::caution[Checkov: S3_MANAGED]
+Choosing `S3_MANAGED` fails the `checkov` security scan with `CKV_AWS_145` on the website and distribution log buckets. The vended `test` target doesn't accept a `--config-file`, but `checkov` auto-discovers a `.checkov.yaml` in its working directory, so drop one in `packages/infra/src`:
 
-`enable_key_rotation` (default `true`) only applies to the automatically created key, i.e. when `encryption` is `"KMS"` (the default) and no `kms_key_arn` is supplied.
+```yaml title="packages/infra/src/.checkov.yaml"
+skip-check:
+  - CKV_AWS_145
+```
+:::
+
+To use your own KMS key instead of one created automatically, pass `kms_key_arn` and set `create_kms_key` to `false`:
+
+```hcl title="packages/infra/src/main.tf" {7-8}
+module "my_website" {
+  source = "../../common/terraform/src/app/static-websites/my-website"
+
+  providers = {
+    aws.us_east_1 = aws.us_east_1
+  }
+  kms_key_arn    = aws_kms_key.website.arn
+  create_kms_key = false
+}
+```
+
+A customer-supplied key must already grant the CloudWatch Logs, S3 and CloudFront service principals the permissions they need in its own key policy.
+
+`enable_key_rotation` (default `true`) only applies to the automatically created key, i.e. when `encryption` is `"KMS"` (the default) and `create_kms_key` is `true`:
+
+```hcl title="packages/infra/src/main.tf" {7}
+module "my_website" {
+  source = "../../common/terraform/src/app/static-websites/my-website"
+
+  providers = {
+    aws.us_east_1 = aws.us_east_1
+  }
+  enable_key_rotation = false
+}
+```
+
+:::caution[Checkov: key rotation disabled]
+Disabling key rotation on the automatically created key fails the `checkov` security scan with `CKV_AWS_7`. Add it to the same `.checkov.yaml`:
+
+```yaml title="packages/infra/src/.checkov.yaml"
+skip-check:
+  - CKV_AWS_7
+```
+:::
 </Fragment>
 </Infrastructure>
 
--- a/docs/src/content/docs/en/guides/react-website.mdx
+++ b/docs/src/content/docs/en/guides/react-website.mdx
@@ -129,58 +129,7 @@ waf -> cloudfront
 cloudfront -> s3
 ```
 
-#### Security Headers
-
-The CloudFront distribution applies a response headers policy that sets `Strict-Transport-Security`, `X-Content-Type-Options`, `X-Frame-Options: DENY`, `Referrer-Policy` and a `Content-Security-Policy` on all responses.
-
-A default `Content-Security-Policy` is enforced. It restricts scripts and framing to mitigate XSS and clickjacking, while permitting HTTPS and WSS connections so the website can call AWS service endpoints (such as API Gateway, Cognito and Bedrock AgentCore) whose URLs are only known at deploy time. To adjust the policy (for example to tighten `connect-src` to your specific origins), edit the `content_security_policy` value in your generated `static-website.ts` (CDK) or `static-website.tf` (Terraform).
-
-`runtime-config.json` is served with `Cache-Control: no-cache` so that browsers always fetch the latest configuration after a redeploy, rather than using a stale cached copy.
-
-#### Custom Domain & TLS
-
-By default the distribution uses the default CloudFront domain name (`*.cloudfront.net`) and its default certificate, which does not support enforcing a minimum TLS version of 1.2. To serve your website from your own domain, supply an [ACM certificate](https://docs.aws.amazon.com/acm/latest/userguide/acm-overview.html) (which must reside in `us-east-1` for use with CloudFront) and your domain names — a minimum TLS version of 1.2 is then enforced for viewers:
-
-<Infrastructure>
-<Fragment slot="cdk">
-Pass the `certificate` and `domainNames` props through in your generated website construct in `packages/common/constructs/src/app/static-websites`:
-
-```ts {6-8}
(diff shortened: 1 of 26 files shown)


packages/nx-plugin/src/utils/website-constructs/files/terraform/core/static-website/static-website.tf.template
```diff
@@ -313,6 +353,8 @@ resource "aws_s3_bucket_policy" "distribution_logs_ssl_policy" {
 
 # WAF Web ACL (must be in us-east-1 for CloudFront)
 resource "aws_wafv2_web_acl" "cloudfront_waf" {
+  count = var.enable_waf ? 1 : 0
```
nx-plugin-for-aws, line 362: **`enable_waf = false` fails at `terraform plan` on any greenfield deployment.**

Generating a website and setting `enable_waf = false` before the first apply fails outright:

```
Error: no change found for aws_wafv2_web_acl.cloudfront_waf in module.web.module.static_website
```

The cause is code this PR doesn't touch — `aws_cloudfront_distribution.website` further down this file (line 540) has:

```hcl
  lifecycle {
    replace_triggered_by = [
      aws_wafv2_web_acl.cloudfront_waf
    ]
  }
```

Once the ACL gains `count`, that reference can't be resolved when the count is 0 **and** the resource has no prior state, so the plan aborts.

Scope, confirmed by bisecting on real workspaces:

- **Fails:** first apply of a website with `enable_waf = false`, and adding a new WAF-disabled website module to an existing configuration.
- **Works:** toggling `enable_waf = false` on an already-deployed website. Terraform then has a delete change object to reference, so it plans and applies cleanly — the distribution is updated in place, no replacement, and the site keeps serving.

This also reaches migrated workspaces: the migration adds `count` but leaves `replace_triggered_by` alone (its spec fixture at `migration.spec.ts:480-484` contains the block).

**Fix that works:** delete the `lifecycle` / `replace_triggered_by` block. With it removed, both `enable_waf = false` and the default configuration plan and apply cleanly from scratch. Neither index form is a workaround — `aws_wafv2_web_acl.cloudfront_waf[0]` gives the same `no change found` error, and `aws_wafv2_web_acl.cloudfront_waf[*]` gives `Invalid replace_triggered_by expression`. Dropping it looks safe: `web_acl_id` already references the ACL ARN directly (line 482), so a replaced ACL still drives a distribution update.

packages/nx-plugin/src/utils/website-constructs/files/terraform/core/static-website/static-website.tf.template
```diff
@@ -46,6 +75,13 @@ locals {
   # Delivery source/destination names have a 60 character maximum. Truncate the
   # website name so the longest delivery name stays within the limit.
   access_logs_name_prefix = substr(lower(var.website_name), 0, 16)
+
+  create_website_key = var.encryption == "KMS" && var.kms_key_arn == null
```
nx-plugin-for-aws: **`kms_key_arn` can't be set to a value that isn't known at plan time.**

The natural way to bring your own key is to create it in the same configuration:

```hcl
resource "aws_kms_key" "website" { ... }

module "my_website" {
  source      = "../../common/terraform/src/app/static-websites/my-website"
  kms_key_arn = aws_kms_key.website.arn
  providers   = { aws.us_east_1 = aws.us_east_1 }
}
```

That fails on a fresh apply:

```
Error: Invalid count argument

  on ../../common/terraform/src/core/static-website/static-website.tf line 95, in resource "aws_kms_key" "website_key":
  95:   count = local.create_website_key ? 1 : 0

The "count" value depends on resource attributes that cannot be determined until apply, so
Terraform cannot predict how many instances will be created.
```

`create_website_key` derives from `var.kms_key_arn == null`, and Terraform can't refine an unknown provider attribute to non-null, so the count is unknown.

Passing an ARN that *is* known at plan time (a literal, or a variable) works — I deployed that path successfully, with the supplied key correctly wired into both buckets and the CloudWatch log group.

Two options: derive `count` from an explicit `create_kms_key` boolean variable instead of from `kms_key_arn`'s nullness, or document the constraint on the `kms_key_arn` variable and in the guide's "use your own KMS key" section (a `-target` two-pass apply is the only alternative for users, which is a rough experience).

nx-plugin-for-aws (commented):
Tested this end to end against `593a4709`: all five permutations deployed on both CDK and Terraform, plus the migration run against a live pre-PR Terraform stack.

Most of it holds up well — the default configuration synthesizes a byte-identical CloudFormation template pre/post PR, the migration output matches fresh generation and is idempotent on both surfaces, the diverged-file reporting works, and the resource-address move for the new `count` really is a clean no-op on a deployed stack (`has moved to`, zero changes to the KMS key, alias and Web ACL).

Four things found, one of which blocks the Terraform side. Details in the line comments.

docs/src/content/docs/en/guides/react-website.mdx
```diff
@@ -231,6 +180,236 @@ export const MyComponent = () => {
...
+    suppressRules(
+      website.cloudFrontDistribution,
+      ['CKV_AWS_68'],
+      'WAF is intentionally disabled for this distribution',
+    );
+  }
+}
+```
+
+:::caution[Suppress CKV_AWS_68 when disabling WAF]
```
nx-plugin-for-aws, line 292: **Three of the four new opt-outs break the vended `checkov` gate; only this one is documented.**

The suppression documented here works verbatim — nice. But the other opt-outs fail too:

| Opt-out | CDK | Terraform |
| --- | --- | --- |
| WAF off | `CKV_AWS_68` — documented here ✅ | none ✅ |
| `S3_MANAGED` | `CKV_AWS_158` on the `AccessLogs` log group | `CKV_AWS_145` ×2 on the buckets |
| rotation off | `CKV_AWS_7` | `CKV_AWS_7` |

This is a hard gate, not just a warning: `checkov` is in the CDK `build` target, and on Terraform `plan` → `test`, so `nx run infra:deploy` refuses to run. A workspace that sets `encryption = "S3_MANAGED"` or `enable_key_rotation = false` can't deploy without knowing the suppression trick.

Two details worth capturing while documenting these:

- The `suppressRules(website.cloudFrontDistribution, ...)` shape shown above doesn't extend to the other two, because neither the access log group nor the auto-created key is a public member of `StaticWebsite`. The predicate form does work:

  ```ts
  suppressRules(
    website,
    ['CKV_AWS_158'],
    'Access log group is not KMS encrypted when encryption is S3_MANAGED',
    (c) => c instanceof LogGroup,
  );
  ```

  Same shape with `(c) => c instanceof Key` and `CKV_AWS_7` for rotation. Both verified — CDK goes from 3 failed to 0 failed / 141 passed.

- On Terraform the `test` target takes no `--config-file`, but checkov auto-discovers a `.checkov.yaml` in its cwd, so dropping one in `packages/infra/src` with `skip-check: [CKV_AWS_7, CKV_AWS_145]` works. Verified — 0 failed / 227 passed.

Also a small correction for the PR description: "The Terraform `checkov` target still reports 0 failed checks" holds for the default configuration, but not once the new opt-outs are set.

packages/nx-plugin/src/utils/website-constructs/files/cdk/core/static-website.ts.template
```diff
@@ -74,6 +74,30 @@ export interface StaticWebsiteProps {
...
+   * Server-side encryption for the website and distribution log buckets.
+   *
+   * @default BucketEncryption.KMS
+   */
+  readonly encryption?: BucketEncryption;
+  /**
+   * KMS key used to encrypt the website and distribution log buckets. Only used when `encryption` is
+   * `BucketEncryption.KMS`. When not provided, a new key is created.
+   */
+  readonly encryptionKey?: IKey;
```
nx-plugin-for-aws, line 96: **An imported `encryptionKey` fails at deploy time with nothing at synth to warn you.**

The Terraform `kms_key_arn` variable description spells out that a customer-supplied key must already grant the CloudWatch Logs, S3 and CloudFront service principals what they need. The same constraint applies here, but neither this JSDoc nor the guide's `encryptionKey: myKey` example mentions it, and the failure mode is worse because it's silent until deploy.

`websiteKey?.addToResourcePolicy(...)` is a no-op for a key created via `Key.fromKeyArn` (imported keys have no policy CDK can mutate, and `addToResourcePolicy` defaults to `allowNoOp: true`). So the access log group is synthesized with `KmsKeyId` pointing at a key that doesn't grant `logs.<region>.amazonaws.com` — `cdk synth` succeeds with no warning, then the deploy fails with:

```
Resource handler returned message: "The specified KMS key does not exist or is not allowed to be
used with Arn 'arn:aws:logs:us-east-1:...:log-group:...'" (Service: CloudWatchLogs,
Status Code: 400) ... HandlerErrorCode: AccessDenied
```

(Verified both halves separately: synth with an imported key emits `KmsKeyId` and zero `AWS::KMS::Key` resources with no annotation; and a two-resource CloudFormation probe with a logs-less key policy fails with exactly the error above.)

A key created in the same app works fine — `addToResourcePolicy` applies there, and I deployed that permutation successfully.

Suggest mirroring the Terraform wording in this JSDoc, and adding a line to the guide's "use your own KMS key" CDK snippet.

cogwirrel (commented):
Looks great, just a few minor comments from my agent after some deployment testing! :)

packages/nx-plugin/src/utils/website-constructs/files/terraform/core/static-website/static-website.tf.template
AlexTo: Fixed, just removed the lifecycle block. It's redundant anyway since web_acl_id already points at the ACL arn, so a replaced ACL still drives a normal update. Tested a greenfield apply with enable_waf = false, works clean now. Added a migration step for existing workspaces too.

packages/nx-plugin/src/utils/website-constructs/files/terraform/core/static-website/static-website.tf.template
AlexTo: Fixed, added a create_kms_key bool instead of inferring it from kms_key_arn being null, same fix as on the dynamodb PR. Tested with a real KMS key resource in the same config, plans clean now. Updated the app module pass-through and the docs too.

docs/src/content/docs/en/guides/react-website.mdx
AlexTo: added docs for both. cdk gets suppressRules examples for CKV_AWS_158 (S3_MANAGED) and CKV_AWS_7 (rotation off), using the predicate form since neither the log group or the key is public on the construct. terraform gets a .checkov.yaml skip-check snippet since test target doesn't take a config file. also fixed the pr description, 0 failed only holds for default config

packages/nx-plugin/src/utils/website-constructs/files/cdk/core/static-website.ts.template
AlexTo: yeah confirmed that, addToResourcePolicy is a no-op on an imported key so synth won't warn you. added a note to the encryptionKey jsdoc and the guide saying an imported key needs its own permissions already, same wording as the terraform kms_key_arn description

cogwirrel (approved):

PR #1108, opened 2026-08-23
feat(ts#website#auth): make MFA configurable on the vended UserIdentity construct/module

AlexTo:
### Reason for this change

`UserIdentity` hard-codes `mfa: Mfa.REQUIRED` and `mfaSecondFactor: { sms: true, otp: true }`. This makes it convenient to configure MFA from the application stack, the same way `enableWaf` already is, rather than needing to edit the generated construct itself.

### Description of changes

- Added `mfa?: Mfa` (default `Mfa.REQUIRED`) and `mfaSecondFactor?: MfaSecondFactor` (default `{ sms: true, otp: true }`) construct props to the CDK `UserIdentity` template, passed straight through to the underlying `UserPool`.
- Added matching `mfa`, `mfa_second_factor_otp`, `mfa_second_factor_sms` variables to the Terraform `user-identity` module.
- **Design note:** confirmed via `cdk synth` that CloudFormation's `EnabledMfas` is independent of `SmsConfiguration`, so in CDK, `mfaSecondFactor.sms = false` disables SMS as an MFA method without affecting phone-number verification. The Terraform AWS provider has no equivalent — SMS-MFA availability is tied to the presence of `sms_configuration`, which is also required for phone verification. There's no clean way to decouple them without fighting the provider's own internal calls to `SetUserPoolMfaConfig` on every `apply`. So in Terraform, `mfa_second_factor_sms = false` removes `sms_configuration` (and the SMS IAM role) via `count`, and correspondingly drops `phone_number` from `auto_verified_attributes` — documented on the variable and in the guide.
- Added an `nx migrate` codemod (GritQL-based, following the existing `enableWaf`/SSRF migration's pattern) so already-scaffolded workspaces pick this up for both CDK and Terraform.
- Documented the new props/variables in the English `react-website-auth` guide (MFA section between "Threat protection" and "WAF"), including the Terraform caveat above. Other-language translations were regenerated via the repo's translation tooling.

- Added unit tests for the migration (CDK + both Terraform files): applies cleanly, is idempotent, and skips + reports a diverged/customised construct without touching it.
- End-to-end, using a local Verdaccio registry:
  - Generated fresh CDK and Terraform workspaces with the changed generator; diffed the output against the templates.
  - CDK: wired `UserIdentity` into a real stack with `mfa: Mfa.OPTIONAL, mfaSecondFactor: { sms: false, otp: true }`, ran `cdk synth`, and confirmed the synthesized template has `EnabledMfas: ['SOFTWARE_TOKEN_MFA']` while `SmsConfiguration`/`AutoVerifiedAttributes` (phone) are untouched.
  - Terraform: wired the module into a real `main.tf` with overrides and ran `terraform init -backend=false` / `terraform validate` successfully (this caught a real bug — see below).
  - Ran a genuine `nx migrate` against a workspace generated from the actual latest published release (`1.0.0-rc.76`), for both CDK and Terraform, confirming the migration transforms real generated files correctly and the workspace still builds/validates afterward.
- Along the way, `terraform validate` caught that `depends_on = aws_iam_role_policy.cognito_sms_policy[*]` (needed once the resource has `count`) is rejected by Terraform ("a static list expression is required"); fixed to `depends_on = [aws_iam_role_policy.cognito_sms_policy]`, which needs no change from the pre-existing line at all.

### Checklist

----


docs/src/content/docs/en/guides/react-website-auth.mdx
```diff
@@ -119,6 +119,40 @@ Set `advanced_security_mode` to `ENFORCED` in the `user_pool_add_ons` block in `
...
+}
+```
+</Fragment>
+</Infrastructure>
+
+`mfa` accepts `Mfa.OFF` / `Mfa.OPTIONAL` / `Mfa.REQUIRED` in CDK, and `"OFF"` / `"OPTIONAL"` / `"ON"` in Terraform. `mfaSecondFactor.sms` / `mfa_second_factor_sms` and the `otp` equivalent enable or disable each second-factor method independently; they have no effect when `mfa` is off.
+
+:::caution[Terraform ties SMS MFA to phone number verification]
+Unlike the CDK construct, the Terraform AWS provider has no way to offer SMS as an MFA method independently of using it for phone number verification. Setting `mfa_second_factor_sms = false` therefore also disables SMS-based phone number verification (`auto_verified_attributes` drops `phone_number`).
+:::
```
cogwirrel: My agent's testing found that the same coupling of sms mfa to phone verification applies for cdk too, I'll get it to reply with more details :)

Also a small nit: please can we move the cdk/tf specific pieces into their tabs above? :)

cogwirrel (commented):
Looks great, just a small comment :)

docs/src/content/docs/en/guides/react-website-auth.mdx
nx-plugin-for-aws: Confirmed — the coupling applies to CDK as well, so the caution box as written isn't accurate. Details from testing below.

### The issue

Any `mfaSecondFactor.sms = false` combination **fails to deploy** when `mfa` is `OPTIONAL` or `REQUIRED`:

```
Resource handler returned message: "Invalid request provided: Can't turn off SMS_MFA
when SMS configuration is set for the user pool with a required or optional
MFAConfiguration." (HandlerErrorCode: InvalidRequest)
```

Note this includes the exact CDK snippet in this section of the guide (`mfa: Mfa.OPTIONAL, mfaSecondFactor: { sms: false, otp: true }`) — it rolls back on deploy.

### What I saw

Deployed 6 CDK permutations to a real account (us-east-1):

| Permutation | Result |
|---|---|
| `default` (no props) | ✅ CREATE_COMPLETE |
| `mfa: OFF` | ✅ CREATE_COMPLETE |
| `REQUIRED` + `{sms: true, otp: false}` | ✅ CREATE_COMPLETE |
| `OPTIONAL` + `{sms: false, otp: true}` | ❌ ROLLBACK_COMPLETE |
| `REQUIRED` + `{sms: false, otp: true}` | ❌ ROLLBACK_COMPLETE |
| `REQUIRED` + `{sms: false, otp: false}` | ❌ ROLLBACK_COMPLETE |

All 20 permutations pass `cdk synth`, and all 30 CI checks are green — CloudFormation only enforces this at create time, so synth can't catch it. That's why the `cdk synth` validation in the PR description came back clean.

**Root cause:** the construct hardcodes `autoVerify: { phone: true }`. CDK's internal `smsConfiguration()` emits `SmsConfiguration` whenever phone is auto-verified (or `signInAliases.phone` is set) — entirely independent of `enabledMfas`. Cognito then rejects `SmsConfiguration` present + optional/required MFA + `SMS_MFA` absent from `EnabledMfas`. Isolated it by synthing a bare `UserPool` across three variants:

| Variant | `EnabledMfas` | `SmsConfiguration` |
|---|---|---|
| `autoVerify.phone: true` (what the construct does) | `[SOFTWARE_TOKEN_MFA]` | **present** → deploy fails |
| `autoVerify.phone: false` | `[SOFTWARE_TOKEN_MFA]` | absent |
| `enableSmsRole: false` | `[SOFTWARE_TOKEN_MFA]` | absent |

Cognito enforces this in *both* directions, so the two settings have to move together — `enableSmsRole: false` on its own also fails:

```
"SMS configuration is required when phone_number is selected for auto verification"
```

### Recommendation

Mirror what the Terraform module in this PR already does: when SMS MFA is off, drop phone from auto-verification too. I deployed this successfully (`CREATE_COMPLETE`) with `mfa: REQUIRED, mfaSecondFactor: { sms: false, otp: true }`:

```ts
const smsEnabled = mfa !== Mfa.OFF && mfaSecondFactor.sms;

autoVerify: { email: true, phone: smsEnabled },
keepOriginal: { email: true, phone: smsEnabled },
```

`keepOriginal` needs the same treatment — `attributesRequireVerificationBeforeUpdate` must be a subset of `autoVerifiedAttributes` or `UpdateUserPool` is rejected.

Two follow-ups while this section is being reworked:

1. **The caution box should cover both.** Rather than framing it as a Terraform-only limitation, it's a Cognito constraint that both IaC providers inherit — SMS MFA can't be decoupled from SMS phone verification in either. Happy to take the nit about moving the cdk/tf specifics into the tabs at the same time.

2. **`REQUIRED` + `{sms: false, otp: false}`** is worth a guard. AWS accepts it (verified live via Terraform) and CDK doesn't validate it, producing a pool that mandates MFA with no available second factor. Cheap to reject at synth time.

<sub>Also flagged separately: on the Terraform side, toggling `mfa_second_factor_sms` true→false on an *existing* pool leaves the live pool advertising SMS MFA against a deleted IAM role, with `plan` reporting no changes. Fresh applies are unaffected — I'll write that up as its own thread since it's independent of this one.</sub>

AlexTo: - same Cognito constraint applies to CDK, autoVerify.phone was hardcoded true so SmsConfiguration stayed set even with SMS_MFA off. now gated on whether SMS is actually usable as a second factor
- added a synth time guard for REQUIRED + no factors
- verified with a real cdk deploy this time, not just synth
- moved cdk/tf bits into the tabs, caution box now covers both providers

cogwirrel (approved):
Thanks @AlexTo 🤩

PR #1109, opened 2026-08-23
feat(dynamodb): make table encryption configurable

AlexTo:
### Reason for this change

`DynamoDBTable` (and its Terraform equivalent) hardcodes `TableEncryption.CUSTOMER_MANAGED` and always creates a dedicated KMS key. Workspaces have no way to use the free AWS-owned key, the shared AWS-managed key, or their own existing key, without hand-editing generated code — and that edit gets clobbered the next time the generator runs. This follows the same pattern already established for `StaticWebsite` (#1107).

### Description of changes

- CDK `DynamoDBTable` construct (`common/constructs/src/core/dynamodb.ts`): new `encryption` and `encryptionKey` props. The KMS key is only created when `encryption` is `TableEncryption.CUSTOMER_MANAGED` and no `encryptionKey` is supplied; `enableKeyRotation` only applies to that auto-created key. `TableEncryption.DEFAULT` (the AWS owned key) already works for free, since the prop passes straight through to CDK's own `Table` construct.
- Terraform `dynamodb` core module and per-table app module: new `encryption` and `kms_key_arn` variables. `encryption` accepts `CUSTOMER_MANAGED`, `AWS_MANAGED`, or `DEFAULT`. The KMS key resource is now conditional (`count`), and `server_side_encryption.enabled` is tied to `encryption != "DEFAULT"` — note that DynamoDB always encrypts at rest regardless of this flag; it only chooses between the AWS-owned key and a KMS-visible key.
- Nx migration (`dynamodb-configurable-encryption`) to bring existing workspaces' vended files up to the new shape: the CDK construct, the Terraform core module, and every per-table Terraform app module. Diverged files are left untouched and reported via `nextSteps`.
- Docs: added an "Encryption" section to the shared `dynamodb/deploying-table` snippet (used by both `ts-dynamodb.mdx` and `py-dynamodb.mdx`), covering all three encryption modes plus bringing your own key. Split the per-mode subsections and the existing "Encryption Key Rotation" section into a new nested `dynamodb/encryption-options` snippet so they render nested under "Encryption" in the sidebar, mirroring the existing `prerequisites.mdx` pattern. Extracted the same structure across all translated locales, using their already-translated text.
- Dungeon-game tutorial: the sandbox `ApplicationStack` now creates `DungeonDb` with `encryption: TableEncryption.DEFAULT`, since a customer-managed KMS key can't be deleted immediately (a mandatory 7-30 day wait) — unnecessary friction for a stack that's regularly torn down and recreated during the tutorial.

- Added `migration.spec.ts` (10 tests) covering vended-shape application, idempotency, and divergence handling across the CDK construct, the Terraform core module, and Terraform app modules.
- Updated `ts#dynamodb`/`py#dynamodb` generator snapshots to match the new vended shape.
- Tested the migration end-to-end against real workspaces generated from the published `@aws/nx-plugin` version, for both `--iac=cdk` and `--iac=terraform`: confirmed `pnpm lint` passes before and after, the migration is idempotent on re-run, and `terraform validate` passes on both the core and app modules.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
+++ b/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
@@ -33,7 +33,7 @@ variable "deletion_protection_enabled" {
 }
 
 variable "encryption" {
-  description = "Server-side encryption for the table. One of CUSTOMER_MANAGED, AWS_MANAGED or DEFAULT (the AWS owned key, at no cost and with no key to manage)."
+  description = "Server-side encryption for the table. One of CUSTOMER_MANAGED, AWS_MANAGED or DEFAULT (the AWS owned key, at no cost and with no key to manage). Changing this on an already-deployed table away from CUSTOMER_MANAGED requires a manual step first - see the ts#dynamodb generator's docs."
   type        = string
   default     = "CUSTOMER_MANAGED"
 
--- a/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
+++ b/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
@@ -44,13 +44,19 @@ variable "encryption" {
 }
 
 variable "kms_key_arn" {
-  description = "ARN of an existing KMS key used to encrypt the table when encryption is CUSTOMER_MANAGED. When not provided, a new key is created. Note that a customer-supplied key must already grant the DynamoDB service the necessary permissions in its own key policy."
+  description = "ARN of an existing KMS key used to encrypt the table when encryption is CUSTOMER_MANAGED. When not provided and create_kms_key is true, a new key is created. Note that a customer-supplied key must already grant the DynamoDB service the necessary permissions in its own key policy."
   type        = string
   default     = null
 }
 
+variable "create_kms_key" {
+  description = "Whether to create a KMS key for the table. Only applies when encryption is CUSTOMER_MANAGED. Set to false when supplying kms_key_arn."
+  type        = bool
+  default     = true
+}
+
 variable "enable_key_rotation" {
-  description = "Whether to enable automatic key rotation on the KMS key used to encrypt the table. Only applies when encryption is CUSTOMER_MANAGED and kms_key_arn is not provided."
+  description = "Whether to enable automatic key rotation on the KMS key used to encrypt the table. Only applies when encryption is CUSTOMER_MANAGED and create_kms_key is true."
   type        = bool
   default     = true
 }
@@ -71,16 +77,17 @@ locals {
     for gsi in var.global_secondary_indexes : [gsi.hash_key, gsi.range_key]
   ]))
 
-  create_table_key = var.encryption == "CUSTOMER_MANAGED" && var.kms_key_arn == null
+  create_table_key = var.encryption == "CUSTOMER_MANAGED" && var.create_kms_key
   table_kms_key_arn = (
-    var.encryption != "CUSTOMER_MANAGED" ? null :
-    local.create_table_key ? aws_kms_key.table[0].arn :
-    var.kms_key_arn
+    var.encryption == "CUSTOMER_MANAGED" ? (local.create_table_key ? aws_kms_key.table[0].arn : var.kms_key_arn) :
+    var.encryption == "AWS_MANAGED" ? "arn:${data.aws_partition.current.partition}:kms:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:alias/aws/dynamodb" :
+    null
   )
 }
 
 data "aws_caller_identity" "current" {}
 data "aws_partition" "current" {}
+data "aws_region" "current" {}
 
 resource "aws_kms_key" "table" {
   count = local.create_table_key ? 1 : 0
@@ -177,6 +184,6 @@ output "table_arn" {
 }
 
 output "kms_key_arn" {
-  description = "ARN of the KMS key used to encrypt the DynamoDB table, if using customer-managed encryption."
+  description = "ARN of the KMS key used to encrypt the DynamoDB table, or null when using the AWS owned key (encryption = DEFAULT)."
   value       = local.table_kms_key_arn
 }
--- a/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
+++ b/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
@@ -32,8 +32,25 @@ variable "deletion_protection_enabled" {
   default     = true
 }
 
+variable "encryption" {
+  description = "Server-side encryption for the table. One of CUSTOMER_MANAGED, AWS_MANAGED or DEFAULT (the AWS owned key, at no cost and with no key to manage)."
+  type        = string
+  default     = "CUSTOMER_MANAGED"
+
+  validation {
+    condition     = contains(["CUSTOMER_MANAGED", "AWS_MANAGED", "DEFAULT"], var.encryption)
+    error_message = "encryption must be one of CUSTOMER_MANAGED, AWS_MANAGED or DEFAULT."
+  }
+}
+
+variable "kms_key_arn" {
+  description = "ARN of an existing KMS key used to encrypt the table when encryption is CUSTOMER_MANAGED. When not provided, a new key is created. Note that a customer-supplied key must already grant the DynamoDB service the necessary permissions in its own key policy."
+  type        = string
+  default     = null
+}
+
 variable "enable_key_rotation" {
-  description = "Whether to enable automatic key rotation on the KMS key used to encrypt the table."
+  description = "Whether to enable automatic key rotation on the KMS key used to encrypt the table. Only applies when encryption is CUSTOMER_MANAGED and kms_key_arn is not provided."
   type        = bool
   default     = true
 }
@@ -53,12 +70,21 @@ locals {
   gsi_attribute_names = distinct(flatten([
     for gsi in var.global_secondary_indexes : [gsi.hash_key, gsi.range_key]
   ]))
+
+  create_table_key = var.encryption == "CUSTOMER_MANAGED" && var.kms_key_arn == null
+  table_kms_key_arn = (
+    var.encryption != "CUSTOMER_MANAGED" ? null :
+    local.create_table_key ? aws_kms_key.table[0].arn :
+    var.kms_key_arn
+  )
 }
 
 data "aws_caller_identity" "current" {}
 data "aws_partition" "current" {}
 
 resource "aws_kms_key" "table" {
+  count = local.create_table_key ? 1 : 0
+
   description         = "KMS key for DynamoDB table ${var.name}"
   enable_key_rotation = var.enable_key_rotation
 
@@ -134,8 +160,8 @@ resource "aws_dynamodb_table" "table" {
   }
 
   server_side_encryption {
-    enabled     = true
-    kms_key_arn = aws_kms_key.table.arn
+    enabled     = var.encryption != "DEFAULT"
+    kms_key_arn = local.table_kms_key_arn
   }
 }
 
@@ -150,6 +176,6 @@ output "table_arn" {
 }
 
 output "kms_key_arn" {
-  description = "ARN of the KMS key used to encrypt the DynamoDB table."
-  value       = aws_kms_key.table.arn
+  description = "ARN of the KMS key used to encrypt the DynamoDB table, if using customer-managed encryption."
+  value       = local.table_kms_key_arn
 }
--- a/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
+++ b/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
@@ -121,6 +121,7 @@ resource "aws_kms_key" "table" {
 }
 
 resource "aws_dynamodb_table" "table" {
+  #checkov:skip=CKV_AWS_119:Encryption is configurable via var.encryption; checkov cannot resolve the conditional server_side_encryption.enabled expression
   name                        = var.name
   billing_mode                = var.billing_mode
   hash_key                    = "pk"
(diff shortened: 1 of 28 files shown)


packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
```diff
@@ -53,12 +70,21 @@ locals {
   gsi_attribute_names = distinct(flatten([
     for gsi in var.global_secondary_indexes : [gsi.hash_key, gsi.range_key]
   ]))
+
+  create_table_key = var.encryption == "CUSTOMER_MANAGED" && var.kms_key_arn == null
```
nx-plugin-for-aws: **`count` derived from `kms_key_arn == null` breaks the bring-your-own-key path**

`create_table_key` depends on `var.kms_key_arn == null`, and `count` can't depend on a value that's unknown at plan time. Any key that comes from a resource in the same configuration — which is the normal way to do this — fails before apply:

```
Error: Invalid count argument
  on ../../common/terraform/src/core/dynamodb/dynamodb.tf line 86, in resource "aws_kms_key" "table":
  86:   count = local.create_table_key ? 1 : 0
The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
```

I reproduced this with 12 lines of HCL against the vended module, and in a generated workspace. It only works if `kms_key_arn` is a hardcoded literal (as in the docs snippet); the sole workaround otherwise is a two-phase `terraform apply -target=aws_kms_key.byo` first. Worth noting the CDK side has no equivalent problem — `encryptionKey` there is fine.

Suggest an explicit boolean rather than inferring intent from the ARN's nullness:

```hcl
variable "create_kms_key" {
  description = "Whether to create a KMS key for the table. Only applies when encryption is CUSTOMER_MANAGED. Set to false when supplying kms_key_arn."
  type        = bool
  default     = true
}
```
```hcl
create_table_key = var.encryption == "CUSTOMER_MANAGED" && var.create_kms_key
```

Verified: with this change the same BYO config plans cleanly (`Plan: 2 to add, 0 to change, 0 to destroy`) in a single phase. Needs a matching pass-through in the app module template and an update to the `encryption-options` docs snippet.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
```diff
@@ -134,8 +160,8 @@ resource "aws_dynamodb_table" "table" {
   }
 
   server_side_encryption {
-    enabled     = true
-    kms_key_arn = aws_kms_key.table.arn
+    enabled     = var.encryption != "DEFAULT"
+    kms_key_arn = local.table_kms_key_arn
```
nx-plugin-for-aws, line 172: **Switching `CUSTOMER_MANAGED` → `AWS_MANAGED` silently leaves the table on a deleted key**

For `AWS_MANAGED`, `table_kms_key_arn` resolves to `null`. Since the provider's `kms_key_arn` is `Optional+Computed`, `null` means "keep the existing value" rather than "clear it" — so the KMS key gets destroyed while the table still references it.

Reproduced live in us-east-1 on a deployed table:

```
Apply complete! Resources: 0 added, 0 changed, 1 destroyed.
```

Note `0 changed`. Afterwards `describe-table` still reported the old CMK ARN, `describe-key` showed `KeyState: PendingDeletion`, and a follow-up `terraform plan` reported **"No changes. Your infrastructure matches the configuration."** Once the 7–30 day window elapses that table is permanently unreadable, with nothing in Terraform's output indicating a problem. `DEFAULT` has the same shape (`enabled = false` alongside a stale ARN).

The fix is to pass the AWS-managed key explicitly so a real diff is produced. It must be the full alias **ARN** — the provider rejects a bare `alias/aws/dynamodb` with `is an invalid ARN: arn: invalid prefix`:

```hcl
table_kms_key_arn = (
  var.encryption == "CUSTOMER_MANAGED" ? (local.create_table_key ? aws_kms_key.table[0].arn : var.kms_key_arn) :
  var.encryption == "AWS_MANAGED" ? "arn:${data.aws_partition.current.partition}:kms:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:alias/aws/dynamodb" :
  null
)
```

Verified: the switch then plans a real `~ kms_key_arn` diff, applies as `0 added, 1 changed, 0 destroyed`, and `describe-table` confirms the table moved to `alias/aws/dynamodb`.

One caveat this surfaces: with the CMK destroyed in the same apply, Terraform may order the delete before the table update and fail with `KMSInvalidStateException: ... is pending deletion`. A `create_before_destroy`/`depends_on` ordering guard (or a documented two-step for that transition) would make this robust — flagging rather than prescribing, since you may prefer to handle it differently.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
```diff
@@ -134,8 +160,8 @@ resource "aws_dynamodb_table" "table" {
   }
 
   server_side_encryption {
-    enabled     = true
-    kms_key_arn = aws_kms_key.table.arn
+    enabled     = var.encryption != "DEFAULT"
```
nx-plugin-for-aws, line 171: **The migration turns the vended `test` target red for every existing Terraform workspace**

Routing the key through `local.table_kms_key_arn` defeats checkov's static resolution of CKV_AWS_119, and it fails even when the user changes nothing and stays on the `CUSTOMER_MANAGED` default. On a workspace generated from published `rc.76`:

- before the migration: `Passed checks: 25, Failed checks: 0`
- after the migration: `Passed checks: 24, Failed checks: 1` → `CKV_AWS_119 FAILED for module.dynamodb_table.aws_dynamodb_table.table`

Since `test` gates `apply`/`deploy`, upgrading alone breaks the build. I tried a direct `local.create_table_key ? aws_kms_key.table[0].arn : ...` ternary at the `server_side_encryption` site and checkov still doesn't resolve it, so a suppression seems like the pragmatic route — matching the existing convention in this repo (e.g. the WAF module's `#checkov:skip=CKV2_AWS_31:...Checkov does not resolve...`):

```hcl
resource "aws_dynamodb_table" "table" {
  #checkov:skip=CKV_AWS_119:Encryption is configurable via var.encryption; Checkov does not resolve the key ARN through local.table_kms_key_arn
```

Verified: `Passed checks: 24, Failed checks: 0, Skipped checks: 8`.

Related on the CDK side: choosing `AWS_MANAGED` or `DEFAULT` fails CKV_AWS_119 too, and `enableKeyRotation: false` / a non-rotating BYO key fails CKV_AWS_7. The PR already hits this in the dungeon tutorial and handles it with `suppressRules`, but the `encryption-options` snippet doesn't warn users who pick those modes — worth a short note there with the `suppressRules` line, since the build failure isn't obvious.

nx-plugin-for-aws (commented):
Tested this end to end, including deployment of all encryption permutations to AWS (us-east-1) and migration testing against workspaces generated from published `@aws/nx-plugin@1.0.0-rc.76`. All test resources have been torn down.

**The CDK side is solid.** All five permutations verified against live tables: default `CUSTOMER_MANAGED` (CUSTOMER key, rotation on), `TableEncryption.DEFAULT` (`SSEDescription: null` = AWS owned key, no key created), `AWS_MANAGED` (`alias/aws/dynamodb`, KeyManager=AWS), BYO `encryptionKey` (exact ARN, no extra key), and `enableKeyRotation: false`. In-place encryption changes apply without table replacement. Build, lint and all 3579 unit tests pass; the red "Translate Documentation" check fails on `Input required and not supplied: token`, unrelated to this PR.

The Terraform permutations are all functionally correct once deployed, and the `count` conditional correctly suppresses key creation (exactly 2 KMS keys in state across 4 tables). But I hit two blocking issues and one migration regression on that side — details in the line comments. Every suggested fix below is one I applied and verified.

Separately, two non-blocking notes: the migration's output doesn't exactly match the templates (the CDK construct is missing the `enableKeyRotation` JSDoc scope sentences, and the Terraform `kms_key_arn` output keeps its old description), and the migrated `.tf` files aren't `terraform fmt`-clean — workspace lint doesn't catch that.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
AlexTo: Fixed, added a create_kms_key bool instead of inferring from kms_key_arn being null. Tested it with a real BYO key resource in the same config and it plans clean in one shot now. Also passed the var through the app module and updated the docs snippet.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
AlexTo: Switched to the explicit alias/aws/dynamodb ARN like you suggested, so it actually produces a real diff instead of a silent no-op. On the ordering problem, I tried a few things but couldn't find a pure HCL way to make terraform destroy the key after the table update, it always wants to destroy the now unreferenced key first. Documented a two step workaround instead (flip the SSE config via CLI first, then apply) and verified it end to end against a live table.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
AlexTo: Added the checkov skip on CKV_AWS_119, same pattern as the WAF module. Ran checkov on a workspace generated from published rc.76 after migrating and it's back to 0 failed checks. Also added suppressRules tips in the docs for AWS_MANAGED, DEFAULT and CKV_AWS_7 on disabled key rotation so people aren't caught off guard by it.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
nx-plugin-for-aws: Confirmed fixed. Retested at 7b89d2d against a real BYO key resource in the same config (`kms_key_arn = aws_kms_key.byo.arn`, `create_kms_key = false`) — plans clean in one shot (`Plan: 17 to add, 0 to change, 0 to destroy`), no `Invalid count`, and applies single-phase with no `-target` needed. Deployed live: the table uses exactly the supplied key ARN, and `terraform state list` shows only 2 KMS keys across 4 tables, so `create_kms_key` correctly suppresses key creation for the AWS_MANAGED / DEFAULT / BYO modes.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
nx-plugin-for-aws: Confirmed fixed. On a workspace generated from published rc.76, checkov goes from `Passed: 25, Failed: 0` before the migration to `Passed: 24, Failed: 0, Skipped: 8` after — CKV_AWS_119 now reports `SKIPPED ... Suppress comment: Encryption is configurable via var.encryption; checkov cannot resolve the conditional server_side_encryption.enabled expression`. Migration is still idempotent (no changes on re-run).

Also verified the CDK-side docs tips work verbatim: a stack using `AWS_MANAGED` + `DEFAULT` + a non-rotating BYO key builds with `Failed checks: 0` using exactly the `suppressRules` calls from the snippet. Thanks for adding the CKV_AWS_7 note too.

And the migration-fidelity nit from my earlier summary is resolved — after lint, the migrated CDK `core/dynamodb.ts` is now byte-identical to a freshly generated one, including the `enableKeyRotation` JSDoc.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
nx-plugin-for-aws: The silent no-op is fixed — the switch now produces a real diff (`~ kms_key_arn = "...key/e087dbd1..." -> "...alias/aws/dynamodb"`, `Plan: 0 to add, 2 to change, 1 to destroy`) instead of the previous `0 changed`. And I verified your documented two-step workaround works end to end on a live table: the CLI `update-table --sse-specification ... KMSMasterKeyId=alias/aws/dynamodb` moves it to the AWS managed key, and the follow-up apply succeeds.

But the explicit alias ARN introduces a **perpetual diff** for `AWS_MANAGED`. The provider reads `server_side_encryption.kms_key_arn` back as `null` for the AWS managed key, so config and state never agree. After a fully successful apply, `terraform plan` still reports `0 to add, 2 to change, 0 to destroy` — I applied three times and it never converged:

```
Apply complete! Resources: 0 added, 2 changed, 0 destroyed.
Plan: 0 to add, 2 to change, 0 to destroy.   # and again, and again
```

```
# terraform state show ...aws_dynamodb_table.table
server_side_encryption {
    enabled     = true
    kms_key_arn = null
}
```

This hits every `AWS_MANAGED` table, not just switched ones — the freshly-created `t_aws_managed` module shows the same `+ kms_key_arn` diff forever. Practical impact: `terraform plan` is never clean, so drift detection is useless and CI gates that assert an empty plan will always fail.

I tried `data.aws_kms_alias.ddb.target_key_arn` (the resolved key ARN rather than the alias ARN) and it has the same problem — the provider still reads back `null`, so that isn't the fix either.

The tension is real and I don't think there's a clean pure-HCL answer: `null` converges but silently no-ops on switches; an explicit ARN fixes switches but never converges. Given you've already documented the two-step workaround for the switch case, **reverting `AWS_MANAGED` to `null` and leaning on that documented procedure may be the better trade** — a permanently dirty plan for all AWS_MANAGED users is arguably worse than a documented manual step for the rarer switch case. `ignore_changes = [server_side_encryption[0].kms_key_arn]` would silence the diff but would also mask genuine drift, so I'd lean away from it.

Either way this is your call on the trade-off — flagging it because I don't think it should ship with plan never converging. Everything else in this PR checks out: all 4 Terraform and all 4 CDK permutations verified live, and the CDK path switches *and* converges cleanly (`no changes` on redeploy), so this is Terraform-specific.

cogwirrel (approved):

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
cogwirrel: I think on balance this is a better approach. Most users should be using the CMK anyway :)

PR #1116, opened 2026-08-24
docs(connection): match the graph builder's agent/gateway constraints to the guards

nx-plugin-for-aws:
### Reason for this change

`CONNECTION_CONSTRAINTS` in `scaffold-catalog.ts` exists for the docs graph builder, which reads it to promise that *a graph it accepts is a graph that scaffolds*. Two of its agent/gateway entries had drifted from the guards the connection generators enforce — in opposite directions, so each broke that promise differently.

**The generators themselves are correct.** Only the catalogue the builder reads was wrong, which is why this is `docs:` rather than `fix:` — no generated code or product behaviour changes.

This is theme (2) from the bug bash reported in #1112. The drift class is only findable by executing both sides: the existing catalogue test checks that a constrained option exists and that its value is in the schema enum, but never that a generator agrees.

**A missing rule.** Both agent → gateway generators require the gateway to serve the `mcp` protocol, but the catalogue declared only the two `auth` rules. The builder validated an agent → `http` gateway graph clean (`issues: []`) and emitted a script whose final `connection` command threw `Gateway 'gw' has protocol='http'. Only MCP-protocol gateways are supported` — *after* the projects had already been created. The bidirectional case was worse, because the builder's own auto-fix sets the gateway to `http` in order to attach the agent, actively steering users into it.

**A rule that does not exist.** The catalogue pinned a gateway's agent target to `auth: iam`, so the builder hard-errored on a Cognito agent behind a gateway (`a: auth must be 'iam' for this connection`) with no auto-fix available. The generator deliberately accepts both, and says so in its own comment: an IAM agent is invoked with the gateway's own role, a Cognito one has the caller's JWT forwarded. The vended `addAgent` backs the implementation, picking `credentials: agent.auth === 'cognito' ? 'jwt-passthrough' : 'gateway-iam'` — so the catalogue was the stale side, and `agentcore-gateway-agent.mdx` already documented Cognito as supported.

### Description of changes

- **`docs(connection)`: catalogue** — adds the `target protocol = mcp` constraint to `AGENT_TO_GATEWAY`, and drops the `target auth = iam` constraint from `GATEWAY_TO_AGENT`. The agent schema's `auth` enum is exactly `iam | cognito` and the generator accepts both, so the constraint is removed rather than rewritten (expressing "one of" would need a shape `ConnectionConstraint` does not have).

  Adding the `mcp` rule makes the two agent↔gateway directions mutually exclusive, which is the real product constraint: **one gateway cannot both front an agent and be called by one.** That is now surfaced while drawing rather than emitted — verified that both the single-edge and bidirectional graphs report an error naming both protocols, and that the auto-fix does not fight it (`mcp` is already the gateway default, so attaching an agent *to* a gateway needs no fix).

- **`docs(connection)`: guides** — the `ts-agent-gateway` / `py-agent-gateway` guides listed only `auth: iam` as the gateway prerequisite, so following either page with a gateway on another protocol failed at the connection step. Both now name the protocol and note that a gateway serves one direction only.

- **`docs: update translations`** — generated locally with `scripts/translate.ts` and committed here, so merging this does not trigger a translation-bot commit and a CI restart.

- **A new test pins both rules by running the generators**, and I confirmed it fails against the current catalogue *before* the change: the 4 generator-behaviour assertions passed while all 4 catalogue assertions failed, which is what establishes the catalogue as the stale side rather than the code. After the change, 8/8 pass.
- **Full plugin suite green: 198 files, 3707 tests.** `docs:test` green: 124 tests. Formatting clean.
- Verified all 16 translated files contain both changes, with code tokens, `<Link>` paths and `:::note` syntax preserved.

Worth flagging for review: dropping the `auth` constraint is the narrow change, but the underlying `ConnectionConstraint` shape cannot express "one of a set", which is why a genuine `iam | cognito` rule has to be modelled as *no* rule. If you would rather it stayed explicit, adding a `oneOf` field is a small change I'm happy to make instead. The finding also suggested a broader integration test that, for every `CONNECTION_CONSTRAINTS` entry, scaffolds the endpoints with a *violating* value and asserts the generator throws — that would catch this class wholesale, but it is a larger change than this, so I have left it out.

### Issue # (if applicable)

N/A — theme (2) of the bug bash reported in #1112.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/connection/scaffold-catalog.ts
+++ b/packages/nx-plugin/src/connection/scaffold-catalog.ts
@@ -339,6 +339,13 @@ const AGENT_TO_GATEWAY: readonly ConnectionConstraint[] = [
     equals: 'iam',
     reason: 'Agent connections require the gateway to authenticate with IAM.',
   },
+  {
+    side: 'target',
+    option: 'protocol',
+    equals: 'mcp',
+    reason:
+      'An agent reaches a gateway as an MCP client, so the gateway must serve the mcp protocol. Note a single gateway cannot both front an agent and be called by one, since fronting requires the http protocol.',
+  },
 ];
 
 const GATEWAY_TO_MCP: readonly ConnectionConstraint[] = [
@@ -364,12 +371,6 @@ const GATEWAY_TO_AGENT: readonly ConnectionConstraint[] = [
     equals: 'http',
     reason:
       'Agent runtime targets can only be attached to an http-protocol gateway.',
-  },
-  {
-    side: 'target',
-    option: 'auth',
-    equals: 'iam',
-    reason: 'The gateway signs requests to its agent targets with IAM.',
   },
 ];
 
--- a/docs/src/content/docs/en/guides/connection/py-agent-gateway.mdx
+++ b/docs/src/content/docs/en/guides/connection/py-agent-gateway.mdx
@@ -21,9 +21,13 @@ The generator wires the agent so it authenticates to the Gateway with IAM SigV4 
 Before using this generator, ensure you have:
 
 1. A Python project with a <Link path="guides/py-agent">Agent</Link> component (`infra: agentcore`)
-2. A <Link path="guides/agentcore-gateway">`agentcore-gateway`</Link> project with `auth: iam`
+2. A <Link path="guides/agentcore-gateway">`agentcore-gateway`</Link> project with `protocol: mcp` and `auth: iam`
 
-The Gateway must use IAM authentication — the agent signs its requests with SigV4 using its own execution role. The generator rejects Cognito-authenticated gateways.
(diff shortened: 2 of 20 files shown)


packages/nx-plugin/src/connection/scaffold-catalog.ts
```diff
@@ -357,6 +364,10 @@ const GATEWAY_TO_MCP: readonly ConnectionConstraint[] = [
   },
 ];
 
+// The agent's `auth` is deliberately unconstrained: a gateway fronts an IAM
+// agent with its own role and a Cognito agent by forwarding the caller's JWT,
+// and the agent schema offers no third value. Pinning it to `iam` would rule out
+// the Cognito topology the generator and the vended `addAgent` both support.
```
cogwirrel: remove this comment

nx-plugin-for-aws: Removed in 5a0c398.

The rule it described stays covered by `catalog-guard-parity.spec.ts`, which asserts the generator actually connects a Cognito agent behind a gateway — so the reason the constraint is absent is pinned by a test rather than a comment.

cogwirrel (approved):

PR #1121, opened 2026-08-26
fix(ts#dynamodb,ts#rdb): add Terraform destroy protection to data-bearing resources

nx-plugin-for-aws:
### Reason for this change

The CDK and Terraform paths had materially different data-loss protection, and Terraform's was weaker by one whole layer.

CDK guards data-bearing resources with **two independent** guards:

| | CDK | Terraform (before) |
|---|---|---|
| DynamoDB table | `deletionProtection = true` **+** `RemovalPolicy.RETAIN` | `deletion_protection_enabled = true` only |
| Aurora cluster | `deletionProtection = true` **+** `RemovalPolicy.RETAIN` | `deletion_protection = true` only |

`grep -rn "prevent_destroy" packages/nx-plugin/src/` returned zero matches. The service-side defaults were correctly at `true` and matched CDK — that part was fine. The gap was that a user who flips that flag to `false` — a one-line, plan-visible, entirely ordinary change ("disable deletion protection so I can rename the table") — destroys the data **in the same apply**. On CDK they would still have to defeat `DeletionPolicy: Retain` separately, in a different place.

### Description of changes

**1. A second, independent guard on the two data-bearing resources.**

`lifecycle { prevent_destroy = true }` on `aws_dynamodb_table.table` and `aws_rds_cluster.database`. The service-side flag is enforced by DynamoDB/RDS; this is enforced by Terraform, so clearing the variable alone can no longer destroy the data — mirroring the CDK posture.

**Design decision — why `prevent_destroy`, and why it is a literal.** `prevent_destroy` **cannot be driven by a variable**; Terraform requires a literal. That constraint shaped the design:

- Scoped to the **two resources that actually hold data**, not all 23 KMS keys or every resource in the modules. A blanket application would make routine key rotation and instance replacement fail plans for no data-safety benefit.
- Vended as a literal with a comment naming the teardown action, and documented in both generator guides. This follows the precedent already in the repo for an "edit this to tear down" flow (the e2e string-patches Cognito `deletion_protection`).
- The alternative — docs/KMS alignment only, no `prevent_destroy` — was rejected because it leaves the actual exposure open: the whole point is that one variable flip should not be sufficient to delete data.

**2. `deletion_window_in_days = 30` on both data resources' KMS keys.**

This is the correction of a premise worth flagging explicitly, because it runs **opposite** to "make the table's CMK consistent with its 22 siblings":

- CDK's `Key` default `pendingWindow` is **30 days** (`aws-cdk-lib` typing: `@default - 30 days`).
- The AWS provider's `aws_kms_key.deletion_window_in_days` default is also **30**: *"If you do not specify a value, it defaults to `30`."*

So the DynamoDB and Aurora keys, which omitted the attribute, were already the **safest** keys and already at CDK parity. The 5 siblings pinning `= 7` are the ones diverging from CDK. Setting the table's key to `7` "for consistency" would have **shrunk** the recovery window 30→7 and moved it *further* from CDK — the opposite of the stated goal. A PITR restore of a CMK-encrypted table requires the original key, so the window matters as much as the backup. I therefore pinned `30` **explicitly** on the two data keys, which is a no-op in behaviour but makes the intent load-bearing and immune to a future provider default change. The 22 unrelated `= 7` keys (buckets, logs, APIs) are left alone as out of scope.

**3. A migration**, per CONTRIBUTING ("assume a migration is needed"), since these files are vended `KeepExisting` and an upgraded workspace would otherwise keep the weaker protection. Deterministic, GritQL-based, pattern-matches the vended shape before writing, reports diverged files via `nextSteps`, preserves a user's own `prevent_destroy = false`, and is idempotent.

**4. Docs** (English only) for both guides: the two-layer model, and the full teardown path including the `prevent_destroy`-is-a-literal caveat and a caution that the core module is shared by every table/database in the workspace.

**Teardown still works — the critical constraint.** `terraform-deploy` and `terraform-deploy-rdb` both destroy these exact resources, so leaking CI resources would be worse than the bug. `e2e/src/smoke-tests/terraform-deploy.spec.ts` now flips the literal to `false`, alongside the existing Cognito patch and immediately before `bootstrap`/`apply`:

```ts
for (const relativeTfPath of ['core/dynamodb/dynamodb.tf', 'core/rdb/aurora/aurora.tf']) {
  // ... .replace(/prevent_destroy\s*=\s*true/g, 'prevent_destroy = false')
}
```

Why this is sufficient:
- The server-side flags are **already** disabled by the wiring templates (`deletion_protection_enabled = false` in `main.tf.template`; `deletion_protection = false` + `skip_final_snapshot = true` in `main-rdb.tf.template`), so the lifecycle literal was the only remaining blocker.
- `prevent_destroy = false` is a valid literal, so no block removal or HCL surgery is needed.
- Run against the **real generated workspace**, the patch matched both files exactly — `remaining true=0, false=1` in each — and the patched config still plans clean.
- **`bootstrap-destroy` is unaffected**: `infra/bootstrap/main.tf` defines only its own `aws_s3_bucket.terraform_state` (+ versioning/encryption/policy) and never consumes the vended data modules. `grep -rn prevent_destroy infra/bootstrap/` → no matches.

**Proof the guard actually works.** Real `terraform apply` then `terraform destroy`:

```
Plan: 0 to add, 0 to change, 1 to destroy.
Error: Instance cannot be destroyed
Resource null_resource.table has lifecycle.prevent_destroy set, but the plan
calls for this resource to be destroyed.
```

`0 destroyed`. After applying the e2e's exact substitution: `Destroy complete! Resources: 1 destroyed.` Both directions verified.

**Real generated workspace + credential-free plan.** Compiled the plugin, generated a Terraform workspace with `ts#dynamodb` (electrodb) and `ts#rdb` (Aurora postgres), and ran `terraform init -backend=false` plus `terraform test` with the mocked-provider pattern from `e2e/src/smoke-tests/terraform-plan-test.ts`:

```
Success! 1 passed, 0 failed.
```

Generated HCL:

```hcl
resource "aws_kms_key" "table" {
  count = local.create_table_key ? 1 : 0

  description         = "KMS key for DynamoDB table ${var.name}"
  enable_key_rotation = var.enable_key_rotation

  # A point-in-time restore of a CMK-encrypted table requires this key, so keep
  # the maximum window in which a scheduled deletion can still be cancelled.
  deletion_window_in_days = 30
```

```hcl
resource "aws_dynamodb_table" "table" {
  # ...
  # Second, independent layer of protection: `deletion_protection_enabled` is
  # enforced by DynamoDB, this is enforced by Terraform, so clearing the
  # variable alone cannot destroy the table. Remove this block to tear down.
  lifecycle {
    prevent_destroy = true
  }
}
```

```hcl
resource "aws_rds_cluster" "database" {
  # ...
  # Second, independent layer of protection: `deletion_protection` is enforced
  # by RDS, this is enforced by Terraform, so clearing the variable alone cannot
  # destroy the cluster. Remove this block to tear down.
  lifecycle {
    prevent_destroy = true
  }
}
```

**Tests.** Added generator unit tests asserting both guards appear in the generated `.tf` (`ts#dynamodb`, `ts#rdb`), and 6 migration tests (applies to each vended shape, no-op on a workspace without the modules, skips + reports a customised file, preserves an existing `prevent_destroy = false`, idempotent with exactly one guard per file).

- `pnpm nx run @aws/nx-plugin:test` — **198 files / 3709 tests pass**. The only snapshot changes were the 4 additive diffs for these templates; each was inspected before updating.
- `pnpm nx run-many --target build --all` — pass.
- `pnpm nx run-many --target lint --all --configuration=fix` — pass, no changes to any file in this PR.

### Issue #

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
+++ b/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
@@ -95,8 +95,6 @@ resource "aws_kms_key" "table" {
   description         = "KMS key for DynamoDB table ${var.name}"
   enable_key_rotation = var.enable_key_rotation
 
-  # A point-in-time restore of a CMK-encrypted table requires this key, so keep
-  # the maximum window in which a scheduled deletion can still be cancelled.
   deletion_window_in_days = 30
 
   policy = jsonencode({
@@ -176,9 +174,6 @@ resource "aws_dynamodb_table" "table" {
     kms_key_arn = local.table_kms_key_arn
   }
 
-  # Second, independent layer of protection: `deletion_protection_enabled` is
-  # enforced by DynamoDB, this is enforced by Terraform, so clearing the
-  # variable alone cannot destroy the table. Remove this block to tear down.
   lifecycle {
     prevent_destroy = true
   }
--- a/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
+++ b/packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
@@ -95,6 +95,10 @@ resource "aws_kms_key" "table" {
   description         = "KMS key for DynamoDB table ${var.name}"
   enable_key_rotation = var.enable_key_rotation
 
+  # A point-in-time restore of a CMK-encrypted table requires this key, so keep
+  # the maximum window in which a scheduled deletion can still be cancelled.
+  deletion_window_in_days = 30
+
   policy = jsonencode({
     Version = "2012-10-17"
     Statement = [
@@ -170,6 +174,13 @@ resource "aws_dynamodb_table" "table" {
   server_side_encryption {
     enabled     = var.encryption != "DEFAULT"
     kms_key_arn = local.table_kms_key_arn
+  }
+
+  # Second, independent layer of protection: `deletion_protection_enabled` is
+  # enforced by DynamoDB, this is enforced by Terraform, so clearing the
+  # variable alone cannot destroy the table. Remove this block to tear down.
+  lifecycle {
+    prevent_destroy = true
   }
 }
 
--- a/docs/src/content/docs/en/snippets/dynamodb/deploying-table.mdx
+++ b/docs/src/content/docs/en/snippets/dynamodb/deploying-table.mdx
@@ -46,7 +46,7 @@ This provisions a DynamoDB table with:
 - On-demand (`PAY_PER_REQUEST`) billing
 - Customer-managed KMS encryption with automatic key rotation
 - Point-in-time recovery enabled
-- Deletion protection enabled
+- Deletion protection enabled, plus a `prevent_destroy` lifecycle guard
 - Table name registered in Runtime Config under the `dynamodb` namespace in AWS AppConfig
 
 The `core/runtime-config/appconfig` module exposes the `dynamodb` namespace by default, so the table name is deployed without further configuration. If you pass `namespaces` to that module explicitly, keep `dynamodb` in the list — otherwise no configuration profile is created for it and the generated table client cannot resolve the table name.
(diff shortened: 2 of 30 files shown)


packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
```diff
@@ -95,6 +95,10 @@ resource "aws_kms_key" "table" {
   description         = "KMS key for DynamoDB table ${var.name}"
   enable_key_rotation = var.enable_key_rotation
 
+  # A point-in-time restore of a CMK-encrypted table requires this key, so keep
+  # the maximum window in which a scheduled deletion can still be cancelled.
```
cogwirrel: remove comment

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
```diff
@@ -171,6 +175,13 @@ resource "aws_dynamodb_table" "table" {
     enabled     = var.encryption != "DEFAULT"
     kms_key_arn = local.table_kms_key_arn
   }
+
+  # Second, independent layer of protection: `deletion_protection_enabled` is
+  # enforced by DynamoDB, this is enforced by Terraform, so clearing the
+  # variable alone cannot destroy the table. Remove this block to tear down.
```
cogwirrel: remove comment

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
nx-plugin-for-aws: Removed. The bare `deletion_window_in_days = 30` is left in place — it's self-documenting, and the rationale (a PITR restore of a CMK-encrypted table needs the original key, so this keeps the maximum cancellable window and matches the CDK `Key` default) now lives only in the PR description.

I removed the equivalent comment on the Aurora key in `aurora.tf.template` in the same push, since it was the same pattern.

packages/nx-plugin/src/utils/dynamodb-constructs/files/terraform/core/dynamodb/dynamodb.tf.template
nx-plugin-for-aws: Removed. The `lifecycle { prevent_destroy = true }` block itself is unchanged — the guard is the point of the PR, and the block reads clearly without prose around it.

The teardown path (remove this block; it can't be a variable, and the module is shared by every table in the workspace) is documented in the `ts#dynamodb` / `py#dynamodb` guides rather than in the vended file. Same comment removed from the Aurora cluster in `aurora.tf.template` in this push. Snapshots updated; full suite green (199 files / 3715 tests), and the generated config still passes the credential-free mocked `terraform test`.

cogwirrel (dismissed):

cogwirrel (approved):

PR #1123, opened 2026-08-26
fix(agentcore-gateway): declare ejs at the workspace root for terraform Cedar rendering

nx-plugin-for-aws:
### Reason for this change

Under `iac=terraform`, an AgentCore Gateway with Cedar policies fails at `terraform apply` because the Cedar rendering script cannot resolve its `ejs` dependency. `cedarPolicy` defaults to `true` for `mcp`-protocol gateways, so this is the default path, not an edge case.

The mechanism:

- `render-cedar.cjs` does `require('ejs')`.
- It is invoked by a `data "external"` block as `program = ["node", "${path.module}/render-cedar.cjs"]`, so it runs from `packages/common/terraform/src/app/gateways/<name>/`.
- That directory belongs to the shared **terraform** project, which has **no `package.json` at all** — only `project.json` and `src/`. Meanwhile `ejs` was declared as a dev dependency of each individual **gateway** package.
- Under pnpm's isolated `node_modules`, Node's resolution from the script's directory walks up to the workspace root and finds no `ejs`. It happens to work under npm, which hoists to the root.

Reproduced in a real generated Terraform workspace (`pnpm install`, then the script invoked exactly as Terraform does):

```
$ cd packages/common/terraform/src/app/gateways/my-gateway
$ echo '{"template":".../policies/permit-all.cedar","gatewayArn":"arn:aws:bedrock-agentcore:us-west-2:123456789012:gateway/MyGateway-1a2b3c4d","accountId":"123456789012"}' | node ./render-cedar.cjs
Error: Cannot find module 'ejs'
Require stack:
- /tmp/gw-before/packages/common/terraform/src/app/gateways/my-gateway/render-cedar.cjs
    at Function._resolveFilename (node:internal/modules/cjs/loader:1430:15)
  code: 'MODULE_NOT_FOUND',
```

Two reasons this went unnoticed, both worth recording:

- **No plan could catch it.** An `external` data source's `result` is deferred to apply — under plan it sits in `after_unknown` — so Terraform never executes `program` during `terraform plan`, nor during the mocked `terraform test` the plan lane runs.
- **The deploy lane was green anyway.** It does apply a Cedar-enabled gateway, but `terraform` runs under `nx apply`, and pnpm's bin shim for `nx` exports a `NODE_PATH` covering pnpm's hoisted virtual store. Nx itself depends on `ejs`, so the script silently resolved nx's `ejs@5.0.1` rather than the catalog's `6.0.1`:

  ```
  NODE_PATH=.../node_modules/.pnpm/nx@23.1.1/node_modules:.../node_modules/.pnpm/node_modules
  RESOLVED .../node_modules/.pnpm/ejs@5.0.1/node_modules/ejs/lib/cjs/ejs.js
  ```

  With `NODE_PATH` unset — a user running `terraform apply` directly — the same script fails.

### Description of changes

**Declare `ejs` at the workspace root on the terraform branch.** Rendering still uses `ejs` exactly as before; only where the dependency is declared changes. The workspace root is the nearest manifest above the script that Node can resolve against, since the shared terraform project has none. It follows the existing `root: true` pattern (`ts#infra` declares `aws-cdk`/`esbuild`/`tsx` that way):

```ts
{ name: 'ejs', when: generatedTerraform, dev: true, root: true },
```

Gated on the terraform branch, so a CDK-only workspace gains nothing it doesn't need — CDK already resolves `ejs` from `packages/common/constructs/package.json`, and that path is untouched. Only added when `cedarPolicy` is on, since a gateway without policies vends no render script.

**Reconciled `AGENT_CORE_CONSTRUCTS_DEPENDENCIES`.** It gated `ejs` / `@types/ejs` on `generatedInfrastructure` (either provider) while only the CDK branch installed them, so the declaration claimed ownership of packages the Terraform branch never placed. Each entry is now gated on the branch that actually installs it — `generatedCdk` for the three that go into shared constructs, `generatedTerraform` for the new root entry.

For newly generated projects the generator no longer adds `ejs` / `@types/ejs` to the gateway project's manifest: nothing there imports either, and that manifest was never on the script's resolution path.

**Migration.** `latest/terraform-gateway-cedar-render-root-ejs` adds the root `ejs` to existing workspaces that have a Terraform gateway with a render script. It is **purely additive** — it never removes anything, since it cannot tell whether a user has started importing `ejs` in a gateway project themselves, and a redundant duplicate declaration is harmless where a removed one they depend on would not be. A CDK gateway is left alone.

No documentation change: the guide already describes the policies as EJS templates, which stays accurate.

**End-to-end in a real generated pnpm workspace.** Compiled the plugin, generated a Terraform workspace with a Cedar-enabled gateway (the default), ran a real `pnpm install`, and invoked the script the way Terraform's `external` data source does — with `NODE_PATH` cleared, so the nx-shim accident above can't mask the result. The script still does `require('ejs')`; the root manifest now carries the dependency:

```
=== ROOT manifest ===
"devDependencies": { ..., "ejs": "catalog:", "make-dir-cli": "catalog:", "tsx": "catalog:" }
=== GATEWAY manifest (no longer carries ejs) ===
"devDependencies": { "@types/express": "catalog:" }

$ ls -la node_modules/ejs
node_modules/ejs -> .pnpm/ejs@6.0.1/node_modules/ejs

$ cd packages/common/terraform/src/app/gateways/my-gateway
$ grep -n "require(" render-cedar.cjs
4:const ejs = require('ejs');
5:const fs = require('fs');
$ echo '{"template":".../policies/permit-all.cedar","gatewayArn":"arn:aws:bedrock-agentcore:us-west-2:123456789012:gateway/MyGateway-1a2b3c4d","accountId":"123456789012"}' | node ./render-cedar.cjs
{"rendered":"// Default policy: permit any IAM caller from the deploying AWS account.\n// The PolicyEngine runs in ENFORCE mode (default-deny) - deleting this file\n// without a replacement permit denies every call. See policies/README.md.\n\npermit (\n  principal is AgentCore::IamEntity,\n  action,\n  resource == AgentCore::Gateway::\"arn:aws:bedrock-agentcore:us-west-2:123456789012:gateway/MyGateway-1a2b3c4d\"\n) when {\n  principal.id like \"arn:aws:*::123456789012:*\"\n};\n"}
exit status: 0
```

**Unit tests.** A `render-cedar.cjs` block in `agentcore-gateway/generator.spec.ts` asserts `ejs` is declared at the root and not in the gateway project, then executes the generated script from a module directory nested under a root manifest that is the only place `ejs` is reachable — mirroring the generated layout — and asserts it renders with every placeholder substituted. The migration has specs for the additive behaviour, the CDK no-op and idempotency.

**Package manager coverage.** A root declaration is the one place every package manager guarantees is on the script's resolution path, because Node walks `node_modules` up the tree from the script and the workspace root is its last stop:

- **pnpm** — the failing case. Isolated `node_modules` means a dependency declared only in a sibling package is unreachable; declared at the root it is symlinked into the root `node_modules` (verified above), which is on the walk.
- **npm** — hoists to the root `node_modules`, and a root-declared dependency lands there directly, so resolution is now guaranteed rather than an incidental side effect of hoisting.
- **yarn** — Nx writes `nodeLinker: node-modules` for berry, and classic uses node_modules too, so a root-declared dependency sits in the root `node_modules` as under npm.
- **bun** — installs workspace dependencies into the root `node_modules`, so the root declaration is on the walk.

**Gate.** `pnpm nx run @aws/nx-plugin:test` — 199 files, 3714 tests pass, no snapshot changes. `pnpm nx run-many --target build --all` — success for 5 projects. `pnpm nx run-many --target lint --all --configuration=fix` — success, no changes to any file in this PR.

**Not run locally:** `terraform` is not installed in this environment, so I did not execute the `terraform` or `terraform-deploy` lanes locally, and did not deploy to AWS. CI covers both.

### Issue # (if applicable)

N/A

### Checklist

----


cogwirrel (changes requested):
No, we should use ejs like before. add this as a root dependency so it resolves. the fix should be simpler, just adding ejs as a root dependency?

nx-plugin-for-aws:
Thanks — switched to the root-dependency approach as you asked.

`render-cedar.cjs` is back to `require('ejs')` exactly as before (reverted verbatim), and the fix is now just the declaration:

```ts
// The Cedar render script resolves this from the workspace root.
{ name: 'ejs', when: generatedTerraform, dev: true, root: true },
```

Gated on the terraform branch so a CDK-only workspace doesn't gain a dependency it never needs — CDK keeps resolving `ejs` from `packages/common/constructs/package.json` and that path is untouched. Only added when `cedarPolicy` is on, since a gateway without policies vends no render script.

Two related things I kept:

- **Reconciled `AGENT_CORE_CONSTRUCTS_DEPENDENCIES`.** It gated `ejs`/`@types/ejs` on `generatedInfrastructure` (either provider) while only the CDK branch installed them, so it claimed ownership of packages Terraform never placed. Each entry is now gated on the branch that actually installs it. The gateway project's own `ejs`/`@types/ejs` are dropped — nothing there imports either, and that manifest was never on the script's resolution path.
- **The e2e coverage**, since nothing exercised this path, which is why it survived. Verified end to end in a real generated pnpm workspace: root manifest now carries `ejs`, `node_modules/ejs -> .pnpm/ejs@6.0.1/...`, and the script renders (`exit status: 0`) where it previously failed with `Cannot find module 'ejs'`.

Worth flagging one thing I found while digging: the terraform-deploy lane *was* already applying a Cedar gateway, but passed because `terraform` runs under `nx apply` and pnpm's `nx` bin shim exports a `NODE_PATH` covering the hoisted virtual store — and nx itself depends on `ejs`, so the script was silently resolving nx's `ejs@5.0.1` rather than the catalog's `6.0.1`. The tests clear `NODE_PATH` so that can't mask it again.

The plan lane's new guard has already run green on this branch:
```
Rendered my-gateway/permit-all.cedar: status=0 stderr=
Rendered parent-gateway/permit-all.cedar: status=0 stderr=
```

Title and description updated to match the simpler fix.

e2e/src/files/terraform-deploy/main.tf.template
```diff
@@ -794,6 +794,18 @@ output "my_gateway_url" {
   value = module.my_gateway.gateway_url
 }
 
+output "my_gateway_arn" {
+  value = module.my_gateway.gateway_arn
+}
+
+# The Cedar policies as rendered at apply time. An `external` data source's
+# result is deferred to apply, so surfacing it as an output is what proves the
```
cogwirrel: remove this comment

e2e/src/smoke-tests/terraform-deploy.spec.ts
```diff
@@ -340,6 +340,24 @@ const runTerraformDeployVariant = (config: TerraformDeployVariant) => {
...
+            outputs.my_gateway_rendered_policies,
+          ) as Record<string, string>;
+          console.log('my_gateway rendered policies:', renderedPolicies);
+          expect(Object.keys(renderedPolicies)).toContain('permit-all.cedar');
+          for (const [file, policy] of Object.entries(renderedPolicies)) {
+            expect(policy, file).toContain('permit');
+            expect(policy, file).toContain(outputs.my_gateway_arn);
+            // An unsubstituted placeholder would reach the Cedar validator.
+            expect(policy, file).not.toContain('<%');
+          }
```
cogwirrel: remove this

e2e/src/smoke-tests/terraform-plan-test.ts
```diff
@@ -2,10 +2,77 @@
...
+ * resolution unproven by the whole Terraform lane: it runs from the shared
+ * terraform project, which has no `package.json`, so under pnpm's isolated
+ * layout the dependency has to come from the workspace root.
+ *
+ * `NODE_PATH` is cleared deliberately. Nx's own bin shim exports one covering
+ * pnpm's hoisted virtual store, and Nx happens to depend on `ejs` itself, so a
+ * script spawned beneath `nx apply` could resolve a copy the workspace never
+ * declared — masking a missing declaration.
+ */
+const runCedarRenderScripts = (projectRoot: string) => {
```
cogwirrel: remove this

packages/nx-plugin/src/migrations/latest/terraform-gateway-cedar-render-root-ejs/migration.ts
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * The redundant copy in each gateway project's manifest is dropped, since
+ * nothing in a gateway project imports `ejs`. A version the user pinned
+ * themselves (a non-`catalog:` specifier) is left alone.
+ */
+
+const RENDER_SCRIPT = 'render-cedar.cjs';
+
+/** Drop `ejs` / `@types/ejs` from a gateway project's manifest. */
+const removeUnusedEjs = (tree: Tree, manifestPath: string): void => {
```
cogwirrel: remove this - we don't know if ejs is used or not, might break users workspaces

docs/src/content/docs/es/guides/agentcore-gateway.mdx
```diff
@@ -195,7 +195,7 @@ Pasa el nombre del rol como una variable de plantilla (ver <a href="#template-va
 
 ### Variables de plantilla
 
-Las políticas son plantillas [EJS](https://ejs.co/) renderizadas en tiempo de synth/plan para que permanezcan portables entre cuentas y redespliegues de Gateway:
+Las políticas son plantillas renderizadas en tiempo de despliegue para que permanezcan portables entre cuentas y redespliegues de Gateway:
```
cogwirrel: undo translation changes

e2e/src/files/terraform-deploy/main.tf.template
nx-plugin-for-aws: Removed.

e2e/src/smoke-tests/terraform-deploy.spec.ts
nx-plugin-for-aws: Removed — cut down to a single line naming what the block asserts.

e2e/src/smoke-tests/terraform-plan-test.ts
nx-plugin-for-aws: Removed. I also swept the rest of the diff for the same pattern and trimmed the narrative comments in `agent-core-constructs.ts`, `shared-constructs-constants.ts`, `generator.spec.ts` and the migration — sorry for making you ask three times.

packages/nx-plugin/src/migrations/latest/terraform-gateway-cedar-render-root-ejs/migration.ts
nx-plugin-for-aws: Agreed, and removed — you're right that this could break a workspace. The migration is now purely additive: it only adds the root `ejs` and never removes anything, since it can't know whether someone has started importing `ejs` in that project themselves. A redundant duplicate declaration is harmless; removing one they depend on isn't. Dropped the two tests that asserted removal and replaced them with one asserting the gateway manifest's dependencies are left untouched.

(The generator still omits `ejs`/`@types/ejs` for newly generated gateway projects, which is safe — there's nothing there to break yet.)

docs/src/content/docs/es/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Undone — reverted all eight translated guides (es, fr, it, jp, ko, pt, vi, zh) to `origin/main`. These came from the translate workflow picking up an English edit I'd since reverted; I should have caught the leftovers.

The branch now ships no documentation change at all: the guide already describes the policies as EJS templates, which stays accurate since the fix keeps `ejs`. `git diff origin/main -- docs/` is empty for these files.

e2e/src/smoke-tests/terraform-deploy.spec.ts
cogwirrel: No this whole check

e2e/src/smoke-tests/terraform-plan-test.ts
cogwirrel: No this whole check

e2e/src/smoke-tests/terraform-deploy.spec.ts
nx-plugin-for-aws: Sorry — misread that as the comment. Removed the whole check now: the `renderedPolicies` parse, the `console.log` and all the assertions are gone, and `terraform-deploy.spec.ts` is byte-identical to `main` again.

Also removed what fed it from `main.tf.template` — both the `my_gateway_rendered_policies` and `my_gateway_arn` outputs (that's what your `main.tf.template` comment was pointing at) — plus the `rendered_policies` output I'd added to the vended gateway `.tf` template. That file is back to untouched too.

e2e/src/smoke-tests/terraform-plan-test.ts
nx-plugin-for-aws: Same — removed the whole `runCedarRenderScripts` helper, its call site, and the `spawnSync`/`existsSync`/`readdirSync` imports it needed. `terraform-plan-test.ts` is byte-identical to `main` again.

The PR is now just the root `ejs` declaration as you asked: `git diff` against the merge base is 7 files — the declaration in `agent-core-constructs.ts`, the `generatedCdk` predicate, the generator no longer adding `ejs` to gateway projects, the additive migration, and specs. Happy to raise the apply-time coverage gap as a separate issue if useful, but I've dropped it from here.

cogwirrel (approved):

PR #1124, opened 2026-08-26
feat(ts#api): support the isolated integration pattern for terraform APIs

nx-plugin-for-aws:
### Reason for this change

Generated APIs default to the `isolated` integration pattern (one Lambda function per operation), but Terraform only ever implemented the `shared` "router" pattern — a single Lambda behind a catch-all `{proxy+}` route. Choosing `--iac=terraform` therefore silently gave you the shared pattern regardless of `--integrationPattern`, and the docs told users that per-operation integrations required hand-editing the generated module.

This closes that gap so both patterns behave consistently across CDK and Terraform.

### Description of changes

The generated Terraform API modules (HTTP and REST) now support both patterns. For `isolated`, the module iterates an operations metadata file and defines a Lambda function, IAM role and policies, integration, route/method, invoke permission and log group per operation via `for_each`.

**Operations metadata.** The module reads `packages/common/terraform/src/generated/<api>/operations.json`, written as part of the API project's build:

- **tRPC** has no OpenAPI spec, so a vended `scripts/generate-operations.ts` walks the router — mirroring `routerToOperations` in the CDK construct — and writes the JSON.
- **FastAPI / Smithy** reuse their OpenAPI spec through a new hidden `open-api#operations-metadata` generator, alongside the existing `open-api#ts-metadata` used by CDK.

The generated file is `.gitignore`d by default, and the shared Terraform project's `build` depends on the producing target so the file exists before plan.

**Details worth calling out:**

- **Resource naming.** Operation names are slugified and suffixed with an 8-char digest. tRPC uses dot notation for nested routers (`nested.deepEcho`) and OpenAPI produces tag-qualified names for duplicated operation ids (`items.echo`); neither is valid in AWS resource names, and names can exceed the 64-character limit, so the digest keeps truncated names unique.
- **REST resource tree.** REST APIs need an `aws_api_gateway_resource` per path segment. Terraform cannot express recursion, so the module declares one resource per depth and chains each level to its parent. Paths deeper than the supported depth fail the plan with a message naming the offending paths.
- **Least-privilege permissions.** Each function's invoke permission is scoped to its own route (e.g. `.../GET/echo`) rather than a wildcard, with path parameters becoming wildcards. Verified against a real deployment.
- **`terraform validate` before the first build.** `file()` errors on a missing path, so reading the operations is guarded with `fileexists` and a `terraform_data` precondition surfaces a genuinely missing (or empty) file at plan time with an actionable message.
- **Outputs.** Under `isolated` the Lambda/IAM/integration/log-group outputs become maps keyed by operation name (`lambda_function_names`, `lambda_execution_role_names`, `integration_ids`, …). The `shared` pattern keeps its existing singular outputs, so it is unchanged.

The `shared` pattern's generated Terraform is byte-for-byte unchanged; all new behaviour is behind the `isolated` branch.

**Unit tests** — full suite passes (3760 tests / 206 files). Added:
- `open-api#operations-metadata` generator tests covering path/method extraction, stable sort ordering, and tag-qualified disambiguation.
- tRPC generator tests asserting the isolated module iterates operations, drops the proxy catch-all, exposes per-operation outputs and wires the `operations` target — plus a test that `shared` still produces the single-router module and no operations target.
- Snapshots updated: Terraform snapshots now reflect the isolated default.

**Real AWS deployment.** Generated a Terraform workspace with tRPC (HTTP + REST, including a nested router and a mutation) and FastAPI, then deployed for real:

- 6 distinct Lambda functions created, one per operation, correctly named per operation.
- All 5 operations invoked successfully with SigV4, returning correct payloads (tRPC HTTP, tRPC REST, nested-router procedures, FastAPI).
- Confirmed genuine isolation: only the invoked operations' log groups had log streams — `createItem`, which was never called, had none.
- Confirmed permission scoping: the `echo` function's resource policy is limited to `arn:aws:execute-api:...:<api>/*/GET/echo`.
- Torn down afterwards — all 126 resources destroyed, verified no Lambdas or APIs left behind.

**Plan-time validation.** `terraform test` with mocked providers (the same credential-free approach as `terraform-plan-test.ts`) passes with assertions on the per-operation Lambda counts. This caught a real plan-time bug during development: gating the additional-policies `for_each` on an apply-time value, now fixed.

**E2E coverage.** The existing terraform smoke and deploy tests now exercise `isolated` by default. Added `my-api-shared` / `my-api-shared-http` projects to the generator matrix and deploy wiring so the `shared` pattern stays covered and both are invoked.

Also verified `nx run-many --target build --all`, lint, and the docs site build (1495 pages).

### Note on Smithy

Smithy's operations metadata generation is verified (its `operations.json` generates correctly from the OpenAPI spec, and its module renders and passes HCL validation), but I was not able to complete a real Smithy deployment: my manually-created test workspace defaulted to CommonJS while Smithy's generated code and its ESM-only `@middy/core@7` dependency require ESM. That is an artefact of the test workspace rather than these changes — the e2e suite creates ESM workspaces via Verdaccio and will cover the Smithy path.

### Checklist

----


docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -11,11 +10,12 @@ The REST/HTTP API CDK constructs are configured to provide a type-safe interface
...
+Terraform modules support both integration patterns:
 
-For explicit per-operation integrations with Terraform, you would need to manually create individual Lambda functions and API Gateway routes. See the [Explicit Integrations](#explicit-integrations) section for examples.
-:::
+- `isolated` creates one Lambda function per operation. This is the default for generated APIs.
+- `shared` creates a single router Lambda which serves every operation.
+
+With the `isolated` pattern, the module reads a generated `operations.json` describing your API's operations, and defines a Lambda function, integration, route and log group for each one. Since Terraform has no equivalent of the type system the CDK constructs use, per-operation resources are addressed by operation name at apply time rather than being type-checked.
 </Fragment>
 </Infrastructure>
```
cogwirrel: Remove this entire Infrastructure block

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -302,315 +392,16 @@ new MyApi(this, 'MyApi', {
...
-For explicit per-operation integrations with Terraform, you should modify the generated app-specific module to replace the default proxy integration with specific integrations for each operation.
+With the `isolated` pattern the generated module already defines an integration per operation, so there is nothing to replace — each operation has its own `aws_lambda_function`, integration, route (or method) and permission, all keyed by operation name.
 
-Edit `packages/common/terraform/src/app/apis/my-api/my-api.tf`:
+If you want a specific operation to use a different kind of integration, see [Overriding Integrations](#overriding-integrations) above.
 
-1. **Remove the default proxy routes** (e.g., `resource "aws_apigatewayv2_route" "proxy_routes"`)
-2. **Replace the single Lambda function** with individual functions for each operation
-3. **Create specific integrations and routes** for each operation, reusing the same ZIP bundle:
+If you generated the API with the `shared` pattern and now want per-operation integrations, re-run the API generator with `--integrationPattern=isolated`, deleting the existing module directory so it is regenerated:
```
cogwirrel: This section should just say replace the `foreach` for `isolated` pattern with explicit instantiations of lambda functions, integrations and permissions. Note that unlike CDK there is no type-safety in terraform that will give a type error if you add a new operation in your application code but forget to declare it in your infra.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -643,20 +434,29 @@ export class MyApi<...> extends ... {
...
+- `shared` creates a single router Lambda and a catch-all proxy route serving every operation.
 
-```hcl
-# Default router pattern - single Lambda function for all operations
-module "my_api" {
-  source = "../../common/terraform/src/app/apis/my-api"
+`isolated` gives you finer-grained permissions and configuration per operation. `shared` reduces Lambda and API Gateway integration sprawl.
 
-  asset_bucket_name = module.asset_bucket.bucket_name
+The pattern is baked into the generated module rather than being a variable, since the two produce different resources. To change it, re-run the API generator with the other `--integrationPattern` and delete the previously generated module so it is rewritten.
```
cogwirrel: change to the following (filling in the RunGenerator component appropriately):

````
Unlike CDK, the integration pattern is baked into the generated module. To change the integration pattern:

<Steps>
  1. Delete the previously generated API module in `packages/common/terraform/src/app/apis`
  2. Re-run the API generator with the other integration pattern:
      <RunGenerator .... />
</Steps>
````

File: docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx [hunk 4]
```diff
@@ -643,20 +434,29 @@ export class MyApi<...> extends ... {
...
+
+```hcl
+# packages/common/terraform/src/app/apis/my-api/my-api.tf
+locals {
+  operations_file = "${path.module}/../../../generated/my-api/operations.json"
+  operations      = fileexists(local.operations_file) ? jsondecode(file(local.operations_file)) : {}
 }
 ```
+
+That file is written by the API project's build — a script walks the tRPC router, or the OpenAPI specification is read for FastAPI and Smithy APIs — so adding an operation to your API adds a Lambda function and route on the next deploy. It is `.gitignore`d by default; remove the entry if you prefer to check it in.
```
cogwirrel: change to:

This file is generated from your API, so you do not need to edit this by hand. Adding an operation to your API application code adds the route and lambda function on the next deploy. It is `.gitignore`d by default; remove the entry if you prefer to check it in.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -643,20 +434,29 @@ export class MyApi<...> extends ... {
...
+
+:::caution
+Since the module reads the operations at plan time, the API project must be built before you plan or apply. The generated wiring makes the shared Terraform project's `build` depend on this, so `nx run infra:deploy` handles it for you. If the file is missing, the plan fails with a message telling you to build the project.
+:::
```
cogwirrel, line 487: remove this caution block and content

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
cogwirrel: Remove this part from both tabs and define above:

```
Generated APIs support two integration patterns:

- `isolated` creates one Lambda function per operation. This is the default and recommended option for APIs.
- `shared` creates a single default router Lambda and reuses it for every operation unless you override specific integrations.

`isolated` gives you finer-grained permissions and configuration per operation, as well as better separation for logs and traces. `shared` reduces the likelihood of encountering cold-starts for low-usage APIs.
```

cogwirrel: The integration pattern can be changed at any time in CDK by updating your API construct. For example, setting `pattern` to `'shared'` creates a single function instead of one per integration:

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -643,20 +434,29 @@ export class MyApi<...> extends ... {
...
 
-You can simply instantiate the default module to get the router pattern:
+- `isolated` creates one Lambda function per operation. This is the default for generated APIs.
+- `shared` creates a single router Lambda and a catch-all proxy route serving every operation.
 
-```hcl
-# Default router pattern - single Lambda function for all operations
-module "my_api" {
-  source = "../../common/terraform/src/app/apis/my-api"
+`isolated` gives you finer-grained permissions and configuration per operation. `shared` reduces Lambda and API Gateway integration sprawl.
```
cogwirrel: Remove this part since we've covered it above the tabs

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
```diff
@@ -127,6 +127,60 @@ resource "random_string" "suffix" {
...
+  route_key = { for op, details in local.operations : op =>
+    "${upper(details.method)} ${local.operation_path[op]}"
+  }
+
+  # Source ARN suffix scoping each function's invoke permission to its own route.
+  # Path parameters become wildcards, since the ARN matches concrete request paths.
+  permission_source_suffix = { for op, details in local.operations : op =>
+    "${upper(details.method)}${replace(local.operation_path[op], "/{[^}]*}/", "*")}"
+  }
+}
```
cogwirrel: remove all these comments in the locals section

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
```diff
@@ -127,6 +127,60 @@ resource "random_string" "suffix" {
   upper   = false
 }
 
+<%_ if (backend.integrationPattern === 'isolated') { _%>
+locals {
+  # Operations are generated from the API definition as part of its build.
+  # Guarded with fileexists so `terraform validate` succeeds before the first
+  # build; the precondition below fails the plan if it is genuinely missing.
+  operations_file = "${path.module}/../../../generated/<%- apiNameKebabCase %>/operations.json"
```
cogwirrel: Change comment to:

```
# Generated at build time from the API definition
```

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
```diff
@@ -188,8 +242,14 @@ resource "aws_vpc_security_group_egress_rule" "api_lambda_https" {
...
 
 # Lambda function
+<%_ if (backend.integrationPattern === 'isolated') { _%>
+# One lambda function per operation, each serving just that operation
+resource "aws_lambda_function" "api_lambda" {
+  for_each = local.operations
+<%_ } else { _%>
 # This configures a single "router" lambda to serve all requests
 resource "aws_lambda_function" "api_lambda" {
+<%_ } _%>
```
cogwirrel: Move the comments inside the resource block so the resource block can go outside the ejs `if`

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/rest/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
```diff
@@ -123,6 +123,106 @@ resource "random_string" "suffix" {
   upper   = false
 }
 
+<%_ if (backend.integrationPattern === 'isolated') { _%>
+locals {
+  # Operations are generated from the API definition as part of its build.
```
cogwirrel: Same comments apply as per the http template above.

packages/nx-plugin/src/utils/api-constructs/operations.ts
```diff
@@ -0,0 +1,44 @@
...
+ * File name the operations metadata is written to. The Terraform modules read
+ * this by name, so both the producing targets and the modules refer to it here.
+ */
+export const OPERATIONS_METADATA_FILE_NAME = 'operations.json';
+
+/**
+ * Serialise operations metadata for writing to disk, sorted by operation name
+ * so the file is stable across runs and does not churn the Terraform plan.
+ */
+export const serialiseOperationsMetadata = (
```
cogwirrel: This looks like it's only used by the metadata generator? define it and the other types/constants in there instead and remove this helper if so.

packages/nx-plugin/generators.json
```diff
@@ -124,6 +124,13 @@
       "metric": "g19",
       "hidden": true
     },
+    "open-api#operations-metadata": {
```
cogwirrel: name this `open-api#json-metadata` (and update/rename files etc)

e2e/src/smoke-tests/generator-matrix.ts
```diff
@@ -67,6 +67,17 @@ export const runGeneratorMatrix = async (
     opts,
   );
 
+  // tRPC APIs using the shared integration pattern — the APIs above take the
+  // default `isolated` pattern, so both are covered.
+  await runCLI(
```
cogwirrel, line 72: These should be added to the test matrix generator too.

packages/nx-plugin/src/open-api/operations-metadata/generator.ts
```diff
@@ -0,0 +1,52 @@
...
+} from '../../utils/api-constructs/operations.js';
+import { buildOpenApiCodeGenerationData } from '../ts-client/generator.js';
+import type { CodeGenData } from '../utils/codegen-data/types.js';
+import type { OpenApiOperationsMetadataGeneratorSchema } from './schema';
+
+/**
+ * Generates an operations metadata JSON file from an OpenAPI specification,
+ * read by the vended Terraform modules to define one integration per operation.
+ */
+export const openApiOperationsMetadataGenerator = async (
```
cogwirrel: expose this in the sdk alongside the others

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/rest/__apiNameKebabCase__/__apiNameKebabCase__.tf.template
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/api-constructs/operations.ts
nx-plugin-for-aws: Done.

nx-plugin-for-aws:
Thanks for the review — all comments addressed, and rebased onto latest `main` (which picked up #1127's provider declarations).

Also raised the REST path depth limit from 8 to **16**, with a new `terraform` + `rest-lambda` filtered **Path Depth Limit** section documenting it. Before changing it I checked a few things empirically, which may be useful context:

- **The unused depth levels create no resources.** With only depths 0–1 populated and 8 blocks declared, Terraform creates exactly 3 resources — an empty `for_each = {}` produces zero instances. So the headroom costs only generated configuration, which is why raising it is cheap.
- **Flattening all path parts into a single `for_each` doesn't work.** Referencing `aws_api_gateway_resource.paths[each.value.parent].id` from within the same block fails with `Error: Cycle: ...`, because Terraform's dependency graph is per-block, not per-instance. It still cycles with *zero* nesting, so it's structural — this is [hashicorp/terraform#26697](https://github.com/hashicorp/terraform/issues/26697), which the new caution block links.
- **A multi-segment `path_part`** (e.g. `"items/{itemId}"`) passes `terraform validate` and `plan`, but real AWS rejects it at apply: `Resource's path part only allow a-zA-Z0-9._-`. Worth knowing as a trap, since CI wouldn't catch it.
- **Recursive Terraform modules** aren't viable either — module sources expand statically before `count` is evaluated, so the base case never applies (it ran away to 42 levels and died on `file name too long`).

For scale: tRPC operations are always depth 1 (dot notation, no slashes), so only OpenAPI-based APIs nest at all — a deep route like `/orgs/{orgId}/teams/{teamId}/members/{memberId}/roles` is depth 7. API Gateway itself allows at least 40 levels, so 16 is comfortably clear of realistic APIs while staying well under the service limit. The failure mode is safe: exceeding it fails at *plan* with a message naming the offending paths, not at apply.

One note on the push: the Translate Documentation workflow had auto-committed translations of the previous English docs to this branch. Since I rewrote those sections, the MDX translations were stale, so I dropped them for the workflow to regenerate (per the "only edit English docs" guideline) and kept the still-valid `schema-translations.json` additions for the new generator. Force-pushed with the lease pinned to that exact commit so nothing was clobbered blindly.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -643,20 +417,69 @@ export class MyApi<...> extends ... {
...
 ```
+
+This file is generated from your API, so you do not need to edit this by hand. Adding an operation to your API application code adds the route and lambda function on the next deploy. It is `.gitignore`d by default; remove the entry if you prefer to check it in.
 </Fragment>
 </Infrastructure>
+
+<OptionFilter when={{ iac: 'terraform', infra: 'rest-lambda' }} description="Raising the REST API path depth limit in Terraform">
+##### Path Depth Limit
+
+A REST API needs an API Gateway resource for each segment of an operation's path, so the generated module declares one `aws_api_gateway_resource` per path depth, up to 16 segments. Levels with no matching paths create no resources.
```
cogwirrel: title `Terraform REST API Path Depth Limit`.

Remove the "A REST API needs..." paragraph.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -643,20 +417,69 @@ export class MyApi<...> extends ... {
...
+
+:::caution
+Due to a limitation with Terraform where instances of resources can't reference one another (see [hashicorp/terraform#26697](https://github.com/hashicorp/terraform/issues/26697)), the resource tree is built one level at a time rather than recursively. If an operation's path exceeds 16 segments, the plan fails with a message naming the paths that are too deep.
+
+To support deeper paths, edit `packages/common/terraform/src/app/apis/<my-api>/<my-api>.tf`:
+
+<Steps>
```
cogwirrel: Put this in a Drawer - `Click here for instructions to support deeper paths should your API require it`

packages/nx-plugin/generators.json
nx-plugin-for-aws: Done.

e2e/src/smoke-tests/generator-matrix.ts
nx-plugin-for-aws: Done.

packages/nx-plugin/src/open-api/operations-metadata/generator.ts
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Done.

cogwirrel (approved):

nx-plugin-for-aws:
Second round addressed — all checks green.

- Renamed `open-api#operations-metadata` → **`open-api#json-metadata`** (directory, schema `$id`/title, exported identifiers, registry entry, target command and the schema translations key), and exposed it from the SDK alongside the other OpenAPI generators.
- Added the shared-pattern APIs (`my-api-shared`, `my-api-shared-http`) to `internal#test-matrix`, so both patterns are covered by the composed in-process path as well as the CLI smoke tests.
- Retitled to **Terraform REST API Path Depth Limit**, dropped the "A REST API needs..." paragraph, and moved the instructions into a `Drawer` with the requested trigger text.

While doing the rename I noticed nothing asserted the OpenAPI-driven operations target, so a rename could have silently broken the FastAPI/Smithy path without failing a test (the snapshots only cover the `.tf` files, not `project.json`). Added two tests covering it: that the target is created with the right generator and output path and is wired into the shared Terraform project's `build`, and that it is absent for the `shared` pattern.

Also apologies for the earlier missed comments — I'd filtered by recency rather than checking unresolved threads, so three from the first round slipped through. Now verified against the unresolved-thread list: 16 threads, all replied to.

Verified: 3789 unit tests, lint, `build --all`, and the docs site (1495 pages) all pass, and I confirmed the new depth section and Drawer render in the built HTML.

One note: the Translate Documentation workflow pushed translations to this branch twice mid-review, both times for English docs I then rewrote. Each time I dropped the stale MDX (per the "only edit English docs" guideline) and force-pushed with the lease pinned to that exact commit so nothing was clobbered blindly. The latest run has now regenerated them against the final English source, and I've fast-forwarded to include that commit — so the branch has up-to-date translations and local matches remote.

PR #1128, opened 2026-08-26
fix(ts#api,py#api): cache HTTP API custom authorizer results on Terraform

nx-plugin-for-aws:
### Reason for this change

For `--auth=custom` HTTP APIs, the Terraform module pinned `authorizer_result_ttl_in_seconds = 0`, disabling authorizer result caching entirely, so **every single request** invoked the authorizer Lambda. CDK's `HttpLambdaAuthorizer` defaults `resultsCacheTtl` to `Duration.minutes(5)`, and both providers' REST authorizers sit on the same 300s default (the Terraform REST template omits the attribute, so the provider default of 300 applies). HTTP-on-Terraform was the lone outlier, which made the divergence look unintentional rather than chosen.

There was a second, related bug in the same block: the Terraform HTTP custom authorizer set **no `identity_sources` at all**. API Gateway uses the identity sources as the cache key and [requires at least one for caching to be enabled](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.caching), so raising the TTL alone would have been inert. (The `identity_sources` already present in that template belongs to the *Cognito* authorizer, a separate resource.) CDK defaults `identitySource` to `["$request.header.Authorization"]`, so the two are set together to match.

### Description of changes

`packages/nx-plugin/src/utils/api-constructs/files/terraform/app/apis/http/.../*.tf.template` — on the `custom_authorizer` resource:

```diff
-  authorizer_result_ttl_in_seconds  = 0
+  authorizer_result_ttl_in_seconds  = 300
+  identity_sources                  = ["$request.header.Authorization"]
```

**Why 300 and not 0 — the security/cost trade-off.** This is a real trade-off, not just a consistency argument, so to state the choice plainly: **300s means a revoked token can keep working for up to 5 minutes.** TTL 0 buys instant revocation, at the cost of a Lambda invocation on every request. I have gone with 300 because:

- **It is what users already get on the other three of four paths.** CDK/HTTP, CDK/REST and Terraform/REST are all 300 today. Someone reasoning about their revocation window from the REST docs, or from the CDK construct, is already reasoning about 5 minutes; leaving one provider silently at 0 means the same generator flags produce a materially different security posture depending on `--iac`, which is the more dangerous failure mode. Divergence is the bug here, in either direction.
- **The vended authorizer denies by default and is a TODO.** The generated `authorizer.ts` returns `isAuthorized: false` until the user writes their logic, so we are picking the default for a *type* of authorizer, not for a known token scheme. 300s is upstream's default for exactly that situation.
- **Cost/latency is not marginal.** At TTL 0 a chatty client pays an authorizer invocation per request — proven below at 8 invocations for 8 requests, versus 1 for 10 once cached.

Anyone who needs instant revocation sets the TTL to 0 in the vended module — it is their file to edit.

**Should it be a module variable?** No — deliberately, to match CDK. `HttpLambdaAuthorizer` exposes `resultsCacheTtl` as a construct prop, but the vended CDK construct does not thread it out to a `TestApiProps` field; users edit the construct. Adding a Terraform `variable` would make Terraform *more* configurable than CDK for this one attribute, re-introducing an asymmetry in the opposite direction. A literal on both sides keeps the two readable side by side. Worth a follow-up to expose it on both providers together if there's appetite.

**No migration.** The generated files are vended `KeepExisting`, and this deliberately ships without a migration: **existing workspaces keep `authorizer_result_ttl_in_seconds = 0` and only newly generated ones get the cached authorizer.** That is the intended trade-off, not an oversight — raising the TTL under an existing deployment silently lengthens its token revocation window, which is a change of security posture better left to an explicit, opt-in edit than applied automatically by `nx migrate`. Users who want it can make the two-line change in their own module.

Documentation: the 5-minute window is user-facing behaviour that wasn't documented, so the existing `:::caution[Custom Lambda Authorizer]` block in the tRPC, FastAPI and Smithy guides now notes it (both the CDK and Terraform variants). English only; the 24 translated files are the `Translate Documentation` workflow's own output.

`pnpm nx run @aws/nx-plugin:test` — **208 files / 3785 tests green** (baseline `main` is 3781, so exactly +4: the four regression cases below). `build --all` and `lint --all --configuration=fix` both pass. Rebased on latest `main` (#1125, #1126, #1127, #1136, #1137 merged since branching); #1127 touched this same HTTP template to add `required_providers`, in a different region of the file, and rebased without conflict. Two Terraform snapshots updated (tRPC + FastAPI), each showing only the two intended lines.

Regression tests pinning the TTL across **both** providers and **both** protocols (`trpc/backend/generator.spec.ts`):

- Terraform/HTTP — pins `300` *and* asserts `identity_sources` is present. The pairing is the point: a TTL without an identity source silently does nothing.
- Terraform/REST — asserts the attribute stays *absent*, so the provider's own 300 default applies.
- CDK, HTTP and REST — asserts the construct does not override `resultsCacheTtl`, since `aws-cdk-lib`'s default is the value we want.

In fairness these overlap substantially with the existing snapshots, which do already capture both the TTL and `identity_sources` on contiguous lines. What the tests add beyond them is the *negative* assertions — that CDK never sets `resultsCacheTtl` and Terraform/REST never sets `authorizer_result_ttl_in_seconds`, so the four paths cannot silently drift apart again by someone adding an override. Happy to drop them if the snapshots are considered sufficient.

#### Deployed to AWS and proven from CloudWatch logs

A synth/plan diff can't demonstrate caching, so both providers were **deployed** using the locally compiled plugin (`pnpm link`), with the vended authorizer instrumented to log `AUTHORIZER_INVOKED` on entry. Requests were sent with an identical `Authorization` header.

**Live authorizer config — identical on both providers:**

| | Terraform (`8ot4lanrgk`) | CDK (`k2it3v0mf5`) |
|---|---|---|
| `AuthorizerResultTtlInSeconds` | `300` | `300` |
| `IdentitySource` | `["$request.header.Authorization"]` | `["$request.header.Authorization"]` |
| `EnableSimpleResponses` | `true` | `true` |
| `AuthorizerPayloadFormatVersion` | `2.0` | `2.0` |

**Terraform, fixed — 10 requests, 1 invocation:**

```
$ ./cachetest.sh https://8ot4lanrgk.execute-api.us-east-1.amazonaws.com "Bearer cache-probe-tf-1" 10
req 1..10 -> 403   (all ten)

$ aws logs filter-log-events --log-group-name /aws/lambda/TestApiAuthorizer-too4qoni \
    --filter-pattern AUTHORIZER_INVOKED --query 'length(events)'
1
```

```json
{"level":"INFO","message":"AUTHORIZER_INVOKED","timestamp":"2026-08-26T14:12:45.795Z",
 "function_name":"TestApiAuthorizer-too4qoni",
 "function_request_id":"61669d1d-4732-447d-853f-1bd97aab92de",
 "token":"Bearer cache-probe-tf-1"}
```

**Same live API with the TTL forced back to 0 (the pre-fix state) — 8 requests, 8 invocations.** This is the contrast that shows the attribute is what's doing the work; it is the *only* thing changed between the two runs:

```
2026-08-26T14:14:29.322Z  Bearer cache-probe-ttl0  946342a5
2026-08-26T14:14:31.042Z  Bearer cache-probe-ttl0  9a793ee5
2026-08-26T14:14:32.742Z  Bearer cache-probe-ttl0  97092eb9
2026-08-26T14:14:34.444Z  Bearer cache-probe-ttl0  b581c287
2026-08-26T14:14:36.185Z  Bearer cache-probe-ttl0  301fd525
2026-08-26T14:14:38.069Z  Bearer cache-probe-ttl0  a42c4583
2026-08-26T14:14:39.752Z  Bearer cache-probe-ttl0  548dad6f
2026-08-26T14:14:41.423Z  Bearer cache-probe-ttl0  43b491a6
```

Eight distinct `function_request_id`s for eight requests bearing the same token.

**CDK, for comparison — 10 requests, 1 invocation:**

```json
{"level":"INFO","message":"AUTHORIZER_INVOKED","timestamp":"2026-08-26T14:21:49.467Z",
 "function_name":"cdkws-infra-sandbox-Appli-TestApiAuthorizerFunctio-HPXKpuebqHHW",
 "function_request_id":"98bcbe77-3b8e-4f39-a10b-5ba9d7ee0e3a",
 "token":"Bearer cache-probe-cdk-1"}
```

**Caching really is keyed on the identity source**, rather than blanket-allowing everything for 5 minutes: a second, different token in the same window produced its own separate invocation (`cache-probe-tf-2`, request id `6e776c0b`). Authorization itself was confirmed working first — `Bearer let-me-in` passed through to tRPC, `Bearer nope` returned 403.

#### Teardown

All deployed infrastructure destroyed and verified: `terraform destroy` completed (`terraform state list` now empty), `cdk destroy` completed (stack `does not exist`), and the CDK-retained authorizer log group deleted by hand. `get-apis` returns `[]` and no `TestApi*`/`cdkws*` Lambdas or log groups remain.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/fastapi.mdx
+++ b/docs/src/content/docs/en/guides/fastapi.mdx
@@ -491,6 +491,8 @@ The `UserIdentity` construct can be generated using the <Link path="/guides/reac
 <OptionFilter when={{ auth: 'custom' }} description="Custom Lambda Authorizer CDK usage">
 :::caution[Custom Lambda Authorizer]
 When using `Custom` auth, the construct creates a Lambda Authorizer internally from the generated `authorizer.py` file, which **denies all requests by default**. You must implement your authorization logic in that file before your API will accept any traffic.
+
+Authorizer results are cached for 5 minutes, keyed on the `Authorization` header, so a revoked token may be accepted until its cached result expires. Reduce the cache duration in the generated infrastructure if you need faster revocation.
 :::
 </OptionFilter>
 </Fragment>
@@ -604,6 +606,8 @@ module "my_api" {
 <OptionFilter when={{ auth: 'custom' }} description="Custom Lambda Authorizer usage with Terraform">
 :::caution[Custom Lambda Authorizer]
 When using `Custom` auth, your API is protected by a Lambda Authorizer that **denies all requests by default**. You must implement your authorization logic in the generated `authorizer.py` file before your API will accept any traffic.
+
+Authorizer results are cached for 5 minutes, keyed on the `Authorization` header, so a revoked token may be accepted until its cached result expires. Reduce the cache duration in the generated infrastructure if you need faster revocation.
 :::
 </OptionFilter>
 </Fragment>
--- a/docs/src/content/docs/en/guides/trpc.mdx
+++ b/docs/src/content/docs/en/guides/trpc.mdx
@@ -573,6 +573,8 @@ export class ExampleStack extends Stack {
 
 :::caution[Custom Lambda Authorizer]
 When using `Custom` auth, the construct creates a Lambda Authorizer internally from the generated `src/authorizer.ts` file, which **denies all requests by default**. You must implement your authorization logic in that file before your API will accept any traffic.
+
+Authorizer results are cached for 5 minutes, keyed on the `Authorization` header, so a revoked token may be accepted until its cached result expires. Reduce the cache duration in the generated infrastructure if you need faster revocation.
 :::
 </OptionFilter>
 
@@ -718,6 +720,8 @@ module "my_api" {
 <OptionFilter when={{ auth: 'custom' }} description="Custom Lambda Authorizer usage with Terraform">
 :::caution[Custom Lambda Authorizer]
 When using `Custom` auth, your API is protected by a Lambda Authorizer that **denies all requests by default**. You must implement your authorization logic in the generated `src/authorizer.ts` file before your API will accept any traffic.
+
+Authorizer results are cached for 5 minutes, keyed on the `Authorization` header, so a revoked token may be accepted until its cached result expires. Reduce the cache duration in the generated infrastructure if you need faster revocation.
 :::
 </OptionFilter>
 </Fragment>
--- a/docs/src/content/docs/en/guides/ts-smithy-api.mdx
+++ b/docs/src/content/docs/en/guides/ts-smithy-api.mdx
@@ -697,6 +697,8 @@ The `UserIdentity` construct can be generated using the <Link path="/guides/reac
 <OptionFilter when={{ auth: 'custom' }} description="Custom Lambda Authorizer CDK usage">
 :::caution[Custom Lambda Authorizer]
 When using `Custom` auth, the construct creates a Lambda Authorizer internally from the generated `src/authorizer.ts` file, which **denies all requests by default**. You must implement your authorization logic in that file before your API will accept any traffic.
+
+Authorizer results are cached for 5 minutes, keyed on the `Authorization` header, so a revoked token may be accepted until its cached result expires. Reduce the cache duration in the generated infrastructure if you need faster revocation.
 :::
 </OptionFilter>
 </Fragment>
@@ -768,6 +770,8 @@ module "my_api" {
 <OptionFilter when={{ auth: 'custom' }} description="Custom Lambda Authorizer usage with Terraform">
 :::caution[Custom Lambda Authorizer]
 When using `Custom` auth, your API is protected by a Lambda Authorizer that **denies all requests by default**. You must implement your authorization logic in the generated `src/authorizer.ts` file before your API will accept any traffic.
+
+Authorizer results are cached for 5 minutes, keyed on the `Authorization` header, so a revoked token may be accepted until its cached result expires. Reduce the cache duration in the generated infrastructure if you need faster revocation.
 :::
 </OptionFilter>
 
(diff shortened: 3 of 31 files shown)


packages/nx-plugin/src/migrations/latest/http-authorizer-result-cache-ttl/migration.ts
```diff
@@ -0,0 +1,144 @@
...
+
+const authorizerBlockContaining = (contains: string) =>
+  hcl(
+    `\`resource "aws_apigatewayv2_authorizer" "custom_authorizer" { $props }\` where { $props <: contains ${contains} }`,
+  );
+
+const divergedMessage = (filePath: string) =>
+  `${filePath}: the HTTP API custom Lambda authorizer has diverged from the generated shape - left untouched. To cache authorizer results rather than invoking the authorizer on every request, set \`authorizer_result_ttl_in_seconds = 300\` and \`identity_sources = ${IDENTITY_SOURCE}\` on it.`;
+
+export default async function migration(
```
cogwirrel: No need for a migration here

cogwirrel (dismissed):

packages/nx-plugin/src/migrations/latest/http-authorizer-result-cache-ttl/migration.ts
nx-plugin-for-aws: Agreed — removed. The migration is gone entirely (`metadata.json`, `migration.ts`, `migration.spec.ts`), and I rebuilt the branch so it leaves no trace in history rather than appearing and then being reverted. `packages/nx-plugin/migrations.json` is gitignored and assembled by folder discovery at build time, so there was no manifest or ranking file to unwind; a repo-wide grep for the migration name now returns nothing.

The trade-off is deliberate and now stated in the PR body: existing workspaces keep `authorizer_result_ttl_in_seconds = 0` and only newly generated ones get the cached authorizer. Raising the TTL under a live deployment would silently lengthen its token revocation window, which is better left to an explicit opt-in edit than applied by `nx migrate`.

What remains is the template fix (TTL 300 + the `identity_sources` it has to be paired with), the regression test, the two snapshots, and a docs note about the 5-minute window.

cogwirrel (approved):

PR #902, opened 2026-07-07
feat(py#rdb): add Python relational database generator

AlexTo:
### Reason for this change

Adds a `py#rdb` generator for Python projects that need a relational database backed by Amazon Aurora, with SQLModel for data modelling, Alembic for migrations, and IAM authentication.

Example project: [https://github.com/AlexTo/aws-nx-poc/tree/py-rdb](https://github.com/AlexTo/aws-nx-poc/tree/py-rdb)

### Description of changes

- New `py#rdb` generator scaffolding `connection.py` (IAM auth via `get_session`/`get_engine`, SSL/TLS support), Alembic migration handler, and create-db-user handler as Docker-based Lambdas; uses SQLModel for data modelling
- Adds connection generators for `py#fast-api`, `py#agent`, and `py#mcp-server` to `py#rdb`
- Updates and restructures SSL requirement docs snippets by language (`ts-lambda-rdb-ssl-requirements`, `py-lambda-rdb-ssl-requirements`, `ts-mcp-server-rdb-ssl-requirements`)

### Description of how you validated changes

- Unit tests
- Manually generated projects and connections to verify they work end-to-end for both PostgreSQL and MySQL engines, including SSL configuration for RDS Cluster and RDS Proxy

### Issue # (if applicable)

N/A

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/awslabs/nx-plugin-for-aws/blob/main/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #1129, opened 2026-08-26
fix(ts#rdb,py#rdb): grant the terraform rds proxy role rds-db:connect on mysql

nx-plugin-for-aws:
### Reason for this change

For MySQL Aurora on Terraform, the RDS Proxy's IAM role was handed the **admin database secret** instead of the `rds-db:connect` permission it actually needs.

The vended proxy role policy (`aws_iam_role_policy.proxy_db_user_connect`) had a MySQL carve-out:

```hcl
<%_ if (engine === 'mysql' && framework !== 'sqlmodel') { _%>
  { Action = ["secretsmanager:GetSecretValue"], Resource = [module.aurora.secret_arn] },
  { Action = ["kms:Decrypt"],                   Resource = [module.aurora.kms_key_arn] }
<%_ } else { _%>
  { Action = ["rds-db:connect"], Resource = [".../dbuser:.../${local.database_runtime_user}"] }
<%_ } _%>
```

So a MySQL workspace's proxy role got read access to the Aurora **admin** credentials and their KMS key, and **no `rds-db:connect` at all**. Postgres already got it right, and the CDK construct calls `this.cluster.grantConnect(proxyRole, this.dbUser)` unconditionally for every engine while never granting the proxy secret access.

The secret grant is **dead permission**, not a MySQL requirement. The vended proxy is created with `default_auth_scheme = "IAM_AUTH"` and **no `auth` block**, i.e. no Secrets Manager secret is registered with it, so there is no mechanism by which it could use those credentials. Confirmed on the deployed proxies:

```
$ aws rds describe-db-proxies --db-proxy-name my-db-proxy-8oyqawg6 \
    --query 'DBProxies[0].{Auth:Auth,Scheme:DefaultAuthScheme}'
{ "Auth": [], "Scheme": "IAM_AUTH" }
```

AWS's own [end-to-end IAM authentication docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy-iam-setup.html) say the same, engine-agnostically: "*this requires configuring your IAM role with `rds-db:connect` permissions*", and the Secrets Manager policy is documented as the alternative for *standard* IAM auth, which this is not.

**Origin of the bug**, confirmed by @AlexTo (who authored the branch) [in review](https://github.com/awslabs/nx-plugin-for-aws/pull/1129#discussion_r3876763801): it is a **leftover**. MySQL migrations genuinely *used* to connect through the proxy with the admin secret, because the Rust driver did not support the `mysql_clear_password` plugin that IAM auth requires — so at the time the grant was load-bearing, and `rds-db:connect` was deliberately not given. #892 moved migrations off that path (they now connect direct-to-cluster as admin on every engine) but missed this Terraform branch: its Terraform hunk rewrote `migration_handler_access`, the security-group rules and the handler's env, and never touched `proxy_db_user_connect`. So the secret grant outlived its reason, and because `rds-db:connect` sat on the `else`, the MySQL proxy was left with neither a working credential path nor the permission it now needs. #902 removed the corresponding `auth { auth_scheme = "SECRETS" }` block and the core module's `proxy_secret_access` policy, which is what made the remaining grant not just unnecessary but inert.

This mattered on two counts: the proxy could not authenticate as the app user, and it held admin credentials — a privilege escalation relative to CDK. So the grant is **replaced**, not supplemented.

**MySQL needs nothing Postgres doesn't here.** Both engines authenticate identically as far as the proxy role is concerned: the vended `create-db-user` handler provisions the app user for IAM auth per engine (`IDENTIFIED WITH AWSAuthenticationPlugin AS 'RDS'` for MySQL, `GRANT rds_iam` for Postgres), and the *client* — not the proxy — mints the auth token (`@aws-sdk/rds-signer` in `prisma.ts.template`, `generate_db_auth_token` in `connection.py.template`), on both engines. Migrations connect directly to the cluster as admin via `DATABASE_SECRET_ARN` on both engines since #892, so nothing in the proxy path wants the secret.

`enable_rds_proxy = false` is unaffected: the whole policy is `count = var.enable_rds_proxy ? 1 : 0`, so it isn't created at all, and clients then grant against `connect_resource_id`, which resolves to the cluster resource id.

### Description of changes

- **`utils/rdb-constructs/.../__nameKebabCase__.tf.template`** — deleted the MySQL carve-out so every engine gets the single `rds-db:connect` statement. `ts#rdb --engine=mysql` and `--engine=postgres` now vend a byte-identical proxy policy.
- **New migration `latest/terraform-rdb-proxy-iam-connect-grant`** — the database app modules are vended `KeepExisting`, so an upgraded MySQL workspace would otherwise keep the admin secret grant. Walks `common/terraform/src/app/dbs/*`, and via GritQL on HCL: skips modules already granting `rds-db:connect` (Postgres, and re-runs), skips modules with no proxy policy, and replaces the statement list only where it still matches the exact two-statement shape the generator vended. A customised policy is left untouched and reported via `nextSteps`.

The second, identically-shaped MySQL carve-out at ~line 713 — `<%_ if (engine !== 'mysql' && framework !== 'sqlmodel') { _%>` gating `db_user` in `null_resource.migration_trigger`'s trigger set — is **deliberate, and left alone**. It is not the same mistake. Since #892 migrations run as admin from the secret on every engine and never reference `dbUser`, so for MySQL the app user genuinely is not an input to the migration handler and re-running migrations when it changes would be spurious. It is only a trigger-invalidation hint, carries no IAM consequence, and the surrounding `depends_on` still orders `migration_trigger` after `create_db_user_trigger`. (Arguably the Postgres side is now the stale one — `env.py.template` still reads `cfg['dbUser']` for sqlmodel, which is why the `framework` clause is there — but that is a separate question from this security fix, so I have not touched it.)

**Second commit — @AlexTo's style point**, that these conditions should say `framework === 'prisma'` rather than `framework !== 'sqlmodel'`. Kept separate from the IAM fix so the security change stays legible next to the mechanical one. Five sites converted:

| File | Site |
|---|---|
| `terraform/app/dbs/…/__nameKebabCase__.tf.template` | `variable "asset_bucket_name"` |
| | `create_db_user_bundle_path` local |
| | `data "archive_file" "create_db_user_zip"` + its `aws_s3_object` |
| `cdk/app/dbs/__nameKebabCase__.ts.template` | the `Code` import |
| | the `Runtime` import |

Each guards the **zip-bundled create-db-user Lambda**, which exists precisely because Prisma needs a Node.js zip handler where SQLModel uses a container image — so `=== 'prisma'` is the honest statement of the requirement, not merely an equivalent spelling. Every one of them is the counterpart of an adjacent `=== 'sqlmodel'` block (or, in CDK, feeds the `else` of a `framework === 'sqlmodel'` split of `createDbUserHandler`), i.e. a strict either/or over the bundling strategy rather than "anything that isn't SQLModel".

**One site deliberately left as `!==`:** `null_resource.migration_trigger`'s `engine !== 'mysql' && framework !== 'sqlmodel'`. It fails that test — the value it guards, `db_user`, is consumed by SQLModel's Alembic `env.py` (`username=cfg['dbUser']`) and by nothing in the Prisma migration handler, so the condition is genuinely SQLModel-negative and rewriting it `=== 'prisma'` would assert something different from what is meant. (Only the framework half; `engine !== 'mysql'` was never in question.)

**Unit tests** — a per-engine regression test in `ts/rdb/generator.spec.ts` (`it.each(['postgres','mysql'])`) asserting the proxy role's actions: `rds-db:connect` present, and `secretsmanager:GetSecretValue` / `kms:Decrypt` absent from the policy body. Plus 6 migration tests (applies, already-granted no-op, no proxy policy, diverged-and-reported, idempotent, empty workspace). Full suite green: **3762 passed**, with the only snapshot change being the MySQL Terraform policy. `build` and `lint --configuration=fix` both clean.

**Deployed to AWS on both engines, with the proxy enabled.** Note the existing `terraform-deploy-rdb` smoke test sets `enable_rds_proxy = false`, so this policy has never actually been exercised by CI — I deployed a workspace with `enable_rds_proxy = true` for both a MySQL and a Postgres Aurora RDB (generated by the locally compiled plugin), which is what makes the proxy role's grant live.

`apply` succeeded in 12m40s. The vended `create-db-user` and migration handlers both ran clean on MySQL (`StatusCode: 200`, no `FunctionError`, nothing in logs). The **deployed** proxy role policies are now identical across engines and hold nothing else:

```
# MySQL — role my-db-aurora-proxy-111bcce350129fea2766170230
{ "Action": ["rds-db:connect"], "Effect": "Allow",
  "Resource": ["arn:aws:rds-db:us-east-1:...:dbuser:cluster-MYFHR7ZQKITT6BQWBD5422JXYQ/db_hdflenq5"] }

# Postgres — role pg-db-aurora-proxy-5609b7ffaed67f519748789bbb
{ "Action": ["rds-db:connect"], "Effect": "Allow",
  "Resource": ["arn:aws:rds-db:us-east-1:...:dbuser:cluster-JROQCGIDMQTPIJDZJ5KOHWWF4U/db_34o7syih"] }
```

A plan diff can't show whether the proxy can still authenticate, so I also deployed a prober Lambda that connects **through each proxy endpoint** as the application db user with an IAM auth token and runs a real query:

```json
{
  "postgres": { "ok": true, "connectedVia": "pg-db-proxy-0ycin2yd.proxy-....rds.amazonaws.com",
    "currentUser": "db_34o7syih", "database": "pg_db",
    "serverVersion": "PostgreSQL 17.7 on aarch64-unknown-linux-gnu...",
    "realQuery": { "math": 2, "backend_ip": "10.0.1.203/32", "ts": "2026-08-26T18:01:25.622Z" } },
  "mysql":    { "ok": true, "connectedVia": "my-db-proxy-8oyqawg6.proxy-....rds.amazonaws.com",
    "currentUser": "db_hdflenq5@%", "database": "my_db", "serverVersion": "8.0.44",
    "realQuery": { "math": 2, "backend_host": "ip-172-16-5-93", "ts": "2026-08-26T18:01:25.000Z" } }
}
```

**MySQL still authenticates end-to-end with the corrected IAM** — proxy-to-cluster on `rds-db:connect` alone, with the admin secret grant removed. `backend_host` / `backend_ip` are the Aurora backend, so the proxy really did complete its own hop. An earlier prober revision additionally confirmed the app user is correctly scoped to CRUD: DDL was refused post-authentication (`CREATE command denied to user 'db_hdflenq5'@'...'`), which is the create-db-user handler's grant working as intended.

Matches CDK, which grants the proxy role only `rds-db:connect` for every engine and no secret access.

**The condition sweep is proven behaviour-neutral rather than argued to be.** `test -u` writes **zero snapshots** — the rendered output is byte-identical for both `ts#rdb` (prisma) and `py#rdb` (sqlmodel), which is the strongest available evidence given both generators snapshot their full Terraform and CDK output. I also generated fresh workspaces on the rebuilt plugin and exercised the SQLModel path explicitly, since my deployment testing only covered Prisma/MySQL:

- `terraform validate` — **Success** on both the prisma module (`ts#rdb`, postgres) and the sqlmodel module (`py#rdb`, mysql).
- Conditional rendering confirmed in both directions: the prisma module renders all 5 zip-staging references; the sqlmodel module renders **0**.
- CDK: prisma renders `Code` / `Runtime` and uses both (`Code.fromAsset`, `Runtime.NODEJS_LATEST`); sqlmodel omits both, so no unused import.

(While checking the CDK path I hit a pre-existing, unrelated bug: `cdk/app/dbs/__nameKebabCase__.ts.template` uses bare `import.meta.url` at two sites that aren't `esm`-guarded the way line 28 is, so `common-constructs` fails to compile in a CJS workspace with `TS1470`. Identical on `main` and untouched by this PR — flagging it rather than fixing it here.)

Everything was torn down (`destroy` exit 0, 24m50s — `prevent_destroy` flipped to `false` first, as `terraform-deploy.spec.ts` does) and teardown verified: zero Aurora clusters, zero proxies, no `rdbproxy-*` VPC, and no `my-db`/`pg-db` secrets remaining.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
+++ b/packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
@@ -198,22 +198,6 @@ resource "aws_iam_role_policy" "proxy_db_user_connect" {
   policy = jsonencode({
     Version = "2012-10-17"
     Statement = [
-<%_ if (engine === 'mysql' && framework !== 'sqlmodel') { _%>
-      {
-        Effect = "Allow"
-        Action = [
-          "secretsmanager:GetSecretValue"
-        ]
-        Resource = [module.aurora.secret_arn]
-      },
-      {
-        Effect = "Allow"
-        Action = [
-          "kms:Decrypt"
-        ]
-        Resource = [module.aurora.kms_key_arn]
-      }
-<%_ } else { _%>
       {
         Effect = "Allow"
         Action = ["rds-db:connect"]
@@ -221,7 +205,6 @@ resource "aws_iam_role_policy" "proxy_db_user_connect" {
           "arn:aws:rds-db:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:dbuser:${module.aurora.cluster_resource_id}/${local.database_runtime_user}"
         ]
       }
-<%_ } _%>
     ]
   })
 }
--- a/packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
+++ b/packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
@@ -138,7 +138,7 @@ variable "create_db_user_docker_image_tag" {
 }
 <%_ } _%>
 
-<%_ if (framework !== 'sqlmodel') { _%>
+<%_ if (framework === 'prisma') { _%>
 variable "asset_bucket_name" {
   description = "Name of the shared asset S3 bucket used to stage the Lambda deployment zip. Instantiate the `core/asset-bucket` module once per deployment and pass its `bucket_name` output here."
   type        = string
@@ -157,7 +157,7 @@ resource "random_string" "suffix" {
 locals {
   config                       = jsondecode(file("${path.module}/../../../../../../../<%= projectRoot %>/config.json"))
   runtime_config_key           = local.config.runtimeConfigKey
-<%_ if (framework !== 'sqlmodel') { _%>
+<%_ if (framework === 'prisma') { _%>
   create_db_user_bundle_path   = "${path.module}/../../../../../../../<%- createDbUserBundleDir %>"
 <%_ } _%>
   migration_function_name      = "<%= nameKebabCase %>-migration-${random_string.suffix.result}"
@@ -209,7 +209,7 @@ resource "aws_iam_role_policy" "proxy_db_user_connect" {
   })
 }
 
-<%_ if (framework !== 'sqlmodel') { _%>
+<%_ if (framework === 'prisma') { _%>
 data "archive_file" "create_db_user_zip" {
   type        = "zip"
   source_dir  = local.create_db_user_bundle_path
--- a/packages/nx-plugin/src/migrations/latest/terraform-rdb-proxy-iam-connect-grant/__snapshots__/migration.spec.ts.snap
+++ b/packages/nx-plugin/src/migrations/latest/terraform-rdb-proxy-iam-connect-grant/__snapshots__/migration.spec.ts.snap
@@ -1,0 +1,23 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`terraform-rdb-proxy-iam-connect-grant migration > should replace the mysql admin secret grant with rds-db:connect 1`] = `
+"resource "aws_iam_role_policy" "proxy_db_user_connect" {
+  count = var.enable_rds_proxy ? 1 : 0
+
+  role = module.aurora.proxy_role_name
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = ["rds-db:connect"]
+        Resource = [
+          "arn:aws:rds-db:\${data.aws_region.current.region}:\${data.aws_caller_identity.current.account_id}:dbuser:\${module.aurora.cluster_resource_id}/\${local.database_runtime_user}"
+        ]
+      }
+    ]
+  })
+}
+"
+`;
--- a/packages/nx-plugin/src/migrations/latest/terraform-rdb-proxy-iam-connect-grant/metadata.json
+++ b/packages/nx-plugin/src/migrations/latest/terraform-rdb-proxy-iam-connect-grant/metadata.json
@@ -1,0 +1,3 @@
+{
+  "description": "Replace the MySQL RDS Proxy role's admin secret access with the rds-db:connect grant it needs for end-to-end IAM auth"
+}
--- a/packages/nx-plugin/src/migrations/latest/terraform-rdb-proxy-iam-connect-grant/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/terraform-rdb-proxy-iam-connect-grant/migration.spec.ts
@@ -1,0 +1,154 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Tree } from '@nx/devkit';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test.js';
+import migration from './migration.js';
+
+const MYSQL_FILE =
+  'packages/common/terraform/src/app/dbs/my-sql-db/my-sql-db.tf';
+const POSTGRES_FILE =
+  'packages/common/terraform/src/app/dbs/postgres-db/postgres-db.tf';
+
+/** The pre-change shape of a vended MySQL database app module. */
+const MYSQL_BEFORE = `resource "aws_iam_role_policy" "proxy_db_user_connect" {
+  count = var.enable_rds_proxy ? 1 : 0
+
+  role = module.aurora.proxy_role_name
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = [
+          "secretsmanager:GetSecretValue"
+        ]
+        Resource = [module.aurora.secret_arn]
+      },
+      {
+        Effect = "Allow"
+        Action = [
+          "kms:Decrypt"
+        ]
+        Resource = [module.aurora.kms_key_arn]
+      }
+    ]
+  })
+}
+`;
+
+/** The vended Postgres shape, which already has the grant. */
+const POSTGRES_BEFORE = `resource "aws_iam_role_policy" "proxy_db_user_connect" {
+  count = var.enable_rds_proxy ? 1 : 0
+
+  role = module.aurora.proxy_role_name
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = ["rds-db:connect"]
+        Resource = [
+          "arn:aws:rds-db:\${data.aws_region.current.region}:\${data.aws_caller_identity.current.account_id}:dbuser:\${module.aurora.cluster_resource_id}/\${local.database_runtime_user}"
+        ]
+      }
+    ]
+  })
+}
+`;
+
+describe('terraform-rdb-proxy-iam-connect-grant migration', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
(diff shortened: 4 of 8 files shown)


packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -206,30 +206,13 @@ resource "aws_iam_role_policy" "proxy_db_user_connect" {
   policy = jsonencode({
     Version = "2012-10-17"
     Statement = [
-<%_ if (engine === 'mysql' && framework !== 'sqlmodel') { _%>
```
AlexTo, line 201: `engine === 'mysql' && framework !== 'sqlmodel'` means Prisma / MySQL, sorry, shouldn't have used `!==` but should be explicit `=== 'prisma'` instead. 
Anyway, it is a left-over bug. MySQL migration used to connect to Proxy using admin secret (instead of IAM user) due to the Rust driver not supporting `mysql_clear_password` required by IAM.
#892 aimed to fix it, but still left this TF part.

nx-plugin-for-aws: Thanks — that background is the missing piece, and much better motivation than the CDK-parity argument I'd built the description on. I'd worked out that the grant was inert *now* (the proxy is `IAM_AUTH` with `Auth: []`, so it has no way to use a secret) but not that it was ever load-bearing. The Rust driver not supporting `mysql_clear_password` explains exactly why MySQL was the odd one out and why `rds-db:connect` was deliberately withheld rather than just forgotten. I've rewritten "Reason for this change" to lead with it as the origin, crediting you and linking this thread.

It lines up with what #892 actually touched: its Terraform hunk rewrote `migration_handler_access`, the security-group rules and the handler env, but never `proxy_db_user_connect` — so the secret grant outlived its reason, and because `rds-db:connect` sat on the `else`, MySQL ended up with neither a working credential path nor the permission it now needs. #902 then removed the `auth { auth_scheme = "SECRETS" }` block and the core module's `proxy_secret_access` policy, which is what tipped it from merely unnecessary to inert.

**On the branch itself:** removed entirely rather than converted — the whole `engine === 'mysql' && framework !== 'sqlmodel'` arm is gone, so both engines now fall through to the single `rds-db:connect` statement and vend a byte-identical policy. Verified against a real deployment with `enable_rds_proxy = true` (which, incidentally, the existing `terraform-deploy-rdb` smoke test doesn't cover — it sets the proxy off, which is why this survived CI): MySQL connects through the proxy endpoint as `db_hdflenq5@%` on 8.0.44 and runs real queries, with the proxy role holding only `rds-db:connect`.

**On `!==` vs `=== 'prisma'`:** done as a second commit, `2233fb45`, kept separate so it doesn't bury the security fix. Five sites converted — `variable "asset_bucket_name"`, the `create_db_user_bundle_path` local, `data "archive_file" "create_db_user_zip"` + its `aws_s3_object`, and the CDK `Code` / `Runtime` imports.

I checked each rather than sweeping mechanically, on the basis that `=== 'prisma'` should only replace `!== 'sqlmodel'` where Prisma is the actual requirement — otherwise it would silently change meaning if a third framework ever landed. All five guard the zip-bundled create-db-user Lambda, which exists because Prisma needs a Node.js zip handler where SQLModel uses a container image, and each is the counterpart of an adjacent `=== 'sqlmodel'` block (in CDK, the `else` of a `framework === 'sqlmodel'` split of `createDbUserHandler`) — a strict either/or over bundling strategy, so Prisma-specific.

**One I left alone:** `null_resource.migration_trigger`'s `engine !== 'mysql' && framework !== 'sqlmodel'`. The `db_user` it guards is read by SQLModel's Alembic `env.py` (`username=cfg['dbUser']`) and by nothing in the Prisma migration handler, so there the negative really is what's meant and `=== 'prisma'` would assert something different. Happy to revisit if you read it otherwise — you'd know better than I would whether that trigger entry is still pulling its weight at all post-#892, since migrations no longer reference `dbUser` on either engine.

Proved the sweep neutral rather than assuming it: `test -u` writes **zero** snapshots (byte-identical output for both generators), and I generated fresh workspaces to exercise the SQLModel path my deployment testing missed — `terraform validate` passes on both modules, the prisma module renders all 5 zip-staging references while sqlmodel renders 0, and CDK omits `Code`/`Runtime` for sqlmodel so there's no unused import.

One unrelated thing I tripped over: `cdk/app/dbs/__nameKebabCase__.ts.template` uses bare `import.meta.url` at two sites that aren't `esm`-guarded the way line 28 is, so `common-constructs` fails `TS1470` in a CJS workspace. Same on `main`, so I've left it out of this PR — worth a separate issue if it isn't known.

cogwirrel (approved):

PR #1130, opened 2026-08-26
feat(terraform#project): add checkov target, unblock bootstrap-destroy, pin required_version

nx-plugin-for-aws:
### Reason for this change

Three related defects in what `terraform#project` vends:

1. **`bootstrap-destroy` hung on an interactive prompt.** The target ran `terraform destroy -state=<dist>/bootstrap.tfstate` with no `-var=aws_region`, while `bootstrap/variables.tf` declares `aws_region` with no default. In any non-TTY context (CI, `nx run` under a non-interactive parent) terraform blocks forever on its input prompt. The sibling `bootstrap` target passes the variable correctly via `scripts/bootstrap.ts`, so this was an omission.

2. **Both `providers.tf` omitted `required_version`.** `src/providers.tf` and `bootstrap/providers.tf` were the only vended `.tf` files without it — ~60 others set `required_version = ">= 1.0"`. `bootstrap/providers.tf` also had no `terraform` block at all, so it pinned no AWS provider version either.

3. **`nx run-many --target checkov` silently skipped Terraform projects.** CDK infrastructure projects name their security scan `checkov`; Terraform projects named theirs `test`. A workspace with both scanned only the CDK app. Terraform projects also shipped no `checkov.yml`, so unlike CDK users they had no central place to configure skips — only per-resource `#checkov:skip` comments.

### Description of changes

**`bootstrap-destroy`** now runs `scripts/bootstrap-destroy.ts`, mirroring how `bootstrap` works: it resolves account + region from the AWS SDK credential chain (`resolveAwsConfig`) and passes `-var=aws_region=<region>` plus `-auto-approve`. It also pulls the bootstrap tfstate from S3 first, so it works from a fresh checkout, and empties the bucket before destroying it — the bucket holds its own state object and is not `force_destroy`, so `DeleteBucket` otherwise fails with `BucketNotEmpty` (verified below). When no state is found it logs and exits cleanly rather than failing.

**`required_version = ">= 1.0"`** added to both `providers.tf`. `bootstrap/providers.tf` gains a full `terraform` block, so it now also pins the AWS provider via the existing `awsProviderVersion` template variable (picked up automatically by the vended-version sync, since it is a `.tf` file).

**Target names.** The security scan moves to a `checkov` target, matching the name CDK infrastructure projects already use. **`test` now runs Terraform's native test framework** (`terraform test`, over any `.tftest.hcl` files) rather than aliasing the scan — `test` should mean tests. The full target set is unchanged apart from the addition:

```
apply bootstrap bootstrap-destroy build checkov deploy destroy fmt init output plan test validate
```

`test` is self-contained — `terraform init -backend=false` then `terraform test`. That standalone property is load-bearing and is preserved deliberately. Both halves were verified rather than assumed:

- `terraform test` without a prior `init` fails with `unknown provider registry.terraform.io/hashicorp/aws`, so the providers must be installed first.
- It must **not** depend on the vended `init` target, which runs `scripts/init.ts` to configure the **S3 remote backend** and fails `NoSuchBucket` until `bootstrap` has run. `-backend=false` installs the providers without touching a backend; mocked tests reach no backend anyway.
- A project with **no** `.tftest.hcl` files is a clean no-op success (`Success! 0 passed, 0 failed.`, exit 0), so a freshly generated project does not break `run-many --target test --all`.

**`build` depends on `fmt`, `checkov` and `test`** (plus the shared Terraform lib for application projects), so the security scan *and* the Terraform tests both run on a normal build rather than only when someone remembers to invoke them.

Two things make putting `test` in `build` safe, both driven by measurement rather than assumption:

- **`TF_DATA_DIR` isolates the tests' `.terraform`** into `dist/{projectRoot}/terraform-test`. `init`, `validate`, `plan` and `destroy` all use `cwd: {projectRoot}/src` and so share one `src/.terraform` and one `.terraform.lock.hcl`; initialising the tests into that same directory races them. I reproduced that race directly — concurrent `init`s give `the cached package for hashicorp/aws 6.61.0 (in .terraform/providers) does not match any of the checksums recorded in the dependency lock file`. `cwd` deliberately stays `{projectRoot}/src`, so `.tftest.hcl` discovery and the relative module sources (`../../common/terraform/src/metrics`) keep resolving — a copied directory or `-chdir` would break those.
- **The cache is modelled so a hit is honest.** `inputs` gains `^production` and `outputs` records the data dir. Both fixed real defects I found: with `inputs: ['default']` alone, editing a consumed shared module served a **stale pass** (verified: `1/1 hit` after changing `metrics.tf`; now correctly `0/1 hit`), and without `outputs` a cache hit reported `Success!` over a directory where `terraform test` fails `Module not installed`. A hit now restores `modules` + `providers`, and running `terraform test` in that restored state genuinely passes.

**`checkov.yml`** is now vended for Terraform projects too, matching the CDK app, and wired in with `--config-file ../checkov.yml`. It is generated with `OverwriteStrategy.KeepExisting` from its own template dir (the rest of the project is vended with `Overwrite`), so a re-run never clobbers a user's curated skips. The CDK template's `output`/`output-file-path` keys are deliberately omitted — the Terraform target already passes those on the command line, and duplicating them in the config file would fight it.

**Migration** (`latest/terraform-project-checkov-target-and-bootstrap-destroy`) carries existing workspaces across all three: it adds `checkov` (copying the scan from the `test` target the pre-fix generator put it on) and wires in `--config-file`, deliberately leaving an existing workspace's `test` target untouched since that target is the user's, repoints `bootstrap-destroy` at the new script, vends `checkov.yml` + `bootstrap-destroy.ts` (`KeepExisting`), and adds `required_version` via GritQL HCL rewrites. It pattern-matches before writing — a diverged `test` or `bootstrap-destroy` target, or a `providers.tf` it doesn't recognise, is left untouched and reported via `nextSteps` — and a `providers.tf` that already declares `required_version` keeps the user's constraint.

Docs updated in `guides/terraform-project.mdx` only (English): the `checkov` target, a new section on writing and running Terraform tests (a `.tftest.hcl` example with a mocked provider, `command = plan` vs `apply`, and a link to the upstream reference), how to configure skips in `checkov.yml` vs per-resource comments, both new files in the file trees, and a note that `bootstrap-destroy` empties the bucket and runs unattended.

`pnpm nx run @aws/nx-plugin:test` (**3773 passed**, no snapshot churn), `run-many --target build --all`, and `run-many --target lint --all --configuration=fix` all pass.

**Regression tests added** — generator: the full target set above is pinned so no name can be renamed or dropped, `build` depends on both `checkov` and `test`, `test` runs `terraform test` and depends on `init`, `bootstrap-destroy` passes the region, both `providers.tf` carry `required_version`, `checkov.yml` is vended and wired in, and user-curated skips survive a re-run. Migration: 14 tests covering each fix, both project types, idempotency, the diverged/skip paths, and that an existing `test` target is left byte-identical.

Then end-to-end in a real generated workspace (`nx 23.1.1`, locally compiled plugin, project name suffixed uniquely because the vended state key has no workspace uniqueness).

**Defect 1 — `bootstrap-destroy`, before.** The exact old command, non-interactive. With stdin at EOF it prompts then errors; with stdin held open (what a CI runner looks like) it blocks until killed:

```
$ terraform destroy -state=.../bootstrap.tfstate      # stdin held open, no TTY
var.aws_region
  AWS Region (us-east-1, us-west-2, ap-southeast-2, etc)

  Enter a value:
=== terraform exit: 124 (124 = blocked forever) ===
```

**After**, the argv the new script builds, same held-open stdin:

```
$ terraform destroy -auto-approve -state=.../bootstrap.tfstate -var=aws_region=us-east-1
Destroy complete! Resources: 0 destroyed.
=== terraform exit: 0 ===
```

And the real target, full cycle against a state bucket in an otherwise-unused region so it was exclusively mine to destroy — both runs with `< /dev/null`, no TTY:

```
$ nx run <project>:bootstrap            # bucket did not exist beforehand (404)
NX   Successfully ran target bootstrap        === exit: 0 ===
   -> bucket created, contains bootstrap.tfstate

$ nx run <project>:bootstrap-destroy
aws_s3_bucket.terraform_state: Destruction complete after 1s
Destroy complete! Resources: 6 destroyed.
NX   Successfully ran target bootstrap-destroy  === exit: 0 (124 = hung) ===
   -> HeadBucket now 404; no leftover buckets
```

The `emptyBucket` step is load-bearing: an earlier run against a non-empty bucket failed with `BucketNotEmpty: The bucket you tried to delete is not empty. You must delete all versions in the bucket.`

**Defect 3 — `run-many --target checkov`.** Before (scan on `test`, as generated today):

```
$ nx show projects --with-target checkov
[]
$ nx run-many --target checkov --all
NX   No tasks were run
```

After:

```
$ nx show projects --with-target checkov
["@org/terraform","@org/tfinfra-b7478f96"]
$ nx run-many --target checkov --all
NX   Successfully ran target checkov for 2 projects
```

`checkov` scans and honours the config file (`uvx --from checkov==3.3.13 checkov --config-file ../checkov.yml ...` → `Passed checks: 1, Failed checks: 0, Skipped checks: 1`), and `nx run <project>:test` still runs the same scan, now via its `checkov` dependency.

**A note on one CI failure this PR went through.** An earlier revision put `test` in `build`'s `dependsOn` and gave `test` a `dependsOn: ['init']`. That failed every package-manager smoke lane at once. The root cause was not the Terraform CLI being absent or offline — it is installed in those lanes, and `terraform init` ran fine. It was the vended `init` target configuring the S3 remote backend:

```
@e2e-test/tf-infra: Error: Failed to get existing workspaces:
  S3 bucket "571480185771-tf-state-us-east-1" does not exist.
  NoSuchBucket: The specified bucket does not exist
Error: Command failed: terraform init -reconfigure -backend-config=bucket=571480185771-tf-state-us-east-1 ...
```

Smoke lanes never run `bootstrap`, so that bucket never exists. Hence `test` now inits with `-backend=false` and is kept out of `build`.

**`test` running `terraform test`.** Verified end to end in the generated workspace, all three cases:

```
# 1. freshly generated project, no .tftest.hcl files
$ nx run <project>:test
> terraform test
Success! 0 passed, 0 failed.
NX   Successfully ran target test        === exit: 0 ===

# 2. after adding the mocked-provider example from the guide
$ nx run <project>:test
Success! 1 passed, 0 failed.            === exit: 0 ===

# 3. same file plus a deliberately failing assertion
$ nx run <project>:test
deliberate failure - this assertion must fail the target
Failure! 1 passed, 1 failed.
NX   Running target test ... failed      === exit: 1 ===
```

Case 3 is the one that matters — a `test` target that always passes would be worse than the one it replaces. All three were re-verified on the final shape with credentials stripped (`AWS_*` unset, `AWS_CONFIG_FILE=/dev/null`). The docs example is credential-free as claimed: with `AWS_*` unset and `AWS_CONFIG_FILE=/dev/null`, `terraform test` still passes, because `mock_provider` means no API calls. (Writing it also caught that the example needs `environment` in its `variables` block, since the generated project declares it with no default — the guide now says so.)

**`test` inside `build`**, verified on the rebased tree in a generated workspace:

```
$ nx run <project>:build          # dependsOn: fmt, checkov, test, @org/terraform:build
Success! 0 passed, 0 failed.      # shared lib (no test files)
Success! 1 passed, 0 failed.      # the project's own test
NX   Successfully ran target build     exit 0

# no .terraform DIRECTORY leaks into src; the isolated dir holds it instead
src/.terraform            -> absent
dist/<project>/terraform-test -> modules providers
```

The one thing `TF_DATA_DIR` does *not* relocate is `.terraform.lock.hcl`, which `terraform` still writes next to the config. That is benign: it is byte-identical to the file the real `init` writes (verified by checksum), and Terraform intends it to be committed.

**The regression this design claims to avoid** — `plan`/`apply` after a `build` — demonstrated rather than argued:

```
$ nx run <project>:build          exit 0
$ nx run <project>:plan           Plan: 3 to add ... Saved the plan to ...   exit 0
$ nx run <project>:apply          Apply complete! Resources: 3 added         exit 0
```

No "backend configuration changed", no `-reconfigure`/`-migrate-state` demand. Those 3 resources were then torn down with `terraform destroy` from the project's own `src` (`Destroy complete! Resources: 3 destroyed`) — never `bootstrap-destroy`, since the state bucket is shared with concurrent sessions.

**A failing test fails the build**, so `test` in `build` isn't vacuous:

```
$ nx run <project>:build          # with a deliberately failing assertion
deliberate failure - build must fail
Failure! 1 passed, 1 failed.
Failed tasks: - <project>:test    exit non-zero
```

With `terraform` removed from `PATH` entirely, `build` fails on **`fmt`** (`terraform fmt`), a target this PR does not touch and which is on `build` in `main` today — so requiring the Terraform CLI for a Terraform project's build is pre-existing behaviour, not introduced here.

`run-many` reaches both Terraform projects for both targets, and the two are independent:

```
$ nx show projects --with-target checkov   ["@org/terraform","@org/tftest-c3f91a"]
$ nx show projects --with-target test      ["@org/terraform","@org/tftest-c3f91a"]
$ nx run-many --target test --all          Successfully ran target test for 2 projects
```

**Defect 2 — `required_version`.** `terraform init` and `validate` both succeed in the bootstrap dir, and the constraint is genuinely enforced — temporarily raising it to `>= 99.0` gives `This configuration does not support Terraform version 1.14.5`, confirming terraform parses it rather than ignoring it.

**Residual issue for a follow-up (not fixed here).** `init`, `validate`, `plan` and `destroy` still share `src/.terraform`, so two of them running concurrently can still corrupt the provider cache against the lock file (the checksum error quoted above; roughly 1 in 11 in my runs). This PR does not make that worse — `test` is now the only target with its own data dir, `init` remains the sole writer of the shared one, and 5/5 concurrent `test`+`validate` runs were clean after isolation. Fixing it properly means giving the read-only targets their own data dirs too, which is a wider change than this PR.

One further thing I noticed but deliberately left alone, as out of scope: the bootstrap state bucket is named from account + region only (`<account>-tf-state-<region>`), so every workspace in an account/region shares one bucket, and `bootstrap-destroy` deletes it along with every project's state in it. In this shared test account that bucket held 24 objects from concurrent sessions, which is why I verified in an unused region instead. Worth a follow-up.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -250,12 +253,66 @@ Format your Terraform code using the `fmt` target:
...
+}
+```
+
+#### Running Terraform Tests
+
+The `test` target runs [Terraform's native test framework](https://developer.hashicorp.com/terraform/language/tests) over any `.tftest.hcl` files in your project:
+
+<NxCommands commands={['test tf-infra']} />
+
+A project with no test files is a no-op success, so you can add tests when you need them. This target is not part of `build`, since running tests downloads the providers — invoke it directly, or from CI where Terraform is available.
```
cogwirrel: Test should be part of build

cogwirrel (approved):

PR #1131, opened 2026-08-26
feat: align, resolve and sync vended lambda runtimes

nx-plugin-for-aws:
### Reason for this change

The same generator with the same options vended a **different Node major depending on `--iac`**, and Terraform also disagreed with itself.

- The CDK branch emitted `Runtime.NODEJS_LATEST`, which resolves to **`nodejs24.x`** against the pinned `aws-cdk-lib`; the Terraform branch hardcoded **`nodejs22.x`**.
- Terraform was self-inconsistent: the REST/HTTP API and DCR proxy modules hardcoded `nodejs22.x`, while the RDB create-db-user handler was already on `nodejs24.x`.
- Python matched at `python3.14` on both, because CDK names it explicitly (`Runtime.PYTHON_3_14`) rather than via a `_LATEST` alias — which is exactly the difference that let Node drift.

Separately, the Python bundle target pinned `--python-platform` but no `--python-version`, so wheels resolved against whichever interpreter the build machine happened to have rather than the Lambda runtime; and the uv project interpreter was set independently of the Lambda Python runtime, so the two could diverge.

### Description of changes

**One shared source.** `LAMBDA_RUNTIME_VERSIONS` in `utils/versions.ts` is the single pin both IaC providers derive from, via `cdkLambdaRuntime` / `terraformLambdaRuntime` and the matching `*LambdaRuntimeVars()` template helpers. A bump is one edit that moves both providers together.

**CDK is pinned explicitly** to `Runtime.NODEJS_24_X` rather than the `NODEJS_LATEST` alias. An alias is resolved by whichever `aws-cdk-lib` is installed, so its value moves on a dependency bump with no edit in this repo — which is how CDK drifted ahead of Terraform in the first place. An explicit member makes the runtime a reviewable choice here.

Every `runtime` assignment in a vended template now derives from the pin:

| Template | Was |
|---|---|
| `utils/function-constructs/function-constructs.ts` (CDK branch) | `Runtime.NODEJS_LATEST` |
| `utils/function-constructs/function-constructs.ts` (Terraform branch) | `nodejs22.x` |
| `api-constructs/files/cdk/app/apis/{rest,http}/…` | `Runtime.NODEJS_LATEST` ×2 each |
| `api-constructs/files/cdk/core/api/rest/api-gateway-account.ts` | `Runtime.NODEJS_LATEST` |
| `api-constructs/files/terraform/app/apis/{rest,http}/…` | `nodejs22.x` ×2 each |
| `rdb-constructs/files/cdk/app/dbs/…` | `Runtime.NODEJS_LATEST` |
| `rdb-constructs/files/terraform/app/dbs/…` | `nodejs24.x` |
| `dcr-proxy-constructs/files/cdk/app/dcr-proxies/…` | `lambda.Runtime.NODEJS_22_X` |
| `dcr-proxy-constructs/files/terraform/app/dcr-proxies/…` | `nodejs22.x` ×2 |
| `agent-core-constructs/files/cdk/core/agentcore-gateway/…` | `lambda.Runtime.NODEJS_22_X` |

The `python3.14` literals in the Terraform API templates come from the same pin, so both runtimes are handled identically.

**Drift fails a test.** `utils/lambda-runtimes.spec.ts` sweeps every `.template` under the plugin source and fails on either mechanism that allowed the drift: any `Runtime.*_LATEST` alias in a CDK template, and any hardcoded runtime on a `runtime` assignment. Because it scans the tree rather than a list, a *new* template that hardcodes a runtime is caught without the spec changing. Both assertions were verified to fail by reintroducing each original bug.

**The pin is resolved weekly.** `scripts/update-versions.ts` gains `getUpdatedLambdaRuntimeVersions`, in the same shape as the existing per-ecosystem resolvers, reported through `writeReport` as a `Lambda Runtimes` group. The source is `Runtime.ALL` from `aws-cdk-lib` — a curated list that omits runtimes still in public preview, so nothing has to infer general availability:

```
=== Runtime.ALL unique identifiers: 50
   includes preview nodejs26.x? false | python3.15? false
=== RESOLVED runtimes: {"node":"24","python":"3.14"}
```

That curation is load-bearing. The botocore `Runtime` enum *does* list `nodejs26.x` and `python3.15` with no preview or deprecation metadata, and live `CreateFunction` accepts both, so neither could distinguish GA from preview; there is no `aws lambda list-runtimes` API. `aws-cdk-lib` (with its `constructs` peer) is a **root devDependency for the scripts only** — absent from the published plugin's `dependencies`, `devDependencies` and `peerDependencies`. `Runtime.ALL` repeats names (`NODEJS_LATEST` aliases `nodejs24.x`), so identifiers are deduped, and entries with no comparable version (`nodejs`, `nodejs4.3-edge`) or another language (`java21`, `provided.al2023`) are skipped rather than mis-parsed.

**A resolution failure is isolated, not fatal.** Every resolver in that script runs before a single `flushChanges`, inside one `try` whose `catch` calls `process.exit(1)` — so a throw here would cancel that week's TypeScript, Python, Terraform, Java and mise bumps too, and the workflow's report and `Create Pull Request` steps would never run. Instead the current pins are kept, a warning is logged, and each unresolved language is emitted as a note in the report so it reaches the PR body rather than only CI logs:

```
TypeScript Dependencies
- zod 4.4.2 -> 4.4.3

Lambda Runtimes
- Found no node runtime in the aws-cdk-lib runtime list, so it keeps its pinned runtime (24)
```

The unrelated bump survives alongside the visible note, and the note means an unresolved runtime cannot be mistaken for "already up to date".

The project Python version reports as its own `Project Python Version` group carrying the full `major.minor.patch` (`3.14.7 -> 3.15.2`, not a bare patch digit), and a uv resolution failure is reported there rather than under Lambda Runtimes.

**The uv interpreter derives from the Lambda Python runtime.** Lambda names only `major.minor` (`python3.14`) and patches the interpreter itself, but uv pins an exact `major.minor.patch` — so `LAMBDA_RUNTIME_VERSIONS.python` supplies the `major.minor` and `PYTHON_RUNTIME_PATCH` supplies only the patch, resolved from `uv python list --all-versions --output-format json` for whichever minor the run lands on (so a minor bump and its patch move together). Pre-releases and free-threaded variants are excluded. `PYTHON_RUNTIME_PATCH` is `'7'`, the value the resolver returns today, so source already sits where the weekly run would put it rather than shipping stale. Ruff derives `target-version` from `requires-python`, so lint targets the deployed version too. The bundle target now pins `--python-version` so wheels resolve against the runtime.

**Existing workspaces are carried across by the version sync.** The vended constructs and modules are `KeepExisting`, so `sync-vended-versions` — the committed `everyMigration` entry that runs on every upgrade — also syncs Lambda runtimes, the root and per-project `.python-version`, `requires-python`, and the bundle's `--python-version`. No separate one-off migration: one mechanism.

The runtime sync lives in `version-upgrade-migration/sync-lambda-runtimes.ts` and the Python language version in `sync-python-language-version.ts`, each beside its spec. It starts at each owned directory, so the provider follows from where a file sits rather than being inferred from it, carried as a `'cdk' | 'terraform'` discriminant.

Detection is GritQL throughout; regex only ever reads a scalar GritQL has already isolated. `captureAllGritQLVariable` (added to `utils/ast.ts`) returns one metavariable's binding per match, so matching `\`runtime: $value\`` yields `Runtime.NODEJS_24_X` alone and nothing re-derives the assignment around it. This matters because a regex cannot tell a property assignment from surrounding text — it matches inside comments and strings — and tests cover a commented-out assignment under both providers plus a non-Lambda `{ runtime: 'edge' }` property, all correctly left alone.

The helper returns `undefined` when a pattern cannot be applied, which an empty array cannot express. That distinction is what stops a parse failure looking like "nothing to do": such a file is reported as diverged rather than silently skipped, which would otherwise make the sync a no-op without failing.

**What the sync will and will not touch.** Scope is `packages/common/constructs/` and `packages/common/terraform/` — a runtime is not a dependency, so there is no per-package ownership to consult, and the scope has to be the directories the plugin owns outright. Within those, the sync reads the runtimes each file declares and **moves anything below the pin**, including a `_LATEST` alias (whose value comes from `aws-cdk-lib`, not this repo). A stale runtime inside a directory the plugin owns is drift, so `NODEJS_18_X` there *is* upgraded. A runtime **at or ahead of the pin** is left as the user's deliberate choice, and a Lambda anywhere outside those two directories is never touched.

**Unit / build / lint** — `@aws/nx-plugin:test` **215 files, 3894 tests pass** (16 in the resolution module, 22 in the runtime sync, 6 in the drift guard). `build --all` and `lint --all --configuration=fix` both green, with no unrelated churn.

**The drift guard catches the original bugs.** Reintroducing `Runtime.NODEJS_LATEST` in the RDB CDK template failed 2 of 3 guard tests; reintroducing `runtime = "nodejs22.x"` in the REST Terraform template failed the third, naming the offending file and value. Restored after.

**Deployed to AWS on both providers.** Two fresh workspaces (`--iac=cdk`, `--iac=terraform`) on the locally compiled plugin, each with a standalone `ts#lambda-function` and a tRPC REST API, both deployed for real. Workspace names carry a `8deb14` suffix so the vended Terraform state key could not collide with a concurrent session.

`aws lambda get-function-configuration` — **CDK**:

```json
{
    "FunctionName": "rt-cdk-8deb14-infra8deb14-san-RuntimeProbeFD7A7467-wo0sjdYpFAtI",
    "Runtime": "nodejs24.x",
    "Handler": "index.handler"
}
{
    "FunctionName": "rt-cdk-8deb14-infra8deb14-Probeapi8deb14echoHandle-o3CyC1JQGibw",
    "Runtime": "nodejs24.x",
    "Handler": "index.handler"
}
```

**Terraform**:

```json
{
    "FunctionName": "fnhost8deb14-runtime-probe-a3324t65",
    "Runtime": "nodejs24.x",
    "Handler": "index.handler"
}
{
    "FunctionName": "Probeapi8deb14Handler-uazl2fiy",
    "Runtime": "nodejs24.x",
    "Handler": "index.handler"
}
```

All four report `nodejs24.x` — the providers agree in deployed reality, not just in the templates. Each invoked with `StatusCode: 200` and no `FunctionError`; both tRPC handlers returned the real payload `{"result":{"data":{"message":"runtime-check"}}}`.

**The Python bundle targets the runtime, not the build machine.** The build machine runs **Python 3.11**, yet the built bundle contains **`cp314`** wheels:

```
=== compiled-extension ABI tags ===
  cpython-314-x86_64-linux-gnu
=== dist-info WHEEL tags (non-pure) ===
  pydantic_core-2.46.4.dist-info: cp314-cp314-manylinux_2_17_x86_64, cp314-cp314-manylinux2014_x86_64
  wrapt-2.3.0.dist-info: cp314-cp314-manylinux_2_5_x86_64, cp314-cp314-manylinux1_x86_64, cp314-cp314-manylinux_2_28_x86_64
```

`--python-version` is what does this: the same requirements resolved with `--python-version 3.12` yield `cp312-cp312-…` wheels, with `3.14` yield `cp314`. A deployed Python Lambda confirms the interpreter that loads them:

```json
{
    "FunctionName": "pyfn8deb14-py-probe-je27htsk",
    "Runtime": "python3.14",
    "Handler": "rt4_8deb14_pyfn8deb14.py_probe.lambda_handler"
}
```

```
INIT_START Runtime Version: python:3.14.mainline.v63
{"level":"INFO","message":"Received event","service":"PyProbe"}
```

Invoked `StatusCode: 200`, no `FunctionError`.

**Teardown verified.** Each deployment was destroyed from its own project (`terraform destroy`; `cdk destroy`) — never `bootstrap-destroy`, which would wipe other sessions' state from the shared bucket. Afterwards `describe-stacks` returns "does not exist", and `list-functions`, `get-rest-apis` and the S3 listing return no `8deb14` resources. Only each workspace's own state key was removed from the shared state bucket, which still holds the other 26 objects.

**Migration tested against published `rc.81`.** A workspace generated on the published version (whose Node lambda really was `nodejs22.x`), with the Python values wound back to 3.13, then migrated with the local build. Every semantic value then matched a freshly generated workspace — the two remaining textual diffs were HCL alignment only, which `terraform fmt` normalises away. Re-running the sync reported **zero changes and no `nextSteps`**.

That comparison earned its keep: it caught two gaps the unit tests missed — the per-project `.python-version` (only the root was synced) and the bundle `--python-version` in an already-generated workspace. Both fixed, with tests.

**Scoping verified against that real workspace:**

| Fixture | Expected | Result |
|---|---|---|
| `packages/my-own-infra/src/my-lambda.tf` on `nodejs22.x` (user project) | untouched | **`nodejs22.x`** — untouched, absent from the change list |
| `common/terraform/.../tsfn8deb14-node-probe.tf` (owned) | synced | `nodejs22.x` → **`nodejs24.x`** |
| `common/terraform/.../pyfn8deb14-py-probe.tf` (owned) | synced | `python3.13` → **`python3.14`** |

The user-owned file was left alone *while an owned file on the very same runtime was rewritten in the same run* — the scoping discriminates on location, not on value.

**A generated Terraform project builds green.** `build` on the shared Terraform project runs `fmt`, `checkov` and `test`, all passing with the vended `nodejs24.x`.

### Issue # (if applicable)

N/A

### Checklist

----


packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -641,6 +655,287 @@ const syncTerraformProviders = async (tree: Tree): Promise<string[]> => {
...
+
+const VENDED_TF_NODE_RUNTIMES = ['nodejs22.x', 'nodejs24.x'] as const;
+
+/** Python runtimes this plugin has vended, in each provider's spelling. */
+const VENDED_CDK_PYTHON_RUNTIMES = [
+  'Runtime.PYTHON_3_13',
+  'Runtime.PYTHON_3_14',
+] as const;
+
+const VENDED_TF_PYTHON_RUNTIMES = ['python3.13', 'python3.14'] as const;
```
cogwirrel: remove this - otherwise we need this to be kept up to date automatically too.

We should still match as strictly as possible, eg `runtime: Runtime.NODEJS_xxxx`

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -346,6 +346,91 @@ export const CONTAINER_REPOSITORIES = {
...
+ * against `nodejs24.x` — so each provider held its own copy and the two drifted:
+ * the same generator with the same options vended a different Node major
+ * depending on `--iac`. Deriving both from the version here makes a bump one
+ * edit that moves them together.
+ *
+ * Pinned to an exact major rather than tracking CDK's `NODEJS_LATEST` alias,
+ * whose value moves with `aws-cdk-lib` and would silently reopen the drift on
+ * the next dependency bump. The cost is that this has to be bumped by hand,
+ * which is the point: the runtime a user's functions execute on is a deliberate,
+ * reviewable choice.
```
cogwirrel: remove this part of the comment, just the first sentence will do.

scripts/update-versions.ts
```diff
@@ -427,6 +429,153 @@ const getUpdatedJavaVersions = async (): Promise<Record<string, string>> =>
     ),
   );
 
+/** The AWS documentation page listing every managed Lambda runtime. */
+const LAMBDA_RUNTIMES_DOC_URL =
+  'https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html';
```
cogwirrel: Docs are too brittle. use CDK's "latest" (making sure that cdk is just a dev dependency in our repo for scripts, not shipped as a dependency in the nx plugin itself):

https://github.com/aws/aws-cdk/blob/b8be853177df0d959662825c0c007bfc54f26171/packages/aws-cdk-lib/aws-lambda/lib/runtime.ts

use Runtime.ALL and look for latest python/node.

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -346,6 +346,91 @@ export const CONTAINER_REPOSITORIES = {
...
+/**
+ * Patch of the CPython release uv pins as the project interpreter.
+ *
+ * A Lambda runtime names only `major.minor` (`python3.14`) because Lambda patches
+ * the interpreter for you, but uv pins an exact `major.minor.patch`. Only the
+ * patch lives here — the `major.minor` comes from
+ * {@link LAMBDA_RUNTIME_VERSIONS}, so the interpreter a project builds against
+ * cannot drift from the runtime it deploys onto.
+ */
+export const PYTHON_RUNTIME_PATCH = '0';
```
cogwirrel: This needs to be managed by the update versions workflow too - always latest patch for the minor from lambda, determined from `uv` itself if we have to specify this in pyproject.toml/.python-version

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Done — removed the hardcoded `VENDED_*_RUNTIMES` lists entirely. The sync now reads the runtimes each file actually declares and moves anything below the pin, so there is nothing left to keep up to date by hand.

Matching stayed strict as you asked, anchored on the assignment:

- CDK: `runtime:\s*((?:lambda\.)?Runtime\.((?:NODEJS|PYTHON)_[A-Z0-9_]+))`
- Terraform: `runtime\s*=\s*"((?:nodejs|python)[0-9][^"]*)"`

So a `Runtime` reference that isn't a runtime assignment is untouched — added a test for exactly that (`export const supported = [Runtime.NODEJS_18_X, Runtime.NODEJS_20_X]` is left byte-identical). A `_LATEST` alias is still pinned, since its value comes from `aws-cdk-lib` rather than from this repo, and a runtime *ahead* of the pin is left alone as the user's choice.

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Done — trimmed to the first point, plus a line noting it's resolved by the version update workflow:

```ts
/**
 * The managed Lambda runtimes generated infrastructure targets, as the single
 * source both IaC providers derive from.
 *
 * CDK and Terraform name the same runtime differently — `Runtime.NODEJS_24_X`
 * against `nodejs24.x` — so deriving both from the version here keeps a bump one
 * edit that moves them together. Resolved by the version update workflow.
 */
```

The "pinned by hand, which is the point" reasoning is gone — it's no longer accurate now that the workflow resolves it.

scripts/update-versions.ts
nx-plugin-for-aws: Agreed and done — the docs scrape is gone. Now reads `Runtime.ALL` from `aws-cdk-lib`.

You were right that this is also a *better* GA signal, not just a less brittle one: `Runtime.ALL` is curated, so preview runtimes simply aren't in it. Verified against the installed `aws-cdk-lib` 2.266.0:

```
=== Runtime.ALL unique identifiers: 50
   includes preview nodejs26.x? false | python3.15? false
=== RESOLVED runtimes: {"node":"24","python":"3.14"}
=== unresolved: []
```

That's a stronger guarantee than the deprecation-date heuristic I had — the raw botocore enum *does* list `nodejs26.x`/`python3.15`, so the curation is doing real work here.

On the dependency placement: `aws-cdk-lib` (plus its `constructs` peer) is a **root devDependency only**. Confirmed it stays out of the shipped plugin:

```
plugin deps has aws-cdk-lib: false
plugin devDeps has aws-cdk-lib: false
plugin peerDeps: {"nx":"23.1.1"}
```

Two details worth flagging:

1. `Runtime.ALL` contains duplicate names — `NODEJS_LATEST` aliases `nodejs24.x`, so it appears twice. Deduped before selecting.
2. Identifiers without a comparable version (`nodejs`, `nodejs4.3-edge`) and other languages (`java21`, `provided.al2023`) are skipped rather than mis-parsed; there are tests for each.

A failure is still isolated rather than fatal (the point from your earlier round): if the import fails, the current pins are kept, a warning is logged, and it's reported in the PR body.

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Done — resolved from `uv` by the same workflow, so it's no longer a hand-maintained value.

`uv python list --all-versions --output-format json` gives structured `version_parts`, and the patch is taken as the highest for whichever minor the run lands on. Verified against real `uv`:

```
=== RESOLVED runtimes: {"node":"24","python":"3.14"}
=== RESOLVED patch for 3.14: 7
=== full interpreter: 3.14.7
```

So the next weekly run bumps `3.14.0` → `3.14.7`, which also removes the known limitation I'd flagged in the description.

Three things I was careful about:

- **Resolved for the minor the run lands on**, not the one it started from — it takes `updatedLambdaRuntimeVersions.python`, so a minor bump and its patch move together in the same PR rather than the patch lagging a week.
- **Pre-releases and free-threaded builds excluded** — uv lists `3.15.0rc1` and `+freethreaded` variants, and a generated project needs the default build of a final release. Tested.
- **`PYTHON_RUNTIME_PATCH` is a bare `const`**, not a table entry, so `applyUpdatedVersions` doesn't reach it; it gets its own GritQL rewrite and its own report line.

Failure is isolated the same way as the runtimes: if `uv` can't be queried or lists no matching release, the current patch is kept with a warning rather than taking the whole version update down.

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Converted to GritQL for detection, with regex only parsing the value GritQL binds:

```ts
const RUNTIME_ASSIGNMENT = '`runtime: $value`';
const TF_RUNTIME_ASSIGNMENT = 'language hcl\n`runtime = $value`';
```

Then `readCdkRuntime` / `readTerraformRuntime` parse the bound value (`Runtime.NODEJS_22_X`, `nodejs22.x`) to get the version out. Both patterns work — I probed them before committing:

```
PATTERN "`runtime: $value`"            => ["runtime: Runtime.NODEJS_22_X", "runtime: lambda.Runtime.PYTHON_3_13"]
HCL PATTERN "language hcl\n`runtime = $value`" => ["runtime = \"nodejs22.x\"", "runtime = \"python3.13\""]
```

Your point about the regex doing structural work it couldn't do was borne out — the probe file included a commented-out `// runtime: Runtime.NODEJS_18_X` and GritQL correctly skipped it, where `\bruntime:\s*` would have matched. I added tests for that under both providers, plus a non-Lambda `{ runtime: 'edge' }` property. All three would have been claimed by the old regex.

On the silent-`[]` risk you flagged: `captureAllGritQL` returns `[]` both for "no runtimes" and "pattern failed", so `declaredRuntimes` distinguishes them — if the file mentions a runtime assignment but nothing matched, it returns `undefined` and the file is reported as diverged rather than skipped. A parse failure can't quietly turn this into a no-op.

Behaviour is unchanged: still anchored on the assignment, `_LATEST` always ours to pin, at-or-ahead-of-pin left alone. The `[Runtime.NODEJS_18_X, Runtime.NODEJS_20_X]` byte-identical guard still passes through the GritQL path — 22 tests green.

packages/nx-plugin/src/utils/versions.ts
nx-plugin-for-aws: Set to the value my own resolver returns, so source and resolver agree by construction. Re-ran the query rather than taking a number on faith:

```
CPython 3.14 patches available: 0,1,2,3,4,5,6,7
resolveLatestPythonPatch => 7
full interpreter => 3.14.7
```

So `PYTHON_RUNTIME_PATCH = '7'`. A freshly generated workspace now writes `3.14.7` to both the root and per-project `.python-version`, with `requires-python` still `>=3.14`.

Re-verified the two claims that matter after the bump:

- **Bundle still resolves `cp314`** — `cpython-314-x86_64-linux-gnu.so`, and `pydantic_core` / `wrapt` at `cp314-cp314-manylinux*`.
- **Deployed function still reports a 3.14 interpreter** — deployed for real, `Runtime: python3.14`, invoked `StatusCode: 200` with no `FunctionError`, and the log confirms `INIT_START Runtime Version: python:3.14.mainline.v63`. Torn down afterwards (19 resources destroyed, `list-functions` empty).

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -641,6 +656,352 @@ const syncTerraformProviders = async (tree: Tree): Promise<string[]> => {
...
+  );
+
+  if (captured.length === 0) {
+    const contents = tree.read(path, 'utf-8') ?? '';
+    const assigns = hcl
+      ? /\bruntime\s*=/.test(contents)
+      : /\bruntime:/.test(contents);
+    // Mentions a runtime assignment but matched none: the pattern did not apply.
+    return assigns ? undefined : [];
+  }
```
cogwirrel: can we just capture the `$value` and then simplify reading to avoid re-parsing `runtime:` or `runtime =`?

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -641,6 +656,352 @@ const syncTerraformProviders = async (tree: Tree): Promise<string[]> => {
...
+ * reason the pin itself does.
+ *
+ * A runtime below the pin is ours to move, as is a `_LATEST` alias — its value is
+ * decided by `aws-cdk-lib`, which is the drift this replaces. A runtime at or
+ * ahead of the pin is the user's and is left. One in a shape the rewrite cannot
+ * reach is reported instead.
+ *
+ * @returns the files changed, and the owned files still holding an older runtime
+ */
+const syncLambdaRuntimes = async (
```
cogwirrel: Refactor this and the helpers to a `sync-lambda-runtimes.ts`

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -641,6 +656,352 @@ const syncTerraformProviders = async (tree: Tree): Promise<string[]> => {
...
+
+/**
+ * Sync the interpreter a uv project pins to the Lambda Python runtime.
+ *
+ * `.python-version` holds the exact interpreter and `[project].requires-python`
+ * its lower bound; Ruff's `target-version` is derived from the latter, so leaving
+ * these behind lints against a different version than the function runs on. Only
+ * moved forward, and only from a value this plugin vended.
+ */
+const syncProjectPythonVersion = (tree: Tree): string[] => {
```
cogwirrel: Refactor this and python bundle version to a `sync-python-language-version.ts` in `version-upgrade-migration`

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -641,6 +656,352 @@ const syncTerraformProviders = async (tree: Tree): Promise<string[]> => {
...
+ */
+const syncLambdaRuntimes = async (
+  tree: Tree,
+): Promise<{ updated: string[]; diverged: string[] }> => {
+  const files: { path: string; hcl: boolean }[] = [];
+
+  visitNotIgnoredFiles(tree, '.', (path) => {
+    if (!isOwnedInfraFile(path)) {
+      return;
+    }
```
cogwirrel: instead of visiting all not ignored files and skipping, can we start in the 2 dirs?

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
```diff
@@ -641,6 +656,352 @@ const syncTerraformProviders = async (tree: Tree): Promise<string[]> => {
...
+    if (path.endsWith('.ts')) {
+      files.push({ path, hcl: false });
```
cogwirrel: if we start in each dir above, we already know which is hcl and which isn't. instead of a boolean let's use `'terraform' | 'cdk'` so it's easier to extend in future

scripts/update-versions.ts
```diff
@@ -540,6 +617,49 @@ const main = async () => {
...
+      await applyGritQL(
+        tree,
+        'packages/nx-plugin/src/utils/versions.ts',
+        `\`export const PYTHON_RUNTIME_PATCH = '${PYTHON_RUNTIME_PATCH}'\` => \`export const PYTHON_RUNTIME_PATCH = '${updatedPythonPatch}'\``,
+      );
+      lambdaRuntimeChanges.push({
+        name: 'python runtime patch',
+        oldVersion: PYTHON_RUNTIME_PATCH,
+        newVersion: updatedPythonPatch,
+      });
```
cogwirrel: Report this separately as Project Python Version, and include the full major.minor.patch

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Done — and you were right that the helper needed extending rather than the caller working around it.

`captureAllGritQL` returns `node.text()` (the whole match), so it couldn't express this. But `node.var()` *is* reachable — `captureGritQLVariable` already uses it for the single-match case — so I added `captureAllGritQLVariable` to `utils/ast.ts` as the plural sibling. Probed it before committing:

```
CDK $value => ["Runtime.NODEJS_22_X", "lambda.Runtime.PYTHON_3_13", "'edge'"]
HCL $value => ["\"nodejs22.x\"", "\"python3.13\""]
```

The binding is the value alone, so `readCdkRuntime` / `readTerraformRuntime` no longer strip a `runtime:` / `runtime = ` prefix. Regex now only ever sees the isolated scalar:

```ts
/^(?:lambda\.)?Runtime\.((?:NODEJS|PYTHON)_[A-Z0-9_]+)$/
/^"((?:nodejs|python)[0-9][^"]*)"$/
```

**On the two content probes:** they're gone entirely. `captureAllGritQLVariable` returns `undefined` when a pattern can't be applied, which `[]` cannot express — so that distinction (which you rightly said to keep) is now carried by the helper's return type rather than by a `/\bruntime:/` text test. No structural decision is left to a regex.

**The `/"([^"]+)"/` at the provider helper was reachable**, so I fixed it too: `providerVersionPattern` already binds `$version`, so `declaredProviderVersion` reads that binding and just unquotes it, instead of re-extracting from the matched attribute text.

Added six tests for the new helper (bindings across matches, sigil-optional variable name, no match inside a comment, another language, and `undefined` for both an inapplicable pattern and a missing file).

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Done — `syncLambdaRuntimes` and its helpers (`DeclaredRuntime`, the GritQL patterns, `cdkMemberVersion`, `readCdkRuntime`, `readTerraformRuntime`, `declaredRuntimes`, `vendedRuntime`) now live in `version-upgrade-migration/sync-lambda-runtimes.ts`, matching the existing `sync-lambda-runtimes.spec.ts`.

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Done — `syncProjectPythonVersion` and `syncPythonBundleVersion` now live in `version-upgrade-migration/sync-python-language-version.ts`.

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Done — `syncLambdaRuntimes` now takes the owned dirs and visits each directly, so `isOwnedInfraFile` is gone:

```ts
const OWNED_INFRA_DIRS: Readonly<Record<VendedIac, string>> = {
  cdk: `${PACKAGES_DIR}/${SHARED_CONSTRUCTS_DIR}`,
  terraform: `${PACKAGES_DIR}/${SHARED_TERRAFORM_DIR}`,
};
```

On the other two sweeps: I checked rather than assumed, and they should stay workspace-wide. `uv` writes a `.python-version` per Python project (verified on a generated workspace — both the root and `packages/<project>/.python-version`), and a `bundle` target's `--python-version` lives in whichever project owns the bundle. Neither is confined to the two infrastructure directories, so scoping them there would miss the files that matter. They're in `sync-python-language-version.ts` now, where the module comment records why the scope differs.

packages/nx-plugin/src/utils/version-upgrade-migration/sync-vended-versions.ts
nx-plugin-for-aws: Done — `hcl: boolean` is now `VendedIac = 'cdk' | 'terraform'`, threaded through `declaredRuntimes` and the read helpers.

And yes, it composes with the previous point exactly as you said: the provider comes from the directory being visited, so it's never inferred from the file. The GritQL patterns are keyed by it too, which reads better than a branch:

```ts
const RUNTIME_ASSIGNMENT: Record<VendedIac, string> = {
  cdk: '`runtime: $value`',
  terraform: 'language hcl\n`runtime = $value`',
};
```

Adding a third provider is now a key in that record plus its file extension, rather than another boolean branch.

scripts/update-versions.ts
nx-plugin-for-aws: Done — its own `Project Python Version` group, reporting the full `major.minor.patch`:

```
Lambda Runtimes
- python 3.14 -> 3.15

Project Python Version
- python 3.14.7 -> 3.15.2
```

On where the failure notes sit — you were right that these split. A uv failure is now reported under the new group, not under Lambda Runtimes:

```
Lambda Runtimes
- Found no node runtime in the aws-cdk-lib runtime list, so it keeps its pinned runtime (24)

Project Python Version
- Could not list Python versions with uv, so the project Python version stays at 3.14.7
```

So `getUpdatedPythonRuntimePatch` returns `{ patch, unresolved? }` rather than a bare string, and both of its failure paths (uv unavailable, and uv listing no release for that minor) name the full version they're keeping rather than the bare patch digit.

cogwirrel (approved):

PR #1132, opened 2026-08-26
fix(ts#rdb,py#rdb): let RDS manage the Aurora admin password on Terraform (no migration)

nx-plugin-for-aws:
### Reason for this change

On Terraform, the Aurora **admin database password was written into Terraform state**. The core module generated it with `random_password` and stored it via `aws_secretsmanager_secret_version`, so both the generated value and the full secret string were persisted in plaintext.

On CDK it never materialised — `Credentials.fromGeneratedSecret(...)` has Secrets Manager mint the password, so it never appears in the template or in CloudFormation state.

This compounded a known weakness: the vended state bucket is `${account}-tf-state-${region}`, shared account-wide and encrypted AES256 only. Anyone with read access to it held admin database credentials. (The bucket's encryption is a separate concern, not changed here.)

Confirmed on a real deployment of the pre-change module — the password literal was present in state:

```
BEFORE migration
  password literal occurrences in Terraform state: 2
  secret-bearing resource types: ['aws_secretsmanager_secret', 'aws_secretsmanager_secret_version', 'random_password']
```

### Description of changes

`manage_master_user_password = true` on `aws_rds_cluster`, with `master_user_secret_kms_key_id` pointing at the module's existing CMK. RDS generates the password, stores it in Secrets Manager and rotates it every 7 days; Terraform only ever sees the secret ARN. This is the direct analogue of CDK's behaviour.

Viability against the three consumers:

1. **The vended migration / create-db-user handlers.** An RDS-managed secret holds **only** `username` and `password` (verified empirically). The handlers previously read `host`/`port`/`dbname` off the secret, so they now take those from their environment, fed by a new `writer_endpoint` output plus the existing `cluster_port` and `database_name`. `writer_endpoint` is the cluster endpoint rather than `database_host` (which resolves to the proxy) because these handlers connect with admin credentials, which the proxy does not broker.
2. **The RDS Proxy.** Unaffected — it uses end-to-end IAM auth and never read the admin secret. Confirmed on a deployed proxy: `"Scheme": "IAM_AUTH", "Auth": []`.
3. **The CloudFormation rotation stack.** Conflicts, and becomes unnecessary. Secrets Manager rejects a rotation Lambda on a service-managed secret: `Rotation Lambda ARN is not supported for a service-managed secret`. Since managed secrets rotate natively, the stack, its two security groups, its rules, the `enable_credential_rotation` variable and the core module's now-unused `lambda_subnet_ids` are removed. That also drops the `CKV2_AWS_57` checkov skip the old secret carried.

The KMS key policy gains a `secretsmanager.amazonaws.com` statement, since Secrets Manager now encrypts the managed secret with the module's CMK.

The `secret_arn` output uses `one(...master_user_secret[*].secret_arn)` rather than a `[0]` index — the list is empty until the cluster is modified, and `[0]` fails the plan there.

CDK is unchanged; it never had the exposure.

### No migration — newly generated workspaces only

**There is deliberately no migration.** I wrote one, then tested it as a real before/after deployment, and it does not transition a live cluster safely.

The cluster change itself is fine: `aws_rds_cluster.database` plans as `updated in-place`, keeps its `cluster_identifier`, and the data survives. But the first `apply` **errors partway and leaves the database unreachable**. It deletes the old secret and switches the cluster, then fails in the same apply:

```
module.pgdb.module.aurora.random_password.master_password: Destruction complete after 0s
module.pgdb.module.aurora.aws_secretsmanager_secret.credentials: Destruction complete after 0s
...
Error: Provider produced inconsistent final plan

When expanding the plan for module.pgdb.aws_iam_role_policy.migration_handler_access
to include new values learned so far during apply, provider "hashicorp/aws" produced
an invalid new value for .policy: was ...\"Resource\":[null]...
but now ...\"Resource\":[\"arn:aws:secretsmanager:...rds!cluster-8ecb8a9a-...\"]...
```

Both handler IAM policies hit this, so their `DATABASE_SECRET_ARN` and secret-read grants are never updated on that apply — neither Lambda was modified. A second `apply` converges (`2 added, 4 changed, 0 destroyed`) and everything works, but between the two the handlers point at a deleted secret. A migration that half-works against a live database is worse than none, so existing workspaces keep the old shape (the vended files are `KeepExisting`).

The vended `utils` templates are back-compatible: the connection overrides are spread over the secret, so a secret that still carries `host`/`port`/`dbname` works unchanged.

Note that removing `enable_credential_rotation` is breaking for a root configuration that passes it (`An argument named "enable_credential_rotation" is not expected here`). With no migration an existing workspace never picks up the new module, so this only affects newly generated ones, which don't pass it.

`pnpm nx run @aws/nx-plugin:test` (3844 passing), `pnpm nx run-many --target build --all` (including the `terraform test` that #1130 added to `build`), and `pnpm nx run-many --target lint --all --configuration=fix` all pass.

**Deployed to AWS twice.**

**1. Newly generated workspace, both engines** (PostgreSQL 17.7 with RDS Proxy enabled, MySQL 8.0 with it disabled): `Apply complete! Resources: 111 added, 0 changed, 0 destroyed.`

Password absent from state — fetching each cluster's real password from its managed secret and grepping:

```
secret: rds!cluster-cb9f85fb-...-7FgbKC
  password length: 28 chars (value withheld)
  occurrences in 'terraform state pull': 0
  occurrences in 'terraform show -json': 0
secret: rds!cluster-0fb31d4d-...-7I8zF9
  occurrences in 'terraform state pull': 0
  occurrences in 'terraform show -json': 0

=== resource types in state that could hold secret material ===
NONE
```

Both handlers ran during the apply on both engines (`StatusCode: 200`), then an in-VPC probe made real queries:

```json
{
  "managedSecretKeys": ["password", "username"],
  "postgresAdmin": { "currentUser": "dbadmin", "serverVersion": "PostgreSQL 17.7", "wroteAndRead": "managed-password-works" },
  "postgresProxyIam": { "currentUser": "db_7km1bb2j", "readThroughProxy": "managed-password-works", "usedPassword": false },
  "mysqlAdmin": { "currentUser": "dbadmin@%", "serverVersion": "8.0.44", "wroteAndRead": "managed-password-works" }
}
```

`postgresProxyIam` reading back the row written by `postgresAdmin` proves the proxy authenticates and reaches the same database.

**2. Before/after transition test on the published rc.84** (the basis for dropping the migration): applied the pre-change module for real (68 resources), seeded a row, ran the migration, applied again. Findings above. For completeness, after the second apply the pre-migration row was intact and the old password was gone:

```
AFTER migration
  pre-migration password literal still in state: 0
  secret-bearing resource types: NONE
  cluster_identifier: pgdb-aurora-klm7bisd   (unchanged - no replacement)
  manage_master_user_password: True
```

```json
{ "rowsSurvivingMigration": [{ "id": 1, "note": "written-before-migration" }],
  "resolvedFrom": { "host": "environment", "port": "environment", "database": "environment" } }
{ "currentUser": "db_iolv32fn", "usedPassword": false,
  "rows": [{"id":1,"note":"written-before-migration"},{"id":2,"note":"written-after-migration"}] }
```

Both deployments were fully destroyed afterwards, including probe functions, roles, security-group rules and the Aurora service-created log groups. Only my own state keys were removed; the shared bootstrap module was left untouched.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/snippets/rdb/admin-credentials.mdx
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+
+An RDS-managed secret holds a JSON object with `username` and `password` only — no connection details. Take the host, port and database name from the module's `writer_endpoint`, `cluster_port` and `database_name` outputs.
+</Fragment>
+</Infrastructure>
+
+:::caution[Rotating manually]
+Rotating the admin password outside of Aurora's schedule is done through the RDS API (`modify-db-cluster --rotate-master-user-password --apply-immediately`), not through Secrets Manager. Attaching your own rotation function to an RDS-managed secret is rejected.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/snippets/rdb/architecture.mdx
```diff
@@ -35,4 +35,7 @@ app -> proxy: SQL (IAM auth)
 proxy -> aurora
 migrations -> aurora: Schema migrations
 migrations -> secrets: Admin credentials
+aurora -> secrets: Generates and rotates\nadmin credentials
 ```
+
+The admin credentials are generated and rotated by Aurora itself in Secrets Manager, so the password is never handled by your infrastructure code — it does not appear in a CloudFormation template or in Terraform state. Your application does not use these credentials: it connects as a least-privilege database user via IAM authentication, and only the migration and create-db-user handlers read the secret.
```
cogwirrel: remove this paragraph

packages/nx-plugin/src/migrations/latest/terraform-aurora-rds-managed-master-password/migration.ts
```diff
@@ -0,0 +1,277 @@
...
+    hcl('`variable "enable_credential_rotation" { $_ }` => ``'),
+  );
+  await applyGritQL(
+    tree,
+    filePath,
+    hcl('`lambda_subnet_ids = var.lambda_subnet_ids` => ``'),
+  );
+};
+
+export default async function migration(
```
cogwirrel: we must test deploy before and after. looks like utils.ts/py aren't changed by the migration?

docs/src/content/docs/en/snippets/rdb/admin-credentials.mdx
nx-plugin-for-aws: Removed. The manual-rotation caution is gone from the English snippet, and I removed the eight bot-generated translations of this file too so the translate bot regenerates them from the corrected English rather than carrying the deleted paragraph.

docs/src/content/docs/en/snippets/rdb/architecture.mdx
nx-plugin-for-aws: Removed — the diagram now stands on its own. The `aurora -> secrets` edge already conveys that Aurora generates and rotates the credentials, so the prose was redundant. The eight translations of this snippet have had the paragraph dropped as well.

packages/nx-plugin/src/migrations/latest/terraform-aurora-rds-managed-master-password/migration.ts
nx-plugin-for-aws: You were right on both counts, and testing the deploy transition is what killed the migration. **I've removed it** — the template change now ships for newly generated workspaces only, same call as #1128 and #1143.

**On `utils.ts`/`utils.py`:** correct, they weren't touched. Those files are `KeepExisting` (`ts/rdb/generator.ts:191`), so post-migration the secret holds only `username`/`password` and `host`/`port`/`dbname` all came back `undefined` — a runtime break in user code invisible to plan/validate. I did extend the migration to carry them (matching on the verbatim-return shape, reporting diverged copies via `nextSteps`), and it worked: `nx migrate --run-migrations` reported `UPDATE packages/pgdb/src/utils.ts` and the result was byte-identical to what the generator vends. So the gap was closable — but the deploy test then found something worse.

**The deploy transition, done for real.** rc.84 workspace → `terraform apply` (68 resources, live cluster, `random_password` literal present in state 2×) → seeded a row → ran the migration → `apply` again.

Two findings:

1. A `[0]` index on `master_user_secret` fails the plan outright on the transition — the list is still empty at plan time. `one(...[*])` fixes that and I've kept it in the template.

2. **The first apply errors and leaves the database unreachable.** With the plan fixed, `apply` deletes the old secret and switches the cluster, then fails in the same apply:

```
module.pgdb.module.aurora.random_password.master_password: Destruction complete after 0s
module.pgdb.module.aurora.aws_secretsmanager_secret.credentials: Destruction complete after 0s
...
Error: Provider produced inconsistent final plan

When expanding the plan for module.pgdb.aws_iam_role_policy.migration_handler_access
to include new values learned so far during apply, provider "hashicorp/aws" produced
an invalid new value for .policy: was ...\"Resource\":[null]...
but now ...\"Resource\":[\"arn:aws:secretsmanager:...rds!cluster-8ecb8a9a-...\"]...
```

Both handler IAM policies hit this, so their `DATABASE_SECRET_ARN` and secret-read grants are never updated on that apply — I confirmed neither `aws_lambda_function.create_db_user` nor `migration_handler` was modified. The old secret is gone and the handlers still point at it. A second `apply` converges (`2 added, 4 changed, 0 destroyed`) and everything then works, but that intermediate window is a broken database, so per your guidance the migration is out rather than reshaped to dodge it.

For the record, the parts that *did* hold, since they matter if this is revisited: no replacement (`aws_rds_cluster.database will be updated in-place`, same `cluster_identifier` before and after — the only replacements were the create-db-user zip and some `null_resource`s); the removed resources left state cleanly (`secret-bearing resource types: NONE`, old password literal `0` occurrences); and after the second apply the pre-migration row was intact, read both as admin (`resolvedFrom: host/port/database = environment`) and as the app user over IAM auth (`currentUser: db_iolv32fn, usedPassword: false`).

**One thing worth flagging for the no-migration path:** removing `enable_credential_rotation` from the module is breaking for anyone whose root `main.tf` passes it — I hit `An argument named "enable_credential_rotation" is not expected here` on `terraform validate`. With no migration, an existing workspace never picks up the new module, so it isn't hit in practice; but if you'd prefer the variable retained as a deprecated no-op for forward-compatibility, say so and I'll add it back.

The vended `utils` templates are back-compatible either way — the overrides are spread over the secret, so a secret that still carries `host`/`port`/`dbname` works unchanged.

cogwirrel (approved):

PR #1138, opened 2026-08-27
chore(ts#api,py#api): avoid empty IAM policies for isolated Terraform APIs

nx-plugin-for-aws:
### Reason for this change

The `terraform-deploy` smoke test has been failing on `main` since #1124 introduced the `isolated` integration pattern for Terraform APIs. Every isolated API module failed to apply:

```
Error: putting IAM Role (MyApi-echo-role-...) Policy (MyApi-echo-...-policies):
operation error IAM: PutRolePolicy, https response error StatusCode: 400,
MalformedPolicyDocument: Syntax errors in policy.

  with module.my_api.aws_iam_role_policy.lambda_additional_policies["echo"],
```

The `isolated` pattern declares one `aws_iam_role_policy` per operation with `for_each = local.operations`. When neither `appconfig_application_arn` nor `additional_iam_policy_statements` is wired, `local.lambda_policy_statements` is empty, so each policy rendered `Statement = []`, which IAM rejects.

This hit 9 API modules in the smoke test (`my_api`, `my_api_http`, `my_api_custom`, `my_api_custom_http`, `my_smithy_api`, `py_api`, `py_api_http`, `py_api_custom`, `py_api_custom_http`). The `shared` pattern was unaffected because it uses `count = length(local.lambda_policy_statements) > 0 ? 1 : 0`, which correctly skips the resource.

### Description of changes

Mirroring the `shared` guard onto `for_each` does **not** work. The AppConfig ARN is only known after apply, so gating the instance keys on the statement count makes them unknown at plan time and Terraform fails with:

```
Error: Invalid for_each argument
  │ local.lambda_policy_statements is a list of object, known only after apply
  The "for_each" map includes keys derived from resource attributes that
  cannot be determined until apply
```

Terraform permits unknown values in resource *attributes* but not in `for_each` *keys*. So the resource is still declared once per operation (keys stay static and plan-known), and the policy **document** falls back to a no-op deny when there are no statements:

```hcl
# IAM rejects an empty statement list, so fall back to a no-op deny
lambda_policy_document_statements = length(local.lambda_policy_statements) > 0 ? local.lambda_policy_statements : [
  {
    Effect   = "Deny"
    Action   = ["appconfig:GetLatestConfiguration"]
    Resource = ["arn:aws:appconfig:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:application/none"]
  }
]
```

This is applied to both the `rest` and `http` API templates. The `shared` branch is untouched and the new local is only emitted for `isolated`, so shared-pattern output is byte-identical.

Unit tests: `pnpm nx run @aws/nx-plugin:test -u` — 3844 passed, 16 snapshots updated. Build and `lint --fix` are clean.

Validated against **real AWS** with a locally compiled + linked plugin, deploying a Terraform workspace with Terraform 1.15.7 (the version CI pins):

1. **Reproduced the CI failure first.** With the pre-fix template, `nx apply infra` failed with the identical `MalformedPolicyDocument` error on `lambda_additional_policies["echo"]`. Separately confirmed the root cause in isolation — `PutRolePolicy` with `{"Version":"2012-10-17","Statement":[]}` returns exactly `MalformedPolicyDocument: Syntax errors in policy`.
2. **Isolated, no appconfig** (the CI case): applies cleanly; role carries the fallback deny; `AWSLambdaBasicExecutionRole` + `AWSXRayDaemonWriteAccess` still attached; SigV4-signed `GET /echo` → `200 {"result":{"data":{"message":"test"}}}`.
3. **Isolated, appconfig wired** (regression check for the plan-time failure): applies cleanly; the inline policy contains the real `appconfig:StartConfigurationSession`/`GetLatestConfiguration` grant scoped to the application ARN — the deny fallback is correctly absent; `GET /echo` → `200`.
4. **All three generator paths deployed together** (76 resources): tRPC isolated, tRPC `shared`, and FastAPI isolated (exercising the `rest` template). All three returned `200`. Verified the shared-pattern role has **no** inline policy (still skipped via `count`) while the FastAPI isolated role carries the fallback.

All provisioned AWS resources were torn down afterwards.

### Issue # (if applicable)

N/A — fixing a failing build on `main`.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/py/fast-api/__snapshots__/generator.terraform.spec.ts.snap
+++ b/packages/nx-plugin/src/py/fast-api/__snapshots__/generator.terraform.spec.ts.snap
@@ -601,7 +601,7 @@ resource "aws_iam_role_policy" "lambda_additional_policies" {
 
   policy = jsonencode({
     Version = "2012-10-17"
-    Statement = local.lambda_policy_statements
+    Statement = local.lambda_policy_document_statements
   })
 }
 
@@ -617,6 +617,15 @@ locals {
       Resource = ["\${var.appconfig_application_arn}/*"]
     }
   ] : [], var.additional_iam_policy_statements)
+
+  # IAM rejects an empty statement list, so fall back to a no-op deny
+  lambda_policy_document_statements = length(local.lambda_policy_statements) > 0 ? local.lambda_policy_statements : [
+    {
+      Effect   = "Deny"
+      Action   = ["appconfig:GetLatestConfiguration"]
+      Resource = ["arn:aws:appconfig:\${data.aws_region.current.region}:\${data.aws_caller_identity.current.account_id}:application/none"]
+    }
+  ]
 }
 
 # CloudWatch Log Group for Lambda
@@ -1417,7 +1426,7 @@ resource "aws_iam_role_policy" "lambda_additional_policies" {
 
   policy = jsonencode({
     Version = "2012-10-17"
-    Statement = local.lambda_policy_statements
+    Statement = local.lambda_policy_document_statements
   })
 }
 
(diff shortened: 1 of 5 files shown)


nx-plugin-for-aws:
Note on CI coverage: `terraform-deploy` is defined in `ci.yml`, which only triggers on push to `main` (and `workflow_dispatch`) — `pr.yml` does not include the deploy smoke tests, so the job that caught this cannot run as a PR check. I dispatched `ci.yml` against this branch to try it, but the deploy role's OIDC trust policy is ref-restricted, so every job that calls `configure-aws-credentials` failed with `Not authorized to perform sts:AssumeRoleWithWebIdentity` before running any test — unrelated to this change.

Instead the fix was validated by deploying to a real AWS account from this branch (Terraform 1.15.7, matching the version CI pins), covering both branches of the conditional and all three generator paths. Details are in the PR description. `terraform-deploy` will exercise this for real on merge.

All 62 PR checks are green.

cogwirrel (approved):

PR #1149, opened 2026-08-28
fix(ts#website,ts#website#auth): assign each website its own dev-server and preview port

AlexTo:
### Reason for this change

Every workspace gets a root `dev` script (`dev: 'nx run-many --target dev'`, part of `CONVENIENCE_SCRIPTS` in `packages/nx-plugin/src/utils/init.ts`) that runs every project's `dev` target concurrently — the standard way to run everything locally, in place since `#806` (2026-06-29). But every website generated by `ts#website` hardcoded `server.port: 4200` and `preview.port: 4300` in `vite.config.mts`, so `pnpm dev` in a workspace with more than one website had them all configured for the same port.

For example, a workspace with a `user-portal` and an `admin-portal`, both with Cognito auth: running `pnpm dev` starts both, and each ends up on its own port. But the Cognito User Pool Client's callback/logout URLs were a hardcoded `['http://localhost:4200', 'http://localhost:4300']` pair regardless of how many websites existed, so Cognito has no idea the second portal is even running on a different port — only whichever portal actually got 4200 can complete the Hosted UI sign-in redirect; the other fails to log in locally.

With this change, `pnpm dev` allow-lists every website's own port, so all of them can sign in locally at the same time.

### Description of changes

- `ts#website` now assigns each website an explicit, unique dev-server port via `assignPort` (same helper already used by API/agent generators for this), and writes it into `vite.config.mts`. The preview port is derived as `devPort + 100` (@nx/react's default relationship between the two) rather than tracked separately.
- `ts#website#auth` reads a website's assigned port(s) back and allow-lists them on the shared identity construct (`LOCAL_CALLBACK_URLS` in `user-identity.ts` for CDK) / module (`local_callback_urls` in `identity.tf` for Terraform), instead of relying on the static two-entry list. Both providers use GritQL for this — the Terraform side via its HCL grammar (`language hcl`).
- Added a migration (`react-website-unique-dev-ports`) that backfills both the per-project port assignment and the callback URL allow-list for existing workspaces, in deterministic (alphabetical) order so re-runs are stable. It pattern-matches before writing and reports anything it can't recognise via `nextSteps` rather than clobbering it.
- Updated the `ts#website#auth` guide and the CORS configuration snippets (CDK/Terraform) to stop stating the callback allowlist / example CORS origins as a fixed `4200`/`4300` pair, since that's no longer true once a workspace has more than one website.

- Unit tests: extended `cognito-auth`'s generator spec with cases covering multiple websites getting distinct ports allow-listed without duplication (CDK and Terraform), and added a full spec for the new migration (applies cleanly, skips + reports diverged shapes, idempotent on re-run).
- Manually validated end-to-end per CONTRIBUTING.md, for both CDK and Terraform:
  - Generated fresh workspaces with 4 websites (all with Cognito auth) against the local build — confirmed distinct ports, correct callback URL allow-listing, and a full `nx run-many --target build --all` pass.
  - Created workspaces on the **published** version, generated the same shape, then linked the local build and ran the migration via a hand-written `migrations.json` (per CONTRIBUTING's guidance for testing unpublished migrations) — confirmed the same correct end state, `pnpm lint` clean, and a full build pass.

### Checklist

----


AlexTo:
Thanks @shsrams for the review. Fixed:

1. Dropped return value: `addIdentityInfra` now returns the ports that couldn't be allow-listed, generator warns instead of failing silently.
2. Terraform couldn't append to an emptied `local_callback_urls` list, fixed with an `or` branch matching the empty array literally (the originally suggested fix didn't actually match the HCL grammar).
3. `includes()` check gave false positives/negatives, replaced with a scoped GritQL match against the array itself.
4. Static 4200/4300 pair: both templates now start with an empty array, ports only get added by real websites.
5. Preview port not reserved in metadata.ports: skipping this one, not worth it until we actually have 100+ websites.
6. No test asserting on actual port values: added one.
7. Comment/description wrongly credited the +100 offset to Vite instead of @nx/react, fixed both.
8. localeCompare isn't deterministic across locales, switched to codepoint comparison.
9. Unused export, removed.

Also tested fresh generation and the migration path end to end for both CDK and Terraform. Ports, callback URLs, idempotency and builds all check out.

shsrams:
The manual translate step (the failing `Translate Documentation` step) only applies to external contributors who add **en-only** docs. This PR already includes complete, up-to-date translations across all 9 locales (en, es, fr, it, jp, ko, pt, vi, zh) for the three changed doc files — Alex did his usual local translation and committed it. And the translated content matches the en change and isn't stale. So, going ahead with the merge

PR #1151, opened 2026-08-29
feat(agentcore-harness): add experimental generator flag and mark agentcore-harness experimental

nx-plugin-for-aws:
### Reason for this change

Some generators land while their shape is still settling — `agentcore-harness` being the current example. Under our normal rule (CONTRIBUTING.md, *What should be a migration*) every change to a generator's output wants a migration so a workspace generated last month converges on what today's generators produce. For a generator we expect to reshape repeatedly while it stabilises, that cost isn't buying users much, but there was no way to say so: nothing in the repo distinguished a settled generator from one still in flux, and nothing told users or agents which was which.

### Description of changes

Adds an `experimental` flag to `generators.json` entries. An experimental generator is **permitted to change without a migration** being published to apply the change to an existing workspace, and `agentcore-harness` is marked as such.

Setting the flag is the whole authoring step — it propagates to every surface that describes generators:

- **`GeneratorInfo`** (`utils/generators.ts`) gains `experimental?: boolean`, read through by `buildGeneratorInfoList` with the same conditional-spread idiom already used for `hidden` / `guidePages`, so the property stays absent on ordinary generators.
- **Docs guide pages** get a caution banner via a new `experimental-banner.astro`, rendered from `markdown-content.astro`. It keys off the page's existing `generator:` frontmatter — the same link that already drives the option-filter bar — so no guide page needed editing, and translated pages pick it up too (localised for all nine locales). Nothing renders when the flag is absent.
- **MCP tools** — `renderGeneratorInfo` emits a `> [!WARNING] Experimental: …` line, which covers both `list-generators` and `generator-guide` since both render their per-generator summary through it. The `> [!WARNING]` form matches the GitHub-style alerts the guide pipeline already emits for `> [!NOTE] Only when …` markers and "Unsupported combination" warnings.
- **Skill / power files** — `scripts/make-skills.ts` appends `(experimental)` in the generators table; `SKILL.md` and `POWER.md` regenerated.

Docs: a new "Experimental generators" section in CONTRIBUTING.md, listed among the exceptions under *What should be a migration* and covering how to set and later remove the flag; a note under *Linking a guide to its generator* that the frontmatter also drives the banner; and a short user-facing note in the upgrading guide.

One thing worth calling out for review: the banner's per-locale strings live inline in the component, because `scripts/translate.ts` only covers MDX content files and `schema-translations.json`. Happy to move them into a translated data file if you'd prefer.

- Two new unit tests: the flag surfaces on `agentcore-harness` and is absent on `ts#project` (`utils/generators.spec.ts`); `renderGeneratorInfo` emits the warning only for experimental generators (`mcp-server/generator-info.spec.ts`). Full suite green — 3847 tests across 212 files.
- `pnpm nx run docs:build` — 1495 pages built, internal link validation passes. Confirmed the banner in the built HTML for `en`, `jp`, `zh` and `fr`, and confirmed it is absent on `agentcore-gateway`.
- Rendered the built site under `astro preview` and checked the banner visually on the harness guide, plus that the option-filter bar is unaffected on a non-experimental guide.
- Rendered `renderGeneratorInfo` directly for an experimental and a non-experimental generator to check the exact MCP text an agent receives.
- `pnpm nx run-many --target lint --all --fix` and the non-docs builds are clean.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/components/experimental-banner.astro
```diff
@@ -0,0 +1,60 @@
...
+  es: `El generador \`${generatorId}\` es experimental. Su comportamiento y la salida generada pueden cambiar en una versión futura sin que se publique una migración para aplicar el cambio a un espacio de trabajo existente, por lo que puede que tengas que aplicar esos cambios tú mismo al actualizar.`,
+  pt: `O gerador \`${generatorId}\` é experimental. Seu comportamento e a saída gerada podem mudar em uma versão futura sem que uma migração seja publicada para aplicar a mudança a um espaço de trabalho existente, portanto você pode precisar reconciliar essas mudanças manualmente ao atualizar.`,
+  zh: `\`${generatorId}\` 生成器为实验性功能。其行为和生成的输出可能在未来版本中发生变化，且不会发布迁移来将变更应用到现有工作区，因此升级时您可能需要自行处理这些变更。`,
+  vi: `Trình tạo \`${generatorId}\` đang ở giai đoạn thử nghiệm. Hành vi và đầu ra được tạo có thể thay đổi trong một bản phát hành tương lai mà không có migration nào được công bố để áp dụng thay đổi vào không gian làm việc hiện có, vì vậy bạn có thể cần tự điều chỉnh những thay đổi đó khi nâng cấp.`,
+};
```
cogwirrel: can we reuse Snippet so that translation will be handled automatically?

nx-plugin-for-aws: Good call — done in 9faa759.

The banner now renders `<Snippet name="experimental-generator" />`, so the wording lives in `docs/src/content/docs/en/snippets/experimental-generator.mdx` and the translation workflow keeps the locale copies in sync (both `translate.config.json`'s `**/*.mdx` include and the workflow's `docs/src/content/docs/en/**` path filter already cover snippets). That drops the component from 60 lines to 28 — it's now just the `generators.json` lookup plus the snippet.

I generated the 8 locale copies by running `pnpm tsx ./scripts/translate.ts` locally rather than hand-writing them, so they came from the same pipeline CI uses.

One consequence worth flagging: snippets are static MDX with no props, so the text can't interpolate the generator id — it reads "This generator is experimental" rather than naming it. That seemed fine given the banner only ever appears on that generator's own guide page, but say the word if you'd rather keep the id and I'll find another way.

The MCP warning in `renderGeneratorInfo` is unchanged — it reads `generators.json` directly rather than going through the rendered page, so it doesn't touch the snippet path.

cogwirrel (approved):

PR #1154, opened 2026-08-30
fix(ts#project): write vitest coverage outside the project directory

nx-plugin-for-aws:
### Reason for this change

`@nx/vitest` writes `reportsDirectory: './test-output/vitest/coverage'` into every generated project's vitest config, which resolves **inside** `{projectRoot}`. Coverage is off by default, so the directory stays dormant — until someone runs tests with `--coverage`.

At that point the coverage HTML reporter writes Istanbul's bundled third-party scripts (`block-navigation.js`, `prettify.js`, `sorter.js`) into the project. The vended `biome.json` `files.includes` excludes `!**/dist`, `!**/out-tsc`, `!**/node_modules` and `!**/generated/**` — but not `test-output`. So the project's `format` target reformats vendored third-party JS and **fails permanently on every subsequent build**. The files also fall under the `default` named input, so no task in the project can ever be cached again.

#### Reproduction

```bash
pnpm create @aws/nx-workspace ws-cov-repro --no-interactive --iac=cdk --no-gitSecrets
cd ws-cov-repro
nx g @aws/nx-plugin:ts#project --no-interactive --name=lib
nx run-many --target build --all                              # passes
nx run @ws-cov-repro/lib:test --skip-nx-cache -- --coverage
nx run-many --target build --all                              # FAILS
```

Before this change:

```
packages/lib/test-output/vitest/coverage/block-navigation.js format ━━━━━
packages/lib/test-output/vitest/coverage/prettify.js         format ━━━━━
packages/lib/test-output/vitest/coverage/sorter.js           format ━━━━━
Found 3 errors.

NX   Running target build for project @ws-cov-repro/lib and 4 tasks it depends on failed
Failed tasks:
  - @ws-cov-repro/lib:format
```

The failure is permanent — a failing target is never cached, so it re-runs and re-fails forever. Coverage output also invalidated the `test` target's cache (100% hit → 0% hit after a coverage run).

### Description of changes

**Generator** (`packages/nx-plugin/src/ts/lib/vitest.ts`) — a new GritQL transform, applied alongside the existing `vitest-pass-with-no-tests` and `ROOT_IMPORT_META_DIRNAME` patterns, repoints coverage at `dist/<projectRoot>/test-output/vitest/coverage`:

```
reportsDirectory: '../../dist/packages/lib/test-output/vitest/coverage',
```

The relative depth is derived from the project root (vitest resolves `reportsDirectory` against the config's `root`, which is `import.meta.dirname`), so a nested root such as `packages/nested/lib` correctly gets `../../../dist/...`. The pattern is anchored to the exact literal `@nx/vitest` generates and scoped to the `coverage` block of the config's own `test` object, so a `reportsDirectory` the user has repointed — or one appearing elsewhere in the file — is left alone.

**No `outputs` change was needed.** `@nx/vitest`'s plugin derives the `test` target's `outputs` from the resolved config, normalising a `..`-prefixed path to `{workspaceRoot}/…`. Verified on a real workspace:

```
outputs: ["{workspaceRoot}/dist/packages/lib/test-output/vitest/coverage"]
```

So Nx still caches and restores coverage output correctly, now keyed outside the project.

**Biome exclude** — `!**/test-output` is added to the vended `biome.json`. I chose to include it: the generator fix only covers the directory *we* generate, and a user who points coverage somewhere else themselves (or any other tool writing test reports) would hit the identical failure. It is cheap, and it also rescues workspaces that already have stale coverage committed or on disk — validated below, where the migrated workspace built successfully **with the old `test-output/` directory still present**.

**Migration** (`latest-vitest-coverage-output-dir`) — brings existing workspaces across. It rewrites `reportsDirectory` only where it still matches the generated shape, reports a repointed directory via `nextSteps` instead of clobbering it, adds the Biome exclude only alongside the excludes we vend, and is idempotent.

It also **deletes the coverage directory an older release already wrote inside the project** and adds `test-output` to the workspace `.gitignore`. Without that, the stale directory stays git-visible — the vended `.gitignore` covers `dist` but not `test-output` — so the reporter's vendored third-party scripts can be committed, and `license#sync` (whose candidate set is built from git-visible files, not from `biome.json`) stamps an Apache-2.0 header onto them. Reproduced end to end: `nx sync` rewrote Istanbul's `sorter.js` to begin with an Amazon copyright header, clobbering its `/* eslint-disable */` line. Files a user keeps elsewhere under `test-output` are preserved.

Two cases that previously passed in silence now emit a `nextSteps` note: a `biome.json` whose `files.includes` has diverged from the vended list (so the exclude could not be placed), and a `reportsDirectory` set on a **target**, which overrides the vitest config and so leaves the project broken despite the rewrite. `CONFIG_FILES` also covers the other extensions vitest accepts, so a renamed config is migrated — safe to widen because the pattern applied inside is anchored to the literal `@nx/vitest` generates.

**Unit tests** — `pnpm nx run @aws/nx-plugin:test`: **3969 passed** (221 files), up from the 3948 baseline. 10 snapshots updated; every diff is `reportsDirectory` moving to the `dist` path or `"!**/test-output"` appearing in `biome.json`, and nothing else.

Added 3 generator tests (dist path, nested-root depth, user-repointed directory preserved) and 18 migration tests, covering the generated shape, nested roots, `vite.config.mts`, the Biome exclude and its position, a repointed directory reported via `nextSteps`, a `reportsDirectory` outside the config left alone, a config with no coverage block, an empty workspace, a rewritten `files.includes` preserved, and idempotency (both file contents and `nextSteps`) — plus deletion of the stale coverage directory, preservation of unrelated files beside it, the gitignore entry and its non-duplication, a renamed config, a target-set `reportsDirectory` reported (and one already outside the project not reported), and the diverged-`biome.json` note.

`pnpm nx run-many --target lint --all --configuration=fix` clean; `pnpm nx run-many --target build --all` passes.

**End-to-end, fresh workspace with the locally built plugin:**

```
reportsDirectory: '../../dist/packages/lib/test-output/vitest/coverage'

$ nx run @ws-cov-after/lib:test --skip-nx-cache -- --coverage
$ find dist -path "*test-output*" -name "*.js"
dist/packages/lib/test-output/vitest/coverage/block-navigation.js
dist/packages/lib/test-output/vitest/coverage/prettify.js
dist/packages/lib/test-output/vitest/coverage/sorter.js
$ find packages/lib -path "*test-output*"      # nothing inside the project

$ nx run-many --target build --all             # passes
$ nx run-many --target build --all             # Cache: 5/5 hit (100%)
$ nx run-many --target build --all             # Cache: 5/5 hit (100%)
```

Build passes after a coverage run, and repeat builds are a **100% cache hit — 0 tasks re-run**.

**End-to-end migration**, on a workspace generated *before* the change (stale coverage output already written into `packages/lib`), via `nx migrate --run-migrations=migrations.json`. Two projects: `lib` with the generated shape, `custom` deliberately repointed to `./my-coverage`.

```
Changes:
  UPDATE biome.json
  UPDATE packages/lib/vitest.config.mts

Next steps
  - @ws-cov-before/custom: `packages/custom/vitest.config.mts` sets its own
    `test.coverage.reportsDirectory`. Point it outside the project (eg
    `../../dist/packages/custom/test-output/vitest/coverage`) …
```

`lib` rewritten, `custom` preserved at `./my-coverage` and reported, `!**/test-output` added after `!**/out-tsc`. A second run changed no files and emitted the same single note (byte-identical `diff`), confirming idempotency.

The previously-broken build then recovered — **with the stale `packages/lib/test-output/` still on disk** (this is the Biome exclude earning its place):

```
$ nx run-many --target build --all    # Successfully ran target build
$ nx run-many --target build --all    # Cache: 10/10 hit (100%)
```

**Docs** — the TypeScript project guide now documents how to collect coverage and where reports land, since the location was previously undocumented.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/typescript-project.mdx
+++ b/docs/src/content/docs/en/guides/typescript-project.mdx
@@ -385,7 +385,7 @@ Pass Vitest's `--coverage` flag after a `--` separator to collect coverage:
 
 <NxCommands commands={['test <project-name> -- --coverage']} />
 
-Reports are written to `dist/<project-root>/test-output/vitest/coverage`, with the HTML report at `index.html`. Coverage output lives under `dist` rather than inside your project so it is not linted, formatted or treated as an input to your project's tasks.
+Reports are written to `dist/<project-root>/test-output/vitest/coverage`, with the HTML report at `index.html`.
 
 ## Linting
 
--- a/docs/src/content/docs/en/guides/typescript-project.mdx
+++ b/docs/src/content/docs/en/guides/typescript-project.mdx
@@ -379,6 +379,14 @@ You can run an individual test or suite of tests using Vitest's `-t` flag. Pass 
 If you are a VSCode user, we recommend installing the [Vitest Runner for VSCode that actually works](https://marketplace.visualstudio.com/items?itemName=rluvaton.vscode-vitest) extension, which allows you to run, watch or debug tests from your IDE.
 :::
 
+### Test Coverage
+
+Pass Vitest's `--coverage` flag after a `--` separator to collect coverage:
+
+<NxCommands commands={['test <project-name> -- --coverage']} />
+
+Reports are written to `dist/<project-root>/test-output/vitest/coverage`, with the HTML report at `index.html`. Coverage output lives under `dist` rather than inside your project so it is not linted, formatted or treated as an input to your project's tasks.
+
 ## Linting
 
 TypeScript projects use [Biome](https://biomejs.dev/) for linting and formatting. Biome is configured in the workspace root `biome.json` file — changes to this apply to all TypeScript projects in your workspace and ensure consistency.
--- a/docs/src/content/docs/es/guides/typescript-project.mdx
+++ b/docs/src/content/docs/es/guides/typescript-project.mdx
@@ -379,6 +379,14 @@ Puedes ejecutar una prueba individual o un conjunto de pruebas usando la bandera
 Si eres usuario de VSCode, recomendamos instalar la extensión [Vitest Runner for VSCode that actually works](https://marketplace.visualstudio.com/items?itemName=rluvaton.vscode-vitest), que te permite ejecutar, observar o depurar pruebas desde tu IDE.
 :::
 
+### Cobertura de Pruebas
+
+Pasa la bandera `--coverage` de Vitest después de un separador `--` para recopilar cobertura:
+
+<NxCommands commands={['test <project-name> -- --coverage']} />
+
+Los informes se escriben en `dist/<project-root>/test-output/vitest/coverage`, con el informe HTML en `index.html`. La salida de cobertura se encuentra bajo `dist` en lugar de dentro de tu proyecto para que no se aplique lint, no se formatee ni se trate como una entrada para las tareas de tu proyecto.
+
 ## Linting
 
 Los proyectos TypeScript usan [Biome](https://biomejs.dev/) para linting y formateo. Biome está configurado en el archivo `biome.json` de la raíz del workspace — los cambios a este se aplican a todos los proyectos TypeScript en tu workspace y aseguran consistencia.
--- a/docs/src/content/docs/es/guides/typescript-project.mdx
+++ b/docs/src/content/docs/es/guides/typescript-project.mdx
@@ -385,7 +385,7 @@ Pasa la bandera `--coverage` de Vitest después de un separador `--` para recopi
 
 <NxCommands commands={['test <project-name> -- --coverage']} />
 
-Los informes se escriben en `dist/<project-root>/test-output/vitest/coverage`, con el informe HTML en `index.html`. La salida de cobertura se encuentra bajo `dist` en lugar de dentro de tu proyecto para que no se aplique lint, no se formatee ni se trate como una entrada para las tareas de tu proyecto.
+Los informes se escriben en `dist/<project-root>/test-output/vitest/coverage`, con el informe HTML en `index.html`.
 
 ## Linting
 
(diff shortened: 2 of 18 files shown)


cogwirrel (dismissed):

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -379,6 +379,14 @@ You can run an individual test or suite of tests using Vitest's `-t` flag. Pass
 If you are a VSCode user, we recommend installing the [Vitest Runner for VSCode that actually works](https://marketplace.visualstudio.com/items?itemName=rluvaton.vscode-vitest) extension, which allows you to run, watch or debug tests from your IDE.
 :::
 
+### Test Coverage
+
+Pass Vitest's `--coverage` flag after a `--` separator to collect coverage:
+
+<NxCommands commands={['test <project-name> -- --coverage']} />
+
+Reports are written to `dist/<project-root>/test-output/vitest/coverage`, with the HTML report at `index.html`. Coverage output lives under `dist` rather than inside your project so it is not linted, formatted or treated as an input to your project's tasks.
```
cogwirrel: Remove "Coverage output lives under `dist` rather than inside your project so it is not linted, formatted or treated as an input to your project's tasks."

cogwirrel (dismissed):

docs/src/content/docs/en/guides/typescript-project.mdx
nx-plugin-for-aws: Done.

cogwirrel (approved):

PR #1152, opened 2026-08-29
perf(init): raise default Nx task parallelism to 8

nx-plugin-for-aws:
### Reason for this change

Nx defaults to **3** concurrent tasks. Workspaces created by these generators have a wide, shallow task graph — each project contributes `lint`, `compile`, `test`, and often `bundle` and `docker` targets — so 3 leaves cores idle during a full build. Nx's own run summary flags it:

```
Recommendations:
  - Increase parallelism to recover up to 19.4s
```

### Description of changes

Sets `parallel: 8` in `nx.json` from `applyWorkspaceInit` (shared by the workspace preset and the `init` generator).

An existing workspace's explicit `parallel` choice is preserved (`nxJson.parallel ?? DEFAULT_PARALLEL`), so running `init` against an established workspace does not override a deliberate setting — the same pattern already used for `analytics`.

8 is used rather than the machine core count because a full build saturates there: the task graph's critical path becomes the limit, and higher values add memory pressure for no gain.

Also documents the setting (and how to lower it) in the workspace guide.

### Description of how you validated changes

Measured on a generated 14-project workspace (`ts#project`, `ts#api`, `py#api`, `ts#website`, `ts#dynamodb`, `ts#agent`, `py#agent`, `ts#mcp-server`, `py#mcp-server`, `terraform#project`, `ts#infra` + connections) — **91 tasks** in `build --all`.

6 alternating pairs, `nx reset` + `rm -rf dist` before each run, Docker layers warm so image builds do not dominate:

| Pair | `parallel=3` | `parallel=8` | Δ |
|---|---|---|---|
| 1 | 46.40s | 33.88s | −27% |
| 2 | 48.17s | 33.06s | −31% |
| 3 | 44.75s | 33.17s | −26% |
| 4 | 52.69s | 32.60s | −38% |
| 5 | 49.14s | 29.65s | −40% |
| 6 | 38.28s | 27.47s | −28% |
| **median** | **46.7s** | **32.9s** | **−30%** |

Confirmed the plateau beyond 8 on the same host (32 vCPU):

| `--parallel` | Cold `build --all` |
|---|---|
| 3 | 46.4s |
| 8 | 33.9s |
| 16 | 34.0s |
| 24 | 34.3s |
| 32 | 34.7s |

A second workspace generated with `--iac=terraform` showed the same direction: 26.8s → 23.5s.

[...]

PR #1155, opened 2026-08-30
perf(terraform#project): share provider downloads via TF_PLUGIN_CACHE_DIR

nx-plugin-for-aws:
### Reason for this change

The generated `test` target for a `terraform#project` points `TF_DATA_DIR` inside `dist`:

```json
"env": { "TF_DATA_DIR": "../../../dist/{projectRoot}/terraform-test" }
```

`dist` gets cleaned, and the target is cache-busted by any change to the project or a consumed module. So on every real (non-cached) run, `terraform init` re-resolved and re-downloaded the entire provider set — ~880MB for the AWS provider alone. That download dominated the target's runtime.

Terraform has a first-class fix for exactly this: [`TF_PLUGIN_CACHE_DIR`](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache). Providers are downloaded once and linked from the cache afterwards.

### Description of changes

Targets that run `terraform init` reuse a persistent provider cache under `.terraform/plugin-cache` in the workspace root, **one directory per project**:

- **`test`** — the target that motivated this, since its `TF_DATA_DIR` is cleaned with `dist`.
- **`init`** — a library's runs `terraform init` inline, so it reads the cache from the target's `env`. An application's delegates to the vended `scripts/init.ts`, which reads it from a new `scripts/env.ts` helper.
- **`validate` / `plan` / `apply` / `destroy` / `output`** — these share `src/.terraform`, populated by `init`, so they inherit the saving without needing the env var themselves.
- **`common/terraform`** — gets it too, via the same generator. Its deeper root means it correctly computes `../../../../`.

`checkov` and `fmt` run no `terraform init`, so they are untouched. `bootstrap` and `bootstrap-destroy` are deliberately excluded — see below.

The `test` target also **declares no outputs**. Both decisions came out of review; the reasoning and measurements are below.

#### One cache per project, not one shared

A shared cache is the obvious design and it is what I originally shipped. It is wrong, and a reviewer caught it. Two `terraform init` runs filling one cache concurrently each compute a different `h1:` for the same provider — `h1:` is a dirhash over the provider directory, and a reader computing it sees partial contents another writer is still filling.

Nx's task parallelism governs cross-target concurrency; `parallel: false` only orders commands *within* a target. **PR #1152 raised the default `parallel` from 3 to 8**, so this is reachable, not latent. Measured in a real workspace with four `terraform#project` applications at `parallel: 8`, committed lock files, cold cache:

| | shared cache | per-project cache |
| --- | --- | --- |
| `nx run-many -t test` hard failures | **2/6 trials** | **0/8 trials** |
| committed lock files churned | **2/5 trials** | **0/6 trials** |

The hard failure is worse than the review predicted (it expected exit 0 always):

```
Error: registry.terraform.io/hashicorp/aws: the cached package for
registry.terraform.io/hashicorp/aws 6.62.0 (in ../../../dist/packages/tf-b/terraform-test/providers)
does not match any of the checksums recorded in the dependency lock file
```

In a synthetic harness the divergence rate scales with concurrency: N=2 1/6, N=3 2/6, N=4 3/6, N=6 3/6, N=8 5/6.

**Why not `parallelism: false`.** It does fix it (0/5 churn, 0/6 failures), but Nx serialises such a task against the *entire workspace* (`tasks-schedule.js:243-251`), not just against its own kind. Interleaved measurements on four projects:

| | parallel | `parallelism: false` |
| --- | --- | --- |
| `run-many -t test` warm | 5.5s | 10.5s |
| `build --all` warm | 10.5s | **24.4s** |

24.4s is **slower than `main`'s 14.0s** — it would have inverted the PR. Per-project caches keep `build --all` at 10.2s.

**Why the disk cost is not a regression.** Per-project caches genuinely do not dedupe (verified: 1 hardlink each, 3.5G for 4 projects). But `main` already wrote **3.5G into `dist`** for the same four projects, because each `TF_DATA_DIR` held a full self-contained provider set. Same total, now in a location that persists and survives `nx reset` instead of being deleted and re-downloaded.

**Why not `-lockfile=readonly`.** It looked ideal (0/6 churn) but **fails hard on a project with no lock file yet** — i.e. a freshly generated one: `Error: Provider dependency changes detected ... the lock file is read-only`. Ruled out.

#### `test` declares no outputs

`TF_DATA_DIR` is terraform's working directory, and its provider entries are now symlinks into the cache. Nx's native copy preserves symlinks rather than dereferencing them, so a cache hit on a machine without the cache restored **dangling links and skipped the commands that would repopulate them**. Reproduced end to end:

```
1) populate the nx cache      -> exit 0
2) rm -rf .terraform dist     (fresh-checkout simulation)
3) re-run                     -> Cache: 1/1 hit (100%)
   DANGLING: .../hashicorp/aws/6.62.0/linux_amd64 -> /workspace/.../plugin-cache/...
4) terraform test             -> Error: missing or corrupted provider plugins:
     there is no package for registry.terraform.io/hashicorp/aws 6.62.0 cached in
     ../../../dist/packages/tf-a/terraform-test/providers
```

On `main` that output was self-contained (883M of real files), so the hazard is new. With `outputs: []` a cache hit restores nothing, leaves no dangling links, and a subsequent miss repopulates correctly. One nuance the review did not have: a cache **miss** self-heals, so the hazard is specifically a *hit* leaving a tree a later consumer reads.

**Unclaimed win, now claimed:** the `test` target's cached output drops from **883M to 60K** per project per hash, so this also stops `.nx/cache` bloating by ~880M per project per input change.

#### Why `bootstrap` / `bootstrap-destroy` are excluded

They run `terraform init` too. They are excluded because `v1.0.0-rc.50/0003-terraform-bootstrap-adopt-existing-bucket`'s spec asserts byte-identity against `bootstrap.ts.template`, and including them would mean editing that released migration's test — which review explicitly ruled out. Both are one-time-per-environment operations, unlike `test` which runs on every build.

#### Nx does not interpolate `{workspaceRoot}` inside `env`

`{workspaceRoot}` expands to the **empty string** in an `env` value. Probed directly:

```
cwd: {projectRoot}/src
env: { PROBE: "{workspaceRoot}/.nx/cache/terraform-plugins" }
-> PROBE=.nx/cache/terraform-plugins        (resolved against cwd!)
```

So the obvious `{workspaceRoot}/.nx/cache/...` resolves *inside the source tree* and is **not gitignored** (`git check-ignore` reports no match). `{projectRoot}` **does** interpolate, which is what makes per-project paths expressible. Every path here is relative to the target's own `cwd`, computed from the project root rather than hardcoded.

`.nx/cache` was also rejected because `nx reset` does `rmSync(cacheDir, ...)` — it would throw the providers away on a routine reset. `.terraform` is gitignored by this generator (which is *also* why it stays out of Nx's file walk — not an independent property) and is not reset-managed.

#### The missing-directory trap

Terraform prints `Error: The specified plugin cache dir ... cannot be opened` and then **exits 0, having downloaded everything anyway** — a silent, exit-0 performance regression. Each target prepends `make-dir` (`forwardAllArgs: false` on that command alone so terraform flags don't leak into it) and the helper calls `mkdirSync(..., { recursive: true })`.

#### Lock files and `provider_installation` semantics

No `-plugin-dir` or lock handling needed. Cold cache, warm cache, and warm-plus-committed-lock all produced **byte-identical** lock files (42 `zh:` hashes across 3 providers, matching a clean run). `plugin_cache_may_break_dependency_lock_file` is deliberately **not** set, so terraform verifies against the lock file rather than trusting the cache.

**Unit tests** — full suite green, no snapshot changes. Coverage for the generator (per-project cache path, `make-dir` first with `forwardAllArgs: false`, `outputs: []`, `env.ts` vended and imported) and 17 migration specs (idempotent; preserves a user-chosen `TF_PLUGIN_CACHE_DIR`, a hand-edited `env.ts`, a customised `init.ts` and customised `outputs`; reports rather than clobbers a diverged `test` target or `init.ts`; converges byte-identically on today's templates).

**Lint** — `nx run-many --target lint --all --configuration=fix` clean.

**Measured, interleaved, in real workspaces** (`main` vs this branch, terraform 1.14.4, `TF_DATA_DIR` deleted before each run to force the real path, `--skip-nx-cache`, exit codes checked):

Single-project `test`, warm cache:

| pair | main | this PR |
| --- | --- | --- |
| 1 | 13444ms | 7251ms |
| 2 | 13036ms | 6589ms |
| 3 | 12994ms | 6757ms |

Four-project `build --all`, warm:

| pair | main | this PR |
| --- | --- | --- |
| 1 | 17539ms | 13645ms |
| 2 | 13982ms | 10233ms |

**~2x faster** on the single-project case and faster on multi-project too, with the after-timings far tighter — the spread on `main` is the provider download.

Also verified:

- **Genuinely cold cache** — `.terraform` deleted entirely: exit 0, no CLI configuration error, 883M populated.
- **`nx run-many -t test` on a cold cache, 8 trials at `parallel: 8`** — 0 hard failures, 0 lock churn.
- **Nx cache-hit safety** — no dangling symlinks restored; a subsequent miss repopulates.
- **Real `.tftest.hcl`** — `1 passed, 0 failed` across all four projects.
- **Ambient `TF_PLUGIN_CACHE_DIR` honoured** — resolves to the user's path when set, per-project path otherwise.
- **Gitignored** — `git check-ignore` matches `.gitignore:.terraform`; `git status` never lists the cache.

**Deployed to AWS and torn down.** On a workspace carried forward by the migration:

- `bootstrap` — 6 resources created (S3 state bucket), exercising the migrated `init.ts` + `env.ts`.
- `deploy` — `init` against the S3 backend, then `plan` + `apply`: 3 resources created, outputs returned.
- Cache reuse confirmed after the deploy: `Using hashicorp/aws v6.62.0 from the shared cache directory`.
- `test` still passed against the deployed state.
- `destroy` — 3 destroyed. `bootstrap-destroy` — 6 destroyed. Confirmed gone: the `us-east-1` state bucket returns 404.

**Migration end-to-end.** On a workspace generated **before** this change, via `nx migrate --run-migrations=migrations.json`:

- Applied cleanly with **no `nextSteps`**, updating both `project.json` files and `init.ts`, and creating `env.ts`.
- Result **byte-identical** to a freshly generated workspace.
- **Idempotent** — re-running reports `No changes were made`.
- All 5 CI `migrate` smoke-test shards passed on the previous push.

**Coordination:** #1162 also touches `terraform/project/generator.spec.ts`. `git merge-tree` shows one conflict there, purely additive adjacent test blocks (`generator.ts` auto-merges cleanly). Happy to rebase once either lands.

### Refuted, or verified unchanged

- The classic `h1:`-only lock-file hazard is **absent** — all `zh:` hashes present, because `plugin_cache_may_break_dependency_lock_file` is not set.
- A divergent `h1:` is **appended**, not substituted, so the cache still serves the provider — verified `Using ... from the shared cache directory` on a churned lock. The harm is git churn and the hard failure above, not a silent re-download.
- Concurrent init into a **warm** cache is safe (0/6 trials churned) — only writers race.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -313,6 +314,18 @@ run "plan_is_valid" {
...
+
+The directory is created automatically before each `terraform init`, and `.terraform` is gitignored. Unlike `.nx/cache`, it is not removed by `nx reset`, so a reset does not throw the providers away. In CI, cache `.terraform/plugin-cache` between jobs to get the same saving there.
+
+Nx's `env` takes precedence over your shell, so setting `TF_PLUGIN_CACHE_DIR` in your environment does not change where these targets look. To point them somewhere else — a location shared between several workspaces, say — change `TF_PLUGIN_CACHE_DIR` in your project's `project.json`, and the path in `scripts/plugin-cache.ts` for the `bootstrap`, `bootstrap-destroy` and `init` targets. Paths in `project.json` are relative to the target's `cwd`, since Nx does not expand `{workspaceRoot}` inside `env`.
+
+:::note[Lock Files]
+The cache does not change which provider versions you get: `.terraform.lock.hcl` still pins them, and Terraform records the same hashes whether a provider came from the cache or the registry. Commit the lock file as usual.
+:::
```
cogwirrel: No need for a whole section of this. Just include this sentence in the appropriate place in the docs:

```
Every target that runs `terraform init` shares a provider cache at `.terraform/plugin-cache` in your workspace root. See the [Terraform documentation](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) for more information.
```

packages/nx-plugin/src/terraform/project/files/application/scripts/plugin-cache.ts.template
```diff
@@ -0,0 +1,20 @@
...
+ * terraform reports an error and falls back to downloading when it is missing.
+ *
+ * These scripts are invoked from the workspace root by their nx targets, which
+ * is where the gitignored `.terraform` sits.
+ */
+export const pluginCacheEnv = (): NodeJS.ProcessEnv => {
+  const dir = join(process.cwd(), '.terraform', 'plugin-cache');
+  mkdirSync(dir, { recursive: true });
+  return { ...process.env, TF_PLUGIN_CACHE_DIR: dir };
+};
```
cogwirrel: Name the generated file `env.ts`.

Remove the file header comment.

Change the pluginCacheEnv comment to "Environment variables for Terraform to cache plugins"

packages/nx-plugin/src/migrations/v1.0.0-rc.50/0003-terraform-bootstrap-adopt-existing-bucket/migration.spec.ts
```diff
@@ -151,6 +152,9 @@ describe('terraform-bootstrap-adopt-existing-bucket migration', () => {
     await generateWithPreFixBootstrap(tree);
 
     const result = await migration(tree);
+    // The script has moved on since this migration, so the later migrations
+    // that touched it run too — as `nx migrate` runs them, in order.
+    await pluginCacheMigration(tree);
```
cogwirrel: do not touch existing migrations

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Done in 0c5b425 — replaced the whole section with your sentence. Also renamed the file-tree entry to `env.ts`.

packages/nx-plugin/src/terraform/project/files/application/scripts/plugin-cache.ts.template
nx-plugin-for-aws: Done in 0c5b425 — renamed to `env.ts`, header comment removed, and the `pluginCacheEnv` comment is now `/** Environment variables for Terraform to cache plugins */`.

packages/nx-plugin/src/migrations/v1.0.0-rc.50/0003-terraform-bootstrap-adopt-existing-bucket/migration.spec.ts
nx-plugin-for-aws: Reverted in 0c5b425 — that spec is untouched now.

To make that possible I dropped `bootstrap.ts` and `bootstrap-destroy.ts` from this change and scoped the script edit to `init.ts`. The rc.50 spec pins `bootstrap.ts.template` byte-for-byte, so touching that template is what forced the test change. Both bootstrap targets are one-time-per-environment, so they gain little from the cache compared to `test`, which runs on every build — the omission is noted in the PR description.

nx-plugin-for-aws: Reverted — this file is now byte-identical to `main` and the PR touches no `v1.0.0-rc.*` file.

Root cause fixed on my side rather than in the released spec: this spec asserts the migrated `bootstrap.ts` matches today's template, and my original change edited that template. The provider cache is now scoped to `init.ts` only, so `bootstrap.ts`/`bootstrap-destroy.ts` are untouched and the assertion holds unmodified. `should produce exactly what the current generator vends` passes (9/9 in that spec, 3965/3965 overall).

`bootstrap`/`bootstrap-destroy` lose nothing meaningful — they are one-time-per-environment operations, unlike `test`, which re-inits on every cache miss.

nx-plugin-for-aws:
Worked through the independent review. Every finding reproduced first; two changed the design.

**M1 (concurrent init churns lock hashes) — reproduced, and worse than reported.** Confirmed at N=6, 3/6 trials, matching the reviewer exactly. Two escalations:

1. The mitigation the review relied on is **gone**: #1152 raised the default `parallel` from 3 to 8. Re-measured in a real 4-project workspace at `parallel: 8` — 2/5 trials churned a committed lock file.
2. It is not always exit 0. It can **fail the build outright**: `Error: ... the cached package for registry.terraform.io/hashicorp/aws 6.62.0 ... does not match any of the checksums recorded in the dependency lock file` — 2/6 `nx run-many -t test` trials.

I tried all three suggested fixes and measured each:

| fix | churn | hard failures | `build --all` (4 projects, warm) |
| --- | --- | --- | --- |
| none (shared cache) | 2/5 | 2/6 | 10.5s |
| `parallelism: false` | 0/5 | 0/6 | **24.4s** |
| per-project cache dirs | 0/6 | 0/8 | 10.2s |
| _`main` for reference_ | 0/3 | 0/3 | 14.0s |

`parallelism: false` works but Nx serialises the task against the *whole workspace*, making `build --all` slower than before the cache existed — it would have inverted the PR. **Went with per-project cache dirs.** The disk objection turns out not to bite: they genuinely do not dedupe (3.5G for 4 projects), but `main` already wrote 3.5G into `dist` for the same projects, so it is the same total in a location that now persists.

Also ruled out `-lockfile=readonly` — 0/6 churn, but it **fails hard on a project with no lock file yet**, i.e. a freshly generated one.

**M2 (outputs contain symlinks pointing outside) — reproduced verbatim**, including your exact error string. `test` now declares `outputs: []`. One nuance you did not have: a cache *miss* self-heals, so the hazard is specifically a *hit* leaving a tree a later consumer reads. This also banks the win you flagged — the cached output drops **883M → 60K** per project per hash.

**L1 — reproduced.** Confirmed the env var beats `~/.terraformrc` (vended dir got 17M, user's stayed empty). Now `process.env.TF_PLUGIN_CACHE_DIR ?? join(...)`, consistent with the migration.

**L2 / Nit — fixed.** Docs cover the per-project rationale and how to point at your own cache. Dropped the implication that staying out of Nx's file walk is independent of the gitignore.

**Partial refutation of M1(b).** A divergent `h1:` is **appended**, not substituted, so the provider is still served from the cache — verified `Using ... from the shared cache directory` on a churned lock. So there is no silent loss of the optimisation; the harm is git churn plus the hard failure above. Also: concurrent init into a **warm** cache is safe (0/6), so only cold population races.

**Deployed to AWS and tore it down** — `bootstrap` (6 resources), `deploy` (3 resources, outputs returned), cache reuse confirmed after the apply, `test` still passing, then `destroy` (3) and `bootstrap-destroy` (6). State bucket now 404s.

Nothing from your "verified-good" list was undone: `zh:` hashes still complete, `make-dir` mitigation intact, relative paths unchanged, migration still idempotent and non-clobbering.

Note #1162 also touches `generator.spec.ts` — one purely additive conflict; happy to rebase once either lands.

cogwirrel (approved):

PR #1156, opened 2026-08-30
perf(py#project): scope python build inputs so test edits do not invalidate the build chain

nx-plugin-for-aws:
### Reason for this change

Editing a Python test file — one that nothing can possibly import, e.g. `packages/py_api/tests/test_main.py` — invalidated **nine** tasks in a generated workspace and cascaded all the way into `infra:synth`:

```
ws_cdk.py_api: format, lint, test, compile, typecheck, openapi, bundle-x86
@ws-cdk/infra:  synth, test
```

Two independent causes, neither fix sufficient alone.

**Cause A — the `production` named input had no Python equivalent.** It excluded TypeScript specs but nothing Python, while the Python build-chain targets read `["default", "^production"]`, where `default` is `{projectRoot}/**/*` — every file in the project. A test file was a build input by construction. The OpenAPI spec targets declared no `inputs` at all, so they fell back to `default` too.

**Cause B — timestamped test reports propagated downstream.** `default` also carried `{ "dependentTasksOutputFiles": "**/*", "transitive": true }`. The Python `test` target writes JUnit and coverage XML that pytest stamps wall-clock time into, so whenever `test` re-ran, a fresh hash propagated to every downstream task.

### Description of changes

**Cause A.** `production` now also excludes `!{projectRoot}/tests/**/*`. `compile`, the Python bundle targets and the OpenAPI spec targets read `["production", "^production"]`. The input is created when a workspace lacks it, since Nx fails hard on an undefined named input.

The exclusion is deliberately anchored to `tests/` rather than matching `test_*.py` anywhere. `@nxlv/python` matches `ignorePaths` against the project root's top-level entries only (`provider.js:530`), so a nested `py_api/test_data_loader.py` **is** copied into the build and packaged into the wheel — verified by extracting it. Excluding it would leave the wheel stale after a fix to it.

**`typecheck` deliberately still reads `default`.** `ty` type checks the tests directory, so scoping it to `production` is a correctness regression, not an optimisation:

```
$ nx run ws_cdk.py_api:typecheck     # after adding `x: int = "not an int"` to a test
NX   Successfully ran target typecheck        # 11ms, 1/1 cache hit  <-- WRONG
$ nx run ws_cdk.py_api:typecheck --skip-nx-cache
Incompatible value of type `Literal["not an int"]` ... Found 1 diagnostic
```

This is why the floor is 4 re-run tasks rather than 3. `test`, `lint` and `format` likewise keep hashing test files.

**Cause B.** The dependent-task-output input becomes `!{reports,coverage}/**` — excluding the timestamped reports rather than allow-listing `dist`. Several generators emit generated sources *outside* `dist` (an OpenAPI client under a website's `src/generated`, the Smithy SSDK, Prisma's client, shared OpenAPI metadata) and `.gitignore` them, and Nx's file map honours `.gitignore`, so they appear in no project fileset:

```
@ws-cdk/web: total=29 files, generated files in map = 0
```

That makes this input the only thing hashing them, so an allow-list of `dist/**` would have removed their last coverage. Verified against `matchGlobPaths` (`Y` = hashed):

```
                          dist reports coverage web-gen constructs-gen agent-gen prisma-gen
"**/*"                     Y     Y       Y        Y        Y             Y         Y
"dist/**"                  Y     .       .        .        .             .         .
"!{reports,coverage}/**"   Y     .       .        Y        Y             Y         Y
```

A migration backfills both changes, seeding `production` when absent so it can never emit a config referencing an undefined input. It rewrites only entries still carrying a shape this plugin vended, and reports divergence via `nextSteps` — a customised glob, user-tuned target `inputs` — rather than clobbering them.

**Unit tests** — 3967 pass across 221 files (baseline 3948). Snapshot diffs reviewed: only the glob change. New tests cover idempotency, preserved user customisations, non-Python projects untouched, `typecheck` staying on `default`, a user-authored `-openapi` target not being rescoped, and — asserting the invariant directly — that whenever a target reads `production`, `production` is defined. Lint clean.

**Measured A/B in real 14-project workspaces** (`py#api` + `py#project` + `ts#infra`), py test file mutated by a script that keeps Ruff formatting valid, full `build --all`, all runs exit 0:

| | tasks re-run | median wall |
|---|---|---|
| baseline | **9** | 18.05s |
| this PR | **4** | **11.48s** |

**-36% wall clock.** The 4 remaining tasks all genuinely read the test file.

Isolated A/B/C, each in its own workspace, confirming both causes are independent and both required — and that the final glob keeps Cause B fixed. Run on the Cause-A-only workspace, where the whole `dist` tree is provably byte-identical before and after:

| | tasks | notes |
|---|---|---|
| baseline | 9 | |
| Cause A only (`**/*`) | 6 | `infra:synth` + `infra:test` survive |
| Cause B only | 7 | py_api build chain survives |
| A + `dist/**` | 4 | over-scoped, see above |
| **A + `!{reports,coverage}/**`** | **4** | shipped |

**No regressions.** A real Python source edit and a real TypeScript source edit each invalidate exactly the same task set as baseline, including `infra:synth`. A new FastAPI route still propagates end to end: openapi spec → gitignored generated client → `web:compile` → `web:bundle`. Editing a nested `test_data_loader.py` correctly re-runs `compile` and the fresh code lands in the wheel.

**Migration verified end-to-end** on a fresh pre-change workspace via `nx migrate --run-migrations=migrations.json` with the locally compiled plugin: updated `nx.json` and all three Python projects, left `typecheck` alone, second run reported "No changes were made" with byte-identical files, and the migrated workspace then measured 4 tasks.

### Issue # (if applicable)

N/A

### Checklist

----


cogwirrel (dismissed):

nx-plugin-for-aws:
Follow-up: a question was raised about whether the second half of this change (the `dependentTasksOutputFiles` narrowing) targets a real mechanism, since an independent sweep concluded the pytest report timestamps were "real but inert". I re-tested it directly. **The mechanism is real and the narrowing is load-bearing.** Evidence below.

### Why the "inert" experiment cannot detect it

The experiment behind the inert conclusion is: force `py_api:test` to re-run, then run `build --all`, and observe 0 downstream tasks. I reproduced exactly that, and it does show 0 — but it is masked by construction:

```
$ nx run ws_cdk.py_api:test --skip-nx-cache
junit md5 after forced run:      b628e7c4dcf8
$ nx run-many --target build --all
downstream tasks re-run:         0
junit md5 AFTER the build:       b3ebd63ea4d5   <-- changed again, mid-build
```

On the second command `test` is a **cache hit**, so Nx *restores* the cached reports over the fresh ones. The timestamps are reverted before anything downstream is hashed, so of course nothing propagates. That experiment cannot observe the mechanism.

### The mechanism, isolated

Run against the **Cause-A-only** workspace, where `compile` / `bundle-x86` / `openapi` already do not re-run on a test edit — so `dist/` is provably untouched and cannot be the carrier. I hashed the entire `dist/packages/py_api` tree before and after:

| | dist tree | junit | tasks re-run |
|---|---|---|---|
| `dependentTasksOutputFiles: "**/*"` | `10f88ab8c1b3` → `10f88ab8c1b3` (**unchanged**) | changed | **6** — incl. `infra:synth`, `infra:test` |
| `dependentTasksOutputFiles: "dist/**"` | `10f88ab8c1b3` → `10f88ab8c1b3` (**unchanged**) | changed | **4** — no `infra` tasks |

Same workspace, same edit, same untouched `dist/`. The only variable is the glob, and `infra:synth` + `infra:test` appear or disappear with it. The only input that changed is the non-`dist` test reports, so they are what carried the invalidation. This also matches the isolated A/B/C in the description: Cause-A-alone leaves `infra:synth` + `infra:test` surviving, and only the narrowing removes them.

So Cause B fires precisely in the scenario that matters — when `test` genuinely re-runs as part of the same build, which is exactly what a source or test edit causes. It is inert only when `test` is cached.

### On excluding `README.md` from the build inputs

I also evaluated excluding `README.md`, which was suggested as an additional win. I confirmed the underlying fact — `pyproject.toml` sets `readme = "README.md"` and hatch embeds the text into `dist-info/METADATA`, verified by extracting the wheel and reading it back.

I deliberately **left it out**, because it is a correctness trade rather than an optimisation. Measured on a README edit with everything else in this PR applied:

- **without** the exclusion: 9 tasks re-run, wheel METADATA contains the new text
- **with** the exclusion: 4 tasks re-run, but the wheel keeps the **stale** README (`Description-Content-Type: text/markdown` section still holds the old text)

Since the README is genuinely part of the built artifact, excluding it makes `compile` skip when its own output is out of date and ship a wheel whose metadata does not match the source. That is the same class of trade I rejected for `typecheck` (where `ty` really does check `tests/`, so scoping it to `production` returns a stale pass on a real type error in a test). A README edit invalidating the wheel is correct behaviour; the wasteful part is the fanout into `infra:synth`, and the `dist/**` narrowing in this PR does not help there because the README legitimately changes `dist`. Making that cheap would need the metadata split out of the wheel input, which is a separate change with a real design question attached, so I have not smuggled it in here.

nx-plugin-for-aws:
Pushed `4e2fe990` addressing the review findings. Both HIGH findings reproduced before fixing; one of the two I am pushing back on in part, with evidence.

## H1 — migration referenced an undefined `production` input · FIXED

Reproduced exactly as reported, by running the real migration on a workspace with no `namedInputs`:

```
INITIAL namedInputs  = undefined
PROBE_PRODUCTION     = undefined
PROBE_COMPILE_INPUTS = ["production","^production"]   <-- references an undefined input
```

`migrateNamedInputs` declined to create the input while `migrateProjectTargets` rewrote targets to read it, with no gate between. The generator half already seeded it, so the two halves disagreed. The migration now seeds `production` as `[default]` exactly as the generator does. After the fix:

```
PROBE_PRODUCTION     = ["default","!{projectRoot}/tests/**/*"]
PROBE_COMPILE_INPUTS = ["production","^production"]
PROBE_NEXTSTEPS      = []
```

The advisory `nextSteps` note is gone, since there is nothing left for the user to do. Two tests added — one asserting the created input *and* the resulting target `inputs` (the gap that let CI stay green), and one asserting the invariant directly: whenever any target reads `production`, `production` is defined.

## H2 — `dist/**` dropped invalidation for non-`dist` outputs · FIXED, via the third option

I verified the core of this and it is a real design flaw, though my reproduction of the *specific* failure differs from the reviewers. Two things I confirmed independently:

1. All six sites write generated sources outside `dist/`, and all are `.gitignore`d by the same generators.
2. **The aggravating factor is the real problem.** Nxs file map honours `.gitignore`, so those files are in no projects fileset:
   ```
   @ws-cdk/web:                total=29 files, generated files in map = 0
   @ws-cdk/common-constructs:  total=36 files, generated files in map = 0
   ws_cdk.py_lib:              total=26 files, generated files in map = 0
   ```
   With `dist/**`, `dependentTasksOutputFiles` was the only thing hashing them, so narrowing it removed their last coverage. That is enough to justify the fix on its own.

**Where I could not reproduce permanent staleness.** In the vended workspace the deployable chain stayed correct under `dist/**` — a new FastAPI route reached the openapi spec, the gitignored client, `web:compile`s `.d.ts` output and `web:bundle`. Deleting or corrupting the generated dir self-healed under **both** globs (Nx restores cached outputs). The reason the vended chain survives is that `web:compile` declares `dependsOn: [^build]`, so the dependencys `dist` outputs change too and invalidate it regardless. So in the *vended* configuration I measured, `dist/**` was not observably stale — but that is incidental protection, and it evaporates for any consumer without that `^build` edge, which is exactly the shape the reviewers minimal repro isolates. Not something to rely on.

Two probe corrections worth recording, since both initially looked like confirmations: `web:compile` emits only `.d.ts`, so grepping it for a *route* string always misses (the model symbol `H2e<N>` is present); and `infra:cdk.out` references assets by hash and never embeds website JS, so my "step 5 = NO" reading was an invalid probe under both globs, not evidence of staleness.

**Resolution — the third option you suggested, which is strictly better than either extreme:**

```
dependentTasksOutputFiles: !{reports,coverage}/**
```

Verified against `matchGlobPaths` (`Y` = hashed):

```
                          dist reports coverage web-gen constructs-gen agent-gen prisma-gen
"**/*"                     Y     Y       Y        Y        Y             Y         Y
"dist/**"                  Y     .       .        .        .             .         .
"!{reports,coverage}/**"   Y     .       .        Y        Y             Y         Y
```

Every generated-source location stays hashed; only pytests timestamped JUnit/coverage XML is excluded. And it **keeps the Cause B fix** — same isolating test as before, on the Cause-A-only workspace where `dist/` is provably untouched:

| glob | dist tree | tasks re-run |
|---|---|---|
| `**/*` | unchanged | **6** — incl. `infra:synth`, `infra:test` |
| `dist/**` | unchanged | **4** |
| `!{reports,coverage}/**` | unchanged | **4** |

So reverting to `**/*` would have reintroduced the cascade; this keeps the win without the over-scoping. Migration treats both `**/*` and `dist/**` as previously-vended and rewrites either.

## M1 — unanchored globs excluded shipped production code · FIXED

Confirmed via minimatch, then end-to-end. `ignorePaths` is matched against `readdirSync(projectRoot)` — top-level only (`provider.js:530`) — so a nested `test_*.py` really is packaged:

```
$ python -c "zipfile ... namelist()"
[ws_cdk_py_api/__init__.py, ws_cdk_py_api/init.py, ws_cdk_py_api/main.py,
 ws_cdk_py_api/test_data_loader.py]   <-- shipped in the wheel
```

Narrowed to just `!{projectRoot}/tests/**/*`. Verified the fix in a real workspace: editing `ws_cdk_py_api/test_data_loader.py` now re-runs `compile` and the wheel carries the new code (`version 2 (fresh): True`), while a `tests/` edit still re-runs only 4 tasks. The docs section now states the exclusion is limited to `tests/` and that a `test_*.py` inside the package directory counts as production code.

## Nits

- **Comment symmetry** — the "Python counterparts" claim is gone; the comment now explains the anchoring and why.
- **`-openapi` false positive** — detection now also requires `nx:run-commands` and an `openapi` output path, since these targets are identified by declaring no `inputs`. Test added for a user-authored `my-fixtures-openapi`.
- **Test-config edit** — agreed the title slightly over-reaches for `pyproject.toml`; the docs already scope the claim to test *files*, so leaving the behaviour as-is (a dependency change must still invalidate the build).
- **L1** — acknowledged; the workspace-global seed is deliberate and append-only, and is what H1 requires.

## Validation

3967 tests pass (221 files), lint clean, migration verified end-to-end on a fresh pre-change workspace including a byte-identical idempotent re-run. Regression suite unchanged: a real Python source edit and a real TypeScript source edit each invalidate exactly the same task set as baseline. Task count on a py test-only edit remains **4** (9 before this PR).

Nothing from the "Refuted / verified-good" list was touched — `typecheck` stays on `default`, `test`/`lint`/`format` keep hashing test files, and the migration still never clobbers user `namedInputs`.

## Sequencing

Agreed this should merge **before #1162**, resolving the three-file conflict in favour of `[production,^production]`. Commented on #1162 to coordinate.

cogwirrel (approved):

PR #1158, opened 2026-08-30
fix(ts#agent,py#agent): stop caching container image build and scan targets

nx-plugin-for-aws:
### Reason for this change

**This is a cache-soundness correctness bug, not a performance nit.** The `*-docker` targets declared `outputs` covering only the copied `Dockerfile` — or nothing at all for the Python ones. But the artefact they actually produce, the image, lives in the container engine's store and is not an `outputs` path. Nx therefore believed it could restore these tasks from cache when it could not.

Reproduced in a 14-project workspace (ts + py agents and MCP servers), on `main`:

```
$ docker rmi -f ws-ts-lib-agent:latest
$ nx run-many --target build --all
> nx run @ws-cdk/ts-lib:agent-docker  [local cache]
 NX   Successfully ran target build for 12 projects
Cache: 77/81 hit (95%)

$ docker image inspect ws-ts-lib-agent:latest
*** ABSENT ***
```

Nx reports the build complete and cached while the image simply does not exist.

The `*-trivy` scan targets inherited the same unsoundness — they were cacheable and keyed on project source, so a scan that passed once kept reporting a cached pass against an image that is gone. Forced to actually run, it fails:

```
$ nx run @ws-cdk/ts-lib:agent-trivy
Cache: 4/4 hit (100%)
> docker save -o .../image-0.tar ws-ts-lib-agent:latest
Error response from daemon: reference does not exist
```

**Proven end-to-end against real AWS.** The blast radius differs by IaC provider, which is worth stating precisely:

- **Terraform reads the image out of the daemon** — `runtime.tf.template` runs `<engine> inspect … --format '{{.Id}}'` via `data "external"`, then tags and pushes that digest. This is where the bug is user-visible and breaking.
- **CDK does not** — it uses `AgentRuntimeArtifact.fromAsset(bundleDir)` and builds its own image, so for CDK workspaces the unsound cache was *latent* rather than breaking.

Deploying a Terraform agent workspace, from the identical starting state of a deleted image:

| | `nx build` | `nx apply` (real AWS) |
|---|---|---|
| pre-fix (`cache: true`) | reports success, `[local cache]`, **image absent** | **FAILS** — `"…tfagentagent_repository_79ca8a56:" is not a valid repository/tag: invalid reference format` (empty digest) |
| post-fix (`cache: false`) | rebuilds the image | **succeeds** — 29 resources applied, ECR tag `127354cb6747…` matching the local digest exactly |

So a missing image silently cleared a build and then broke a deploy. Torn down afterwards: 29 + 6 resources destroyed, AgentCore runtimes / ECR repos / state buckets / AppConfig apps all verified gone.

### Description of changes

Mark both the image build targets and the scan targets non-cacheable, via a shared `IMAGE_BUILD_CACHE` constant in `utils/docker.ts` so every generator emitting container build commands stays consistent. The now-pointless `Dockerfile` output on the TypeScript targets is dropped — it was a copy made *into* the build context, an input rather than a product. The scan target's `inputs` are dropped too: Nx only hashes a cacheable task, so they were dead configuration (verified — an uncached target with `inputs` re-runs every time regardless).

Affected generators: `ts#agent`, `py#agent`, `ts#mcp-server`, `py#mcp-server`, `ts#rdb`, `py#rdb`.

**The trade-off, and why this option.** The container engine's own layer cache is the sound cache here: it is keyed on the build context, lives in the same daemon as the image, and therefore cannot claim a hit for an image that is not there. A warm rebuild is ~0.35s (TypeScript) / ~0.85s (Python) per image, so correctness costs very little on `build`.

Two alternatives were measured and rejected:

- **A digest/stamp file (or a `runtime` input probing the engine) declared as `outputs`.** Not sound, and I verified this rather than reasoning about it: the "image absent" probe value gets its own cache entry, so deleting the image lands straight back on that entry and the task is skipped again. The image stayed missing.
- **`docker save` the image as the cached output.** Sound, but 462MB per image per cache entry against a sub-second rebuild. Not a reasonable trade.

I also swept the other generated targets for the same class of problem. Everything else cacheable writes a real directory under `dist/` — `bundle`, `synth`, `checkov`, `compile`, `terraform-test`. (Separately noted for follow-up, outside this PR's scope: `terraform#project`'s `output` and `validate` targets are `cache: true` with no `outputs` and read remote state — a different unsoundness class, external mutable state.)

The docs recommended the unsound pattern, so `docker-bundling.mdx` and the `trivy-image-scan` snippet are updated with the reasoning, the real cost, and the network implication.

**Cost on `build --all`.** Wall clock, `NX_DAEMON=true`, warm Docker layers, 10 interleaved before/after rounds, 14-project workspace, 4 images: median **8523ms → 11786ms**, i.e. **+3.2s (~0.8s/image)**. Min-vs-min agrees (+3205ms).

**Cost on `nx trivy` — much larger, and now stated in the docs.** The scan does a full `docker save` plus Trivy run per image on every invocation, where a cache hit was previously free. Measured with the exact vended command shape: `docker save` 462MB taking 24–53s, and the Trivy scan 43s cold / 49–67s on repeat (this host is heavily contended and slower than a typical runner; an independent review measured ~26s/image total, same conclusion). Either way it is **tens of seconds per image, every run** — acceptable because `trivy` is deliberately outside `build` and runs as its own CI job, but the earlier "+3.2s" understated it by an order of magnitude for `nx trivy`.

**Behaviour change worth knowing:** the scan now refreshes Trivy's vulnerability DB on every run (confirmed: `[vulndb] Need to update DB` → `Downloading vulnerability DB...`), so it needs network access where a cache hit previously let it pass offline. That refresh is the point — a cached scan result *is* the stale-CVE hazard. For isolated runners the docs note mounting a persistent `/root/.cache/trivy` plus `--skip-db-update`; measured, that mount takes the scan from ~60s to ~2s, but it changes the vended command shape and needs cross-engine validation, so it is left as a documented option rather than done here.

**Tests.** `pnpm nx run @aws/nx-plugin:test -u` → **3959 passed / 221 files**, no snapshot changes. Existing assertions that encoded the old buggy shape (`cache: true`, the `Dockerfile` output, the dead `inputs`) now assert the new contract. Eleven tests cover the migration.

**Migration** (`latest-docker-target-cache-soundness`), verified end-to-end on a pre-change workspace via `nx migrate --run-migrations=migrations.json`: fixed all 8 generated docker/trivy targets, left a hand-customised `docker push` target untouched and reported it, and re-running is a clean no-op.

Following review, the migration was **narrowed so it cannot un-cache a target that is soundly cacheable**. Matching on name plus any engine command also caught targets whose artefact genuinely is on disk — an `export-docker` whose whole product is a declared `docker save` tarball, or an `e2e-docker` building a fixture image and reporting to JSON. Both were flipped silently; both now stay cacheable and are reported via `nextSteps`. The vended `Dockerfile` output is also matched by its full generated path rather than any path ending in `/Dockerfile`, so a user's own output is no longer stripped.

`pnpm nx run-many --target lint --all --configuration=fix` and `pnpm nx run-many --target build --all` both clean.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -182,6 +181,12 @@ Add a `docker` target which:
 }
 ```
 
+:::caution[Do not let Nx cache an image build]
+A built image lives in the container engine's store, not under any `outputs` path, so Nx has nothing to restore on a cache hit — it would report the build complete while the image is missing (say after a `docker rmi` or a `docker system prune`), and a `trivy` target depending on it would report a cached pass against an image that isn't there.
+
+Setting `"cache": false` keeps this sound at almost no cost: the container engine's own layer cache is keyed on the build context and lives alongside the image, so an unchanged context rebuilds in well under a second and always leaves the image present.
+:::
```
cogwirrel: Change to:

```
:::caution[Do not cache image builds]
Set `"cache": false` on image build targets. A built image is not written to an `outputs` path in the filesystem, so Nx has nothing to restore on a cache hit. Container engines cache image layers regardless, so there is no performance penalty.
:::
```

docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -296,8 +301,10 @@ The scan is intentionally not wired into `build` since this can introduce unnece
...
+:::tip[Why the scan is deliberately not cached]
+The image the scan reads is only reachable through the container engine, so a restored cache entry would report a pass for an image that may no longer be there. With `"cache": false` the scan always runs against the real image, and fails loudly if it is missing. Pin the Trivy image version (e.g. <code>trivy:<TrivyVersion /></code>) so scans are reproducible.
+
+This means every `trivy` run does a full `docker save` plus scan per image — expect tens of seconds per image rather than a free cache hit, which is why the scan belongs in its own CI job rather than in `build`. It also refreshes Trivy's vulnerability database on each run, so it needs network access; that refresh is the point, since a scan is only as good as the CVE data behind it. On a network-isolated runner, mount a persistent cache directory at `/root/.cache/trivy` and add `--skip-db-update`, having seeded the database in an earlier step.
 :::
```
cogwirrel: can we use a cache input of the hash for a docker image? I think nx supports a cli command to return something to use as a cache input.

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Applied verbatim.

One adjustment to the surrounding prose so "no performance penalty" is not read as covering the Trivy scan too: the scan block further down (`Why the scan is deliberately not cached`) now opens with "Unlike the image build above — where the engine's layer cache absorbs the cost — this one is not free". The distinction matters because uncaching the *scan* genuinely does cost tens of seconds per image (full `docker save` + scan, and a vulnerability DB refresh), whereas uncaching the *build* is absorbed by the layer cache exactly as you describe — I measure a warm rebuild at ~0.5s for a trivial image and ~0.35s/0.85s (TS/Python) for the vended ones.

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Good instinct, and Nx does support it — `{"runtime": "<command>"}` as an input. I tried exactly this first, because it keeps caching, and it turns out to be **self-defeating in the naive form**. Minimal two-target probe, runtime input `docker images -q rtprobe:latest`:

```
start: image ABSENT
$ nx run p:img
  nx run p:img
  image PRESENT
$ docker rmi -f rtprobe:latest && nx run p:img
  nx run p:img  [existing outputs match the cache, left as is]
  image ABSENT
$ docker rmi -f rtprobe:latest && nx run p:img
  nx run p:img  [existing outputs match the cache, left as is]
  image ABSENT   <-- build reported done, image never rebuilt
```

The mechanism: when the image is gone the probe returns empty, and **that empty value gets its own cache entry**. From then on "image absent" is a cache *hit*, so Nx replays it and never rebuilds. A `--format '{{.Id}}'` digest variant has the identical hole — any value derived from the missing image is itself cacheable. It looks like a fix and silently is not, which is the worst failure mode for a soundness bug.

**I did find a sound variant**, so this is not a dead end on the facts — it just loses on the trade. Make the absent case hash to something *unique* rather than to a stable empty string, so it can never match a prior entry:

```
{ "runtime": "docker image inspect --format '{{.Id}}' img:tag 2>/dev/null || date +%s%N" }
```

That is genuinely sound — 3/3 delete-then-build cycles rebuilt the image, and it is even marginally faster than `cache: false` in steady state (~0.9s vs ~1.4s per invocation on my probe). I still do not think it is worth shipping:

1. **What the cache hit buys is a sub-second no-op.** A hit only occurs when the image is already present *and* unchanged — precisely the case where `docker build` is already a warm layer-cache hit (~0.5s on the probe). We would be adding machinery to skip work that is nearly free.
2. **The fallback is not portable.** `2>/dev/null` and `date +%s%N` are POSIX-shell-isms; the vended generators have to work under `cmd.exe`, and the repo runs Windows smoke tests.
3. **It makes the input nondeterministic**, so every absent-image build mints a fresh cache entry (I watched `.nx/cache` grow one per cycle) and invalidates anything downstream keyed on it.
4. **It fails open when the engine is down** — with `DOCKER_HOST` pointed at nothing the target still exits 0, so the sound-looking config still cannot distinguish "no image" from "no daemon".

For completeness, the other two options I measured and rejected: **`docker save` as a declared output** — sound, but 462MB and 24–53s per image against that ~0.5–2s warm rebuild; and **splitting context-prep into its own cached target**, keeping only `docker build` uncached — I A/Bd it and it showed no reliable win, since restoring a 188MB Python context from the Nx cache costs about what re-copying it does.

So `cache: false` is the one that is sound, portable, deterministic, and free. Happy to revisit if you would rather take the fail-closed runtime input despite the Windows constraint — the code change is small.

docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -296,8 +299,10 @@ The scan is intentionally not wired into `build` since this can introduce unnece
...
+:::tip[Why the scan is deliberately not cached]
+The image the scan reads is only reachable through the container engine, so a restored cache entry would report a pass for an image that may no longer be there. With `"cache": false` the scan always runs against the real image, and fails loudly if it is missing. Pin the Trivy image version (e.g. <code>trivy:<TrivyVersion /></code>) so scans are reproducible.
+
+Unlike the image build above — where the engine's layer cache absorbs the cost — this one is not free: every `trivy` run does a full `docker save` plus scan per image, so expect tens of seconds per image rather than a cache hit. That is why the scan belongs in its own CI job rather than in `build`. It also refreshes Trivy's vulnerability database on each run, so it needs network access; that refresh is the point, since a scan is only as good as the CVE data behind it. On a network-isolated runner, mount a persistent cache directory at `/root/.cache/trivy` and add `--skip-db-update`, having seeded the database in an earlier step.
 :::
```
cogwirrel: Just remove the tip altogether

cogwirrel (dismissed):

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Removed. The `Run Trivy in CI` tip above it and the `Suppressing findings` note below are untouched.

Two details in that block were arguably user-actionable rather than rationale, and they went with it — flagging in case you want either back as a plain sentence:

- Pinning the Trivy image version for reproducible scans. The vended target already pins it, and the JSON snippet on this page shows the pinned tag, so this was advice for someone hand-rolling their own target.
- For network-isolated runners: mount a persistent `/root/.cache/trivy` and add `--skip-db-update`, since the scan now refreshes the vulnerability DB on every run.

Removing the block also left the `TrivyVersion` import unused, so that is gone too.

nx-plugin-for-aws: Done.

cogwirrel (approved):

PR #763, opened 2026-06-05
docs(ts#rdb): remove Docker-specific mention from docs

AlexTo:
docs(ts#rdb): remove Docker-specific mention from docs

chore(ts#rdb): rename `docker-start.ts` and `docker-pull.ts` to `start-container.ts` and pull-image.ts

fix(ts#rdb): fix start-container.ts to remove container correctly on exit for both Docker and Finch

### Description of how you validated changes

Unit Tests

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/awslabs/nx-plugin-for-aws/blob/main/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #745, opened 2026-06-01
feat(ts#dynamodb): add DynamoDB support

AlexTo:
### Reason for this change

nx-plugin-for-aws provides generators for APIs and agents but had no support for DynamoDB as a managed data store. Setting up DynamoDB Local for local development, wiring infrastructure for CDK and Terraform, and connecting it to other project types involves a number of non-trivial decisions:

* Running DynamoDB Local in a container (Docker or Finch) with shared-container coordination across multiple table packages
* Keeping local table schema (GSIs) in sync automatically on each serve-local run
* Generating CDK/Terraform infrastructure with encryption, deletion protection and billing mode configured by default
* Connecting DynamoDB to tRPC APIs, Smithy APIs, TypeScript agents and MCP servers via connection generators

### Description of changes

* New `ts#dynamodb` generator producing a TypeScript DynamoDB project (ElectroDB, GSIs, local dev scripts, CDK/Terraform infrastructure)
* Docker and Finch container engine support for local dev scripts, resolved from aws-nx-plugin.config.mts
* Connection generators for tRPC, Smithy, TypeScript agent, and MCP server
* Full documentation guide localised into 9 languages

### Description of how you validated changes

* Unit tests with snapshot coverage for Docker/Finch, CDK/Terraform, and all four connection generators
* E2e smoke tests iterating over iacProvider × containerEngine (CDK/Terraform × docker/finch)
* Manual testing: generated tables with CDK and Terraform using both Docker and Finch, verified serve-local lifecycle and deployed to AWS to confirm end-to-end connectivity

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/awslabs/nx-plugin-for-aws/blob/main/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #1159, opened 2026-08-30
fix(ts#dynamodb,ts#rdb): depend on pull-image from the dev target

nx-plugin-for-aws:
### Reason for this change

The `dev` target of `ts#dynamodb` and `ts#rdb` starts the local container but does not declare `dependsOn: ['pull-image']`, unlike its `py#dynamodb` and `py#rdb` counterparts, which do.

This is a regression rather than a deliberate difference. Both TS generators authored the dependency when they were first added (#745 for `ts#dynamodb`, #552 for `ts#rdb`, where it was spelled `docker-pull` and later renamed to `pull-image` by #763); it was dropped when `serve-local` was renamed to `dev` in #806, which kept it for the Python generators:

```
-  projectConfig.targets['serve-local'] = {
+  projectConfig.targets['dev'] = {
-    dependsOn: ['pull-image'],
```

The impact is not merely cosmetic. For `ts#dynamodb`, `dev` runs `start-container` and `create-local-table` **in parallel**, so on a machine without the image the table script races the pull: `create-local-table` starts connecting immediately and exhausts its 30-attempt retry budget before DynamoDB Local is listening, then exits with `ECONNRESET`. The container comes up but the table is never created — and because `dev` is a continuous target, `nx dev` still prints `Successfully ran target dev`, so the failure is silent.

`ts#rdb` runs a single command, so an inline image check in `start-container` was previously masking the missing dependency there; for that generator this change is a consistency/observability fix (the pull is reported as its own task rather than happening mid-`dev`).

With `pull-image` now gating `dev`, the inline `ensureImage()` in the vended `start-container.ts` was doing work that had already happened, so it is removed — fetching the image is expressed in exactly one place and 15 lines leave every user's workspace. Thanks to @AlexTo for [pointing this out](https://github.com/awslabs/nx-plugin-for-aws/pull/1159#issuecomment-5468721814). The two shared templates cover the TypeScript and Python DynamoDB and relational database generators alike. A standalone `start-container.ts` still works — the container engine pulls an absent image implicitly on `run` (verified below), so no replacement guard is needed.

**No migration for the script change.** `start-container.ts` is vended `KeepExisting` into `packages/common/scripts/`, so it is user-owned and may carry local edits. The leftover `ensureImage()` in an existing workspace is redundant but harmless — it inspects an image that `pull-image` has already fetched — so rewriting a user's script to delete it would risk clobbering their changes for no functional gain. Existing workspaces keep their copy; new ones get the simpler script. The `dependsOn` change *is* migrated, since that lives in generator-owned `project.json`.

### Description of changes

- `ts#dynamodb` and `ts#rdb` now author `dependsOn: ['pull-image']` on `dev`, matching the Python generators.
- New `dev-target-depends-on-pull-image` migration backfills existing workspaces. It only touches projects whose `metadata.generator` is `ts#dynamodb`/`ts#rdb` and whose `dev` target still runs the generated `start-container.ts`; a customised `dev` target is left alone and reported via `nextSteps`. It preserves any existing `dependsOn` entries, normalises target key order, and is a no-op on re-run.

The migration dedupes with `addDependencyToTargetIfNotPresent`, so a hand-written `dependsOn: [{ target: 'pull-image' }]` (a shape the Nx project schema permits) is recognised rather than duplicated, and a `dependsOn` that is not an array is reported via `nextSteps` rather than rewritten.

I also swept the rest of the generators for the same class of bug — a `dev`/`serve` target needing a container or downloaded artefact without depending on the target that fetches it. These four were the only affected cases: no other dev/serve target starts a container, and the remaining artefact-producing targets (`docker`, `copy-ssdk`) are already wired to their consumers.

Docs needed no change — `snippets/dynamodb/local-dev-start.mdx` already documents `dev` as pulling the image first, which was only true for Python before this change.

Unit tests: updated the `ts#dynamodb`/`ts#rdb` generator specs (including the MySQL `dev` assertion) and added 13 migration tests covering both generators, key order, preserved `dependsOn`, the customised-target `nextSteps` path, other generators being left alone, a project with no `pull-image` target, and idempotency. One fixture is derived from `ts#dynamodb` itself — it runs the generator, strips `dependsOn` from the result, and asserts the migration converges on exactly what the generator vends, so the fixture cannot drift from the generated shape. Full suite green: 3961 tests / 221 files. Lint clean.

End-to-end, in a generated workspace using the locally built plugin with the image absent:

| | image absent, before | image absent, after |
|---|---|---|
| task order | `dev` only | `pull-image` then `dev` |
| `Created table` / `Created GSI` | 0 | 3 |
| `ECONNRESET` / `TimeoutError` | yes, `create-local-table` crashed | none |
| `nx dev` reported | `Successfully ran target dev` | `Successfully ran target dev` |

After removing `ensureImage()`, re-verified in a fresh workspace built from this branch with a timestamped probe in all three vended scripts:

- **Cold (`docker rmi` first):** `PULL_START 0ms -> PULL_END 6820ms`, then `start-container` at `+255ms` and `create-local-table` at `+386ms` after `PULL_END`. Both parallel commands are gated behind the pull, table and 2 GSIs created, zero `ECONNRESET`/`TimeoutError`.
- **Image already present:** `pull-image` completes in 18ms with zero registry traffic, and the full `dev` run creates the table with no pull attempts — so `nx dev` still works with no network.
- **Standalone `tsx start-container.ts`, image absent:** still succeeds. `docker run` pulls the image implicitly and DynamoDB Local starts, so removing the inline pull does not strand this path and no error-message guard is warranted.

Migration verified end-to-end on a pre-change workspace via `nx migrate --run-migrations=migrations.json`: it updated both `project.json` files, produced `dependsOn: ["pull-image"]` in Nx's key order, and a second run reported `No changes were made`.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/dev-target-depends-on-pull-image/metadata.json
+++ b/packages/nx-plugin/src/migrations/latest/dev-target-depends-on-pull-image/metadata.json
@@ -1,0 +1,3 @@
+{
+  "description": "Add dependsOn pull-image to the dev target of ts#dynamodb and ts#rdb projects"
+}
--- a/packages/nx-plugin/src/migrations/latest/dev-target-depends-on-pull-image/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/dev-target-depends-on-pull-image/migration.spec.ts
@@ -7,12 +7,22 @@ import {
   type ProjectConfiguration,
   readProjectConfiguration,
   type Tree,
+  updateProjectConfiguration,
 } from '@nx/devkit';
 import { PY_DYNAMODB_GENERATOR_INFO } from '../../../py/dynamodb/generator.js';
-import { TS_DYNAMODB_GENERATOR_INFO } from '../../../ts/dynamodb/generator.js';
+import {
+  TS_DYNAMODB_GENERATOR_INFO,
+  tsDynamoDBGenerator,
+} from '../../../ts/dynamodb/generator.js';
 import { TS_RDB_GENERATOR_INFO } from '../../../ts/rdb/generator.js';
+import { resolveContainers } from '../../../utils/containers.js';
+import { readProjectConfigurationUnqualified } from '../../../utils/nx.js';
 import { createTreeUsingTsSolutionSetup } from '../../../utils/test.js';
 import migration from './migration.js';
+
+vi.mock('../../../utils/containers', () => ({
+  resolveContainers: vi.fn(),
+}));
 
 const pullImageTarget = (scriptsDir: string) => ({
   executor: 'nx:run-commands',
@@ -126,7 +136,60 @@ describe('dev-target-depends-on-pull-image migration', () => {
 
     expect(
       readProjectConfiguration(tree, 'my-table').targets.dev.dependsOn,
-    ).toEqual(['pull-image', '^build']);
+    ).toEqual(['^build', 'pull-image']);
+  });
+
+  it('should treat the object form of the pull-image dependency as already migrated', async () => {
+    const project = dynamoDbProject();
+    project.targets.dev.dependsOn = [{ target: 'pull-image' }];
+    addProjectConfiguration(tree, 'my-table', project);
+
+    const { nextSteps } = await migration(tree);
+
+    expect(nextSteps).toHaveLength(0);
+    expect(
+      readProjectConfiguration(tree, 'my-table').targets.dev.dependsOn,
+    ).toEqual([{ target: 'pull-image' }]);
+  });
+
+  it('should add pull-image alongside an unrelated object form dependency', async () => {
+    const project = dynamoDbProject();
+    project.targets.dev.dependsOn = [{ target: 'build', projects: ['other'] }];
+    addProjectConfiguration(tree, 'my-table', project);
+
+    await migration(tree);
+
+    expect(
+      readProjectConfiguration(tree, 'my-table').targets.dev.dependsOn,
+    ).toEqual([{ target: 'build', projects: ['other'] }, 'pull-image']);
+  });
+
+  it('should add pull-image when the object form targets another project', async () => {
+    const project = dynamoDbProject();
+    project.targets.dev.dependsOn = [
+      { target: 'pull-image', projects: ['other'] },
+    ];
+    addProjectConfiguration(tree, 'my-table', project);
+
+    await migration(tree);
+
+    expect(
+      readProjectConfiguration(tree, 'my-table').targets.dev.dependsOn,
+    ).toEqual([{ target: 'pull-image', projects: ['other'] }, 'pull-image']);
+  });
+
+  it('should report a dependsOn that is not an array rather than rewriting it', async () => {
+    const project = dynamoDbProject();
+    (project.targets.dev as any).dependsOn = '^build';
+    addProjectConfiguration(tree, 'my-table', project);
+
+    const { nextSteps } = await migration(tree);
+
+    expect(nextSteps).toHaveLength(1);
+    expect(nextSteps[0]).toContain('my-table');
+    expect(
+      readProjectConfiguration(tree, 'my-table').targets.dev.dependsOn,
+    ).toEqual('^build');
   });
 
   it('should skip and report a customised dev target', async () => {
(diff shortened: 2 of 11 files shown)


AlexTo:
I think the `ensureImage` function in `start-container.ts` introduced in #806 is also redundant.  `pull-image` needs to happen synchronously before both `start-container` and `create-local-table` can start.

nx-plugin-for-aws:
Thanks @AlexTo — I tested both halves of this. The ordering point is right (and already holds), but I'd push back on removing `ensureImage()`, because it's still the only protection on a path that doesn't go through `pull-image`.

### `pull-image` gating both commands — confirmed

I instrumented all three vended scripts with timestamps and ran `nx dev` cold (`docker rmi` first, isolated tag so as not to disturb anything else). `dependsOn: ['pull-image']` does gate **both** parallel commands, not just the target as a whole:

```
event                      t+ms
PULL_START                    0
PULL_END                  11353   <- 11.4s cold pull completes
STARTCONTAINER_ENTRY      11613   <- +260ms after PULL_END
CREATETABLE_ENTRY         11760   <- +407ms after PULL_END
```

Both commands start *after* `PULL_END`, and `ENSUREIMAGE_PULLING` never fires on this path — so with the fix in place, `ensureImage()` is indeed doing no work when you go through `dev`. That's the redundancy you're describing, and it's real for that path.

Mechanically it's guaranteed rather than lucky: `pull-image` has no `continuous` flag, so `create-task-graph.js` routes it to the hard `dependencies` map (not `continuousDependencies`), and `tasks-schedule.js` won't schedule `dev` until its dependencies complete. `parallel: true` then applies *within* `dev`, after that gate.

### Why I'm keeping `ensureImage()`

It's reachable without `pull-image`. `start-container.ts` is vended into `packages/common/scripts/` as a user-owned file (`OverwriteStrategy.KeepExisting`) and documented in the guides as a standalone script. Run directly from the project directory with the image absent:

```
__PROBE STARTCONTAINER_ENTRY
__PROBE ENSUREIMAGE_PULLING     <- fired; image afterwards: PRESENT
```

Without `ensureImage()` that path becomes `docker run` against a missing image. It's also what makes the script self-contained for anyone composing it into their own tooling — the two scripts don't have to be invoked as a pair.

The cost of keeping it is one `docker image inspect` (~18ms measured) on a path where the image is already guaranteed present, and it's what preserves the offline property: with the image local, `pull-image` finished in 18ms and `ensureImage` made no network call, so `nx dev` still works with no registry.

So I'd frame the two as complementary rather than duplicative: `dependsOn` makes the pull a visible, ordered task for the normal `nx dev` path; `ensureImage()` keeps the script correct on its own. Happy to remove it if you'd rather the scripts be strictly target-driven — it'd want a migration note either way, since existing workspaces own their copy and wouldn't pick up a template change.

One correction to the issue text while I'm here: the dependency wasn't introduced by #806 — #806 is where it was *dropped* for the TS generators (it kept it for the Python ones). It came from #745 (`ts#dynamodb`) and #552 (`ts#rdb`, as `docker-pull`, renamed by #763). `ensureImage()` was added in #806, which is plausibly why the TS regression went unnoticed for so long: it masked the missing dependency in the single-command `ts#rdb` case.

nx-plugin-for-aws:
You're right, and I was wrong to push back — I've removed `ensureImage()` in a43831e8.

My reasoning weighed the ~18ms runtime cost and treated the standalone-script path as justification for keeping it. That was the wrong frame: vended code is read and maintained by every user who receives it, so redundant defensive code has a real ongoing cost even when it's fast. Fetching the image now lives in exactly one place, and 15 lines leave every user's workspace.

Two things I'd got wrong beyond the principle:

- I claimed removing it would strand a standalone `tsx start-container.ts`. It doesn't — I retested cold and `docker run` pulls an absent image implicitly, so the container starts fine. My earlier test only showed `ensureImage()` *firing*, which I over-read as it being load-bearing. Since the path still works, no replacement error-message guard is needed either.
- The `ts#rdb` case I'd used to argue for keeping it is actually the better argument against: `ensureImage()` was *masking* the missing `dependsOn` there, which is plausibly why the regression survived from #806 until now.

It's a two-line change in the source but covers all four generators, since the DynamoDB and RDB `start-container.ts` templates are shared between the TypeScript and Python variants.

Retested with a timestamped probe in all three vended scripts:

- **Cold** (`docker rmi` first): `PULL_START 0ms → PULL_END 6820ms`, then `start-container` at `+255ms` and `create-local-table` at `+386ms`. Both parallel commands still gated behind the pull; table and 2 GSIs created; zero `ECONNRESET`.
- **Image present**: `pull-image` returns in 18ms with no registry traffic, full `dev` creates the table with no pull attempts — still works offline.
- **Standalone, image absent**: succeeds as described above.

On docs: I checked all four guides plus the dungeon-game tutorial, and none instruct running these scripts directly — they're `<FileTree>` descriptions, and the local-dev sections already point at `nx dev`. `snippets/dynamodb/local-dev-start.mdx` still accurately says the `dev` target pulls the image first via `pull-image`. So nothing was made wrong by this change and I've left the docs alone rather than adding a caveat for a path we don't document.

**No migration for the script removal**, deliberately: `start-container.ts` is vended `KeepExisting`, so it's user-owned and may carry local edits. A leftover `ensureImage()` is redundant but harmless — it inspects an image `pull-image` has already fetched — so rewriting someone's script to delete it risks clobbering their changes for no functional gain. It ages out as workspaces are recreated. The `dependsOn` change *is* migrated, since that lives in generator-owned `project.json`. Happy to add one if you'd rather it be cleaned up actively.

cogwirrel (approved):

PR #1160, opened 2026-08-30
perf(open-api#ts-client,open-api#ts-hooks): run codegen targets in-process

nx-plugin-for-aws:
### Reason for this change

Generated OpenAPI codegen targets shelled out to a nested `nx g` through `nx:run-commands`. Every codegen step therefore paid a **second full Nx bootstrap** — project graph construction, plugin resolution, generator schema validation — for work the Nx process running the target had already done.

Measured on a 14-project workspace (ts + py APIs, agents, MCP servers), `NX_DAEMON=true`, `--skip-nx-cache`, 12 randomised-order interleaved reps:

| arm | min | p50 |
|---|---|---|
| `bench-noop` (`node -e ''` through `nx:run-commands` — the outer bootstrap floor) | 1450ms | 2321ms |
| shelled-out codegen | 4435ms | 5419ms |
| in-process codegen | 2448ms | 3875ms |

So the shelled-out target cost **~3.0s over the bootstrap floor** while the same generator run in-process costs **~1.0s** — roughly 2s per target was the nested bootstrap, not the code generation. Timed from inside the executor, the actual work is ~200ms of generator plus ~200ms of module loading; a bare `node` script doing the identical generation took 1.3s against the nested `nx g`'s 5.2s in the same shell.

This compounds: `generate:<api>-metadata` lands on the shared constructs project's `compile`, so N APIs serialise N nested bootstraps into one project's compile.

### Description of changes

**Inventory** — four target families shelled out to a nested `nx g` (verified by grepping `packages/nx-plugin/src` for `commands` containing `nx g`; these are the only four):

| family | generator | vended by |
|---|---|---|
| `generate:<api>-client` | `open-api#ts-hooks` | `utils/connection/open-api/react.ts` |
| `generate:<api>-metadata` | `open-api#ts-metadata` | `utils/api-constructs/open-api-metadata.ts` (CDK) |
| `generate:<api>-operations` | `open-api#json-metadata` | `utils/api-constructs/open-api-metadata.ts` (Terraform) |
| `<agent>-generate-client` | `open-api#ts-client` | `py/agent/generator.ts` |

The `watch-generate:*` targets also invoke `nx`, but they wrap `nx watch -- nx run <project>:<codegen target>`, i.e. they invoke the codegen *target* rather than the generator. They are unchanged and out of scope.

**Changes:**

- New `@aws/nx-plugin:open-api-codegen` executor (`src/open-api/codegen/executor.ts`, registered in `executors.json`). It dynamically imports the requested generator, runs it against an `FsTree` rooted at `workspaceRoot`, then `printChanges` + `flushChanges` — the same sequence `nx g` itself performs. Generators are loaded on demand so a run only pays for the code generation it asks for.
- `src/utils/open-api-codegen-target.ts` builds the target configuration, so all four call sites agree on `cache`/`inputs`/`outputs`/`dependsOn` and on Nx's key order (via the existing `normalizeTargetKeyOrder`, keeping generation idempotent).
- Migration `latest-open-api-codegen-executor` rewrites existing targets.

**Constraints handled:**

- **Tree flushing** — `tree.lock()` then `flushChanges(workspaceRoot, tree.listChanges())`, mirroring `nx/src/command-line/generate`.
- **`--dry-run`** — declared in the executor schema so Nx normalises both `--dry-run` and `--dryRun` to `dryRun`; changes are printed and nothing is written. Verified: `--dry-run` printed the three `CREATE` lines and left the output directory absent.
- **Caching** — `cache`, `inputs` and `outputs` are unchanged, so hashing semantics are identical. Verified in a real workspace: repeat run `2/2 hit (100%)`, and deleting the output directory restored it from cache rather than re-running.
- **Formatting** — the generators call `formatFilesInSubtree` themselves; the executor does not bypass it. A test asserts the flushed files are formatted.
- **Individually runnable** — `nx run '<project>:generate:<api>-client'` and `nx run-many -t 'generate:<api>-client' -p <project>` both work. (Bare `nx run web:generate:py-api-client` fails to parse the colon in the target name — confirmed identical on `main`, so unchanged by this PR.)

**Alternatives considered:**

- **`nx:run-commands` with a `tsx` entrypoint importing the generator** — same in-process saving, but pays a `tsx`/TypeScript startup per step, needs a script vended into every workspace (more generated surface to migrate and to keep in step with the plugin), and loses schema validation. Rejected.
- **Batching several codegen steps into one invocation** — would amortise the bootstrap further, but the steps live on *different projects* (`web`, `common-constructs`, the agent project), so batching means either a synthetic owning project or cross-project outputs. Both weaken Nx's caching granularity: one changed spec would invalidate every client. Rejected; the per-target fix already removes the bootstrap, which was the actual cost.
- **`nx exec`** — still spawns a new Nx process, so it does not address the bootstrap.

**Known risk: the deep import into Nx internals.** The executor imports `FsTree`, `flushChanges` and `printChanges` from `nx/src/generators/tree.js`, which is not a documented Nx entry point — it resolves via nx's `"./src/*"` export wildcard. `@nx/devkit` re-exports `FsTree` only from its `internal` entry and does not export `flushChanges` at all, so there is no public path to the tree primitives a generator-running executor needs; this is exactly what `nx`'s own `nx g` command does internally.

Stating the exposure plainly for reviewers: this is the **first use of that deep import in shipped runtime code** (the pattern already exists in this repo, but only in `*.spec.ts`). `nx` is a pinned `peerDependencies: "23.1.2"`, so it cannot drift under an existing install — but a future Nx release that narrowed the `./src/*` export would break the executor **in user workspaces at run time**, rather than failing here at build time. Mitigations considered and rejected as worse than the risk: vendoring a copy of `FsTree` (duplicates Nx-version-sensitive behaviour we would then have to keep in step), or a runtime `try`/fallback to shelling out (reintroduces the cost this PR removes, on a path that would then be almost never exercised and so never tested). The import site carries a comment recording why it is there, so the constraint is visible to whoever next touches it.

**Out of scope: dev-server startup.** `web:dev` startup was measured before/after (7 interleaved reps, daemon on): before p50 7274ms, after p50 7381ms — **no change**, and it is not the same problem. `dev` runs `vite` directly through `nx:run-commands`, so there is no nested `nx g` on its path; its `dependsOn` codegen does get faster, but startup is dominated by the API dev servers starting in parallel, which sit on the critical path either way. The remaining gap between `nx run ... dev` and raw `npx vite` is the *outer* Nx bootstrap, which is Nx's own cost and not something this plugin can remove. Left alone deliberately.

**Byte-identical output.** For each of the four families, generated into the same directory from the same spec, once via the shelled-out `nx g` the old targets ran and once via the new executor, then `diff -r`:

```
web              (ts-hooks):      IDENTICAL (3 files)
common-constructs (ts-metadata):  IDENTICAL (1 files)
py_lib           (ts-client):     IDENTICAL (2 files)
tf               (json-metadata): IDENTICAL (1 files)
```

**Before/after, real workspace, both daemon modes.** `before` = unmodified plugin, `after` = locally compiled plugin with the migration applied. Randomised-order interleaved A/B in one window so load hits both arms equally; wall clock, `--skip-nx-cache`, all runs exit 0.

`generate:py-api-client`, `NX_DAEMON=true`, n=12 — **6718ms → 2987ms p50 (-3731ms, 56% faster)**

| rep | before | after | saved |
|---|---|---|---|
| 1 | 5202 | 2894 | 2308 |
| 2 | 7064 | 3468 | 3596 |
| 3 | 6803 | 2849 | 3954 |
| 4 | 6606 | 2910 | 3696 |
| 5 | 6536 | 2866 | 3670 |
| 6 | 6950 | 2728 | 4222 |
| 7 | 7442 | 3267 | 4175 |
| 8 | 7527 | 3108 | 4419 |
| 9 | 7110 | 3516 | 3594 |
| 10 | 6634 | 3064 | 3570 |
| 11 | 6448 | 2889 | 3559 |
| 12 | 6595 | 3143 | 3452 |

`generate:py-api-client`, `NX_DAEMON=false`, n=12 — **5416ms → 4000ms p50 (-1416ms, 26% faster)**

| rep | before | after | saved |
|---|---|---|---|
| 1 | 5927 | 4370 | 1557 |
| 2 | 6586 | 4378 | 2208 |
| 3 | 5507 | 4139 | 1368 |
| 4 | 5272 | 3999 | 1273 |
| 5 | 5258 | 4163 | 1095 |
| 6 | 5647 | 3699 | 1948 |
| 7 | 5127 | 3938 | 1189 |
| 8 | 5410 | 3892 | 1518 |
| 9 | 5422 | 3797 | 1625 |
| 10 | 5784 | 4253 | 1531 |
| 11 | 5360 | 4001 | 1359 |
| 12 | 5201 | 3879 | 1322 |

All three codegen targets together (daemon on, n=10) — **7414ms → 3054ms p50 (-4360ms, 59% faster)**; and `common-constructs:compile`, where the metadata target sits (daemon on, n=8) — **10307ms → 4548ms p50 (-5759ms)**.

Note the daemon-off numbers are smaller: without the daemon the *outer* bootstrap is itself much more expensive, so it dominates a larger share of the total. The saving is the nested bootstrap either way.

**Cache still works** (real workspace, post-migration): populate → `1/2 hit`; repeat → `2/2 hit (100%)`, nothing re-run; delete the output directory → `2/2 hit`, outputs restored from cache.

**Migration** applied to a real 14-project workspace updated exactly the three codegen targets, `nextSteps: []`, and `nx run-many -t build` for the affected projects passed afterwards. It is idempotent, and reports rather than clobbers a target that has diverged (extra flag, chained command, extra option) — 11 unit tests cover all four generators, cross-project rewriting, the untouched watch target, each divergence shape, unrelated targets, and idempotency.

**Suites:** `pnpm nx run @aws/nx-plugin:test` — **3967 passed / 222 files**, including 8 new executor tests. 7 pre-existing specs asserted on the old `options.commands` string and were updated to assert the executor and its options; no snapshot changes. `pnpm nx run-many --target lint --all --configuration=fix` clean (exit 0).

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/open-api-codegen-executor/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/open-api-codegen-executor/migration.ts
@@ -1,0 +1,129 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+  getProjects,
+  type MigrationReturnObject,
+  type TargetConfiguration,
+  type Tree,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import type { OpenApiCodegenTarget } from '../../../open-api/codegen/executor-schema';
+import { formatFilesInSubtree } from '../../../utils/format.js';
+import { normalizeTargetKeyOrder } from '../../../utils/nx.js';
+import { OPEN_API_CODEGEN_EXECUTOR } from '../../../utils/open-api-codegen-target.js';
+
+/**
+ * Codegen targets ran `nx g @aws/nx-plugin:open-api#<generator>` through
+ * `nx:run-commands`, paying a second full Nx bootstrap — project graph, plugin
+ * resolution and schema validation — for every codegen step. They now use the
+ * `open-api-codegen` executor, which runs the same generator in the Nx process
+ * that is already running.
+ *
+ * Only the command shape the generators produced is rewritten. A target that has
+ * been customised — an extra flag, a chained command, options beyond `commands` —
+ * is the user's, so it is left as it is and reported.
+ *
+ * How to write a migration:
+ * - https://nx.dev/docs/kb/migration-generators
+ * - What `nextSteps` means: https://nx.dev/docs/reference/devkit/MigrationReturnObject
+ */
+
+/** The generators the executor can run, by the collection name they had. */
+const GENERATORS: Record<string, OpenApiCodegenTarget> = {
+  'open-api#ts-client': 'ts-client',
+  'open-api#ts-hooks': 'ts-hooks',
+  'open-api#ts-metadata': 'ts-metadata',
+  'open-api#json-metadata': 'json-metadata',
+};
+
+/**
+ * Matches a whole generated codegen command, binding the generator name and its
+ * two paths. Anchored, so a command carrying anything else does not match and is
+ * reported instead of rewritten.
+ */
+const CODEGEN_COMMAND =
+  /^nx g @aws\/nx-plugin:(?<generator>[\w#-]+) --openApiSpecPath="(?<specPath>[^"]+)" --outputPath="(?<outputPath>[^"]+)" --no-interactive$/;
+
+interface CodegenOptions {
+  readonly generator: OpenApiCodegenTarget;
+  readonly openApiSpecPath: string;
+  readonly outputPath: string;
+}
+
+/**
+ * The command a target runs, when it is a single-command `nx:run-commands` — the
+ * shape the generators produced. A target carrying several commands, or options
+ * beyond `commands`, has diverged.
+ */
+const singleCommand = (target: TargetConfiguration): string | undefined => {
+  if (target.executor !== 'nx:run-commands') return undefined;
+  const options = target.options as Record<string, unknown> | undefined;
+  if (!options || Object.keys(options).length !== 1) return undefined;
+  const commands = options.commands;
+  if (!Array.isArray(commands) || commands.length !== 1) return undefined;
+  return typeof commands[0] === 'string' ? commands[0] : undefined;
+};
+
+/** Parses a generated codegen command, or undefined if it is not one. */
+const parseCodegenCommand = (command: string): CodegenOptions | undefined => {
+  const groups = CODEGEN_COMMAND.exec(command)?.groups;
+  if (!groups) return undefined;
+  const generator = GENERATORS[groups.generator];
+  if (!generator) return undefined;
+  return {
+    generator,
+    openApiSpecPath: groups.specPath,
+    outputPath: groups.outputPath,
+  };
+};
+
+/**
+ * Whether a target invokes an OpenAPI generator without being the shape the
+ * generators produced, so it must be reported rather than rewritten. The `watch`
+ * targets that wrap `nx run <project>:<codegen target>` are excluded: they
+ * invoke the codegen target rather than the generator, so they stay as they are.
+ */
+const hasCustomisedCodegenCommand = (target: TargetConfiguration): boolean =>
+  target.executor === 'nx:run-commands' &&
+  JSON.stringify(target.options ?? {}).includes('@aws/nx-plugin:open-api#');
+
+export default async function migration(
+  tree: Tree,
+): Promise<MigrationReturnObject> {
+  const nextSteps: string[] = [];
+
+  for (const [projectName, project] of getProjects(tree)) {
+    let changed = false;
+
+    for (const [targetName, target] of Object.entries(project.targets ?? {})) {
+      const command = singleCommand(target);
+      const options = command ? parseCodegenCommand(command) : undefined;
+
+      if (!options) {
+        if (hasCustomisedCodegenCommand(target)) {
+          nextSteps.push(
+            `${projectName}:${targetName} has diverged from the generated shape — left untouched. Switch it to the '${OPEN_API_CODEGEN_EXECUTOR}' executor to avoid a nested Nx invocation.`,
+          );
+        }
+        continue;
+      }
+
+      project.targets[targetName] = normalizeTargetKeyOrder({
+        ...target,
+        executor: OPEN_API_CODEGEN_EXECUTOR,
+        options,
+      });
+      changed = true;
+    }
+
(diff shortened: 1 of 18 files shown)


packages/nx-plugin/src/migrations/latest/open-api-codegen-executor/migration.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  for (const [projectName, project] of getProjects(tree)) {
+    let changed = false;
+
+    for (const [targetName, target] of Object.entries(project.targets ?? {})) {
+      const command = singleCommand(target);
+      const options = command ? parseCodegenCommand(command) : undefined;
+
+      if (!options) {
+        if (hasCustomisedCodegenCommand(target)) {
```
cogwirrel, line 114: We should distinguish and not report next steps for watch targets

nx-plugin-for-aws: Good catch — fixed in 58ac5ea.

It doesn't fire on the *vended* watch target: `watch-generate:<api>-client` wraps `nx watch … -- nx run <project>:"generate:<api>-client"`, which references the target name rather than the generator id, so `hasCustomisedCodegenCommand` never matched it. Confirmed by running the migration against a generated workspace — `nextSteps: []`.

But it does fire on a **user-customised** watch target that inlines the generator id, e.g. `nx watch … -- nx g @aws/nx-plugin:open-api#ts-hooks …`. Reproduced, and the report was actively wrong advice: `nx watch` runs until interrupted, so it cannot become a single executor invocation. Same for any continuous target mentioning the generator id.

Fixed by skipping `continuous` targets. I chose `continuous` over the `watch-` name prefix because it is the property that actually makes a target un-convertible — the prefix is only a convention. A diverged target *named* `watch-` but not continuous is still reported, and there's a test pinning that distinction.

Also checked the other `nx watch` call site (`smithy/ts/api/generator.ts` → `watch-copy-ssdk`) — references `copy-ssdk`, so unaffected. And swept the full history of the vended watch target across all its former paths: `continuous: true` has always been set, so the check is sound for old workspaces too.

Three tests added (both continuous shapes skipped, plus the non-continuous `watch-` control). Verified they fail without the fix.

cogwirrel (approved):

PR #1161, opened 2026-08-30
perf(ts#project,ts#infra,terraform#project): add an assemble target for the deploy path

nx-plugin-for-aws:
### Reason for this change

`deploy-sandbox`, `deploy` and `synth` all declared `dependsOn: ["^build", "compile"]`. `build` is deliberately the "does everything" target — it produces a project's deployable artifacts **and** runs its quality gates — so depending on `^build` meant a deploy also ran every upstream project's lint, format, test and typecheck.

Measured in a 14-project CDK workspace:

| chain | tasks | of which a deploy does not need |
| --- | --- | --- |
| `infra:deploy-sandbox` | 71 | 33 (10x lint, 10x format, 10x test, 3x typecheck) |
| `tf-app:plan` | 41 | 15 |

`deploy-ci` was already lean (it consumes a prebuilt `cdk.out`); that split is correct and is unchanged.

### Description of changes

Adds **`assemble`** — the artifact-only sibling of `build` — and points the deploy family at `^assemble`.

```jsonc
"assemble": { "dependsOn": ["compile", "bundle", "docker"] }   // ts#/py#project with agent+mcp
"assemble": { "dependsOn": ["compile", "bundle"] }             // ts#trpc-api
"assemble": { "dependsOn": ["compile", "bundle", "openapi"] }  // py#fast-api
"assemble": { "dependsOn": ["compile", "<consumer>:assemble"] } // common/constructs
```

The name is Gradle's, where `assemble` means precisely this: produce the outputs without running the verification tasks. **`package` is deliberately not used** — it has a conventional meaning (packaging an artifact for publication to a package manager) and this repo's own publishable projects (`@aws/nx-plugin`, `@aws/nx-plugin-mcp`, `@aws/create-nx-workspace`) already use a `package` target for exactly that. Those are untouched, `ts#nx-plugin` still vends its `package` target, and the migration explicitly leaves any `package` target alone.

- **`ts#infra`** — `synth`, `deploy`, `deploy-sandbox`, `destroy`, `destroy-sandbox` depend on `^assemble`. `synth` is registered on `assemble` (it is the cloud assembly); `checkov` stays a `build`-only gate.
- **`terraform#project`** — `plan` depends on `assemble`. `checkov` stays on `build`. A Terraform library's `assemble` is a no-op carrying whatever consumers register on it.
- **`ts#react-website`** — `compile` depends on `^compile` rather than `^build`. Without this, `^assemble` still drags 8 lint/test tasks back in via the website.
- Two helpers, `addArtifactDependencyToTargets` and `addArtifactProjectToTargets`, register artifact targets and consumed projects on `build` and `assemble` together, so the two cannot drift.

**`build` is completely untouched**, root and per-project. It remains the target that does everything, and `nx.json`'s `targetDefaults.test.dependsOn: ["^build"]` means `infra:build -> infra:test -> ^build` still pulls the full upstream chain — that indirection is what makes narrowing `synth` safe.

Result: CDK **71 -> 34** tasks (33 -> 0 non-critical); Terraform **41 -> 21** (15 -> 0).

### Two judgement calls

**1. `deploy` as well as `deploy-sandbox`?** Both are lean. Both are interactive developer commands run against the developer's own credentials; the CI gate is `deploy-ci`, which is untouched. Giving them different dependency sets would be a surprising inconsistency for no safety gain.

The honest risk: **a developer can now deploy code that fails lint or unit tests.** That is a real behaviour change, documented with a `:::caution` in the TypeScript, Python, infrastructure and Terraform guides. Reviewers who disagree can make `deploy` lean and `deploy-sandbox` not by changing one `dependsOn` in `infra/app/generator.ts`.

**2. Terraform `checkov`.** **Runs from `build` only**, consistent with CDK where the scan sits outside the deploy path. I originally kept it on the Terraform artifact path, arguing it is a security control rather than a code-quality gate; @cogwirrel decided otherwise on review and that is the shape here. `checkov` is still wired into `build` for both application and library projects, so the scan is not removed — only the chain it sits in changed. `plan` therefore produces the artifacts the modules reference without running the scan or the module tests.

Unit tests: **3985 pass** (35 new — 6 for the helpers, 3 for `ts#infra`'s assemble target, 3 for Terraform's, 23 for the migration). All snapshot changes reviewed individually.

**Byte-identical correctness proof**, in one workspace so paths are identical, re-run after the rename:
1. `nx reset`, `rm -rf dist`, deleted all generated docker images, ran lean `infra:synth`, snapshotted `cdk.out`
2. `nx reset`, `rm -rf dist`, ran `run-many -t build --all`, snapshotted `cdk.out`
3. `diff -r -x performance-counters.json` -> **exit 0, zero differences**

All 4 `dockerImages` hashes identical across both snapshots, each `source.directory` present with a Dockerfile.

**Staleness probe** (re-run after the rename): mutated the agent's `systemPrompt` in `src/agent/agent.ts` — reachable from the bundle entrypoint, so rolldown cannot tree-shake it — and exactly one docker asset hash changed, with the new string present in the built bundle.

**`^assemble` is the correct minimum:** narrowing `synth` to `["^compile", "compile"]` fails loudly with `«CannotFindAsset» Cannot find asset at .../dist/packages/ts-api/bundle`, rc=1 — not a silent stale deploy.

**Regression checks:** `run-many -t build --all` loses **zero** tasks on both CDK (89 -> 96) and Terraform (36 -> 36), gaining only the new `assemble` nodes. `infra:build` retains every lint/format/test/typecheck/checkov task, and `tf-app:build` still schedules both `checkov` tasks. Terraform `validate` succeeds against `assemble` artifacts alone.

**A/B measured** — two workspaces differing only in their `project.json` files, alternating order per pair, warm Nx (`NX_DAEMON=true`) and warm Docker, wall clock, every exit code checked:

*CDK `infra:synth`, fully cached (n=8):* baseline median **5.93s** -> lean **2.70s** (**-54.5%**, faster 8/8)
*CDK `infra:synth`, after editing code on the deploy-critical path (n=6):* baseline **33.44s** -> lean **27.46s** (**-17.9%**, faster 6/6)
*Terraform, the work `plan` must do first — baseline `build` vs lean `assemble` (n=6):* baseline **14.60s** -> lean **7.62s** (**-47.8%**, faster 6/6)

Lean was faster in **20/20** pairs. Spread is wide because the box ran many concurrent agents (load average 6-30); medians reported, no outlier dropped. **No cold-cache win** is claimed — docker build dominates there.

### Migration

`latest-add-assemble-target` backfills `assemble` and rewires the `dependsOn`. It derives each project's `assemble` from what its own `build` declares rather than from the generator that created it, so a project whose build gained an extra artifact target keeps producing it. Anything it cannot classify is reported via `nextSteps` and left untouched — including a `package` target the user already authors.

Verified against real workspaces: 14 CDK projects and 6 Terraform projects migrate with no `nextSteps`; a second run changes nothing; and the migrated CDK workspace passes the same byte-identical `cdk.out` proof. Re-verified after `checkov` moved off `assemble`: byte-identity still exit 0, staleness probe still changes exactly one docker asset hash.

**Correcting an earlier claim in this description:** "all projects migrate with no `nextSteps`" holds for workspaces of the shapes above, but is *not* general. A workspace whose `build` declares something the classifier does not recognise will always emit `nextSteps` — by design, since guessing would be worse.

#### A stale-deploy bug this found, verified against AWS

Review found that the classifier did not know `bundle-migration` or `bundle-create-db-user`, which `py#rdb` registers on `build` **directly**. The consequence was severe and silent: the rdb project got no artifact target, while `common-constructs` had *already* been repointed at it, and **Nx silently skips a `project:target` dependency whose target does not exist** (`projectHasTarget`, no error, no warning). A deploy would ship the *previous* migration image.

I reproduced it in a real workspace (`py#rdb --infra=aurora` plus `ts#infra`) and closed it against AWS:

- **Before the fix:** `nextSteps` named `ws_rdb.db`, `common-constructs:package` referenced `ws_rdb.db:package`, that target did not exist, and `infra:synth` scheduled **4 tasks with zero `ws_rdb.db` tasks** — the migration bundles never built.
- **After the fix:** no `nextSteps`, and `infra:synth` schedules **10 tasks** including `bundle-migration` and `bundle-create-db-user`.
- **Staleness probe:** added a real Alembic migration under `migrations/versions/`; exactly one docker asset hash changed (`ac35a733…` -> `91da6316…`) and the file was present in the built image.
- **Deployed to AWS** with `nx run infra:deploy-sandbox`: the deploy ran `bundle-migration` and pushed image `91da63161c51a155…` to ECR — the hash containing the new migration, confirmed present in the repository, with the stale hash never pushed. The stack then rolled back on an unrelated pre-existing issue (`DatabaseName db cannot be used. It is a reserved word for this engine`, because I named the test project `db`), which does not affect the artifact evidence. All stacks and resources were torn down and verified gone in both regions.

Three further migration faults from the same review are fixed, each with tests: a `build` with no `dependsOn` (`ts#astro-docs`) or no `build` target at all is now a clean no-op rather than a spurious `nextSteps`; the object form of `dependsOn` is classified rather than rejected; and narrowing a hand-authored `compile` is reported rather than done silently.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/python-project.mdx
```diff
@@ -139,6 +139,18 @@ Or use the shorthand command:
 
 <PackageManagerShortCommand commands={["build"]} />
 
+### Assembling
+
+Your project also has an `assemble` target, which is the artifact-only sibling of `build`: it produces whatever your project contributes to a deployment (its compiled output, and any bundle, container image or generated specification it declares) without running the lint, test or type-check gates.
```
cogwirrel: change bracketed text to `(for example its compiled or bundled output)`

same for ts project docs

docs/src/content/docs/en/guides/python-project.mdx
```diff
@@ -139,6 +139,18 @@ Or use the shorthand command:
...
+
+<NxCommands commands={['assemble <project-name>']} />
+
+The deploy targets depend on `assemble` rather than `build`, so deploying builds only what it is about to deploy. `build` is unchanged and still does everything, so it remains what you run to check a project — and what you should run in CI.
```
cogwirrel: remove

```
`build` is unchanged and still does everything, so it remains what you run to check a project — and what you should run in CI.
```

same for ts project docs

docs/src/content/docs/en/guides/python-project.mdx
```diff
@@ -139,6 +139,18 @@ Or use the shorthand command:
...
+
+:::caution
+Because `assemble` skips the quality gates, a deploy no longer fails on a lint error, a failing unit test or a type error. Run `build` (or your CI pipeline) to check your code.
+:::
```
cogwirrel: change to:

```
:::caution[Use `build` in your pipeline]
`assemble` is used for faster iteration during development. Run `build` in your CI pipeline to ensure your project passes quality gates prior to a production deployment.
:::
```

same for ts project docs

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -259,6 +259,8 @@ Run security checks on your infrastructure using Checkov with the `checkov` targ
 
 You will find your security test results in the root `dist` folder, under `dist/packages/<my-terraform-project>/checkov`.
 
+Checkov runs as part of both `build` and `assemble`, so an interactive `plan` scans your infrastructure before you apply it.
```
cogwirrel: Remove this line and change the behaviour so that checkov does not run as part of assemble.

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -326,6 +328,14 @@ Before applying changes, you can see what Terraform will do by running the `plan
 
 This will create a plan file in `dist/packages/<my-terraform-project>/terraform/dev.tfplan`.
 
+`plan` depends on `assemble` rather than `build`, so it produces the artifacts your modules reference — the Lambda bundles and generated operations metadata — and runs Checkov, without also running every project's lint, test and type-check gates.
+
+:::caution
+Because `plan` no longer runs the Terraform module tests or your projects' unit tests, run `build` to check your code before applying anything you care about, and gate your pipeline on it:
```
cogwirrel: Docs cannot talk about previous state - do not write anything like "no longer" because users reading don't know or care what came before. change to:

```
:::caution[Use `build` in your pipeline]
`assemble` is used for faster iteration during development. Run `build` in your CI pipeline to ensure your project passes quality gates prior to a production deployment.
:::
```

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
```diff
@@ -383,6 +385,14 @@ Your project has three deploy targets, each suited to a different situation:
 | `deploy`         | Deploying any stage, by naming the stage or stacks you want.                      |
 | `deploy-ci`      | Deploying from a CI/CD pipeline, using a pre-synthesized cloud assembly.          |
 
+`deploy-sandbox` and `deploy` depend on `^assemble`, so they build the artifacts they are about to deploy and nothing else. `deploy-ci` deploys an assembly your pipeline already built, so it has no build dependencies at all.
+
+:::caution
+Because the deploy targets skip the lint, test and type-check gates, deploying will not fail on a lint error or a failing unit test. Run `build` before you deploy anything you care about, and gate your pipeline on it:
+
+<NxCommands commands={['run-many --target build --all']} />
+:::
```
cogwirrel: Change title to ``Use `build` in your pipeline``, and remove "anything you care about"

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Done — `checkov` now runs from `build` only, matching CDK. Removed from `assemble` on both the application and library Terraform targets (the library's `assemble` is now `nx:noop`), and the migration no longer classifies it as a Terraform artifact target so existing workspaces converge on the same shape. It remains wired into `build` for both project types, and I verified `tf-app:build` and `run-many -t build --all` still schedule both `checkov` tasks — the scan moved chain, it wasn't dropped.

New Terraform task count: **41 -> 21** (was 23 with `checkov` in the chain), still 15 -> 0 non-critical. `build --all` loses zero tasks (36 -> 36).

Re-ran the proofs after removing the node, since that's where an edge can silently vanish: the byte-identity `cdk.out` check is still exit 0 with zero differences, and the staleness probe still changes exactly one docker asset hash. `terraform validate` passes against `assemble` artifacts alone, and the migration reports no `nextSteps` on a real 6-project Terraform workspace.

docs/src/content/docs/en/guides/python-project.mdx
nx-plugin-for-aws: Done — applied to both the Python and TypeScript project guides.

docs/src/content/docs/en/guides/python-project.mdx
nx-plugin-for-aws: Done — sentence removed from both the Python and TypeScript project guides.

docs/src/content/docs/en/guides/python-project.mdx
nx-plugin-for-aws: Done — applied verbatim to both the Python and TypeScript project guides.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Done. Applied the caution block verbatim here too, so all four blocks across the guides now share one wording.

On the general rule: I grepped my whole diff for transition language and rewrote every instance to describe the end state. Removed "rather than `^build`" / "rather than `build`" (4 places), "`build` is unchanged and still does everything" (2), "a deploy no longer fails on..." (2), "Because `plan` no longer runs the Terraform module tests" (1), "without **also** running" (2), "for how `assemble` relates to `build`" (1), and "anything you care about" (2). The only remaining "no longer" anywhere in these guides is a pre-existing line about the state bucket that this PR doesn't touch.

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
nx-plugin-for-aws: Done — retitled to `Use `build` in your pipeline` and dropped "anything you care about". This block now matches the other three.

cogwirrel (approved):

PR #1162, opened 2026-08-30
fix(terraform#project,ts#project,py#agent): declare cache inputs correctly on fmt and cacheable targets

nx-plugin-for-aws:
### Reason for this change

Two targets we vend declare their cache inputs wrongly, so Nx invalidates work it should reuse.

**1. `terraform fmt` can never cache-hit — it rewrites its own inputs.**

The format target ran `terraform fmt` (the *writing* form) over its own declared `inputs: ["default"]`. Mutating its own inputs changes the hash it was just computed from, so the target could never register a cache hit — which is why Nx flags it as flaky on every run.

Reproduced on a 14-project workspace: badly-format `packages/tf-lib/src/main.tf`, then

```
run1 exit=0  Cache: 0/1 hit (0%)
run2 exit=0  Cache: 0/1 hit (0%)   <- immediate re-run, still a miss
run3 exit=0  Cache: 1/1 hit (100%)  <- only once it had converged
```

The TypeScript and Python `format` targets are the model: the base target **checks** and only the `fix` configuration writes.

**2. Several cacheable targets declare no `inputs`, which is broader than an explicit list.**

A target with no `inputs` gets Nx's implicit `DEFAULT_INPUTS = ["default", "^default"]` (`nx/dist/src/hasher/task-hasher.js:109`). That `^default` reads each dependency's **whole project directory** — its README, its dev scripts — which is broader than the explicit inputs its siblings carry. That is why a sibling `compile` correctly ignores a dependency's README while `bundle` does not.

I enumerated the affected targets rather than trusting a list, via `nx show project <p> --json` across every project plus a sweep of every `cache: true` site in the generators:

```
@ws-cdk/terraform:checkov      @ws-cdk/ts-lib:bundle          ws_cdk.py_api:openapi
@ws-cdk/tf-lib:checkov         @ws-cdk/ts-lib:agent-docker    ws_cdk.py_lib:agent-openapi
@ws-cdk/ts-api:bundle          @ws-cdk/ts-lib:mcp-server-docker
ws_cdk.py_lib:agent-docker     ws_cdk.py_lib:mcp-server-docker
```

The sweep also turned up `bundle-migration`, `bundle-create-db-user` (`py#rdb`) and `operations` (`ts#trpc-api`), which the reported list did not include.

### Description of changes

**Defect 1** — `terraform#project`, which vends both user Terraform projects and the shared `common/terraform`:

- The target is named **`format`**, matching TypeScript and Python projects, and runs `terraform fmt -check -diff`, failing on an unformatted file. `-diff` names what to fix.
- The writing form lives in a `fix` configuration, and `skip-lint` is a cross-platform no-op, matching `ts/lib/biome.ts` and `py/project`.
- A `lint` target (`dependsOn: ['format']`) means `run-many --target lint` reaches Terraform projects, and `--configuration=fix` / `--configuration=skip-lint` propagate through the edge. Verified that Nx does propagate a configuration across `dependsOn`.
- The migration renames the key for existing workspaces and repoints every `dependsOn` naming it — a stale `dependsOn` is not an error in Nx, it silently drops the edge, so `build` would stop checking formatting. It accepts either name as its starting point, so a workspace part-way through converges.

**The check covers the scope the writing target wrote** — `{projectRoot}/src`, non-recursive. I first tried adding `-recursive`, and CI was right to reject it: widening to nested modules surfaced **34 pre-existing unformatted files across 8 vended template families** (`app/agents`, `app/apis`, `app/dbs`, `app/mcp-servers`, `app/static-websites`, `core/api`, `core/static-website`, `core/user-identity`) that the writing target never reached. Their alignment width depends on which EJS branch renders — `CREATE_DB_USER_DOCKER_IMAGE_TAG` only appears under `framework === 'sqlmodel'` and widens the whole block when it does — so they cannot be statically aligned. That is a real latent issue, but separate from the caching defect, so it is left for its own change. Verified by rendering all 64 `.tf` files `internal#test-matrix` produces on `--infra=terraform` and checking each project's actual target scope: **both terraform projects are `fmt`-clean**.

Two things *are* in scope and needed fixing:

- The application `providers.tf` `backend "s3"` block was not `fmt`-clean. The per-run write is what had kept it formatted, so the template is realigned and the migration realigns existing copies.
- The `lint` target claims the license-check wiring itself (`addLicenseCheckToLintTarget`, as `ts#project` and `py#project` do). The idempotency smoke test caught this: without it the generator re-vended a bare `lint`, the `license` generator re-added its dependency on the next pass, and that diff made the matrix non-idempotent. The generator also now calls `formatFilesInSubtree`, since `updateProjectConfiguration` re-serialises `project.json` with every inline array expanded.

**Defect 2** — the right explicit inputs depend on whether the target has a `dependsOn`, because that is what `default`'s transitive `dependentTasksOutputFiles` entry resolves against:

- **`["default"]`** for `bundle` (`utils/bundle/bundle.ts`, covering `ts#project` and `ts#smithy-api`), `operations` (`ts#trpc-api`) and the `py#rdb` bundle targets. All carry `dependsOn: ['compile']`, so the dependency's build output is still hashed through that upstream task.
- **`["default", "^production"]`** for the three Python `openapi` targets (`py#fast-api`, `py#agent`, `py#agent#react-connection`) and for `checkov`. The `openapi` targets have **no `dependsOn`**, so that entry resolves against nothing and `default` alone would collapse to the project's own files — see H1 below. `checkov` resolves the relative Terraform modules a project consumes.
- **Not narrowed:** `ts#astro-docs`'s `build`, with a comment explaining why — `astro build` resolves imports from a dependency's *source* and declares no `dependsOn`, so `^default` is the only thing keeping a dependency's content in its hash.

**Two migrations** (`latest/terraform-fmt-check`, `latest/declare-cacheable-target-inputs`), split across commits so the ordering CONTRIBUTING describes is explicit. Both deterministic, idempotent, and shape-checking before writing.

#### Review follow-ups

Addressed cogwirrel's four comments:

- **Renamed the format target to `format`** for consistency with ts/py. This is a code change, not just docs: the exported constant, the target key, `build.dependsOn` in both target sets, `lint.dependsOn`, the specs, and the migration (which renames the key *and* repoints every `dependsOn`). Grepped exhaustively — the only remaining `fmt` occurrences are the literal `terraform fmt` command, the migration's `LEGACY_TARGET` constant, and the spec's pre-migration fixtures.
- **Rewrote the guide's section as `Linting`**, mirroring the TypeScript and Python guides' subsection structure and wording, including both `Shorthand Command` tips. Confirmed the `lint` and `build:skip-lint` shorthands `init` vends apply to terraform projects.
- **Dropped the two tips** flagged in `terraform-project.mdx` and `docker-bundling.mdx`.

Re-verified end to end in a migrated 14-project workspace: `build --all` runs `format` for both terraform projects; base `lint` fails on unformatted input; `--configuration=fix` reformats; `--configuration=skip-lint` passes leaving the file byte-identical; `run-many -t lint --all --configuration=fix` covers 14 projects; repeat `format` on formatted input is a **1/1 cache hit**; and the migration re-run reports "No changes were made". `nx run docs:build` passes with the new `PackageManagerShortCommand` import.

#### Earlier review fixes

An independent review found two HIGH faults in the input-narrowing half. I reproduced both before fixing.

**H1 — the three Python `openapi` targets served a stale spec.** They have no `dependsOn`, so `default`'s transitive `dependentTasksOutputFiles` resolves against no upstream task (`check-task-files.js:189-199` collects upstream task ids from the task graph) and `default` alone collapses to the project's own files.

Reproduced in a real 14-project workspace on the genuine `ws_cdk.py_lib -> ws_cdk.agent_connection` Python edge, with a Pydantic response model owned by the dependency and serialised into the spec by `app.openapi()`. Two targets differing *only* in `inputs`, after editing **only** the dependency:

```
Cache: 1/2 hit
  inputs:['default']               -> field_v1        <== STALE
  inputs:['default','^production'] -> field_v2_new    <== correct
```

Fixed to `["default", "^production"]`. Re-verified on the real `agent-openapi` target: cache miss on the dependency edit, spec regenerated, zero stale content. **I also found the same bug in my own migration** — it was narrowing `*openapi` targets to `["default"]`, so it would have *introduced* this into existing workspaces. Corrected and covered by a test.

On `["default","^production"]` vs #1156's `["production","^production"]`: I measured the difference against real Nx. `["production","^production"]` hard-fails with `"production" is an invalid fileset` whenever the named input is absent, whereas `["default","^production"]` only fails once the project also has a dependency. Nothing in this plugin creates that named input — **#1156's generator does**. So this PR uses the value that is correct *and* standalone-safe; once #1156 lands, moving to `["production","^production"]` is a safe one-line follow-up, and this migration skips any target already declaring `inputs`, so the two compose in either merge order.

**H2 — `realignProviders` corrupted customised workspaces.** It hardcoded the alignment width to the two vended arguments and rewrote unconditionally, but `terraform fmt` aligns `=` to the widest key in the whole block. Reproduced with a real terraform binary: a `backend "s3"` block with `bucket`/`dynamodb_table` pinned in-file is aligned to 14 and **fmt-clean**; running the migration over it pushed `encrypt`/`use_lockfile` back to 12 and `terraform fmt -check` returned **exit 3** — so `build` (which `dependsOn: ['format']`) failed immediately after a clean `nx migrate`, on a file the user never touched.

The block is now parsed before writing: one holding exactly the vended arguments is realigned, any other is left alone and reported via `nextSteps` pointing at `format --configuration=fix`, and a block already at the target width is a no-op. Verified with real terraform — **both the vended and the customised file exit 0**, where the customised one previously exited 3.

While fixing this I found the original fixture was wrong in a way that mattered: it was already padded, so the "realign" test would have passed against a file needing no work. The real pre-change template is *over*-padded (width 15, needs 12), confirmed from git history and by running real `terraform fmt -check` over it.

**M1 — `matchesVendedShape` could never match a real `operations` target.** It required every output to contain `/dist/`, but `operations` writes generated Terraform sources to `packages/common/terraform/src/generated/<api>`. Confirmed directly: `operations inputs after migration = undefined`. The check now requires declared outputs without constraining their location, and the test uses the shape the generator actually vends.

**L1 — user `configurations` were clobbered.** Spread order put vended values last, so a user's `fix: terraform fmt -recursive` became `terraform fmt`. Their configurations now win, matching how the same code already preserves `options.cwd`.

#### Scoping notes on concurrent PRs

- **`agent-docker` / `mcp-server-docker` are deliberately left alone.** #1158 is setting `cache: false` on exactly these targets for a cache-soundness bug (their `outputs` don't cover the image), which makes their `inputs` moot.
- **#1156** narrows some of the same Python targets to `["production", "^production"]`. See H1 above — the two compose in either merge order.
- **#1155** (`TF_PLUGIN_CACHE_DIR`) touches `terraform/project/generator.ts` too, but only the env of the `init` / bootstrap scripts, so the diffs do not overlap.
- **#1161** renames `package` → `assemble` in the same terraform target block and touches the `build.dependsOn` line this PR also edits. This diff is scoped to the rename, so the overlap on that line is a one-word reconciliation.
- `smithy#project`'s `compile` / `generate-ssdk` are **intentionally not narrowed**: `imports` in `smithy-build.json` point at another project's *source*, so `^default` is load-bearing there.

`pnpm nx run @aws/nx-plugin:test` — **222 files / 3983 tests pass** (3965 on main; +29 new tests, −3 obsolete `py#rdb` snapshots, reviewed). `lint --all --configuration=fix`, `build --all` and `docs:build` all clean.

Unit tests cover both generator changes and both migrations, including the negative cases: a customised format target is reported not rewritten, a foreign generator's project is skipped, an already-scoped `inputs` is preserved, a target with no declared outputs is skipped, the rename repoints `dependsOn`, a part-way-migrated workspace converges, and both migrations are asserted idempotent. A test also asserts *every* cacheable target the terraform generator vends declares `inputs`, so this class of defect cannot silently reappear there.

**Defect 1 proven end-to-end** on a migrated 14-project workspace:

| | before | after |
|---|---|---|
| unformatted input | exit 0, silently rewrote | **exit 1** (check mode) |
| `--configuration=fix` | n/a | exit 0, reformats |
| repeat on formatted input | `0/1`, `0/1`, `1/1` | `0/1`, **`1/1`**, **`1/1`** |

`run-many --target lint --all --configuration=fix` covers **14 projects** and reformats `tf-lib` (md5 confirmed changed); `run-many --target build --all --configuration=skip-lint` passes leaving the file byte-identical.

**Defect 2 proven, and proven not to under-invalidate.** Editing a *dependency's* README (`common/agent-connection/README.md`) — the case where `^default` differs:

| | tasks re-run | wall clock (3 runs, median) |
|---|---|---|
| before | **7** (incl. `ts-lib:bundle`) | 8.6s / 8.3s / 8.1s → **8.4s** |
| after | **6** (`bundle` gone) | 6.0s / 6.1s / 5.9s → **6.0s** |

And the safety check — a **real** source edit to that dependency (`src/index.ts`) still correctly re-runs `ts-lib:bundle` (11 tasks). A fix that under-invalidates would be worse than the bug.

**Honest note on wall clock:** the 8.4s → 6.0s is real but is dominated by dropping one warm rolldown bundle on a loaded box, not by anything structural. Editing a README *within* the same project still re-runs all 7 tasks and is unaffected by this PR — `default` legitimately includes the project's own README. The win is task count, cold-layer cost and CI, plus removing a permanently-flaky task; please don't read these seconds as a headline speedup.

**Migrations verified end-to-end** against a pre-change workspace via `nx migrate --run-migrations=migrations.json` with the local build swapped into `node_modules`. Both applied, the diff is what today's generators produce, `build --all` passes afterwards, `format` is a cache hit, and a second run reports **"No changes were made"** with an empty `git diff`.

### Issue # (if applicable)

N/A

### Checklist

----


nx-plugin-for-aws:
Cross-PR coordination from #1156 (Python test-input scoping), which touches the same three files: `py/agent/generator.ts`, `py/agent/react-connection/generator.ts`, `py/fast-api/react/open-api.ts`.

An independent review of both PRs found a **semantic** conflict on the OpenAPI spec targets, not just a textual one:

```
<<<<<<< #1156      inputs: [production, ^production]
=======  #1162      inputs: [default]
```

Per #1162's own H1 finding, these targets have no `dependsOn`, so `['default']` alone under-invalidates and can serve a stale spec — resolving this conflict casually, or in favour of whichever merges later, silently ships that bug.

Suggested resolution, which I believe satisfies both PRs' goals:

- **#1156 merges first**, and the conflict resolves to `['production', '^production']`.
- `^production` gives you the cross-project scoping #1162 is after (a dependency's test files no longer contribute), while `production` keeps the spec correctly invalidated by the project's own source. So this should be a superset of the `['default']` change rather than a conflict with it.

Worth noting #1156 also narrows the workspace-wide `default` named input's `dependentTasksOutputFiles` to `!{reports,coverage}/**` (keeping all generated sources hashed, excluding only pytest's timestamped reports), so if you are reasoning about dependency-output hashing anywhere in #1162, that is the value it will land on.

Happy to rebase #1156 or adjust if you would rather sequence it the other way — flagging it so it is a deliberate decision either way.

docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -96,6 +97,14 @@ Configure a `bundle` target that invokes Rolldown. If you are starting from a <L
 }
 ```
 
+:::tip[Declare `inputs` on a cacheable target]
+A cacheable target that declares no `inputs` falls back to Nx's implicit `["default", "^default"]`. That `^default` reads each dependency's **whole project directory** — its README, its dev scripts, everything — so editing a dependency's README re-runs the bundle even though its output cannot change.
+
+`["default"]` is enough **here** because this target declares `dependsOn: ["compile"]`. `default` carries a transitive `dependentTasksOutputFiles` entry, which Nx resolves against the task's upstream tasks — so with `compile` upstream, the dependency's build output is still hashed.
+
+A cacheable target with **no** `dependsOn` has no upstream task for that entry to resolve against, so `default` collapses to the project's own files. There, dropping `^default` removes the last edge to the dependency and the target will serve stale output — use `["default", "^production"]` instead.
+:::
```
cogwirrel: Remove this tip section

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -247,10 +247,26 @@ You can validate your Terraform configuration using the `validate` target:
 
 #### Formatting your Code
 
-Format your Terraform code using the `fmt` target:
+The `fmt` target checks that your Terraform code is formatted, and fails if it is not. It runs as part of `build`:
```
cogwirrel: Let's name this `format` for consistency with ts/py

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -247,10 +247,26 @@ You can validate your Terraform configuration using the `validate` target:
 
 #### Formatting your Code
```
cogwirrel: Rewrite this section as `Linting` and mirror exactly the ts/py documentation including shorthand command tips.

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -247,10 +247,26 @@ You can validate your Terraform configuration using the `validate` target:
...
+
+<NxCommands commands={['fmt tf-infra --configuration=fix']} />
+
+A `lint` target orchestrates the format check, so `run-many` reaches your Terraform projects alongside your TypeScript and Python ones:
+
+<NxCommands commands={['run-many --target lint --all --configuration=fix']} />
+
+:::tip[Why the check is the default]
+`terraform fmt` rewrites the files it reads — the very files the `fmt` target declares as its cache inputs. Writing from the base target would change its own cache key on every run, so it could never register a cache hit. Checking instead keeps the target cacheable, mirroring how the TypeScript and Python `format` targets work.
+:::
```
cogwirrel: No need for this, should be dropped when changing the whole section to be the same as ts/py linting section

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Renamed to `format`. This rippled through code, not just docs:

- `TERRAFORM_FMT_TARGET` → `TERRAFORM_FORMAT_TARGET`, and the target key in `libTargets`
- `build.dependsOn` in both `libTargets` and `applicationTargets`, and `lint.dependsOn` → `['format']`
- the migration now renames the key for existing workspaces **and repoints every `dependsOn` naming the old one**. That mattered: a `dependsOn` pointing at a target that does not exist is not an error in Nx, it silently drops the edge, so `build` would have stopped checking formatting entirely. It accepts either name as its starting point, so a workspace that already ran an earlier form of this migration converges too — covered by a dedicated test.
- specs updated; no snapshot changes were needed.

Grepped exhaustively for `'fmt'`, `"fmt"`, `:fmt`, `fmt:`, `targets.fmt` and `TERRAFORM_FMT` across `packages/nx-plugin/src`, `e2e/` and `docs/.../en/`. The only remaining occurrences are the literal `terraform fmt` command string, the migration's `LEGACY_TARGET = 'fmt'` constant, and the migration spec's pre-migration "before" fixtures — all deliberate.

**Propagation verified end to end** in a real 14-project workspace, migrated with the local build:

| | result |
|---|---|
| `build --all` | both terraform projects run `format`; edge intact |
| `lint` on unformatted input | fails (check mode, through `lint → format`) |
| `lint --configuration=fix` | passes and reformats (`format:fix` runs; md5 changed) |
| `lint --configuration=skip-lint` | passes, file byte-identical |
| `run-many -t lint --all --configuration=fix` | 14 projects, reformats terraform |
| repeat `format` on formatted input | miss, then **1/1 hit**, **1/1 hit** |

`build` also correctly fails via the `format` edge on unformatted input, and the migration re-run reports "No changes were made".

Worth flagging: #1161 is concurrently renaming `package` → `assemble` in this same target block and touches the `build.dependsOn` line. I have kept this diff to the `fmt`→`format` rename only, so the two should reconcile as a one-word overlap on that line.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Rewritten as a `Linting` section mirroring the TypeScript and Python guides. Same subsection structure and wording:

- **Running the Linter** → `lint tf-infra`
- **Fixing Lint Issues** → `lint tf-infra --configuration=fix`, then `run-many --target lint --all --configuration=fix`, with the `Shorthand Command` tip for `lint`
- **Skipping Lint Issues** → `run-many --target build --configuration=skip-lint`, with the `Shorthand Command` tip for `build:skip-lint`

Both shorthands are the workspace-wide ones `init` already vends in `CONVENIENCE_SCRIPTS`, so they apply to terraform projects unchanged — I checked rather than assuming.

Added the `PackageManagerShortCommand` import the tips need, and ran `nx run docs:build` to confirm it resolves — a missing import broke that build on a sibling PR, so I verified rather than relying on the translations regenerating cleanly.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Dropped, as part of rewriting the whole section to match the ts/py Linting sections.

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Removed. The `inputs` line stays in the `bundle` code sample above it, which is where the guidance is actually load-bearing.

Kept this change confined to the tip so it does not tangle with #1158's edits to the same guide.

cogwirrel (dismissed):

cogwirrel (approved):

PR #1164, opened 2026-08-31
chore(agentcore-harness): mark experimental generators on every catalogue surface

nx-plugin-for-aws:
### Reason for this change

#1151 added the `experimental` flag to `generators.json`. It propagates to the docs banner, the MCP `list-generators` / `generator-guide` tools and the generated `SKILL.md` / `POWER.md` tables — but two surfaces that describe a generator to a user were missed, so they disagree with the generated tables for the same data:

- **The README vended into every new workspace** (`utils/init.ts`) listed generators from `generators.json` with `description: v.description`, dropping `experimental`. The adjacent line already respected the `hidden` filter, and `make-skills.ts` was updated, so this was the one consumer left behind. A developer reading their own workspace's "Available generators" list — the closest-to-hand catalogue they have — saw no indication the harness is experimental.
- **The generator tables in `README.md` and `packages/nx-plugin/README.md`**. The latter is the npm-rendered README for `@aws/nx-plugin`, a very common first stop: `agentcore-harness` appeared identically to the stable generators, so a user could adopt it in a production workspace and be surprised on upgrade when no migration is published.

`CONTRIBUTING.md` also asserted the flag "needs nothing else authored", which was not quite true for those tables.

### Description of changes

`describeGeneratorForCatalogue` is now shared by the vended workspace README and `make-skills.ts`, so the two *generated* catalogues cannot drift apart. `SKILL.md` / `POWER.md` regenerate byte-identically.

The README table is deliberately **not** generated: it is a curated subset (20 of 29 visible generators) with hand-written marketing descriptions and markdown links that differ from every `generators.json` description, so generating it would rewrite all 20 rows and lose the curation. Instead the marker is added by hand and a unit test fails if an experimental generator's row lacks it, with a message naming the file to edit. `CONTRIBUTING.md` now documents that one manual step instead of claiming there is none.

Fixed while in the area:

- `experimental-banner.astro` indexed the JSON-inferred object literal type with a `string`, which is `error TS7053` under the repo's own `astro/tsconfigs/strict` (which `docs/tsconfig.json` extends). It compiles today only because `astro build` transpiles without type-checking and **`astro check` is not wired into any target or CI workflow** — `@astrojs/check` isn't even a dependency. Both this and `markdown-content.astro` now use a typed `findGeneratorsJsonEntry` accessor. Wiring `astro check` in would need its own PR: it is a new dependency plus a new target, and would want a look at whatever else it surfaces across the docs site.
- The **"copy as markdown"** button built `resolvedMarkdown` without the banner, which is injected as a rendered component. A user copying a guide into an agent lost the caveat entirely. It now carries the same warning string the MCP tools emit.
- **`generator-guide` skipped the warning on the unsupported-combination path**, which returns `guide.content` early and never calls `renderGeneratorInfo`. Latent for `agentcore-harness` today (single guide page, no `when:` frontmatter) but wrong for the next experimental generator with variants.

`buildGeneratorInfoList` loses its `Record<string, any>` cast, so the conditional spreads are type-checked rather than relying on `'x' in info` guards.

**Every finding was reproduced before being fixed.**

- **TS7053** reproduced under the repo's own tsc with `astro/tsconfigs/strict`, then A/B'd in one harness: old form → `error TS7053`, new form → clean.
- **Vended README (the actual user-visible outcome)** verified end-to-end in a real Nx 23.1.2 workspace using the locally compiled plugin. `agentcore-harness` renders as `Generate an AgentCore Harness project (experimental)`, exactly one marker appears, and stable generators — including the adjacent `agentcore-gateway` — stay undecorated. (First attempt showed no marker: pnpm had resolved a hard-copied store entry rather than the symlink, so the stale build ran. Replacing the store copy confirmed the fix.)
- **`docs:build`** passes: exit 0, 1513 pages, "All internal links are valid", and `git status` clean so CI's mutation gate at `pr.yml:70-71` is satisfied. Verified in the **built HTML** that the harness page's `data-copy-markdown` attribute now contains the `[!WARNING] Experimental` text (previously 19213 chars with zero occurrences) while `agentcore-gateway`'s does not, and that the rendered banner still appears — checked for `en`, `jp`, `zh` and `ko`.
- **`npx tsx scripts/make-skills.ts`** reports no drift; sanity-checked by perturbing `SKILL.md` and watching it restore byte-identically.
- **Unit tests**: new coverage for `findGeneratorsJsonEntry`, `describeGeneratorForCatalogue`, the README-table guard, and the unsupported-combination path (with a negative control asserting a stable generator gets no warning). The "surfaces the experimental flag" test now iterates every entry in `generators.json` rather than hardcoding `agentcore-harness`, so it no longer fails the day the harness graduates to stable — the exact moment the flag is *correctly* removed. The 3 affected spec files pass 85/85, and the one snapshot change is the single intended vended-README line.
- **Full suite**: 284 failed / 3673 passed vs a documented `main` baseline of 284 failed / 3664 passed — same failure count, same 25-file set, same environmental root cause (`Cannot read properties of null (reading 'version')` from `@nx/js` resolving `@nx/vitest`, which fails identically on `main` in this container), no failure in any file touched here, and +6 net new passing tests matching the additions. Upstream CI is the real gate.
- `tsc -p packages/nx-plugin/tsconfig.lib.json --noEmit` and the docs tsconfig both typecheck clean, and `nx run-many --target lint --all --configuration=fix` exits 0 with no mutation.

### Issue # (if applicable)

N/A — follow-up to #1151.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/README.md
+++ b/packages/nx-plugin/README.md
@@ -77,28 +77,28 @@ pnpm nx g @aws/nx-plugin:ts#infra
 
 ## Available Generators
 
-| Generator               | Description                                                                                                       |
-| ----------------------- | ----------------------------------------------------------------------------------------------------------------- |
-| `ts#project`            | TypeScript library                                                                                                |
-| `ts#api`                | TypeScript API (tRPC or Smithy) with API Gateway + Lambda + [Powertools](https://github.com/aws-powertools/powertools-lambda-typescript) |
-| `ts#rdb`                | Relational databases with Aurora RDS                                                                              |
-| `ts#website`            | React app (Vite)                                                                                                  |
-| `ts#website#auth`       | Add Cognito auth to a website                                                                                     |
-| `ts#infra`              | AWS CDK infrastructure project                                                                                    |
-| `ts#lambda-function`    | TypeScript Lambda with type-safe event sources                                                                    |
-| `ts#mcp-server`         | MCP server (TypeScript)                                                                                           |
-| `ts#agent`              | [Strands Agent](https://strandsagents.com/) (TypeScript)                                                          |
-| `ts#nx-generator`       | Nx generator scaffold                                                                                             |
-| `smithy#project`        | Smithy model project — a service model, or a shape library shared between Smithy projects                          |
-| `py#project`            | Python project (uv)                                                                                               |
-| `py#api`                | Python API (FastAPI) with API Gateway + Lambda + [Powertools](https://github.com/aws-powertools/powertools-lambda-python)      |
-| `py#lambda-function`    | Python Lambda with type-safe event sources                                                                        |
-| `py#mcp-server`         | MCP server (Python)                                                                                               |
-| `py#agent`              | [Strands Agent](https://strandsagents.com/) (Python)                                                              |
-| `agentcore-harness`     | [AgentCore Harness](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) agent loop (experimental) |
-| `connection`            | Connect projects together (e.g. frontend to API)                                                                  |
-| `terraform#project`     | Terraform project                                                                                                 |
-| `license`               | Manage LICENSE files and source headers                                                                           |
+| Generator            | Description                                                                                                                              |
+| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| `ts#project`         | TypeScript library                                                                                                                       |
+| `ts#api`             | TypeScript API (tRPC or Smithy) with API Gateway + Lambda + [Powertools](https://github.com/aws-powertools/powertools-lambda-typescript) |
+| `ts#rdb`             | Relational databases with Aurora RDS                                                                                                     |
+| `ts#website`         | React app (Vite)                                                                                                                         |
+| `ts#website#auth`    | Add Cognito auth to a website                                                                                                            |
+| `ts#infra`           | AWS CDK infrastructure project                                                                                                           |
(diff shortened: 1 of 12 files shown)


packages/nx-plugin/README.md
```diff
@@ -95,7 +95,7 @@ pnpm nx g @aws/nx-plugin:ts#infra
 | `py#lambda-function`    | Python Lambda with type-safe event sources                                                                        |
 | `py#mcp-server`         | MCP server (Python)                                                                                               |
 | `py#agent`              | [Strands Agent](https://strandsagents.com/) (Python)                                                              |
-| `agentcore-harness`     | [AgentCore Harness](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) agent loop        |
+| `agentcore-harness`     | [AgentCore Harness](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) agent loop (experimental) |
```
cogwirrel: Let's pad this better

nx-plugin-for-aws: Done in 2728241 — repadded the whole table rather than just this row. The right edge was already ragged before this change (rows ranged 143–166 chars because `ts#api`, `py#api` and `smithy#project` overflowed the header width), and adding `(experimental)` made this row a fourth ragged one. Both columns are now sized to their widest cell, so every row is a uniform width. Copied to the root `README.md` as usual.

cogwirrel (approved):

PR #1165, opened 2026-08-31
feat(ts#agent,py#agent,ts#mcp-server,py#mcp-server): deploy code to AgentCore Runtime by default

nx-plugin-for-aws:
### Reason for this change

Hosting an agent or MCP server on Bedrock AgentCore Runtime required building an arm64 container image, pushing it to ECR and deploying from there. That is a slow build and deploy cycle: a cold image build takes minutes, and every code change pays an image build plus an ECR push before it can be deployed.

AgentCore Runtime supports direct code deployment for both Node.js and Python, which packages built code as a zip and runs it on a managed language runtime — no Dockerfile, no ECR repository and no image push.

### Description of changes

`infra` gains an `agentcore-ecr` value and `agentcore` now means code packaging, so the faster path is the default:

- `agentcore` (default) assembles a code package and deploys it via `AgentRuntimeArtifact.fromCodeAsset` (CDK) or `code_configuration` (Terraform).
- `agentcore-ecr` builds and hosts a container image, unchanged from before, including its Trivy image scan.
- `none` is unchanged.

**Terraform modules are composed rather than duplicated.** `core/agent-core` holds the runtime, its role and its network configuration and knows nothing about packaging; `core/agent-core-code` and `core/agent-core-container` each own only their own artifact plumbing and delegate the runtime to it. The app module's packaging choice is the `source` path it points at. Packaging-specific IAM arrives through the existing `additional_iam_policy_statements` input, so the generic module needs no packaging-aware code. A workspace mixing both packagings vends one copy of the runtime.

Code packaging stages its zip in the existing shared `core/asset-bucket`, the same way Lambda deployment packages are staged, rather than provisioning a bucket of its own.

ADOT observability is preserved on both paths. AgentCore rejects a code package with no OpenTelemetry dependencies present, so the TypeScript package target vendors the ADOT distribution alongside the bundle and both languages use the `opentelemetry-instrument` entry point prefix.

Python agents and MCP servers get a root `main.py` entry point. AgentCore runs the entry point by file path, and Python gives a script run that way no parent package, so the module's own relative imports would raise `ImportError`; the root script imports the module absolutely instead.

AgentCore runtime versions are pinned in `AGENT_CORE_RUNTIME_VERSIONS` and resolved by the version update workflow from the `AgentCoreRuntime` members `aws-cdk-lib` publishes — the same mechanism as the Lambda runtimes, but a separate list since AgentCore offers fewer runtimes and a Lambda bump would otherwise select one it rejects.

A migration reshapes existing workspaces onto the composed modules, **keeping them on container deployment** — switching to code packaging stays a deliberate `--infra=agentcore` re-run. `moved` blocks carry state across, so the deployed runtime, its role and its generated name all survive; a module that has diverged from the generated shape is left untouched and reported through `nextSteps`.

Unit tests cover the new default, the package targets, the vended entry points, the composed module layout, the AgentCore runtime resolution and the migration (including state preservation and idempotency). `agentcore-ecr` and `agentcore` variants are added to both the internal test matrix and the e2e generator matrix, and the CDK and Terraform deploy tests each invoke a container-packaged agent for both languages.

Validated end to end against AWS in us-east-1 with locally compiled packages, deploying and invoking every permutation of generator, language, protocol and IaC provider with SigV4-authenticated requests, then tearing everything down:

- **CDK — 10 runtimes, all HTTP 200 with real Bedrock responses:** ts/py agents over HTTP, A2A and AG-UI (Strands and LangChain), ts/py MCP servers, plus container-packaged ts and py agents. Confirmed from the deployed runtime configuration that 8 are `codeConfiguration` (`NODE_22` / `PYTHON_3_14`, `opentelemetry-instrument` entry point) and 2 are `containerConfiguration`.
- **Terraform — 6 runtimes, all HTTP 200:** ts/py agents over HTTP and A2A, ts MCP server, plus container-packaged ts and py agents. Both packagings coexisted in one workspace over the shared `core/agent-core`.
- Code and container packaging returned identical agent replies, and `checkov` passed with 0 failures on both providers.
- Confirmed ADOT is active on the code runtimes from the runtime logs, on both providers and both languages.
- Ran the migration against a real pre-refactor workspace: it rewrote exactly 3 files, the result validated, and a state-level test confirmed the deployed runtime keeps its identity (same resource id, same generated name) rather than being replaced.

Measured the cycle time this targets: a cold container build took 2m32s against 13s to a deployable code package, and an incremental code change 21s against 10s.

### Checklist

----


nx-plugin-for-aws:
The `Translate Documentation` check is failing, but the cause is pre-existing and unrelated to this PR — flagging it here rather than working around it.

All 18 failures are the translator's own validator rejecting its output for two files with:

```
guides/py-agent.mdx → <lang>: duplicate import of "os"
get_started/tutorials/dungeon-game/1.mdx → <lang>: duplicate import of "uuid"
```

Those duplicate lines are `import os` / `import uuid` **inside fenced Python code blocks** (two framework variants of the same snippet, under `<Tabs>`). They are on `main` unchanged, and the already-committed `jp` translation of `py-agent.mdx` contains the same pair at lines 656/675 — so the validator's "acorn rejects a redeclared identifier" check is misreading fenced Python as MDX ESM imports.

This PR does not touch those code blocks; its changes to both files are prose and JSX only. `nx run docs:build` passes locally on the English sources, and every required check on this PR is green (including `Smoke Tests - dungeon-adventure`). I re-ran the workflow once in case the earlier `security token ... expired` errors were transient; the credential errors cleared and only the validator failures remained.

Happy to split out a separate fix that renames one variable per tab pair (or teaches the validator to skip fenced blocks) if you'd like — it seemed out of scope to bundle here.

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -485,15 +487,23 @@ In order to build your Agent for Bedrock AgentCore Runtime, a `bundle` target is
 - Exports your Python dependencies to a `requirements.txt` file using `uv export`
 - Installs dependencies for the target platform (`aarch64-manylinux_2_28`) using `uv pip install`
 
+<OptionFilter when={{ infra: 'agentcore' }} description="Code package target">
+A `<your-agent-name>-package` target is also added, which assembles the deployable code package: the `aarch64` dependency bundle, your Python module tree, and a root `main.py` entry point. The infrastructure uploads this directory as a `.zip` using `AgentRuntimeArtifact.fromCodeAsset`.
+
+The entry point sits at the package root rather than inside your module, because AgentCore runs it by file path and Python gives a script run that way no parent package — so it imports your agent absolutely, which resolves your module's own relative imports normally.
```
cogwirrel: remove this line

docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -485,15 +487,23 @@ In order to build your Agent for Bedrock AgentCore Runtime, a `bundle` target is
...
 
 ### Image Scanning
 
 <Snippet name="trivy-image-scan" parentHeading="Image Scanning" />
+</OptionFilter>
 
 ### Observability
 
-Your agent is automatically configured with observability using the [AWS Distro for Open Telemetry](https://aws.amazon.com/otel/) (ADOT), by configuring auto-instrumentation in your `Dockerfile`.
+Your agent is automatically configured with observability using the [AWS Distro for Open Telemetry](https://aws.amazon.com/otel/) (ADOT). Under `agentcore` the ADOT distribution is packaged alongside your code and loaded via the `opentelemetry-instrument` entry point; under `agentcore-ecr` it is installed in the image and loaded by the `Dockerfile`'s command.
```
cogwirrel: Just keep this only as ``Your agent is automatically configured with observability using the [AWS Distro for Open Telemetry](https://aws.amazon.com/otel/) (ADOT).``

docs/src/content/docs/en/guides/py-mcp-server.mdx
```diff
@@ -169,11 +171,19 @@ In order to build your MCP server for Bedrock AgentCore Runtime, a `bundle` targ
 - Exports your Python dependencies to a `requirements.txt` file using `uv export`
 - Installs dependencies for the target platform (`aarch64-manylinux_2_28`) using `uv pip install`
 
+<OptionFilter when={{ infra: 'agentcore' }} description="Code package target">
+A `<your-server-name>-package` target is also added, which assembles the deployable code package: the `aarch64` dependency bundle, your Python module tree, and a root `main.py` entry point. The infrastructure uploads this directory as a `.zip` using `AgentRuntimeArtifact.fromCodeAsset`.
+
+The entry point sits at the package root rather than inside your module, because AgentCore runs it by file path and Python gives a script run that way no parent package — so it imports your server absolutely, which resolves your module's own relative imports normally.
```
cogwirrel: Same comments on py-agent docs page apply to py-mcp-server, ts-mcp-server, ts-agent.

docs/src/content/docs/en/snippets/agentcore-hosting.mdx
```diff
@@ -0,0 +1,25 @@
...
+```
+ERROR: Cannot connect to the Docker daemon at unix://path/to/docker.sock.
+```
+:::
+
+:::caution[Switching between the two]
+The `infra` option you pass is always applied, so re-running the generator with `agentcore` on a project previously generated with `agentcore-ecr` switches it to code packaging, deleting the `Dockerfile` and its image build and scan targets.
+
+The generated infrastructure is yours to edit and so is never overwritten. When switching, delete the generated construct or module for the component and re-run the generator so it is rewritten for the new packaging.
+:::
```
cogwirrel: remove this part of the hosting guide.

packages/nx-plugin/src/py/mcp-server/generator.ts
```diff
@@ -162,62 +173,126 @@ export const pyMcpServerGenerator = async (
...
+      };
+    } else {
+      // The managed runtime loads the code from a zip, so no Dockerfile or
+      // image build is involved. Remove one left by a previous container run.
+      removeContainerArtifacts(tree, {
+        project,
+        sourceDir: targetSourceDir,
+        targetPrefix: mcpTargetPrefix,
+      });
+
```
cogwirrel: Ok to leave a dockerfile on a rerun

packages/nx-plugin/src/ts/agent/generator.ts
```diff
@@ -262,65 +273,112 @@ export const tsAgentGenerator = async (
...
+
+      artifact = {
+        type: 'container',
+        dockerImageTag,
+        outputDir: bundleOutputDir,
+      };
+    } else {
+      // The managed runtime loads the code from a zip, so no Dockerfile or
+      // image build is involved. Remove one left by a previous container run.
+      removeContainerArtifacts(tree, {
```
cogwirrel: same

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/core/agent-core/runtime.tf.template
```diff
@@ -159,6 +184,91 @@ resource "aws_ecr_repository_policy" "agent_core_ecr_policy" {
...
+  #checkov:skip=CKV2_AWS_61:Lifecycle configuration not required for code packages
+  #checkov:skip=CKV2_AWS_62:Event notifications not required for code packages
+  #checkov:skip=CKV_AWS_144:Cross-region replication not required for code packages
+  #checkov:skip=CKV_AWS_18:Access logging not required for code packages
+  #checkov:skip=CKV_AWS_145:AES256 (S3-managed) encryption is sufficient for build artefacts
+  bucket        = "${lower(replace(var.agent_runtime_name, "_", "-"))}-code-${random_id.unique_suffix.hex}"
+  force_destroy = true
+
+  tags = var.tags
+}
```
cogwirrel: We already have an artifact bucket - see how lambdas for apis etc are uploaded.

packages/nx-plugin/src/utils/agent-core-packaging.ts
```diff
@@ -0,0 +1,304 @@
...
+
+/**
+ * The AgentCore managed runtime for the Node version generated projects target,
+ * e.g. `NODE_22`.
+ *
+ * AgentCore only publishes a Node 22 runtime, so the major is pinned here
+ * rather than derived from {@link LAMBDA_RUNTIME_VERSIONS}: those track Lambda,
+ * which offers newer majors AgentCore would reject.
+ */
+export const agentCoreNodeRuntime = (): string => 'NODE_22';
```
cogwirrel: We should follow the same pattern as lambda for agentcore runtime versions, ensuring they're updated in the sync versions migration, and kept up to date in the version update workflow

packages/nx-plugin/src/utils/agent-core-packaging.ts
```diff
@@ -0,0 +1,304 @@
...
+
+/**
+ * The AgentCore managed runtime for the Python version generated projects
+ * target, e.g. `PYTHON_3_14`, derived from the same pin the bundle resolves
+ * wheels against so the two cannot drift.
+ */
+export const agentCorePythonRuntime = (): string =>
+  `PYTHON_${LAMBDA_RUNTIME_VERSIONS.python.replace('.', '_')}`;
```
cogwirrel: Not necessarily the same as lambda, same as above, manage via update version workflow which uses cdk.

packages/nx-plugin/src/utils/agent-core-packaging.ts
```diff
@@ -0,0 +1,304 @@
...
+/**
+ * Remove the container build artifacts a previous `agentcore-ecr` run left
+ * behind, so re-running with `agentcore` fully switches the component to code
+ * packaging rather than leaving an orphaned `Dockerfile` and image build.
+ *
+ * The aggregate `docker` and `trivy` targets are dropped only once no component
+ * target remains under them, since a project can host several components and
+ * only some may use containers.
+ */
+export const removeContainerArtifacts = (
```
cogwirrel: per other comment, let's remove this - no need to "undo" changes if regenerating with a different infra value

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/core/agent-core/runtime.tf.template
```diff
@@ -6,10 +6,17 @@ terraform {
       source  = "hashicorp/aws"
       version = "<%- awsProviderVersion %>"
     }
+<%_ if (container) { _%>
```
cogwirrel: Let's figure out how we can share tf code between container/code better so that there's not too much duplication when users use a mix of containers/code. ie:

- core/agent-core/runtime.tf
- core/agent-core/runtime-container.tf
- core/agent-core/runtime-code.tf

cogwirrel (commented):
We need variants with agentcore-ecr and agentcore included in the generator matrix and the internal test matrix generator for e2e test coverage.

cdk/tf deploy tests should include at least 1 ecr variant for both ts and py to ensure this works at runtime (default will switch so new path will be covered).

nx-plugin-for-aws:
Thanks — all comments addressed, and the Terraform duplication is now solved by composition rather than a second copy of the module.

**Terraform structure** (your `runtime.tf` / `runtime-container.tf` / `runtime-code.tf` suggestion, as modules so each vends independently):

```
core/agent-core/            # runtime + role + network. No packaging awareness at all.
core/agent-core-code/       # archive + S3 object  → module "runtime" { source = "../agent-core" }
core/agent-core-container/  # ECR + digest + push  → module "runtime" { source = "../agent-core" }
```

The app module's packaging choice is just the `source` path. Two things made this cleaner than templating both into one file:

- **`additional_iam_policy_statements` already existed**, so the ECR and S3 grants flow through it via `concat` and the generic module needs no packaging-aware IAM.
- **Only the packaging in use is vended**, so a code-only workspace carries no ECR resources and no `external` provider — and a mixed workspace vends one runtime. I verified a mixed workspace plans correctly (15 resources, each wrapper owning only its own plumbing).

**Other comments:**
- **Artifact bucket** — now stages the zip in the existing `core/asset-bucket` under `agent-core/<name>/`, matching how Lambda packages are staged. Its own bucket is gone.
- **Runtime versions** — new `AGENT_CORE_RUNTIME_VERSIONS`, resolved by the version update workflow from the `AgentCoreRuntime` members `aws-cdk-lib` publishes, mirroring `resolveLambdaRuntimes`. Kept separate from the Lambda pins deliberately: AgentCore offers fewer runtimes, so a Lambda bump would otherwise select one it rejects at create time.
- **Dockerfile on rerun / `removeContainerArtifacts`** — both removed; a rerun no longer undoes the previous `infra` value.
- **Docs** — the flagged line and the switching section are gone, and the observability line is now just the ADOT sentence, across all four guides.
- **Test matrices** — `agentcore-ecr` variants added to the internal test matrix and the e2e generator matrix for all four generators, and both the CDK and Terraform deploy tests now invoke a container-packaged agent for **both** ts and py.

**Migration** (`v1.0.0-rc.89/0002`): reshapes existing workspaces onto the composed modules and **keeps them on container** — switching to code stays a deliberate `--infra=agentcore` re-run. `moved` blocks carry state across, which I verified at the state level: the runtime keeps the same resource id and the same generated name rather than being replaced. A diverged module is left untouched and reported via `nextSteps`. One honest caveat in the migration's own comment: the ECR repository's name embeds the suffix whose state moves into the shared module, so it is replaced and the image re-pushed on the next apply — reusing the moved suffix instead would be a dependency cycle (the shared module consumes the repository URL), which I confirmed with `terraform validate`.

**Retested end to end on AWS**, deploying and invoking every permutation with SigV4 and observing real responses, then tearing it all down:

| | runtimes | result |
|---|---|---|
| CDK | 10 (8 code / 2 container) | all HTTP 200, real Bedrock output |
| Terraform | 6 (4 code / 2 container) | all HTTP 200 |

Confirmed from the deployed runtime config that the code ones carry `NODE_22`/`PYTHON_3_14` with the `opentelemetry-instrument` entry point, ADOT is live on both providers and both languages, code and container return identical replies, and `checkov` passes with 0 failures.

CI: all 8 required checks green, all 10 unit shards green. `Smoke Tests - local-dev` failed once and **passed on re-run** (unrelated chat-stream timeout; my last commit was snapshot-only). `Translate Documentation` still fails for the pre-existing reason in my earlier comment — the translator's validator misreads fenced Python `import os`/`import uuid` as MDX imports; those blocks are unchanged on main and the committed `jp` translation has the same pair. Happy to fix that separately if you'd like.

docs/src/content/docs/en/snippets/agent/bedrock-deployment.mdx
```diff
@@ -31,14 +31,18 @@ This construct uses the [`aws-cdk-lib/aws-bedrockagentcore` module](https://docs
 <Fragment slot="terraform">
 A Terraform module is generated for you, named based on the `name` you chose when running the generator, or `<ProjectName>-agent` by default.
 
-Pass the shared <Link path="/guides/runtime-config#writing-configuration">`runtime_config_appconfig`</Link> module's outputs into the agent module:
+Pass the shared <Link path="/guides/runtime-config#writing-configuration">`runtime_config_appconfig`</Link> module's outputs into the agent module. Under the default `agentcore` packaging the agent's code is staged in the shared asset bucket, so pass that too — instantiate the `core/asset-bucket` module once per deployment, as the Lambda and API modules already do:
 
 ```terraform
 module "my_project_agent" {
```
cogwirrel, line 41: Include declaring the asset bucket module above. same for mcp docs

docs/src/content/docs/en/snippets/agentcore-hosting.mdx
```diff
@@ -0,0 +1,19 @@
...
+- **`none`** generates no infrastructure at all, so the project can only be run locally.
+
+Both hosted options are configured for observability with the [AWS Distro for OpenTelemetry](https://aws.amazon.com/otel/) (ADOT).
+
+:::note[Docker is only required for agentcore-ecr]
+`agentcore` needs no container engine, so you do not need [Docker](https://www.docker.com/) installed to build or deploy. Select `agentcore-ecr` if you would rather build a container image, and make sure Docker is installed and running to avoid errors such as:
+
+```
+ERROR: Cannot connect to the Docker daemon at unix://path/to/docker.sock.
+```
```
cogwirrel: remove these parts, just keep the first paragraph and the 3 bullet points above.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/core/agent-core-container/runtime.tf.template
```diff
@@ -0,0 +1,331 @@
...
+
+moved {
+  from = aws_vpc_security_group_egress_rule.agent_core_runtime_https
+  to   = module.runtime.aws_vpc_security_group_egress_rule.agent_core_runtime_https
+}
+
+moved {
+  from = aws_bedrockagentcore_agent_runtime.agent_runtime
+  to   = module.runtime.aws_bedrockagentcore_agent_runtime.agent_runtime
+}
```
cogwirrel: This doesn't make sense for newly generated workspaces. Let's drop the migration altogether.

nx-plugin-for-aws:
All four addressed:

- **`moved` blocks / migration** — you're right, both are wrong here. The `moved` blocks are meaningless in a freshly generated workspace, and since `infra` was never recorded in metadata before, there is nothing for a migration to read or correct. Both are gone; no migration ships.
- **`infra` in metadata** — now recorded by all four generators (`agentcore` / `agentcore-ecr` / `none`), matching what `trpc#backend` and friends already do, so a consumer can tell a code-packaged runtime from a container one without re-deriving it. Verified per-component in a real workspace:

  ```
  agent          infra='agentcore'      iac='terraform'
  ecr-agent      infra='agentcore-ecr'  iac='terraform'
  local-agent    infra='none'           iac=None
  ```

- **Asset bucket declaration** — the `core/asset-bucket` module is now declared in the example, for both the agent and MCP deployment snippets, matching how the API and Lambda guides show it.
- **Hosting snippet** — trimmed to the first paragraph and the three bullets.

Also rebased on main, which picked up your translate fix (#1166) — translations now generate cleanly, including the new hosting snippet.

One consequence worth flagging: dropping the `moved` blocks means an existing Terraform workspace that upgrades and re-runs a generator will see the runtime move into the nested module as a **replacement** rather than a move. That is only reachable by re-running a generator (the vended modules are `KeepExisting`, so an upgrade alone changes nothing), so I have not tried to paper over it — but say the word if you would rather I keep the blocks purely for that path.

Docs build passes, plugin compiles, and the only local test failures are the pre-existing `vite`-in-virtual-tree ones that pass in CI.

cogwirrel (dismissed):

nx-plugin-for-aws:
Good catch, and my earlier justification for it was wrong. I had read your "ok to leave a dockerfile on a rerun" as covering this, but that was about not *undoing* artifacts when regenerating — not about vending an unused one on a fresh code-packaged generation.

The cause: `ts#mcp-server` was the only one of the four with `Dockerfile.template` in its always-generated `files/` root, then deleting it again for `infra: none`. The other three keep it under `files/deploy/` and vend it only on the container path.

Fixed by matching them — `files/app/` for the server sources, `files/deploy/` for the Dockerfile, vended inside the `container` branch. The `infra: none` delete is gone too, since there is nothing to delete now. Verified in a real workspace:

```
code-mcp: no Dockerfile      http.ts index.ts resources server.ts stdio.ts tools
ecr-mcp:  Dockerfile PRESENT http.ts index.ts resources server.ts stdio.ts tools
none-mcp: no Dockerfile      http.ts index.ts resources server.ts stdio.ts tools
```

Also restored the unit assertion I had weakened on that wrong premise (it now asserts the Dockerfile is absent again), fixed the docs line to match the other three guides, and updated the template-fidelity test that pins the Dockerfile path so the version workflow keeps upgrading it.

Full `src/ts` + `src/py` + `src/utils` + `src/migrations` run: 162 failures, all of them the one pre-existing `vite`-in-virtual-tree error, zero genuine. Workspace builds clean and docs build passes.

cogwirrel (approved):

PR #1167, opened 2026-09-01
feat(terraform#project): share one asset ECR registry across container infra

nx-plugin-for-aws:
### Reason for this change

Container-packaged Terraform infrastructure provisioned an ECR repository per container. `agent-core-container` created one per AgentCore Runtime, and each relational database module created one for its Lambda handler images. A workspace with several containers accumulated several repositories, each holding a single image, and every new container added another.

The Lambda deployment packages already had the answer to this: one shared `core/asset-bucket` stages every zip in the workspace. Container images had no equivalent.

### Description of changes

Vend a shared `core/asset-ecr` registry — the container analogue of `core/asset-bucket`, alongside it in the same shared Terraform project, so a workspace gets both without a generator opting in.

- **`core/asset-ecr`** owns the one repository, its policy, its KMS key and its outputs (`repository_url`, `repository_arn`). Images are encrypted with a customer managed key (rotation on by default, via `enable_key_rotation`) rather than ECR-managed AES256. `repository_url` is routed through the repository policy, so a consumer that pushes to it waits until the policy admitting that push exists — which is what the per-module `depends_on` used to do locally.
- **`core/agent-core-container`** takes `asset_ecr_repository_url` / `_arn` as inputs, publishes the image there and scopes the runtime's pull grant to it. The app-level agent module threads the pair through, gated on container packaging, the same way it gates the asset bucket pair on code packaging. The dropped `ecr_repository_url` output is replaced by `container_uri`, which names the published image rather than a repository the module no longer owns.
- **The database modules** publish the migration image — and, for SQLModel, the create-db-user image — to the shared registry via one new `asset_ecr_repository_url` input.

**Sharing one repository means image tags must not collide**, so each tag is namespaced by its owner alongside the content digest it already carried (`agent-core-<runtime>-<digest>`, `<db>-migration-<digest>`, `<db>-create-db-user-<digest>`). Tags stay content-addressed, so the repository keeps `IMMUTABLE` tag mutability and an existing tag always already names the image it would be overwritten with. Database names are truncated so the tag stays inside Docker's 128 character limit; `agent_runtime_name` is already validated to 43 characters, so the agent path is within it by construction.

The registry policy admits the Lambda service principal, which image-packaged database handlers pull through, carried over from the repository the database modules used to own. Consumers need no KMS permissions of their own: ECR creates its own grants on the key when the repository is created and image pulls decrypt through those, so the runtime and Lambda roles are unchanged.

Per the request, **no migration is included** — this follows the same stance as the parent PR, which dropped its own migration.

Unit tests cover the shared registry being vended once (including its KMS encryption), both container agents and both database flavours threading it through rather than declaring a repository, and the per-owner tag namespacing that keeps them from colliding. The existing test asserting the code and container modules stay separate is retargeted at what now distinguishes them (`archive_file` + asset bucket vs `docker_publish` + asset registry), since neither provisions the shared resource it targets any more. Full suite: **4164 passed**, and `nx run-many --target build --all` passes including the docs site.

Validated end to end in a generated workspace with locally compiled packages, exercising a workspace that mixes both packagings — three container agents (ts x2, py), a code-packaged agent, and both a Prisma and a SQLModel database:

- `terraform fmt -check` clean; `terraform validate` clean; the credential-free plan-time `terraform test` (full module graph, mocked providers) **passes**.
- `checkov`: **0 failures**, with `CKV_AWS_136` now passing on merit rather than via a skip.
- Confirmed exactly one module in the whole generated tree declares an `aws_ecr_repository`, and all five image URIs resolve through the shared registry.

Deployed to AWS in us-east-1 — the shared registry plus two container-packaged agents — then tore everything down. Ran twice: once before the KMS change, and again after it to confirm KMS-encrypted push and pull both work.

- One repository (`IMMUTABLE`, scan-on-push) held **both** agents' images under distinct namespaced tags, confirmed from `describe-images`; `describe-repositories` confirmed `encryptionType: KMS` against the generated customer managed key.
- Both runtimes reached `READY` with `containerUri` pointing at that one repository, proving the scoped pull grant works and that no extra KMS grant is needed on the runtime roles.
- Both agents returned real Bedrock responses via the vended tRPC client (`3 * 5 / 4 = 3.75`), and the runtime logs confirmed ADOT started and the server was serving.
- A re-plan showed no churn in the registry, the image pushes or the runtimes — no re-push and no runtime replacement.
- Verified afterwards that no repositories, runtimes or session buckets were left behind.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -379,10 +379,12 @@ Use the `DockerImageAsset` with any AWS construct that accepts a container image
 Terraform's AWS provider does not have a first-class "build and push a Docker image" resource. The pattern used by the generators is:
 
 1. The project's `build` target runs `docker build`, producing a local image tagged `my-scope-my-project:latest`.
-2. An `aws_ecr_repository` to hold the image.
+2. The shared `core/asset-ecr` repository holds the image. Every container in the workspace is published here, so nothing provisions a repository of its own — instantiate the module once per deployment and thread its `repository_url` into each module that publishes an image.
```
cogwirrel: Change to ``2. A shared ECR repository using the `core/asset-ecr` module``

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/core/agent-core-container/runtime.tf.template
```diff
@@ -158,15 +112,18 @@ data "external" "docker_digest" {
 }
 
 locals {
-  # Content-based, immutable image tag derived from the local image digest
-  image_tag = replace(data.external.docker_digest.result.digest, "sha256:", "")
+  # Content-based, immutable image tag derived from the local image digest, and
+  # namespaced by runtime so runtimes sharing the registry never collide. The
+  # runtime name is capped at 43 characters, keeping this within Docker's 128
+  # character tag limit.
+  image_tag = "agent-core-${lower(var.agent_runtime_name)}-${replace(data.external.docker_digest.result.digest, "sha256:", "")}"
```
cogwirrel: Undo comment change

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
```diff
@@ -0,0 +1,110 @@
...
+  length  = 8
+  special = false
+  upper   = false
+}
+
+# Shared asset container registry. Every container image the workspace deploys is
+# pushed here under a tag namespaced by the image's owner, so a workspace needs
+# one registry rather than one per container.
+resource "aws_ecr_repository" "assets" {
+  #checkov:skip=CKV_AWS_136:AES256 (ECR-managed) encryption is sufficient for build artefacts
```
cogwirrel: Given we're centralising this let's use a kms key.

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
```diff
@@ -0,0 +1,110 @@
...
+
+resource "random_string" "suffix" {
+  length  = 8
+  special = false
+  upper   = false
+}
+
+# Shared asset container registry. Every container image the workspace deploys is
+# pushed here under a tag namespaced by the image's owner, so a workspace needs
+# one registry rather than one per container.
```
cogwirrel: remove `, so a workspace needs one registry rather than one per container.`

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
```diff
@@ -0,0 +1,110 @@
...
+resource "aws_ecr_repository" "assets" {
+  #checkov:skip=CKV_AWS_136:AES256 (ECR-managed) encryption is sufficient for build artefacts
+  name = "${var.repository_name_prefix}-${random_string.suffix.result}"
+
+  # Consumers tag by image content digest, so an existing tag already names the
+  # image it would be overwritten with.
```
cogwirrel: remove this comment

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
```diff
@@ -0,0 +1,110 @@
...
+  })
+}
+
+output "repository_name" {
+  description = "Name of the shared asset repository"
+  value       = aws_ecr_repository.assets.name
+}
+
+# Routed through the repository policy so a consumer that pushes to this URL
+# waits until the policy allowing that push exists.
```
cogwirrel: remove this comment

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -240,78 +245,31 @@ module "add_rdb_to_runtime_config" {
...
   program = ["sh", "-c", "echo '{\"digest\":\"'$(<%= containerEngine %> inspect ${var.migration_docker_image_tag} --format '{{.Id}}')'\"}' "]
 }
 
 locals {
-  # Content-based, immutable image tag derived from the local image digest
-  image_tag = replace(data.external.docker_digest.result.digest, "sha256:", "")
+  # Content-based, immutable image tag derived from the local image digest, and
+  # namespaced by database and handler so images sharing the registry never
+  # collide. The name is truncated so the tag stays within Docker's 128
+  # character limit; the digest is what makes it unique.
```
cogwirrel: same, undo comment change

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/core/agent-core-container/runtime.tf.template
nx-plugin-for-aws: Done — restored the original one-line comment.

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
nx-plugin-for-aws: Done. Added a customer managed key with rotation on (`enable_key_rotation`, defaulting true) and `encryption_configuration` on the repository, which also drops the `CKV_AWS_136` skip — checkov now passes it on merit.

Worth noting consumers need no extra IAM: ECR creates its own KMS grants when the repository is created and pulls decrypt through those, so the runtime and Lambda roles are unchanged. Verified by redeploying both container agents against the KMS-encrypted registry — both reached READY and returned real Bedrock responses.

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
nx-plugin-for-aws: Done.

packages/nx-plugin/src/utils/files/terraform/src/core/asset-ecr/asset-ecr.tf.template
nx-plugin-for-aws: Done — removed the comment. Kept the `depends_on` itself, since it's what makes a consumer wait for the push policy to exist (it replaces the per-module `depends_on` the packaging modules used to have on their own repository policy).

packages/nx-plugin/src/utils/rdb-constructs/files/terraform/app/dbs/__nameKebabCase__/__nameKebabCase__.tf.template
nx-plugin-for-aws: Done — restored the original one-line comment.

cogwirrel (approved):

PR #1171, opened 2026-09-01
chore(scripts): declare version pins that must move in lockstep

nx-plugin-for-aws:
### Reason for this change

Some version pins are only correct at *matching* versions, but their bumps arrive
independently — so a weekly run takes one, leaves the others behind, and the
mismatch fails the build. There were three such couplings, handled inconsistently:

- **`@ag-ui/*`** — the packages ship from one release train and are only mutually
  resolvable at the same version. They also have to stay on the release
  `@copilotkit/react-core` depends on, since CopilotKit pins them exactly and a
  second copy of `@ag-ui/client` makes the two `AbstractAgent` declarations
  structurally incompatible (private `_debug`):

  ```
  error TS2322: Type 'Record<string, .../@ag-ui+client@0.0.59/...AbstractAgent>' is not
  assignable to type 'Record<string, .../@ag-ui+client@0.0.57/...AbstractAgent>'.
  ```

  This had **no** protection and took out ~20 jobs per run on two separate
  `version-update` runs.

- **`@biomejs/wasm-nodejs` / `@biomejs/biome`** — the wasm bindings format
  generated TypeScript/JSON in-process, while the CLI is what a generated
  project's `format`/`lint` targets run *and* the version its vended `biome.json`
  `$schema` points at. Also had **no** protection. Worth noting the CLI has stable
  releases the bindings never got (`2.2.3` exists for `@biomejs/biome` but not for
  `@biomejs/wasm-nodejs`), so the CLI genuinely has to wait for the bindings.

- **`ruff` / `@astral-sh/ruff-wasm-nodejs`** — same shape, but the bindings are on
  npm and `ruff` is on PyPI. This had a bespoke `holdRuffToWasmBuild`.

### Description of changes

- New `packages/nx-plugin/src/utils/version-lockstep/lockstep.ts` —
  `holdGroupsInLockstep(groups, ...tables)`. A group is just a list of dependency
  names, held at the **lowest version proposed across it**, so it only moves once
  every member can. Applied to the *proposed* versions before anything is written,
  so a held pin is never recorded, and every hold is reported in the update PR
  body. Comparison and range coercion use `semver` (already a dependency, already
  used in `dependencies.ts` / `migration-versions.ts`), so `==0.16.5` and `0.16.5`
  compare equal and each pin keeps the range syntax it was written with.

  ```js
  const LOCKSTEP_GROUPS = [
    ['@ag-ui/client', '@ag-ui/core', '@ag-ui/encoder'],
    ['@biomejs/wasm-nodejs', '@biomejs/biome'],
  ];
  ```

- The wasm bindings are dependencies of this repo rather than vended pins, so
  their proposed versions are resolved alongside `TS_VERSIONS`, and a held binding
  is written back to the manifests that declare it (`npm-check-updates` bumps
  those before this script runs, so without the writeback the repo and the vended
  pin would diverge the other way).

- `ruff` keeps an explicit hold, because its two sides are resolved by different
  tools against different registries with different cooldowns — not comparable as
  a group.

- Adds `BIOME_WASM_VERSION` plus a test asserting it matches
  `TS_VERSIONS['@biomejs/biome']`, mirroring the existing ruff assertion.

The module lives under `packages/nx-plugin/src/utils/` rather than `scripts/`
because the root project has no `test` target, so a spec under `scripts/` would
never run in CI.

**Unit tests** — 8 cases covering holding to the lowest version, no-op when a
group already agrees, spanning the TypeScript and Python tables (with `==`
preserved), numeric-not-lexical ordering (`0.0.9` < `0.0.10`), non-comparable pins
(`workspace:*`) leaving the group alone, absent members, and independent groups.
Run with the existing formatter assertions:

```
✓ src/utils/version-lockstep/lockstep.spec.ts (8 tests)
✓ src/utils/ruff.spec.ts (23 tests)
✓ src/utils/format.spec.ts (26 tests)
Tests  57 passed (57)
```

**Biome group against the real release gap** — with the bindings at `2.5.11` and
the CLI proposing `2.6.0`, the CLI is held to the bindings' release:

```
@biomejs/biome held at 2.5.11 (2.6.0 available): must move in step with @biomejs/wasm-nodejs
```

**End-to-end against the live registry** — ran the real
`scripts/update-versions.ts`. `@ag-ui/{client,core,encoder}` moved together to
`0.0.59` (correct: `@copilotkit/react-core` 1.70.0 now depends on `0.0.59`),
biome and ruff were already in step so their paths were no-ops, and no manifest
was touched. The `versions.ts` mutation from that verification run was reverted
and is not part of this PR.

Typecheck and lint pass.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
+++ b/packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
@@ -6,15 +6,17 @@
 /**
  * Groups of pins that must move together.
  *
- * Some packages are only correct at matching versions: a Python formatter and the
- * wasm build of it that actually runs, or an npm library and a second library
- * that depends on an exact release of it. The bumps behind those pins arrive
- * independently, so left alone a run takes one and leaves the other behind and
- * the mismatch fails the build.
+ * Packages published as a family are often only mutually resolvable at the same
+ * version, but their bumps arrive independently — so left alone a run takes one
+ * and leaves the rest behind, and the mismatch fails the build.
  *
- * A group nominates one member as the leader. Every other member is rewritten to
- * whatever the leader resolves to, so a group only ever moves as a unit — one
- * place to declare the constraint instead of a bespoke hold per pair.
+ * A group is held at the lowest version proposed across its members, so it only
+ * moves once every member can. Whichever member is furthest behind is the
+ * constraint; the rest are taken on a later run, once it catches up.
+ *
+ * This assumes the members share a version line, which is what makes "lowest"
+ * meaningful. A package coupled to something on an unrelated line (a dependant
+ * that pins it exactly, say) needs its own hold instead.
  */
 
 /** A note explaining a pin the run deliberately did not take. */
@@ -23,85 +25,49 @@ export interface LockstepNote {
 }
 
 /**
- * How a member's version is written, for groups that span registries or
- * manifests. `plain` is a bare version (npm); `pep440` is the `==x.y.z` form
- * `PY_VERSIONS` uses.
+ * The pins in one group, named as they appear in the versions tables. Members may
+ * live in different tables (e.g. the TypeScript and Python pins).
  */
-export type VersionFormat = 'plain' | 'pep440';
+export type LockstepGroup = readonly string[];
 
-export interface LockstepMember {
-  /** Key in the versions table this member lives in. */
-  readonly name: string;
-  /** Defaults to `plain`. */
-  readonly format?: VersionFormat;
-  /**
-   * Resolve the version to hold against, for a constraint that lives somewhere
-   * other than one of our own pins — e.g. a dependant's dependency range, read
-   * from the registry. Receives the merged proposed versions. Returning
-   * `undefined` leaves the group untouched and reports why.
-   */
-  readonly resolve?: (versions: Record<string, string>) => string | undefined;
-}
+/** Strip a range prefix (`==`, `^`, `~`, …) to leave a bare version. */
+const bare = (version: string): string =>
+  version.replace(/^[=~^><!\s]+/, '').trim();
 
-export interface LockstepGroup {
-  /** Shown in the report to explain why the group is coupled. */
-  readonly reason: string;
-  /**
-   * The member the rest follow. Chosen as whichever is the harder constraint to
-   * move — typically the one already installed and running, or the one whose
-   * dependant pins it exactly.
-   */
-  readonly leader: LockstepMember;
-  readonly followers: readonly LockstepMember[];
-}
-
-const stripFormat = (version: string): string =>
-  version.replace(/^[=~^><!]+/, '').trim();
-
-const applyFormat = (
-  version: string,
-  format: VersionFormat = 'plain',
-): string =>
-  format === 'pep440' ? `==${stripFormat(version)}` : stripFormat(version);
+/** The range prefix a pin was written with, so it can be preserved. */
+const prefixOf = (version: string): string =>
+  version.slice(0, version.indexOf(bare(version)));
 
 /**
- * Read a member's version from whichever of the supplied tables declares it, so
- * a group can span the TypeScript and Python pins.
+ * Compare bare `major.minor.patch[...]` versions segment by segment. Numeric
+ * segments compare numerically; anything else (prereleases) compares as a
+ * string, which is enough to order the pins these tables hold.
  */
-const readMember = (
-  member: LockstepMember,
-  tables: readonly Record<string, string>[],
-): string | undefined => {
-  for (const table of tables) {
-    const version = table[member.name];
-    if (version !== undefined) return version;
+export const compareVersions = (a: string, b: string): number => {
+  const as = bare(a).split(/[.\-+]/);
+  const bs = bare(b).split(/[.\-+]/);
+  for (let i = 0; i < Math.max(as.length, bs.length); i++) {
+    const x = as[i] ?? '';
+    const y = bs[i] ?? '';
+    if (x === y) continue;
+    const nx = Number(x);
+    const ny = Number(y);
+    if (Number.isNaN(nx) || Number.isNaN(ny)) return x < y ? -1 : 1;
+    return nx - ny;
   }
-  return undefined;
-};
-
-const writeMember = (
-  member: LockstepMember,
-  version: string,
-  tables: readonly Record<string, string>[],
-): boolean => {
-  for (const table of tables) {
-    if (member.name in table) {
-      table[member.name] = version;
-      return true;
-    }
-  }
-  return false;
+  return 0;
 };
 
 /**
- * Hold every follower in each group at the version its leader resolves to.
+ * Hold every member of each group at the lowest version proposed across it.
  *
  * Applied to the *proposed* versions before they are written, so a held pin is
- * never recorded. Followers are taken next run, once the leader catches up.
+ * never recorded. Each pin keeps the range prefix it was written with, so a
+ * `==x.y.z` Python pin stays in that form.
  *
  * @param groups the coupling declarations to enforce
- * @param tables the proposed-version tables to read and rewrite, in precedence
- *   order — a member is looked up in the first table that declares it
+ * @param tables the proposed-version tables to read and rewrite; a member is
+ *   looked up in the first table that declares it
  */
 export const holdGroupsInLockstep = (
   groups: readonly LockstepGroup[],
@@ -109,34 +75,30 @@ export const holdGroupsInLockstep = (
 ): LockstepNote[] => {
   const notes: LockstepNote[] = [];
 
-  const merged = Object.assign({}, ...[...tables].reverse()) as Record<
-    string,
-    string
-  >;
-
   for (const group of groups) {
-    const leaderVersion = group.leader.resolve
-      ? group.leader.resolve(merged)
-      : readMember(group.leader, tables);
-    if (leaderVersion === undefined) {
+    // A group can only be held where the run actually proposed something for at
+    // least two of its members; a single pin has nothing to stay in step with.
+    const members = group.flatMap((name) => {
+      const table = tables.find((candidate) => name in candidate);
+      return table ? [{ name, table }] : [];
+    });
+    if (members.length < 2) continue;
+
+    const lowest = members
+      .map(({ name, table }) => table[name])
+      .reduce((min, version) =>
+        compareVersions(version, min) < 0 ? version : min,
+      );
+
+    for (const { name, table } of members) {
+      const proposed = table[name];
+      const held = `${prefixOf(proposed)}${bare(lowest)}`;
+      if (proposed === held) continue;
+      table[name] = held;
       notes.push({
-        note: `could not resolve ${group.leader.name} to hold its group against — ${group.followers
-          .map((f) => f.name)
-          .join(', ')} left as proposed`,
-      });
-      continue;
-    }
-
-    for (const follower of group.followers) {
-      const proposed = readMember(follower, tables);
-      if (proposed === undefined) continue;
-
-      const required = applyFormat(leaderVersion, follower.format);
-      if (proposed === required) continue;
-
-      if (!writeMember(follower, required, tables)) continue;
-      notes.push({
-        note: `${follower.name} held at ${required} (${proposed} available): ${group.reason}`,
+        note: `${name} held at ${held} (${proposed} available): must move in step with ${group
+          .filter((other) => other !== name)
+          .join(', ')}`,
       });
     }
   }
--- a/packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
+++ b/packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
@@ -1,0 +1,145 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Groups of pins that must move together.
+ *
+ * Some packages are only correct at matching versions: a Python formatter and the
+ * wasm build of it that actually runs, or an npm library and a second library
+ * that depends on an exact release of it. The bumps behind those pins arrive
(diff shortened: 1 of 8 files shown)


nx-plugin-for-aws:
Retargeted onto `chore/ci-flake-local-dev-chat` (#1169) rather than `main`.

This branch's `Smoke Tests - local-dev` leg failed twice on the chat flake that #1169 fixes — same signature both times (prompt renders `Type a message`, the typed message is never repainted, so it was never consumed):

```
Error: Chat target local_dev_test.py_project:my-py-agent-chat did not stream "The answer is 42" within 120000ms
```

Nothing here touches the e2e suite, so stacking on the fix keeps this PR's CI meaningful. Its diff is now the lockstep change alone; will retarget to `main` once #1169 merges.

nx-plugin-for-aws:
Correction to my earlier comment: this is back on `main`, not stacked on #1169.

Retargeting to a non-`main` base silently stopped CI altogether — `pr.yml` only triggers on `pull_request: branches: [main]` — so stacking wasn't viable. Rebased back onto `main`; the diff is the three lockstep files only.

The `local-dev` flake this branch hit twice is fixed by #1169, so once that merges a rebase here will pick it up.

packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
```diff
@@ -0,0 +1,107 @@
...
+/** The range prefix a pin was written with, so it can be preserved. */
+const prefixOf = (version: string): string =>
+  version.slice(0, version.indexOf(bare(version)));
+
+/**
+ * Compare bare `major.minor.patch[...]` versions segment by segment. Numeric
+ * segments compare numerically; anything else (prereleases) compares as a
+ * string, which is enough to order the pins these tables hold.
+ */
+export const compareVersions = (a: string, b: string): number => {
```
cogwirrel: we already have a dependency on semver

packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
```diff
@@ -0,0 +1,92 @@
...
+/** A note explaining a pin the run deliberately did not take. */
+export interface LockstepNote {
+  note: string;
+}
+
+/**
+ * The pins in one group, named as they appear in the versions tables. Members may
+ * live in different tables (e.g. the TypeScript and Python pins).
+ */
+export type LockstepGroup = readonly string[];
```
cogwirrel: type this properly

scripts/update-versions.ts
```diff
@@ -183,34 +254,61 @@ const getUpdatedPythonVersions = (tmpDir: string): Record<string, string> => {
...
+  // train and cannot duplicate against CopilotKit, so they are left free.
+  ['@ag-ui/client', '@ag-ui/core', '@ag-ui/encoder'],
+  // `@biomejs/wasm-nodejs` formats generated TypeScript/JSON in-process here,
+  // while `@biomejs/biome` is the CLI a generated project's `format`/`lint`
+  // targets run and the version its vended `biome.json` `$schema` points at. Both
+  // are npm packages published together (minutes apart, same version line), so
+  // whichever the run proposes lower holds the other back — out of step, generated
+  // files are formatted by one release and checked against another.
+  ['@biomejs/wasm-nodejs', '@biomejs/biome'],
+];
```
cogwirrel: way too much in the comments, make all the comments you add succinct

scripts/update-versions.ts
```diff
@@ -183,34 +254,61 @@ const getUpdatedPythonVersions = (tmpDir: string): Record<string, string> => {
...
- *
- * The wasm build leads because it is the one already installed and running here;
- * the held-back ruff is taken next run, once the bindings catch up.
+ * Unlike the other couplings this one spans registries: the bindings are on npm
+ * and `ruff` is on PyPI, resolved by a different tool with no cooldown. So one
+ * side's proposed version is not comparable to the other's as a group — the
+ * bindings simply lead, and the held-back `ruff` is taken next run once they
+ * catch up.
  */
 const holdRuffToWasmBuild = (
```
cogwirrel: this doesn't seem unique though, our algorithm picks min of updated-to versions right?

scripts/update-versions.ts
```diff
@@ -72,6 +77,66 @@ interface ChangeGroup {
...
 
+/**
+ * npm packages this repo depends on directly that a lockstep group couples to a
+ * vended pin: the wasm formatter bindings, which format generated files here
+ * while the vended pin is the CLI a generated project runs.
+ */
+const REPO_LOCKSTEP_MEMBERS = [
+  '@astral-sh/ruff-wasm-nodejs',
+  '@biomejs/wasm-nodejs',
+] as const;
```
cogwirrel: can you investigate whether it works more generically if we just merge together the nx plugin package deps with the deps we vend in versions.ts? That might make this simpler and mean fewer hardcoded lists to maintain.

packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
nx-plugin-for-aws: Done — switched to `coerce` + `compare` from `semver` (matching how `dependencies.ts` and `migration-versions.ts` use it). That also removed my hand-rolled range stripping: `==0.16.5` and `0.16.5` now compare equal via `coerce`.

packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
nx-plugin-for-aws: Typed: `LockstepGroup` is now a `readonly [Name, Name, ...Name[]]` tuple so a single-member group doesn't compile, and the tables are `ProposedVersions<Name>` (keyed by member name) rather than `Record<string, string>`. Verified with a `@ts-expect-error` probe that a one-element group is rejected.

scripts/update-versions.ts
nx-plugin-for-aws: Trimmed these down.

scripts/update-versions.ts
nx-plugin-for-aws: You're right — I had this wrong. `ruff` and `@astral-sh/ruff-wasm-nodejs` are published from the same repo on every release, so they share a version line and taking the min gives the same answer:

```
coerce('==0.17.0') -> 0.17.0, coerce('0.16.5') -> 0.16.5, min = 0.16.5
```

The registry difference only affects *when* each bump shows up, not whether the versions are comparable. Removed the bespoke hold; it's now just `['@astral-sh/ruff-wasm-nodejs', 'ruff']` and all three couplings are plain groups.

scripts/update-versions.ts
nx-plugin-for-aws: Investigated — it does work more generically, so both hardcoded lists are gone. The dummy project ncu resolves is now:

```ts
packageJson.dependencies = {
  ...Object.assign({}, ...REPO_MANIFESTS.map(readManifestDependencies)),
  ...TS_VERSIONS,   // vended pin wins where both declare a package
};
```

Checked the overlap first: 12 packages are declared in both the plugin's manifest and `TS_VERSIONS` (`@nx/*`, `typescript`, `vite`, `vitest`, `zod`, `minimatch`, …) and all 12 already agree, so the merge doesn't change any resolved version. Confirmed on a real run that `@astral-sh/ruff-wasm-nodejs`, `@biomejs/wasm-nodejs` and `@biomejs/biome` all reach the resolver with nothing named explicitly.

One thing worth flagging: a held pin still has to be corrected in `package.json`, since ncu bumps that *before* this script runs. I first tried applying every resolved version back to the manifests, but that re-did ncu's job (24 lines of churn across both manifests) and would have left the lockfile stale, because `pnpm i` has already run by then. So the writeback is scoped to just the names a group actually held — `LockstepNote` now carries the dependency name for that. Verified a run with no holds touches no manifest.

packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
```diff
@@ -0,0 +1,107 @@
...
+export type ProposedVersions<Name extends string = string> = Partial<
+  Record<Name, string>
+>;
+
+/**
+ * The pins in one group, named as they appear in the versions tables. At least
+ * two, since one pin has nothing to stay in step with. Members may live in
+ * different tables (e.g. the TypeScript and Python pins).
+ */
+export type LockstepGroup<Name extends string = string> = readonly [
```
cogwirrel: These should be typed as ts/py/java/etc deps from versions.ts too

scripts/update-versions.ts
```diff
@@ -72,6 +75,50 @@ interface ChangeGroup {
...
+      if (manifest[field]?.[name] && manifest[field][name] !== held) {
+        manifest[field][name] = held;
+        changed = true;
+      }
+    }
+  }
+  if (changed) {
+    tree.write(manifestPath, `${JSON.stringify(manifest, null, 2)}\n`);
+  }
+};
```
cogwirrel, line 162: Can we instead remove the ncu run for this repo, do everything in one pass and then write the updates to the manifests rather than rolling back?

packages/nx-plugin/src/utils/version-lockstep/lockstep.ts
nx-plugin-for-aws: Done — typed as `IVersionedDep`, a union of the key types the version tables already export (`ITsDepVersion | IPyDepVersion | IJavaVersion | IMiseVersion`). `LockstepGroup` and `ProposedVersions` are keyed off that, so a group naming something no table pins doesn't compile.

scripts/update-versions.ts
nx-plugin-for-aws: Done, and this is much better — the rollback is gone entirely.

The workflow's `npx npm-check-updates` step is removed; `update-versions.ts` now resolves this repo's manifests in the same pass as the vended pins and writes both. Since nothing has pre-bumped the manifests by then, a held pin is simply never written, rather than written and corrected.

The workflow step is now:

```
pnpm i
tsx ./scripts/backfill-migration-versions.ts
tsx ./scripts/update-versions.ts
pnpm i --no-frozen-lockfile   # manifests are final here
pnpm nx reset
...
```

Two things the removed ncu run covered that I had to add explicitly: `packageManager` (`pnpm@x.y.z`) and `peerDependencies`. Both are resolved now, and manifest bumps get their own "Repo Dependencies" section in the report.

Tested by running the real sequence in a throwaway worktree rather than just the script: `@nx/*` and `nx` bumped 23.1.2 → 23.1.3 across the root manifest, the plugin's deps *and* its `peerDependencies`, `pnpm i --no-frozen-lockfile` then updated `pnpm-lock.yaml` cleanly, and the vended `TS_VERSIONS` moved in the same run. Worth noting I had not been running the full workflow sequence before — only the script — which is exactly why the ordering problem you're pointing at went unnoticed.

scripts/update-versions.ts
```diff
@@ -78,45 +78,73 @@ interface ChangeGroup {
 /** Manifests whose npm dependencies are resolved alongside the vended pins. */
 const REPO_MANIFESTS = ['package.json', 'packages/nx-plugin/package.json'];
```
cogwirrel: other packages in this workspace need to be updated too (create workspace and mcp server)

nx-plugin-for-aws: Good catch — fixed, and made discovery dynamic rather than adding them to a list.

The manifests now come from `pnpm-workspace.yaml`'s own `packages` globs:

```ts
const repoManifests = (): string[] => {
  const workspace = parseDocument(readFileSync('pnpm-workspace.yaml', 'utf-8'));
  const patterns = (workspace.toJS().packages ?? []) as string[];
  return ['package.json', ...fastGlob.sync(patterns.map(p => `${p}/package.json`), { onlyFiles: true }).sort()];
};
```

That resolves to all four — root, `create-nx-workspace`, `nx-plugin-mcp`, `nx-plugin` — and a package added later is picked up with no change here. It also matches the glob the removed ncu run used (`{package.json,packages/**/package.json}`).

Note `create-nx-workspace` and `nx-plugin-mcp` currently declare no dependencies at all, so this is a no-op for them today — but it means they won't be silently skipped once they do.

Verified with a real run: manifest bumps still land correctly under the "Repo Dependencies" report section.

cogwirrel (dismissed):

scripts/update-versions.ts
```diff
@@ -183,33 +277,15 @@ const getUpdatedPythonVersions = (tmpDir: string): Record<string, string> => {
...
-    },
-  ];
-};
+const LOCKSTEP_GROUPS: readonly LockstepGroup[] = [
+  // A duplicate `@ag-ui/client` fails every generated website with TS2322.
+  ['@ag-ui/client', '@ag-ui/core', '@ag-ui/encoder'],
+  // The wasm bindings format generated files; the CLI checks them.
+  ['@biomejs/wasm-nodejs', '@biomejs/biome'],
+  ['@astral-sh/ruff-wasm-nodejs', 'ruff'],
+];
```
cogwirrel: move this to versions.ts and add a unit test which makes sure that the value in versions.ts for each of these are all equal

nx-plugin-for-aws: Done — `LOCKSTEP_GROUPS` now lives in `versions.ts` next to the versions it names, and `versions.spec.ts` asserts each group agrees:

```
✓ pins every member of [@ag-ui/client, @ag-ui/core, @ag-ui/encoder] at the same version
✓ pins every member of [@biomejs/wasm-nodejs, @biomejs/biome] at the same version
✓ pins every member of [@astral-sh/ruff-wasm-nodejs, ruff] at the same version
```

The test resolves each member from wherever it's pinned — `TS_VERSIONS`, `PY_VERSIONS`, or the plugin's manifest for the wasm bindings, which aren't vended — and normalises the range prefix so `ruff`'s `==0.16.5` compares equal to `0.16.5`. It also fails if a member isn't pinned anywhere, which catches a typo'd group entry.

Verified it actually catches a divergence rather than passing vacuously: forcing `@ag-ui/core` to `0.0.58` fails the ag-ui case, and it passes again once reverted.

cogwirrel (approved):

PR #1172, opened 2026-09-02
feat(ts#infra): use CloudFormation express mode for deploy and deploy-sandbox

nx-plugin-for-aws:
### Reason for this change

CDK express mode was enabled by default for the infra `deploy` target in #878, then made opt-in again in #909 because it raced IAM policy propagation.

The specific failure: the AgentCore Gateway runs a creation-time authorization check (`GenesisPolicyEngineCheck` → `bedrock-agentcore:AuthorizeAction`). The gateway construct grants that permission and orders it before the gateway via `addDependency`, but `addDependency` only orders the *CloudFormation* operations — it does not wait for IAM's authorization data plane to propagate the freshly-created inline policy. With express mode returning early, the check could fire before propagation completed → 403 AccessDenied → `CREATE_FAILED`, at roughly **1 in 20** deploys.

That propagation race has since been fixed on the service side, so express mode is safe to re-enable for development deploys.

### Description of changes

- **`deploy-sandbox` target** (`infra/app/generator.ts`): passes `--express`. The flag trails the stage pattern so the pattern stays the first positional argument, which is where `parseStageName` reads the stage from for stage-credential resolution.
- **`deploy` target**: passes `--express` only when the project is generated *without* `stageConfig`, where the only stage `main.ts` declares is the sandbox one. With `stageConfig`, `deploy` takes a stage name that may well be beta or prod, so it stays on full resource stabilization.
- **Migration** (`latest-infra-deploy-express-mode`): backfills the flag on existing ts#infra projects, following the same rule — always `deploy-sandbox`, and `deploy` only when it is not the stage-config variant. Idempotent (skips targets that already pass it), honours an explicit `--no-express` opt-out, and reports customised commands via `nextSteps` rather than rewriting them.
- **`deploy-ci` and all destroy targets**: unchanged. `deploy-ci` never uses express mode and always waits for full stabilization.
- **`buildCdkCommand`** (`cdk-command.ts.template`): left untouched — express mode comes from the target commands, not from a default in the script. Users can still opt out with `--no-express`, since it is a yargs boolean and the later flag wins.
- **Docs** (`typescript-infrastructure.mdx`, dungeon-game tutorial): document `deploy-sandbox`'s express mode with the `--no-express` opt-out, note that `deploy` waits for full stabilization under `stageConfig` (with `--express` available to opt in per stage), and state that `deploy-ci` is the target for CI/CD and production deployments.

**Race verification (the main question this PR turns on).** Built an isolated repro stack containing only the resources on the race path — L2 `agentcore.Gateway`, `CfnPolicyEngine` attached in `ENFORCE` mode, the `AuthorizeAction` grant, and the `addDependency` ordering — matching the `AgentCoreGateway` construct's wiring line-for-line. Deployed it **130 times** with `--express`, each a fresh CREATE with a unique stack name, on `aws-cdk@2.1139.0` / `aws-cdk-lib@2.267.0` in `us-west-2`:

- **130/130 `CREATE_COMPLETE`**, 0 `CREATE_FAILED`, 0 `AuthorizeAction`/`AccessDenied` failures.
- All 130 stacks tore down cleanly (0 `DELETE_FAILED`); no resources left behind.
- Detection read CloudFormation stack events rather than CLI output, and matched any `CREATE_FAILED` reason across all resources, so an unexpected failure mode could not be silently classified as a pass.

**Control — confirming the race path is actually exercised.** A null result is only meaningful if the check still fires, so this was verified independently via CloudTrail: every deploy produced exactly one `AuthorizeAction` call from its own `GenesisPolicyEngineCheck` session on that stack's gateway service role, and **every one returned `OK`** (50 distinct gateway roles in the sampled window, zero `AccessDenied`). At the historical ~1/20 rate, 130 clean runs put the probability of the defect still being present at well under 0.1%.

**Flag precedence.** Confirmed against a real deploy that `--express --no-express` deploys *without* express mode, so the hardcoded `--express` in the target command remains overridable by users.

**Unit tests.** Added 9 migration tests, including a `stageConfig=false`/`stageConfig=true` pair that strips `--express` from a freshly generated project, runs the migration, and asserts the targets converge byte-identically on what the generator vends. Updated the infra generator target/snapshot tests. Full suite green: **4184 tests passing**, snapshots updated, build and docs build pass.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/typescript-infrastructure.mdx
```diff
@@ -425,7 +429,11 @@ Use the `deploy-ci` target if you are deploying to AWS as part of a CI/CD pipeli
 
 <NxCommands commands={['deploy-ci <my-infra> my-stage/*']} />
 
-This target differs slightly from the regular `deploy` target in that it deploys a pre-synthesized cloud assembly rather than synthesizing on the fly. This avoids potential non-determinism from package version changes, ensuring that every pipeline stage deploys using the same cloud assembly.
+This target differs from the `deploy` and `deploy-sandbox` targets in two ways. First, it deploys a pre-synthesized cloud assembly rather than synthesizing on the fly, avoiding potential non-determinism from package version changes and ensuring that every pipeline stage deploys using the same cloud assembly. Second, it does not use express mode, so it waits for every resource to fully stabilize before reporting success.
+
+:::caution[Always use `deploy-ci` for production]
+`deploy` and `deploy-sandbox` use express mode and return before resources have finished stabilizing, which makes them unsuitable for CI/CD and production deployments. Use `deploy-ci` for those.
+:::
```
cogwirrel: remove this caution block.

packages/nx-plugin/src/utils/files/common/scripts/src/infra/stage-credentials/cdk-command.ts.template
```diff
@@ -6,9 +6,8 @@
  * respect their choice and don't add the default. `cdk destroy` has no
  * --require-approval option, so nothing is added for it.
  *
- * Pass --express to opt into CloudFormation express mode for faster
- * iteration; it is not enabled by default so deploys wait for full
- * resource stabilization.
+ * Deploys default to --express (CloudFormation express mode) for faster
+ * local iteration. Pass --no-express to wait for full resource stabilization.
  */
 export function buildCdkCommand(
```
cogwirrel, line 13: Let's leave this untouched

cogwirrel (commented):
needs a migration that adds `--express` for `deploy-sandbox` and `deploy` targets for ts#infra projects.

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
nx-plugin-for-aws: Removed.

packages/nx-plugin/src/utils/files/common/scripts/src/infra/stage-credentials/cdk-command.ts.template
nx-plugin-for-aws: Reverted - this template is now byte-identical to `main`, along with its spec.

Instead of defaulting the flag in `buildCdkCommand`, the `deploy` and `deploy-sandbox` targets pass `--express` in their commands directly. It trails the stage pattern (`... infra-deploy.ts packages/infra "proj-infra-sandbox/*" --express`) so the pattern stays the first positional argument, which is where `parseStageName` reads the stage from for stage-credential resolution.

Users can still opt out with `--no-express`: it's a yargs boolean, so the later flag wins. I confirmed this against a real deploy — `--express --no-express` deployed without the express-mode warning.

nx-plugin-for-aws:
Added `latest-infra-deploy-express-mode`, which adds `--express` to the `deploy` and `deploy-sandbox` targets of existing ts#infra projects (matched on `INFRA_APP_GENERATOR_INFO`).

It handles both target shapes — the direct `cdk deploy` command and the stage-config `infra-deploy.ts` script — and appends the flag after the stage pattern so the pattern stays the first positional argument. `deploy-ci` and all destroy targets are left alone.

Guardrails:
- **Idempotent** — skips targets that already pass `--express`, so re-running is a no-op.
- **Honours opt-outs** — a command with an explicit `--no-express` is left as-is rather than flipped back.
- **No clobbering** — a customised deploy command is reported via `nextSteps` instead of being rewritten.
- **Converges on the generated shape** — a test strips `--express` from a freshly generated project, runs the migration, and asserts the targets are byte-identical to what the generator vends.

8 migration tests, and the full suite is green at 4183 passing.

nx-plugin-for-aws:
Dropped `--express` from the stage-config `deploy` target — that one takes a stage name that may well be beta or prod, so it stays on full stabilization. Only `deploy-sandbox` opts in there.

`deploy` keeps the flag when the project is generated *without* `stageConfig`, since the only stage `main.ts` declares in that case is the sandbox one. The migration follows the same rule and skips the `deploy` target of stage-config projects.

Resulting matrix, verified against generated projects in both modes:

| Target | stageConfig=false | stageConfig=true |
| --- | --- | --- |
| `deploy-sandbox` | express | express |
| `deploy` | express | full stabilization |
| `deploy-ci` | full stabilization | full stabilization |
| all `destroy*` | n/a | n/a |

The convergence test now runs for both modes, asserting the migration's output is byte-identical to what the generator vends in each. Suite is green at 4184 passing.

cogwirrel (approved):

PR #1173, opened 2026-09-02
perf(ts#project): copy with Node's fs.cpSync instead of ncp

nx-plugin-for-aws:
### Reason for this change

`FsCommands.cp()` emits `ncp` into generated Nx targets. `ncp` is a pure-JS recursive copy that predates a built-in equivalent, and it is slow: copying the 110MB / 94-package `node_modules` tree that a code-packaged AgentCore runtime vendors takes **~2.1s**, which is paid on every change to an agent's code. `fs.cpSync` has been in Node since v16 and does the same work in **~0.29s**.

### Description of changes

`cp()` now emits a `node -e` one-liner calling `fs.cpSync`, and `ncp` is dropped from `FS_DEPENDENCIES` and `TS_VERSIONS`:

```
node -e "require('fs').cpSync(process.argv[1],process.argv[2],{recursive:true,verbatimSymlinks:true})" "<src>" "<dst>"
```

**`verbatimSymlinks: true` is load-bearing.** Without it `cpSync` rewrites a relative symlink target to an absolute host path, which would silently corrupt any bundled `node_modules` tree. With it, the emitted link is byte-identical to what `ncp` produced.

Paths are passed through `process.argv` rather than interpolated into the script, so a path can never break out of the expression; each is quoted so paths with spaces survive. There is existing precedent for this shape — `ts/react-website/app/generator.ts` already ships a `node -e` copy with single-quoted `'fs'`, parens and braces, and passes Windows CI today.

Two behaviour differences, both improvements, neither reachable from any call site:

- `ncp` **silently exits 0 and copies nothing** when a file's destination is an existing directory. `cpSync` throws. Every call site names an explicit file destination, so nothing relies on the old behaviour — but it was a latent silent-failure mode.
- Copying a directory into its own subtree: `ncp` "succeeds", `cpSync` throws.

`cpGlobToFile` is untouched — it is the only glob-based copy, has one caller, and keeps using `cpy-cli`.

**Speed**, interleaved runs to cancel drift (110MB / 94-package tree):

| pair | `ncp` | `cpSync` | speedup |
|---|---|---|---|
| 1 | 2112ms | 291ms | 7.2x |
| 2 | 2268ms | 279ms | 8.1x |
| 3 | 2123ms | 291ms | 7.2x |

**Correctness** — a standalone harness compared `ncp` against `cpSync` case by case, covering every shape the call sites use: file→file, file→renamed file (`Dockerfile.migration`→`Dockerfile`), file→existing file (overwrite), dir→new dir, dir→existing empty dir, **dir→existing populated dir (merge)**, missing intermediate parents, empty subdirectories, relative/directory/broken symlinks, mode preservation (`run.sh` 0755), dotfiles, spaces in paths, zero-byte files, read-only source, read-only destination overwrite, missing source, 8-level nesting, and repeated-run idempotency. Behaviour matches on all of them.

The `py#rdb` command sequence was also run end to end through `shell: true` (rm → mkdir → directory merge → file rename), confirming merge, rename and relative-symlink preservation together.

**Shell portability** — the payload contains `(`, `)`, `,` and `{}`, so it must be quoted; the unquoted form was confirmed to fail. The shipped form was verified in sh, bash, dash and zsh, and through Nx's exact spawn shape (`spawn(cmd, [], { shell: true })`, which is `cmd.exe /d /s /c` on Windows).

**Tests** — full `src/ts`, `src/py`, `src/utils`, `src/smithy` and `src/migrations` run: 186 failures, **every one** the pre-existing `@nx/vitest` `reading 'version'` error that also reproduces with this change stashed. Zero failures reference `cpSync` or `ncp`. Snapshots were hand-edited rather than regenerated with `-u`; `git diff --numstat` shows exactly `1 1` per snapshot file, so nothing was dropped.

### Not verified

**This was not executed on Windows or macOS.** Confidence rests on the quoting analysis, the POSIX-shell testing above, and the existing `node -e` precedent already green on Windows CI. Windows coverage for these paths comes from the `windows_standalone` leg.

Existing user workspaces keep a now-unused `ncp` devDependency. No migration strips it — it is harmless, and rewriting every workspace's manifest for a dead dev dependency seemed the worse trade. Happy to add one if you would rather.

### Checklist

----


nx-plugin-for-aws:
Closing in favour of #1175, which replaces `ncp` (and `rimraf`, `make-dir-cli` and `cpy-cli`) with a single `shx` dependency rather than an inline `node -e` `fs.cpSync` call. `shx` is a little slower than `cpSync` on a large tree (~0.5s vs ~0.3s on 110MB / 80 packages) but still 3-4x faster than `ncp`, and consolidating four CLIs into one is the better trade.

PR #1175, opened 2026-09-02
perf(ts#project): replace the filesystem CLIs with shx

nx-plugin-for-aws:
### Reason for this change

The generators ran four single purpose CLIs for filesystem work: `ncp`, `rimraf`, `make-dir-cli` and `cpy-cli`. One `shx` covers every operation they vend, and its `cp` is substantially faster than `ncp` on a large tree — which the agent code packaging targets copy on every build.

Interleaved A/B/C runs over a 110MB / 80 package tree:

| op | before | shx |
|---|---|---|
| dir copy | `ncp` 1754–1791ms | **483–568ms** |
| `rm -rf` | `rimraf` 171–173ms | 222–230ms |
| `mkdir -p` | `make-dir` 70–73ms | 72–73ms |

`rm` and `mkdir` are a wash; the copy is the win, and consolidating four dependencies (11M / 42 packages) into one (3.4M / 46) is the other.

### Description of changes

`FsCommands.cp` is split into `cpFile` and `cpDir`. This is load bearing: `shx cp -R` follows POSIX `cp`, which **nests** the source inside an existing destination rather than merging into it. Every call site `mkdir`s the destination first, so a naive `shx cp -R src dst` would land a Python bundle at `package/bundle-arm/...` instead of `package/...` — an exit-0 build producing a zip that fails at runtime. A directory copy therefore uses `shx cp -R <src>/. <dst>`, which merges the contents, and unlike a `<src>/*` glob still includes dotfiles (`node_modules/.bin` depends on them). A file copy must **not** have the `/.`, so each call site names which it is.

`cpGlobToFile` gains a `shx mkdir -p` for its destination: `cpy` created a missing parent directory implicitly and `shx cp` does not, and the Smithy OpenAPI copy targets `build/openapi`, which nothing else creates.

`terraform#project`'s three direct `make-dir` calls move to `shx mkdir -p` as well.

A migration rewrites the vended commands and swaps the dependency. Whether an `ncp` copied a file or a directory is not recoverable from the command alone, so it matches the shapes the generators produce (Python bundles, module trees, the generated SSDK, `prisma`, Alembic `migrations`; `Dockerfile` variants are recognised as extensionless files) and **skips anything it cannot classify**, reporting it via `nextSteps` rather than guessing. It only drops a package once nothing in the workspace still runs its CLI, so a command it declined to rewrite keeps its dependency installed.

**Semantics — a 9 case matrix** (destination nonexistent / empty / populated × dotfiles, nested dirs, relative symlinks) confirming `shx cp -R "src/."` is byte-for-byte equivalent to `ncp` in all of them. Separately confirmed the two forms that are *not* equivalent and would have silently broken things: bare `shx cp -R src dst` nests, and `shx cp -R "src/*" dst` drops dotfiles.

**Migration against a real workspace** (a generated agent/MCP project on the pre-change plugin) via the real `Tree`/`flushChanges` path: every command rewritten correctly, `Dockerfile` and `.trivyignore` both classified as file copies, deps swapped, no `nextSteps`. Then rebuilt and diffed the resulting package tree against the pre-migration build — **identical**, ADOT present. `nx run m:ecr-mcp-docker` (exercising the `shx cp` Dockerfile copy) passes.

**The exact vended command strings executed for real**, including the shapes the migrated workspace didn't contain: the Python package sequence (bundle merged at root, module tree under its name, entry point at root, dotfiles preserved) and the Smithy sequence (`model.json` file copy, glob copy into a not-yet-existing `build/openapi`, SSDK directory copy).

**Test suite baseline-diffed** against clean `main` (this environment has ~293 pre-existing failures from an unrelated `@nx/js`/vite issue, so raw counts are meaningless): **0 new failures, 0 newly passing**. 13 new migration tests, including no-op-on-second-run, object-form commands, and the skip-and-report path. One snapshot file updated, exactly `20 20`, containing only command rewrites.

Rewriting the spec expectations with the same classifier the migration uses left **0 of 25 `ncp` occurrences unclassifiable**, independently confirming the migration covers every shape actually vended.

Docs updated (`docker-bundling.mdx`), including the `/.` requirement.

Note: `nx run m:ecr-mcp-trivy` fails in this container, but it fails identically on unmigrated `main` — a rootless-Docker mount issue, not a regression; both `.trivyignore` files land correctly.

### Checklist

----


docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -186,7 +186,9 @@ Set `"cache": false` on image build targets. A built image is not written to an
 :::
 
 :::tip[Cross-platform file copy]
-The `ncp` package provides a cross-platform file/directory copy command, avoiding `cp` (unavailable on Windows). Install it at the root of your workspace with `pnpm add -D -w ncp`.
+The `shx` package provides cross-platform `cp`, `rm` and `mkdir` commands, avoiding the shell builtins (unavailable on Windows). Install it at the root of your workspace with `pnpm add -D -w shx`.
+
+Copying a directory needs `-R` and a trailing `/.` on the source — `shx cp -R src/. dst` — which merges the contents into `dst`. Without the `/.` the source is nested inside `dst` instead, and unlike a `src/*` glob it still includes dotfiles.
```
cogwirrel: This line is pointless in this context, remove

docs/src/content/docs/en/guides/docker-bundling.mdx
```diff
@@ -186,7 +186,9 @@ Set `"cache": false` on image build targets. A built image is not written to an
 :::
 
 :::tip[Cross-platform file copy]
-The `ncp` package provides a cross-platform file/directory copy command, avoiding `cp` (unavailable on Windows). Install it at the root of your workspace with `pnpm add -D -w ncp`.
+The `shx` package provides cross-platform `cp`, `rm` and `mkdir` commands, avoiding the shell builtins (unavailable on Windows). Install it at the root of your workspace with `pnpm add -D -w shx`.
```
cogwirrel: Use `<InstallCommand>` instead of pnpm specific `pnpm add`

packages/nx-plugin/src/migrations/latest/shx-fs-commands/migration.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  const rimraf = RIMRAF.exec(trimmed);
+  if (rimraf) return `shx rm -rf ${rimraf[1]}`;
+
+  const makeDir = MAKE_DIR.exec(trimmed);
+  if (makeDir) return `shx mkdir -p ${makeDir[1]}`;
+
+  const cpy = CPY.exec(trimmed);
+  if (cpy) {
+    return `shx mkdir -p ${cpy[2]} && shx cp "${cpy[1]}" ${cpy[2]}/${cpy[3]}`;
```
cogwirrel: && is bash and defeats the whole purpose of using shx in the first place!

packages/nx-plugin/src/utils/fs.ts
```diff
@@ -35,29 +32,49 @@ export class FsCommands<D extends DependencyDeclaration> {
...
+   * the glob matches nothing, or when it matches more than one file.
+   *
+   * `dstDir` is created first: `shx cp` writes through to a named destination
+   * file and fails if its parent is missing.
    */
   public cpGlobToFile(srcGlob: string, dstDir: string, dstFileName: string) {
-    this.add('cpy-cli');
-    return `cpy "${srcGlob}" ${dstDir} --flat --rename=${dstFileName}`;
+    this.add('shx');
+    return `shx mkdir -p ${dstDir} && shx cp "${srcGlob}" ${dstDir}/${dstFileName}`;
```
cogwirrel: No && (see other comment).
We should refactor callers to mkdir first

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Removed. The `-R` / trailing `/.` detail is an implementation note for the generators, not something a reader of this guide needs — the example above it already shows the correct form.

docs/src/content/docs/en/guides/docker-bundling.mdx
nx-plugin-for-aws: Done — now `<InstallCommand pkg="shx" dev />` with the `@components/install-command.astro` import added, matching the convention in `typescript-project.mdx`. Verified with `nx run docs:build` (1522 pages, clean).

packages/nx-plugin/src/migrations/latest/shx-fs-commands/migration.ts
nx-plugin-for-aws: Agreed — dropped the `&&`. A `cpy` now rewrites to two separate entries in the target's `commands`:

```
"shx mkdir -p dist/packages/model/build/openapi",
"shx cp \"dist/packages/model/smithy/source/openapi/*.openapi.json\" dist/packages/model/build/openapi/openapi.json"
```

`rewrite` returns `string[] | null` so one command can become several. Two consequences handled:

- a target using the single `command` option becomes `commands` when its command splits, since `command` only runs one;
- the target is pinned to `parallel: false` when a split happens, so the `mkdir` cannot race the `cp`. An object entry keeps its other keys on both halves.

Covered by four new cases in `migration.spec.ts` (two entries, `parallel` pinned, `command` -> `commands`, object entry preserved); the no-op-on-re-run case still passes.

packages/nx-plugin/src/utils/fs.ts
nx-plugin-for-aws: Done. `cpGlobToFile` now returns only `shx cp "<glob>" <dstDir>/<dstFileName>`, and the caller in `smithy/project/generator.ts` emits `cmd.mkdir(`${BUILD_DIR}/openapi`)` as its own step just before it.

Worth noting the `mkdir` is genuinely required rather than defensive: the target already did `cmd.mkdir(BUILD_DIR)`, but the destination is the `build/openapi` **subdirectory**, which nothing else creates, and `shx cp` will not create a missing parent. Verified against a real `shx`:

- with the new two-step sequence and `dist/build/openapi` absent beforehand, both steps exit 0 and the file lands at `dist/build/openapi/openapi.json` with the right content;
- `cp` alone with the parent missing fails `exit=1` (`copyFileSync: could not write to dest file (code=ENOENT)`) and writes nothing — hence the explicit `mkdir`.

The deliberate fail-loudly behaviour is preserved: a glob matching nothing exits 1 (`cp: no such file or directory`), and a glob matching more than one exits 1 (`cp: dest is not a directory (too many sources)`).

The migration does the same split, so the two stay consistent.

cogwirrel (approved):

PR #1176, opened 2026-09-02
feat(deps): add support for pnpm 12

nx-plugin-for-aws:
### Reason for this change

pnpm 12 is out, and neither this repo nor the workspaces the plugin generates were verified
against it. Two things made that worth doing now:

**1. Generated workspaces pin no `packageManager`**, so a user who already has pnpm 12 on their
PATH gets pnpm 12 whether or not it's supported. Anything that breaks under 12 is breaking for
them today.

**2. It narrows a macOS CI flake.** `macOS Smoke Tests - standalone (N/6)` fails intermittently
while a generator runs:

```
[EINVAL] EINVAL: invalid argument, open '.../standalone-ts-dynamodb/proj/node_modules/@aws/nx-plugin/package.json'
 NX   Command failed: pnpm install --no-frozen-lockfile
```

pnpm's `removeDirectDependency` reads a dependency's `package.json` and deletes that same
directory in one `Promise.all`, and only tolerates `ENOENT` from the read. When the delete wins
the race the read throws whatever errno the partially removed path yields — `EINVAL` on macOS —
and that aborts the whole install. The dependency is pruned in the first place because
`@aws/nx-plugin` peers on `nx`, giving it a ~500-character peer-suffixed lockfile reference, so
**any** generator that adds a root devDependency changes that string and trips the removal.

Verifying pnpm 12 turned up two real bugs, both of which affect users on pnpm 12 regardless of
this PR. Those are the bulk of the change; the version bump itself is a few lines.

### Description of changes

**Support pnpm 12**

- `package.json` — `packageManager` to `pnpm@12.2.1`. `pnpm/setup` reads this field, so every
  CI job that doesn't pin its own package manager runs pnpm 12 too.
- `.github/actions/init-monorepo`, `.github/workflows/on-release.yml` — install pnpm with
  `pnpm/setup` (supersedes `pnpm/action-setup`), taking the version from `packageManager`.
  `install: 'false'` because both callers install further down, once the pnpm store cache has
  been restored.
- `e2e/src/smoke-tests/pnpm-12.spec.ts` — new variant, plus the matrix entry in both `pr.yml`
  and `ci.yml`. `pnpm-10`, `pnpm-11` and `pnpm-12` each pin their own major through corepack,
  so all three supported lines stay independently covered.
- `pnpm-lock.yaml` — additive only; pnpm 12 records itself under a new
  `packageManagerDependencies` importer key, with integrity hashes for the `@pnpm/exe.*`
  binaries. `lockfileVersion` stays `9.0`, so there is no lockfile migration.
- `packages/nx-plugin/src/utils/pnpm-workspace.ts`, `utils/init.ts` — comments that described
  `allowBuilds` and the `strictDepBuilds=true` default as "pnpm 11's" now say pnpm 11 and above,
  since both are unchanged in 12.

Confirmed generated workspaces are pnpm-12 clean: every key the generated
`pnpm-workspace.yaml` contains is recognised — which matters, because pnpm 12 hard-fails on
unrecognised workspace settings (`ERR_PNPM_UNRECOGNIZED_WORKSPACE_SETTINGS`) when the project
pins a pnpm version the running binary satisfies. Both `allowBuilds` and pnpm 10's
`onlyBuiltDependencies` are still accepted, so the dual-key allowlist the generators write needs
no change.

**Fix: pnpm's temp files invalidate the Nx project graph**

pnpm creates `_tmp_<pid>_<hash>` files at the workspace root while it materialises packages. Nx
watches the workspace, and an unignored file appearing and vanishing invalidates the project
graph — so a watch-mode `dev` target running while any install happens rebuilds the graph
continuously and never starts the command it wraps. No error, just a port that never opens. That
is what failed the `dungeon-adventure` and `local-dev` smoke tests, whose `dev` cascades install
as they come up (51,722 watcher events in one run).

- `packages/nx-plugin/src/utils/init.ts` — the generated `.gitignore` covers the pattern, inside
  the pnpm branch since only pnpm writes these.
- `migrations/latest/ignore-package-manager-temp-files/` — the same for existing workspaces,
  gated on `pnpm-lock.yaml`.
- `.gitignore` — this repo needs it too: it has a watch-mode `docs:start` target.

**Fix: an undeclared optional peer split `@strands-agents/sdk` in two**

`@aws-sdk/client-s3` is one of 22 optional peers of `@strands-agents/sdk`. The shared
`agent-connection` project's own code never imports it, so it was never declared — pnpm
auto-installed its own copy there (`3.1119.0`) while a project that does declare it got the
catalog's (`3.1121.0`). Two peer resolutions produce **two peer-suffixed instances of the same
`@strands-agents/sdk@1.15.0`**, and TypeScript treats each copy's classes as distinct nominal
types, so anything importing both projects fails to compile:

```
Type 'McpClient' is not assignable to type 'Agent | ToolList | Tool | McpClient'.
  Types have separate declarations of a private property '_clientName'.
```

That was the `migrate (5/5)` failure. Declaring the peer keeps its version on the catalog and
collapses both projects onto one instance.

- `utils/agent-connection/agent-connection.ts` — declared on the strands templates
  specifically, so it only lands where a strands client does.
- `migrations/latest/agent-connection-pin-strands-peers/` — the same for existing workspaces;
  no-ops when the project has no strands client.

This one is **latent rather than new** — nothing about it is pnpm-12-specific. A from-scratch
install happens to converge on one version, which is why it only surfaces on the incremental
upgrade path the migrate hop exercises.

**Keeping pnpm's supply-chain gates on**

Two pnpm behaviours surface in the e2e suite under 12. Both are handled without weakening a
check or writing anything into a user's workspace:

- `e2e/src/smoke-tests/license-check.spec.ts` installs with `--ignore-scripts`, since only
  license metadata is read. pnpm then treats those packages' build scripts as unreviewed and
  fails the next install a generator runs. It records an explicit `allowBuilds: <pkg>: false`
  decision per package — the same "reviewed, declined" mechanism the generators themselves use —
  rather than relaxing `strictDepBuilds`. New `declinePnpmBuilds` helper in `e2e/src/utils.ts`
  edits `pnpm-workspace.yaml` through the yaml document API, so existing entries and comments
  survive.
- `e2e/src/global-setup.ts` sets `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0`, because smoke tests
  install a build published to verdaccio seconds earlier and pnpm's release-age cutoff hides it.
  Scoped to the e2e process — which every spawned command inherits, including the `spawn`/PTY
  call sites that bypass `runCLI` — so the cutoff still applies everywhere else. A `.npmrc`
  entry would not work: pnpm 12 reads only auth and registry settings from `.npmrc`, so
  `minimum-release-age` there is silently ignored.

**Measured the window.** Polling the pruned path every 2ms while the install runs,
`node_modules/@aws/nx-plugin` is absent for ~254ms under pnpm 11 and ~2ms under pnpm 12.

**Confirmed pnpm 12 works on this workspace.** `pnpm@12.2.1 install --frozen-lockfile` succeeds;
the lockfile stays `9.0` and passes this repo's supply-chain policy check;
`pnpm-workspace.yaml` needs no changes. `@aws/nx-plugin:test` passes 4224 tests across 237
files, and compile and lint pass.

**Confirmed which pnpm each lane actually runs**, rather than assuming the bump propagated —
read back from `Done in … using pnpm vX` in all 39 smoke-test job logs. Every lane runs
**12.2.1** except the three deliberately pinned ones: `pnpm-10` → `10.34.5`, `pnpm-11` →
`11.25.0`, `pnpm-12` → `12.3.0` (each resolves its major's current latest through corepack, so
`pnpm-12` tests one release *ahead* of the repo's own pin). In those three lanes the `12.2.1`
that also appears is only the repo's own harness install, before the test activates its pin.

**Confirmed the dep-builds gate is still enforced**, rather than merely quiet: in a generated
workspace on pnpm 12, removing one package's `allowBuilds` decision and reinstalling fails with
`ERR_PNPM_IGNORED_BUILDS` (exit 1); restoring it passes.

**Confirmed the graph-thrash fix both ways.** Running the generated `dev` target with pnpm 12
active, the Nx daemon log shows the watcher firing on `_tmp_*` deletes and re-entering graph
construction; the spinner reprints ~1000 times in 120s and no port binds. Adding the ignore
takes that to 0 reprints with the port bound in 4s; removing it restores the hang.

**Confirmed the peer fix against a real workspace.** Scaffolded `internal#test-matrix` on rc.90
(31 projects) and bumped the catalog the way a migration does, which reproduced the two variants
CI reports **hash for hash** (`_1e213acde…` / `_b671c015…`); diffing the two virtual-store
directories' resolved peer versions showed `@aws-sdk/client-s3` as the only difference.
Declaring the peer collapses them — verified end to end on a generated workspace under pnpm 12:
one `@aws-sdk/client-s3` copy, both projects linked to the same SDK.

CI is green: 65/65 jobs, no reruns.

### Honest limits of this change

The macOS `EINVAL` flake is **narrowed, not closed**. The underlying pnpm defect is still
present in 12 — probed behaviourally, since 12 ships a native binary rather than a JS bundle:

| pnpm | forcing a non-`ENOENT` errno on that read |
|---|---|
| 12.2.1 | `ERR_PNPM_PACKAGE_MANAGER_PRUNE_DIRECT_DEPS_READ_MANIFEST (os error 21)` |
| 11.25.0 | `[EISDIR] EISDIR: illegal operation on a directory, read` |

Same defect, better error code. So this should make that flake rare rather than impossible, and
a green CI run here does **not** prove that part — the flake is intermittent and only reproduces
on macOS runners. The real fix belongs upstream (await the read before the delete, or widen the
errno guard); I could not find an existing pnpm issue for this signature, so it appears
unreported and is worth filing.

There is no negative test for the init generator's pnpm gate: with no `pnpm-workspace.yaml`,
`detectWorkspacePackageManager` falls through to the real filesystem at `tree.root` — this repo,
which is pnpm — so the assertion cannot be isolated. The migration's gate reads a tree file, so
that case is covered in its spec.

Two things deliberately left out of scope:

- `local-dev` hit `Recursive task invocation detected: postgres-db:pull-image` on one run and
  passed on the next. The `wait-for-db -> dev -> pull-image` chain arrived with #1165, so it
  predates this branch — worth its own fix rather than folding an unrelated change in here.
- Nx 23.1.2's cooldown handling treats any pnpm `>=12.0.0` as `ambiguous` and stops writing
  `minimumReleaseAgeExclude` entries, since its behaviour table ends at pnpm 11. That affects
  `nx migrate` in a pnpm 12 workspace picking up a freshly published package, and belongs
  upstream in Nx.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/.github/actions/init-monorepo/action.yml
+++ b/.github/actions/init-monorepo/action.yml
@@ -78,8 +78,6 @@ runs:
     - name: Use PNPM
       uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
       with:
-        # Dependencies are installed further down, once the pnpm store cache
-        # has been restored.
         install: 'false'
     - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
       if: ${{ runner.os != 'Windows' && inputs.setup-bun == 'true' }}
--- a/.github/actions/init-monorepo/action.yml
+++ b/.github/actions/init-monorepo/action.yml
@@ -78,7 +78,8 @@ runs:
     - name: Use PNPM
       uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
       with:
-        # This action installs later, once the pnpm store cache is restored.
+        # Dependencies are installed further down, once the pnpm store cache
+        # has been restored.
         install: 'false'
     - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
       if: ${{ runner.os != 'Windows' && inputs.setup-bun == 'true' }}
--- a/.github/actions/init-monorepo/action.yml
+++ b/.github/actions/init-monorepo/action.yml
@@ -76,7 +76,10 @@ runs:
       with:
         registry: public.ecr.aws
     - name: Use PNPM
-      uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
+      uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
+      with:
+        # This action installs later, once the pnpm store cache is restored.
+        install: 'false'
     - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
       if: ${{ runner.os != 'Windows' && inputs.setup-bun == 'true' }}
       with:
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,4 @@ Thumbs.db
 .cache-loader/
 vite.config.mts.timestamp-*
 .grit
-# Written at the workspace root while pnpm materialises packages. Left unignored
-# they invalidate the Nx project graph on every install.
 _tmp_*
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ Thumbs.db
 .cache-loader/
 vite.config.mts.timestamp-*
 .grit
+# Written at the workspace root while pnpm materialises packages. Left unignored
+# they invalidate the Nx project graph on every install.
+_tmp_*
--- a/packages/nx-plugin/src/migrations/latest/ignore-package-manager-temp-files/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/ignore-package-manager-temp-files/migration.ts
@@ -10,16 +10,18 @@ import { updateGitIgnore } from '../../../utils/git.js';
 const TEMP_FILE_PATTERN = '_tmp_*';
 
 /**
- * pnpm writes `_tmp_<pid>_<hash>` files at the workspace root while it
- * materialises packages into the virtual store.
- *
- * Nx watches the workspace, and an unignored file appearing and vanishing
- * invalidates the project graph. So a watch-mode `dev` target running while any
- * install happens — the `dev` cascade installing a dependency, or a generator
- * run in another terminal — rebuilds the graph continuously and never starts
- * the command it wraps.
+ * pnpm creates `_tmp_<pid>_<hash>` files at the workspace root while it
+ * materialises packages. Left unignored, Nx's watcher invalidates the project
+ * graph on every install, so a watch-mode `dev` target running at the time
+ * rebuilds the graph continuously and never starts the command it wraps.
  */
 export default async function migration(tree: Tree): Promise<void> {
+  // Only pnpm writes these, so key off its lockfile rather than the workspace
+  // file — a workspace on another package manager may still carry one.
+  if (!tree.exists('pnpm-lock.yaml')) {
+    return;
+  }
+
   updateGitIgnore(tree, '.', (patterns) =>
     patterns.includes(TEMP_FILE_PATTERN)
       ? patterns
--- a/packages/nx-plugin/src/migrations/latest/ignore-package-manager-temp-files/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/ignore-package-manager-temp-files/migration.ts
@@ -1,0 +1,30 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Tree } from '@nx/devkit';
+import { formatFilesInSubtree } from '../../../utils/format.js';
+import { updateGitIgnore } from '../../../utils/git.js';
+
+/** The temporary files pnpm writes while materialising packages. */
+const TEMP_FILE_PATTERN = '_tmp_*';
+
+/**
+ * pnpm writes `_tmp_<pid>_<hash>` files at the workspace root while it
+ * materialises packages into the virtual store.
+ *
+ * Nx watches the workspace, and an unignored file appearing and vanishing
+ * invalidates the project graph. So a watch-mode `dev` target running while any
+ * install happens — the `dev` cascade installing a dependency, or a generator
+ * run in another terminal — rebuilds the graph continuously and never starts
+ * the command it wraps.
+ */
+export default async function migration(tree: Tree): Promise<void> {
+  updateGitIgnore(tree, '.', (patterns) =>
+    patterns.includes(TEMP_FILE_PATTERN)
+      ? patterns
+      : [...patterns, TEMP_FILE_PATTERN],
+  );
+
+  await formatFilesInSubtree(tree);
+}
--- a/packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
+++ b/packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
@@ -33,7 +33,16 @@ import {
   withVersions,
 } from '../versions.js';
 
-/** TypeScript dependencies a caller must declare to emit agent-connection clients. */
+/**
+ * TypeScript dependencies a caller must declare to emit agent-connection clients.
+ *
+ * `@aws-sdk/client-s3` is an optional peer of `@strands-agents/sdk` that this
+ * project's own code never imports. It is declared anyway so the catalog pins the
+ * version: left undeclared, pnpm auto-installs its own copy here while a project
+ * that does declare it gets the catalog's, and the two resolutions give
+ * `@strands-agents/sdk` two peer-suffixed instances of the same version. Anything
+ * importing both projects then sees two nominal identities of every SDK type.
+ */
 export const AGENT_CONNECTION_DEPENDENCIES = [
   { name: '@aws-lambda-powertools/parameters' },
   { name: '@aws-sdk/credential-providers' },
@@ -41,6 +50,7 @@ export const AGENT_CONNECTION_DEPENDENCIES = [
   { name: '@modelcontextprotocol/sdk' },
   { name: '@a2a-js/sdk' },
   { name: '@strands-agents/sdk' },
+  { name: '@aws-sdk/client-s3' },
 ] as const satisfies readonly { name: ITsDepVersion }[];
 
 /** Python dependencies a caller must declare to emit agent-connection clients. */
@@ -310,10 +320,14 @@ const TS_TEMPLATE_DEPS: Record<
     '@modelcontextprotocol/sdk',
   ],
   'core-a2a': ['@aws-sdk/credential-providers', '@a2a-js/sdk'],
-  'core-strands/base': ['@strands-agents/sdk'],
-  'core-strands/mcp': ['@strands-agents/sdk'],
-  'core-strands/gateway': ['@strands-agents/sdk'],
-  'core-strands/a2a': ['@strands-agents/sdk'],
+  // `@aws-sdk/client-s3` rides along with every strands client: it is an
+  // optional peer of `@strands-agents/sdk` that this project never imports, and
+  // declaring it keeps its version on the catalog (see
+  // AGENT_CONNECTION_DEPENDENCIES).
+  'core-strands/base': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
+  'core-strands/mcp': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
+  'core-strands/gateway': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
+  'core-strands/a2a': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
 };
 
 const emitTs = (
--- a/packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
+++ b/packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
@@ -33,16 +33,7 @@ import {
   withVersions,
 } from '../versions.js';
 
-/**
- * TypeScript dependencies a caller must declare to emit agent-connection clients.
- *
- * `@aws-sdk/client-s3` is an optional peer of `@strands-agents/sdk` that this
- * project's own code never imports. It is declared anyway so the catalog pins the
- * version: left undeclared, pnpm auto-installs its own copy here while a project
- * that does declare it gets the catalog's, and the two resolutions give
- * `@strands-agents/sdk` two peer-suffixed instances of the same version. Anything
- * importing both projects then sees two nominal identities of every SDK type.
- */
+/** TypeScript dependencies a caller must declare to emit agent-connection clients. */
 export const AGENT_CONNECTION_DEPENDENCIES = [
   { name: '@aws-lambda-powertools/parameters' },
   { name: '@aws-sdk/credential-providers' },
@@ -320,10 +311,6 @@ const TS_TEMPLATE_DEPS: Record<
     '@modelcontextprotocol/sdk',
   ],
   'core-a2a': ['@aws-sdk/credential-providers', '@a2a-js/sdk'],
-  // `@aws-sdk/client-s3` rides along with every strands client: it is an
-  // optional peer of `@strands-agents/sdk` that this project never imports, and
-  // declaring it keeps its version on the catalog (see
-  // AGENT_CONNECTION_DEPENDENCIES).
   'core-strands/base': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
   'core-strands/mcp': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
   'core-strands/gateway': ['@strands-agents/sdk', '@aws-sdk/client-s3'],
--- a/packages/nx-plugin/src/utils/init.ts
+++ b/packages/nx-plugin/src/utils/init.ts
@@ -62,10 +62,11 @@ export const INIT_DEPENDENCIES = [
 ] as const satisfies readonly { name: ITsDepVersion }[];
 
 // Built dependencies whose install scripts the generated workspace trusts.
-// `onlyBuiltDependencies` is the pnpm 10 key (silently ignored by pnpm 11);
-// pnpm 11 reads `allowBuilds` instead. Any dep NOT in this allowlist will
-// have its install scripts skipped with a warning — matching pnpm 10's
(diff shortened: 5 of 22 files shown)


.github/actions/init-monorepo/action.yml
```diff
@@ -75,8 +75,14 @@ runs:
       uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
       with:
         registry: public.ecr.aws
+    # pnpm/setup rather than pnpm/action-setup: pnpm 11+ ships a native binary,
+    # which action-setup installs as an npm package without running its build
+    # script, leaving a placeholder in place of the executable.
     - name: Use PNPM
```
cogwirrel: No transient comments - readers of the new code won't know or care about pnpm/action-setup. remove this comment altogether

cogwirrel (dismissed):

.github/actions/init-monorepo/action.yml
nx-plugin-for-aws: Removed. The remaining comment on `install: 'false'` describes why it's set rather than what it replaced, so I've left that one.

cogwirrel (dismissed):

nx-plugin-for-aws:
Picked this up and drove it as far as I can verify. Summary of where it stands:

**Fixed, with the hacks removed**

- `strictDepBuilds=false` in `license-check` is gone. It now records an explicit `allowBuilds: <pkg>: false` — the same "reviewed, declined" decision the generators themselves write — so the gate stays on. Verified both ways: remove one decision and the install fails `ERR_PNPM_IGNORED_BUILDS` (exit 1); restore it and it passes.
- The blanket `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0` is off the two CI workflows (whose other jobs install from public npm) and set on the e2e process instead, which every spawned command inherits. Note `.npmrc` is not an option here — pnpm 12 reads only auth/registry settings from it, so `minimum-release-age` there is silently ignored.
- Nothing is written into a generated workspace except the `.gitignore` entry below, and no check is disabled in one.
- Dropped the transient `pnpm/action-setup` comment.

**Root-caused the two dev-server failures**

`dungeon-adventure` and `local-dev` were failing on every run, and it was not flake. pnpm writes `_tmp_<pid>_<hash>` files at the workspace root while materialising packages; Nx watches the workspace, so each one invalidates the project graph. A watch-mode `dev` target running while any install happens rebuilds the graph forever and never starts its command — no error, just a port that never binds. The thousands of `Creating project graph nodes` lines are `DelayedSpinner` reprints in a non-TTY, i.e. repeated graph construction.

Ignoring `_tmp_*` takes the rebuild count from 1027 to 0 with the port bound in 4s; removing the ignore restores the hang. Added for new workspaces, with a migration for existing ones, plus this repo's own `.gitignore`. Both jobs now pass.

**Still red, and I'd treat it as a blocker: `migrate (5/5)`**

Same failure on three consecutive runs, passes on `main`. Under pnpm 12 the migrated `ts-project` resolves **two peer-suffixed copies of the same `@strands-agents/sdk@1.15.0`**, so TypeScript sees two nominal identities and reports 20 errors ("Types have separate declarations of a private property `_clientName`"). The two suffixes differ only past the truncation point — transitive optional peers, of which the SDK declares 22.

I could not reduce it: from-scratch pnpm 12 install, pnpm 12 over a pnpm 11 lockfile, a catalog bump, divergent peer versions across projects, the dep declared at both root and a workspace package, and `pnpm dedupe` all give a single linked copy. Stale virtual-store dirs do appear after a version bump, but pnpm 11 leaves them too and nothing links them. It seems to need the full ~35-project matrix the hop scaffolds.

Worth flagging why this matters beyond CI: generated workspaces pin no `packageManager`, so a user already on pnpm 12 gets the same resolution. I did not touch `a2a-connection`'s `root: true` placement — it's the one generator declaring `@strands-agents/sdk` at the root rather than on the project, which is a plausible lead, but I couldn't show it causes this and changing shipped generator behaviour on an unverified hunch seemed worse than reporting it.

Everything else is green: 64 jobs pass. `@aws/nx-plugin:test` passes 4219 tests, and lint is clean.

nx-plugin-for-aws:
Root-caused `migrate (5/5)`, and it turned out to be a latent bug rather than anything pnpm 12 introduced.

`ts-project` was resolving **two peer-suffixed instances of the same `@strands-agents/sdk@1.15.0`**, so TypeScript saw two nominal identities of `McpClient`/`Agent`/`Plugin` — hence the 20 "separate declarations of a private property `_clientName`" errors.

The reason: `@aws-sdk/client-s3` is an optional peer of `@strands-agents/sdk` that the shared `agent-connection` project's own code never imports, so it was never declared. pnpm auto-installs its own copy there, while a project that *does* declare it gets the catalog's — `3.1119.0` vs `3.1121.0`. Those two resolutions split the SDK, and anything importing both projects gets both copies.

How I pinned it down: scaffolded `internal#test-matrix` on rc.90 (31 projects) and bumped the catalog the way a migration does. That produced the two variants CI reports **hash for hash** (`_1e213acde…` / `_b671c015…`), and diffing the two virtual-store directories showed `@aws-sdk/client-s3` as the single difference between them. Declaring the peer collapses both projects onto one instance — verified end to end on a generated workspace under pnpm 12.

Worth noting it isn't pnpm-12-specific: a from-scratch install happens to converge on one version, which is why it only shows up on the incremental upgrade path the migrate hop exercises. Since generated workspaces pin no `packageManager`, a user on pnpm 12 could hit it too.

Fixed by declaring the peer in the agent-connection project (new workspaces) plus a migration (existing ones). I attached it to the strands templates specifically, so it only lands where a strands client actually does, and the migration no-ops when there's no strands client.

I'd have preferred to leave `a2a-connection`'s `root: true` placement out of it entirely, and did — it was a red herring.

nx-plugin-for-aws:
Status after the peer fix: **all three original failures are green**, confirmed across consecutive runs.

- `dungeon-adventure` — passing (was failing every run)
- `local-dev` — the graph thrash is gone: 0 spinner reprints, dev servers bind, 21/22 tests pass
- `migrate (5/5)` — passing on two consecutive runs, along with all four other migrate shards

Two unrelated flakes have each surfaced once and passed on the next run:

- `standalone (4/6)` — `failed to export requirements txt with exit code 2` from a concurrent `uv export` in the Python build. Passed on the following run.
- `local-dev` — `Recursive task invocation detected: postgres-db:pull-image`. The `wait-for-db -> dev -> pull-image` chain came in with #1165, so this predates the branch; it's a race between `wait-for-db` and a `dev` already holding `pull-image`, not a cycle in the config. Worth a separate look — happy to raise an issue rather than fold it in here.

Currently rerunning to confirm that second one is transient. Nothing else outstanding: 64 jobs pass, `@aws/nx-plugin:test` passes 4223 tests across 237 files, lint clean.

.github/actions/init-monorepo/action.yml
nx-plugin-for-aws: Removed — that one and the equivalent in `on-release.yml`. The `install: 'false'` line stands on its own.

nx-plugin-for-aws:
Green — 65/65 jobs, run conclusion `success`, with no retries or reruns in the mix.

Both flakes I'd flagged earlier passed on this run (`standalone (4/6)` and `local-dev`), so neither was related to the branch. The `postgres-db:pull-image` recursive-invocation race is still worth its own look — the `wait-for-db -> dev -> pull-image` chain arrived with #1165 — but it's out of scope here and I've left it alone rather than folding an unrelated fix in.

Also dropped the two `pnpm/setup` comments outright per the review, rather than rewording them.

Recap of what this ended up being, beyond the pnpm bump itself:

1. **`_tmp_*` graph thrash** — pnpm writes temp files at the workspace root while materialising packages; Nx watches them and rebuilds the graph forever, so watch-mode `dev` targets never start. Fixed for new workspaces + a migration. This was the `dungeon-adventure`/`local-dev` failure.
2. **Undeclared optional peer** — `@aws-sdk/client-s3` is a `@strands-agents/sdk` peer the shared `agent-connection` project never imports, so it went undeclared and resolved to a different version than the project that does declare it. That split the SDK into two peer-suffixed copies of one version, and TypeScript saw two nominal identities of every type. Fixed for new workspaces + a migration. This was `migrate (5/5)`, and it was latent rather than pnpm-12-specific.
3. **No hacks left** — `strictDepBuilds` stays on (explicit per-package `allowBuilds: false` decisions instead), and the verdaccio release-age override lives on the e2e process rather than the CI workflows or any generated workspace.

.gitignore
```diff
@@ -43,4 +43,7 @@ Thumbs.db
 .docusaurus/
 .cache-loader/
 vite.config.mts.timestamp-*
-.grit
\ No newline at end of file
+.grit
+# Written at the workspace root while pnpm materialises packages. Left unignored
```
cogwirrel: Remove comment (and in users workspaces if applicable)

packages/nx-plugin/src/utils/init.ts
```diff
@@ -166,6 +168,13 @@ const setUpPnpmWorkspace = (tree: Tree, catalogs: boolean) => {
  * manager reads the `workspaces` field of the root `package.json`.
  */
 const setUpWorkspaces = (tree: Tree, catalogs: boolean) => {
+  // pnpm writes `_tmp_<pid>_<hash>` files at the workspace root while it
```
cogwirrel: Too verbose, 1 line saying pnpm creates these while it materialises packages

packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
```diff
@@ -33,14 +33,24 @@ import {
   withVersions,
 } from '../versions.js';
 
-/** TypeScript dependencies a caller must declare to emit agent-connection clients. */
+/**
```
cogwirrel: Revert comment change

packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
```diff
@@ -310,10 +320,14 @@ const TS_TEMPLATE_DEPS: Record<
     '@modelcontextprotocol/sdk',
   ],
   'core-a2a': ['@aws-sdk/credential-providers', '@a2a-js/sdk'],
-  'core-strands/base': ['@strands-agents/sdk'],
-  'core-strands/mcp': ['@strands-agents/sdk'],
-  'core-strands/gateway': ['@strands-agents/sdk'],
-  'core-strands/a2a': ['@strands-agents/sdk'],
+  // `@aws-sdk/client-s3` rides along with every strands client: it is an
```
cogwirrel: Remove comment

packages/nx-plugin/src/migrations/latest/ignore-package-manager-temp-files/migration.ts
```diff
@@ -0,0 +1,30 @@
...
+ * materialises packages into the virtual store.
+ *
+ * Nx watches the workspace, and an unignored file appearing and vanishing
+ * invalidates the project graph. So a watch-mode `dev` target running while any
+ * install happens — the `dev` cascade installing a dependency, or a generator
+ * run in another terminal — rebuilds the graph continuously and never starts
+ * the command it wraps.
+ */
+export default async function migration(tree: Tree): Promise<void> {
+  updateGitIgnore(tree, '.', (patterns) =>
```
cogwirrel, line 25: Only required for pnpm (check for pnpm lock)

packages/nx-plugin/src/utils/init.ts
```diff
@@ -166,6 +168,13 @@ const setUpPnpmWorkspace = (tree: Tree, catalogs: boolean) => {
...
+  // materialises packages. Nx watches the workspace, and an unignored file
+  // appearing and vanishing invalidates the project graph — so a watch-mode
+  // `dev` target running while any install happens would otherwise rebuild the
+  // graph continuously and never start its command.
+  updateGitIgnore(tree, '.', (patterns) => [...patterns, '_tmp_*']);
```
cogwirrel: Only required for pnpm

.gitignore
nx-plugin-for-aws: Removed. Generated workspaces never had one — the `updateGitIgnore` helper only appends the pattern, so there was nothing to strip there.

packages/nx-plugin/src/utils/init.ts
nx-plugin-for-aws: Cut to one line: `// pnpm creates these at the workspace root while it materialises packages.`

packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
nx-plugin-for-aws: Reverted to the original one-line doc comment.

packages/nx-plugin/src/utils/agent-connection/agent-connection.ts
nx-plugin-for-aws: Removed.

packages/nx-plugin/src/migrations/latest/ignore-package-manager-temp-files/migration.ts
nx-plugin-for-aws: Gated on `pnpm-lock.yaml` and added a test for the non-pnpm case. Used the lockfile rather than `pnpm-workspace.yaml` per your wording — a workspace on another package manager can still carry the workspace file, so the lockfile is the signal that pnpm is actually in use.

packages/nx-plugin/src/utils/init.ts
nx-plugin-for-aws: Moved inside the `detectWorkspacePackageManager(tree) === 'pnpm'` branch, next to `setUpPnpmWorkspace`.

No negative test here, unlike the migration: with no `pnpm-workspace.yaml` in the tree, `detectWorkspacePackageManager` falls through to `detectPackageManager(tree.root)` — the real filesystem, i.e. this repo, which is pnpm — so the assertion can't be isolated. The migration's gate reads a tree file, so that case is covered in its spec.

cogwirrel (approved):

PR #1180, opened 2026-09-02
fix(ts#project): exclude rolldown's transient config bundle from format and lint

nx-plugin-for-aws:
### Reason for this change

CI has been failing intermittently on `format` targets with a diff against a file nobody wrote, eg in [this run](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33601818715/job/100159421323?pr=1177):

```
@trpc/trpc-iam-rest-shared: packages/trpc-iam-rest-shared/rolldown.config.OC4dhQiW.js format ━━━
@trpc/trpc-iam-rest-shared:   × Formatter would have printed the following content:
@trpc/trpc-iam-rest-shared:      1    │ - import·"node:module";
@trpc/trpc-iam-rest-shared:      2    │ - import·{·defineConfig·}·from·"rolldown";
@trpc/trpc-iam-rest-shared:      3    │ - var·rolldown_config_default·=·defineConfig([{
```

Rolldown has no loader registered for TypeScript configs, so it bundles one to load it — `bundleTsConfig` in rolldown's `load-config.ts` writes `rolldown.config.[hash].js` into the config's **own directory**, `import`s it, then unlinks it in a `finally`:

```js
entryFileNames: `rolldown.config.[hash]${path.extname(configFile).replace('ts', 'js')}`
```

A project's `bundle`, `format` and `lint` targets have no dependency edge between them, so Nx runs them concurrently. For as long as a `bundle` takes, that machine-generated file sits in the project — and a `format` sweeping the same project sees it, reports it unformatted, and fails the target. Nothing about the project's own sources is wrong, which is why a re-run goes green.

The file is also git-visible for that window, so it can be committed if `bundle` is interrupted, and `license#sync` — which builds its candidate set from git-visible files rather than from `biome.json` — would stamp a copyright header onto it.

Not specific to one PR. Same failure across recent runs on unrelated branches:

| Run | Job | Branch |
| --- | --- | --- |
| [33601818715](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33601818715) | Smoke Tests | `perf/agent-core-vendor-target` (#1177) |
| [33493107324](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33493107324) | Smoke Tests - trpc-api, local-dev | `feat/tf-shared-ecr` |
| [33493101423](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33493101423) | Smoke Tests - local-dev | `main` (CI) |

### Description of changes

- **`utils/format.ts`** — the vended `biome.json` excludes `**/rolldown.config.*.*js`, placed next to the other build-output excludes. `.*js` rather than `.js` because the extension follows the config's own: a `.mts` config bundles to `.mjs`, a `.cts` one to `.cjs`.
- **`utils/bundle/bundle.ts`** — `addTypeScriptBundleTarget` adds the same glob to the project's `.gitignore`, so the transient file can neither be committed nor reach `license#sync`.
- **`migrations/latest/rolldown-config-bundle-transient`** — applies both to existing workspaces. The Biome edit is anchored to the excludes we vend; a workspace whose `files.includes` has diverged keeps its own list and is told via `nextSteps` instead. The git ignore only goes to projects that actually carry a `rolldown.config.ts`.
- **Docs** — the manual bundle setup in `typescript-project.mdx` now mentions the ignore.

The repo's own `rolldown.config.mjs` files are unaffected — a JS config is imported directly and never bundled — so the root `biome.json` needs no change.

**Reproduced the race in a standalone workspace** (rolldown 1.2.6, biome 2.5.11), racing `rolldown -c rolldown.config.ts` against `biome format` the way Nx does. With the pre-fix excludes, `format` failed on `rolldown.config.<hash>.js`. With the exclude added, 60 rounds produced **0** failures, while `rolldown.config.ts` itself and other sources were still checked and still reported genuine formatting errors — the exclude is narrow enough not to mask real ones.

**Unit tests** — 8 new migration tests (Biome exclude placement, git ignore scoped to projects with a config, no duplicate entries, diverged-config reporting, no-biome-config, idempotency) and 2 for the generator's git ignore.

**Full suite green** — `pnpm nx run @aws/nx-plugin:test` 4224 passed / 236 files; `pnpm nx run-many --target build --all` and `lint --all` both pass. 4 snapshots updated for the new exclude.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -302,6 +302,14 @@ export default defineConfig([
 Note that in the above target, we chose `src/index.ts` as our entrypoint for the bundle, which means that code exported from this file will be included in the bundle, along with all of its dependencies.
 :::
 
+Finally, add the following to your project's `.gitignore`:
```
cogwirrel: Remove docs changes (including translations )

nx-plugin-for-aws: Done — dropped the docs change, and the translations the bot had pushed for it went with it. The PR is now code + migration + tests only.

cogwirrel (approved):

PR #1181, opened 2026-09-02
feat(shadcn): adopt shadcn 4.0's monorepo support

AlexTo:
### Reason for this change

This repo picked up shadcn 4.x in `52308dde` (2026-03-27, "feat: update dependencies (#513)", `3.8.5` → `4.1.0`) and is currently vending `4.19.0`. shadcn 4.x added monorepo support to the CLI, and it breaks `pnpm dlx shadcn add` for this plugin's generated workspaces: the CLI now resolves an alias's tsconfig-declared path relative to whichever directory it's invoked from rather than where the alias was declared. That doubles the path whether you run it with `-c packages/common/shadcn` from the workspace root (as its own monorepo preflight tells you to), or `cd` into `packages/common/shadcn` and run `shadcn add dialog` directly — both write new components under `packages/common/shadcn/packages/common/shadcn/...` instead of `packages/common/shadcn/src/...`.

### Description of changes

shadcn 4.0's [documented monorepo pattern](https://ui.shadcn.com/docs/monorepo) avoids tsconfig aliases entirely — `components.json`'s aliases become local `#...` package imports (resolved via `package.json#imports`), and other workspaces resolve the package through `package.json#exports`. This PR moves the shared shadcn package onto that shape:

- `shared-shadcn.ts` now vends `imports`/`exports` on the shared package's `package.json` instead of a `tsconfig.base.json` `paths` entry, and every `ux=shadcn` consumer (`ts#website`, the `connection` generator's shadcn theme) declares a real `workspace:*` dependency on it so `exports` actually resolves.
- `components.json.template` and all vendored `.tsx.template` component files now self-reference via local `#...` specifiers instead of the public `<scope>/common-shadcn/...` alias.
- `scripts/update-versions/shadcn.ts` (the weekly dependency-refresh script) configures its scratch scaffold with the same `#...` aliases, so future refreshes keep emitting the correct shape without a separate rewrite step.
- A new deterministic migration (`shadcn-package-imports-exports`) carries an existing workspace onto this shape: rewrites `components.json` and every internal component/lib/hooks import (discovered from files actually present, so a component a user added by hand is covered too), adds `imports`/`exports` to the package's manifest, drops the now-dead `paths` entry, and adds the `workspace:*` dependency to each consumer. Guarded and idempotent - a customised `components.json` or import form is left alone and reported via `nextSteps` rather than rewritten.

- `migration.spec.ts` (7 tests): applies to the vended shape, keeps user-added bindings, skips and reports a customised import form or a customised `components.json`, no-ops on an already-migrated workspace, is idempotent, and no-ops on a workspace with no shared shadcn package.
- Existing generator suite (`ts#website`, `connection`/agui, `ts#website#auth`) still green — 123 tests, unchanged.
- Manual end-to-end validation in two scratch workspaces:
  - **Migration path**: generated a workspace on the published `1.0.0-rc.94`, reproduced the doubled-path bug for real (`shadcn add dialog -c packages/common/shadcn`), linked the local build, hand-ran the migration, and confirmed: the diff matches exactly what's described above, re-running the migration reports "No changes were made", `nx sync` + `pnpm install` produces a real `node_modules` symlink for the shared package, `nx run-many --target compile --all` is clean, a real `nx run website:build` (Vite production build) succeeds, and `shadcn add dialog -c packages/common/shadcn` now lands at the correct path with correctly-aliased imports.
  - **Generator path**: generated a second workspace from scratch against the local build - its `components.json` and `package.json` imports/exports are byte-identical to the migrated workspace's, it compiles cleanly, and `shadcn add -c packages/common/shadcn` from the workspace root, and `shadcn add` run directly from inside `packages/common/shadcn`, both work correctly.

### Issue # (if applicable)

N/A - found via manual testing, no tracked issue.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/migrations/latest/shadcn-package-imports-exports/metadata.json
+++ b/packages/nx-plugin/src/migrations/latest/shadcn-package-imports-exports/metadata.json
@@ -1,0 +1,3 @@
+{
+  "description": "Move the shared shadcn package from a tsconfig paths alias to package.json imports/exports, matching the shadcn CLI's own monorepo resolution"
+}
--- a/packages/nx-plugin/src/migrations/latest/shadcn-package-imports-exports/migration.spec.ts
+++ b/packages/nx-plugin/src/migrations/latest/shadcn-package-imports-exports/migration.spec.ts
@@ -1,0 +1,275 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { addProjectConfiguration, readJson, type Tree } from '@nx/devkit';
+import { REACT_WEBSITE_APP_GENERATOR_INFO } from '../../../ts/react-website/app/generator.js';
+import { createTreeUsingTsSolutionSetup } from '../../../utils/test.js';
+import migration from './migration.js';
+
+const SHADCN_ROOT = 'packages/common/shadcn';
+const COMPONENTS_JSON = `${SHADCN_ROOT}/components.json`;
+const PACKAGE_JSON = `${SHADCN_ROOT}/package.json`;
+const BUTTON = `${SHADCN_ROOT}/src/components/ui/button.tsx`;
+const SIDEBAR = `${SHADCN_ROOT}/src/components/ui/sidebar.tsx`;
+const UTILS = `${SHADCN_ROOT}/src/lib/utils.ts`;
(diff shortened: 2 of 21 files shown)


cogwirrel (approved):
Thanks @AlexTo ! Great catch! I'll add an e2e test to make sure this gets caught if it happens again! :)

AlexTo:
> Thanks @AlexTo ! Great catch! I'll add an e2e test to make sure this gets caught if it happens again! :)

make sense, for a few months I didn't need to add any new Shadcn components so everything was fine until yesterday :D

PR #1183, opened 2026-09-02
feat(agentcore-harness#trpc-connection): stream AG-UI over a regional REST API for all auth modes

waltmayf:
### Reason for this change

An AgentCore Harness chatbot fronted by this plugin's tRPC API truncates long streaming responses. A REST API is **edge-optimized by default**, so even with `ResponseTransferMode.STREAM` the CloudFront layer fronting an edge-optimized endpoint severs the connection after ~30s of idle time. A long generation — e.g. "make a table of the best 100 plants to grow in Austin TX" — streams partway (~row 80) and then the in-flight message disappears; it only reappears on refresh (from durable Memory history). The idle timeout, not the model, is cutting the stream.

### Description of changes

This adds a new **server-side** connection generator, `ts#trpc-api → agentcore-harness`, that turns a tRPC API into a standard AG-UI endpoint. It touches **only** the new connection — no existing connector (`website→trpc`, `react→agui`) is modified, and a plain tRPC API with no connected Harness gains no AG-UI-specific members.

The connection adds a streaming `POST /agui` route plus a `history` tRPC query to the API:

- **Regional REST streaming for all auth modes** (IAM / Cognito / none): `endpointConfiguration: { types: [EndpointType.REGIONAL] }` + `ResponseTransferMode.STREAM` (5-min idle / 15-min max). A regional endpoint has no CloudFront layer, so the stream is held for the full window instead of being cut at ~30s. The handler emits an SSE `: keepalive` comment every 15s to stay inside the 5-min idle limit.
- The `/agui` route sits under the API's root resource, so it inherits the API's existing authorizer (IAM or Cognito) and CORS automatically. Authentication terminates at the API; the Harness is invoked with the Lambda's own execution role and the caller's token is never forwarded. Per-user scoping is enforced by the handler (a server-derived `actorId`), never from the request body.
- The AG-UI handler and `history` resolve the harness/memory ARNs **only** from runtime config (`agentcore.harnesses.<name>` → `{ arn, memoryArn }`); there are no dedicated env vars. Both the CDK harness construct and the Terraform harness module publish `{ arn, memoryArn }` under that key. Memory-read (`ListEvents`) is granted to the tRPC handlers **unconditionally** — mirroring the CDK construct — so the `history` procedure can reconstruct a conversation; `agui_harness_memory_arn` is therefore a **required** Terraform input. (Gating the grant's `count`/`for_each` on `agui_harness_memory_arn != null` fails `terraform plan`, since a Harness's managed-memory ARN is only known after apply.)
- The handler is bundled with **rolldown** via the owning API project's own nx-cached `bundle` target (the same target the API's router handler uses), not CDK's `NodejsFunction`; both the CDK and Terraform paths package the same output.
- **Terraform** support via the connection: the `/agui` Lambda + API Gateway resource/method/integration (`response_transfer_mode = "STREAM"`) and the Memory-read grant are appended to the generated module, wired from the harness module's `harness_arn` / `memory_arn` outputs. The appended `/agui` method + integration are also wired into the module's existing `aws_api_gateway_deployment` (both the flat proxy trigger list and the per-operation `concat(...)` form) so a fresh `terraform apply` redeploys the stage with the route present — otherwise the route is created but not served until an unrelated change forces a redeployment.
- **The website's chat wiring is documented, not generated.** The connection is deliberately server-side only: because `/agui` speaks standard AG-UI over SSE, any AG-UI client (a stock CopilotKit `<CopilotChat>` via an `@ag-ui/client` `HttpAgent`, or a bring-your-own frontend) connects by pointing at the deployed API's `/agui` URL and authenticating exactly as it would for any other request to the API. The guide walks through this end to end.

- Generator unit + snapshot tests for the new connection pass (31 tests), including: IAM/Cognito/none all stream over the regional REST endpoint (`endpointConfiguration: { types: [EndpointType.REGIONAL] }`, `ResponseTransferMode.STREAM`) and assert **no** `FunctionUrl` / `InvokeMode` / `aws4fetch` remain; the Terraform path emits `response_transfer_mode = "STREAM"` and wires the `/agui` method + integration into the API Gateway deployment's redeployment trigger and `depends_on` (asserted against **both** the flat proxy and `concat(...)` deployment forms); import-presence assertions on the CDK construct patch (`EndpointType` / `Grant` / harness-construct imports); idempotency (a diffed second run against both CDK and Terraform APIs is a no-op); and a generated-output snapshot.
- Confirmed the shared connectors (`website→trpc`, `react→agui`, `useSigV4`) are byte-for-byte identical to `main` — this PR adds only the new connection.
- `pnpm nx run-many --target build --all` and lint: green.
- Validated **end-to-end against a deployed workspace** (us-east-1): a large response streamed **to completion at ~30.3s** (`elapsedMs=30294`, `finalTextLen=5883`) over the regional REST `/agui` route — past the edge/CloudFront ~30s cutoff, which is the point of the change — through the full Cognito + TOTP-MFA login. `/agui` returned `200`/`204` (no 5xx), and a `history` read reconstructed the full transcript (text plus tool-call/tool-result linkage). Separately confirmed AgentCore Memory `ListEvents` accepts both a bare memoryId and a full ARN.
  - The deploy caught a real generator bug (fixed here): the api-construct patch dropped the `EndpointType` / `Grant` / harness-construct imports it added, so the generated construct did not compile. Added import-presence spec assertions + refreshed snapshot.

### Issue # (if applicable)

N/A

### Checklist

----

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/src/content/docs/en/guides/connection/trpc-agentcore-harness.mdx
```diff
@@ -0,0 +1,595 @@
...
+import GeneratorParameters from '@components/generator-parameters.astro';
+import NxCommands from '@components/nx-commands.astro';
+import CreateNxWorkspaceCommand from '@components/create-nx-workspace-command.astro';
+
+The `connection` generator wires a <Link path="guides/trpc">tRPC API</Link> to an <Link path="guides/agentcore-harness">AgentCore Harness</Link>, adding a streaming `POST /agui` route to the API that authenticates the caller, invokes the Harness on their behalf, and translates the response into a standard [AG-UI](https://docs.ag-ui.com/) server-sent-event stream. Because the translation happens server-side, the API becomes a drop-in AG-UI endpoint — the generated CopilotKit chat UI connects with a stock `@ag-ui/client` `HttpAgent`, and so can any AG-UI client.
+
+A Harness is never exposed to the browser directly. Per [AWS's guidance](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness-security.html), an untrusted caller should only reach it through a trusted application layer. The generated `/agui` route is that layer: it reads only `messages` and `threadId` from the request and pins the system prompt, model, tools and actor id server-side.
```
cogwirrel, line 17: Nit: we're AWS :) maybe `Per [the documentation](...),`

docs/src/content/docs/en/guides/connection/trpc-agentcore-harness.mdx
```diff
@@ -0,0 +1,595 @@
...
+- IAM grants (plus the same env vars) on every existing tRPC operation handler so the `history` procedure can read Memory independently of the `/agui` Lambda.
+
+:::caution[Streaming duration limits]
+The `/agui` integration uses `ResponseTransferMode.STREAM`, so API Gateway's 29-second integration timeout does **not** apply — the SSE stream runs until the Lambda's 180-second timeout. The REST API is edge-optimized, which would drop the connection after a 30-second idle gap, so the handler emits an SSE keepalive comment every 15 seconds. A turn that outruns the Lambda timeout keeps running server-side, and the client falls back to `history` to read the rest. To keep longer turns streaming live, raise the `AguiHandler` timeout (up to API Gateway's 15-minute ceiling) in your `Api` construct; `history`/Memory remains the durable read path either way.
+:::
+
+Only CDK is wired by this generator. If your API's `iac` is `terraform`, the AG-UI application code is still generated, but you'll need to wire the Lambda, the API Gateway route, and the Harness invoke grant into your Terraform configuration by hand.
+
+## Deploying the full application
```
cogwirrel, line 97: This is a bit too much of an end to end tutorial - would be great if we could generally make the guild page consistent with the style of the other ones :)

packages/nx-plugin/src/agentcore-harness/trpc-connection/generator.ts
```diff
@@ -0,0 +1,166 @@
...
+ * endpoint any `@ag-ui/client` (or bring-your-own AG-UI client) can connect
+ * to by URL alone. Also generates an optional `history` tRPC procedure that
+ * reconstructs a conversation from AgentCore Memory.
+ *
+ * The Harness itself is never exposed to the browser: `InvokeHarness`
+ * accepts per-invocation overrides for the system prompt, model, tools and
+ * actor id, so a trusted server-side caller (this API) is required per the
+ * AgentCore Harness security guidance.
+ */
+export const trpcAgentCoreHarnessConnectionGenerator = async (
```
cogwirrel, line 60: I think the generator should be for a `website -> agentcore-harness` connection, rather than `agentcore-harness -> tRPC` - I think it's ok for that to generate a tRPC api of its own (ideally updating the harness project to be a typescript project itself with src/api containing a the trpc api application code for example), and it vends a construct in common/constructs which already includes the agui endpoint etc so there's no `addAguiRoute` needed?

Then we also don't need to modify the website -> trpc connection at all, but we can potentially share code from it.

So I think from a user perspective the rough experience would be:

Run the connection generator with `source: website` and `target: harness`, and it gives you:

- a trpc api in the harness project, plus the agui route (or noop's this part if it already exists - since multiple sites might connect to the same harness and can reuse the trpc api)
- a trpc api construct in common/constructs - happy for the existing rest api to be reused or for this to be a lambda function url if this works better
- a type-safe client/hooks for invoking the trpc api (what the website -> trpc connection generator vends)
- a copilotkit client for invoking the agui route (what the website -> agui generator vends)

Then instantiate the harness trpc api, eg:

```ts
const harnessApi = new MyHarnessApi(this, 'MyHarnessApi');
harnessApi.grantInvoke(identity.identityPool.authenticatedRole);
```

Then in the website you can integrate your copilot chat straight away:

```tsx
<CopilotChat agentId='my-harness' />
```

Keen to hear your thoughts! :)

packages/nx-plugin/src/connection/harness-trpc-config.ts
```diff
@@ -0,0 +1,308 @@
...
+export const addAguiRouteToApi = async (
+  tree: Tree,
+  options: HarnessTrpcConfigOptions,
+): Promise<void> => {
+  if (options.iac !== 'cdk') {
+    logger.warn(
+      `agentcore-harness#trpc-connection does not yet wire the /agui route into ${options.iac ?? 'this'} infrastructure for '${options.apiNameKebabCase}'. The AG-UI application code was still generated under src/agui — wire the Lambda, API Gateway route and Harness invoke grant manually.`,
+    );
+    return;
+  }
```
cogwirrel, line 66: We should ideally support terraform too - I think this will be easier if we vend a bespoke trpc api as part of the website -> harness connection generator since we won't need to mutate existing terraform with gritql :)

packages/nx-plugin/src/connection/harness-trpc-config.ts
```diff
@@ -0,0 +1,308 @@
...
+   * Adds an AG-UI streaming endpoint backed by ${options.harnessNameClassName},
+   * translating its Converse-style stream into AG-UI server-sent events.
+   * Only 'messages' and 'threadId' are read from the request; every other
+   * Harness field (systemPrompt, model, tools, allowedTools, skills,
+   * actorId) is pinned server-side by the generated handler.
+${endpointDoc}
+   */
+  public addAguiRoute(harness: ${options.harnessNameClassName}): void {
+    const rc = RuntimeConfig.ensure(this);
+    const aguiHandler = new NodejsFunction(this, 'AguiHandler', {
```
cogwirrel, line 212: We should avoid using `NodejsFunction` and instead bundle with rolldown as part of the application code's `bundle` target like we do with the trpc router for example :) This will mean faster builds as CDK sequentially runs esbuild every synth, whereas including it in the other project's build means it can be cached by nx and run in parallel with other tasks :)

packages/nx-plugin/src/connection/harness-trpc-config.ts
```diff
@@ -0,0 +1,308 @@
...
+      bundling: {
+        format: OutputFormat.ESM,
+        mainFields: ['module', 'main'],
+        banner:
+          "import { createRequire as __aguiCreateRequire } from 'module'; const require = __aguiCreateRequire(import.meta.url);",
+      },
+      environment: {
+        HARNESS_ARN: harness.harnessArn,
+        MEMORY_ARN: harness.harness.attrMemoryManagedMemoryConfigurationArn,
+      },
```
cogwirrel, line 226: Ideally the agui handler should only use the RUNTIME_CONFIG_APP_ID environment variable and pull the harness and memory arn from runtime config (see how eg agent -> mcp connections do this).

The `harnessArn` is already published to runtime config under `agentcore.harnesses.<name>`, but the memory arn isn't. We might therefore need to update the structure in the harness construct to be a bit more like:

```ts
    rc.set('agentcore', 'harnesses', {
      ...rc.get('agentcore').harnesses,
      <%- nameClassName %>: {
        arn: this.harness.attrArn,
        memoryArn: this.harness.attrMemoryManagedMemoryConfigurationArn,
      },
    });
```

cogwirrel (commented):
Thanks for this @waltmayf ! I have a few design comments before I dig into this in too much depth :)

waltmayf:
Thanks for the thorough steer, @cogwirrel — this reframing is better than what I had, and I want to reshape the PR around it. Here's the plan:

## Flip to a `website → agentcore-harness` connection generator

Agreed — instead of injecting AG-UI wiring into the existing website→trpc connection, the harness vends a self-contained construct. Rough shape matching your sketch:

- Scaffold a **tRPC API inside the harness project** (`src/api`), no-op if one already exists (multiple sites can reuse one harness API).
- Vend a **harness API construct in `common/constructs`** that owns both the tRPC operations and the AG-UI streaming route — so `addAguiRoute` goes away entirely and the website→trpc connection is left untouched (sharing code, not mutating it).
- Reuse the existing website→trpc **type-safe client/hooks** for the control plane and the website→agui **copilotkit client** for the stream.
- Target UX: `new MyHarnessApi(...)` + `grantInvoke(...)`, then `<CopilotChat agentId='my-harness' />`.

On the "why a tRPC API if the stream isn't tRPC" point: the stream stays a raw route (SSE is a poor tRPC fit), but the tRPC API is the home for the **non-streaming control plane** — the `history` procedure this PR already ships (conversation rehydration from AgentCore Memory), thread listing, etc. — and lets us reuse the mature hooks/client generation rather than hand-rolling fetch. The AG-UI route just rides alongside it in the same vended construct.

## Streaming transport: switch Function URL → **regional** REST API

Digging into the original bug: the truncation wasn't "API Gateway can't stream" — REST APIs stream fine via `responseTransferMode: STREAM` (already what we use). The culprit was specifically the **edge-optimized** endpoint, i.e. CloudFront's ~30s idle timeout. A **regional** REST endpoint has a **5-min idle timeout** and a **15-min max stream lifetime**, so it fixes the cutoff *without leaving API Gateway* — and keeps us on the project's standard `RestApi` construct.

The bigger reason to prefer it over the Function URL: regional REST keeps **Cognito JWT authorizers, usage plans, and SigV4 over `execute-api`**, which dissolves this PR's two biggest compromises:

- The **"IAM-only, for now"** limitation only existed because Function URLs are `AWS_IAM`/`NONE`-only. On regional REST the fix applies to **all auth modes** uniformly — Cognito/none stop being a follow-up.
- The **aws4fetch signing special-case** (signing for the `lambda` service off the `*.lambda-url` host) disappears — the hook signs for `execute-api` exactly like the normal tRPC client.

Trade-offs I'll handle: add an SSE heartbeat (`: keepalive`) under the 5-min idle window, and decide whether the streaming route rides on a regional-switched shared tRPC API or its own regional API in the vended construct (leaning toward the latter so the base tRPC API stays edge-optimized). Streaming disables API GW caching/compression/VTL, which is irrelevant for SSE.

## The other comments

- **Terraform** — agreed, and this falls out of the above: a vended bespoke harness construct means no gritql mutation of existing Terraform.
- **Drop `NodejsFunction`, bundle with rolldown** via the app's `bundle` target — agreed, will do (nx-cached, parallel, no per-synth esbuild); the vended-construct project is the natural home for that target.
- **Handler reads ARNs from runtime config** (only `RUNTIME_CONFIG_APP_ID`), matching agent→mcp — agreed, including publishing `memoryArn` alongside `harnessArn` under `agentcore.harnesses.<name>`.
- **Docs** — will fix the "we're AWS" attribution and trim the guide page to match the house style of the other connection guides.

## Scope

I'll roll it all into this PR as the new vended-construct generator — one coherent change: the `website → agentcore-harness` connection generator, the regional-REST streaming route, the runtime-config/bundling cleanups, Terraform support, and the docs pass. I'll re-request review once it's reshaped.

cogwirrel:
Thanks @waltmayf! This sounds good! :) Just one small thing to clarify:

> Drop NodejsFunction, bundle with rolldown via the app's bundle target — agreed, will do (nx-cached, parallel, no per-synth esbuild); the vended-construct project is the natural home for that target.

Just to clarify re this, the `bundle` target which runs rolldown should be in the agentcore harness project rather than the common/constructs project :)

PR #1186, opened 2026-09-03
test(e2e): cover the release publish, and move the repo back to pnpm 12

nx-plugin-for-aws:
> [!IMPORTANT]
> **The new `release` smoke test is expected to FAIL on this PR.** That failure is the test doing its job — it reproduces the exact `v1.0.0-rc.95` publish failure, which only disappears once nx 23.2 lands via #1182. See [Expected CI state](#expected-ci-state) below.

### Reason for this change

The `v1.0.0-rc.95` release tagged and cut a GitHub release, then failed to publish anything to npm:

```
pnpm publish ".../dist/packages/nx-plugin" --json --"@aws:registry=https://registry.npmjs.org/" --tag=latest --no-git-checks
error: unexpected argument '--@aws:registry' found
  tip: a similar argument exists: '--registry'
```

nx 23.1.2 gates that flag spelling on `>=9.15.7 <10.0.0 || >=10.5.0` — an unbounded upper range. pnpm 11 renamed the option to `--config.@scope:registry` and pnpm 12 dropped the old spelling, so moving the repo to pnpm 12 in #1176 broke every publish.

**Nothing in CI caught this**, and that is the real defect. The release publish had no test coverage at all:

- the other smoke tests publish with plain `npm publish` in `global-setup`, which never exercises `nx release publish`
- unit tests never invoke the release-publish executor
- the command nx builds is *derived from the active pnpm version*, so it can only be validated by running it

The first signal was a failed release on main, **after** the git tag and GitHub release had already been created. That left a `v*` tag with nothing on npm, which then broke migrate shard 1 on every subsequent run, since migrate resolves start versions from git tags.

#1185 was the stopgap that put the repo back on pnpm 11 to unblock releases. This PR reverts that and adds the missing coverage.

### Description of changes

**A `release` smoke test** (`e2e/src/smoke-tests/release.spec.ts`) that publishes the local build to verdaccio and asserts all three packages land. Added to the `smoke_tests` matrix in `ci.yml` and `pr.yml`.

It calls the release script's own `publishWithRetry` rather than re-implementing the command, so the thing under test is the thing that actually releases. `scripts/release.ts` changes to support that:

- `publishWithRetry` is exported and takes optional overrides (`tag`, `additionalArgs`, `attempts`, `cwd`); defaults are unchanged, so the release path behaves exactly as before
- `main()` is guarded behind `import.meta.url === file://${process.argv[1]}`, matching the existing pattern in `scripts/stamp-migrations.ts`, so importing the module doesn't run a release

**Three things stop the test passing vacuously** — this was the main design concern, since a release test that silently publishes nothing is worse than no test:

1. **It publishes a version the registry has never served.** nx short-circuits a package whose version already carries the requested dist-tag, exits `0`, and never runs the publish command. `global-setup` has already published `0.0.0`, so reusing that version would pass green on *any* package manager. The test publishes `9.9.9-release-e2e.<pid>` and **asserts the precondition** before publishing.
2. **It publishes under a `release-e2e` dist-tag**, so `latest` keeps pointing at the `0.0.0` build every other smoke test resolves.
3. **It queries the registry afterwards** to confirm each package is really served at that version.

**`packageManager` moves back to `pnpm@12.2.1`**, reverting #1185. `pnpm-lock.yaml` is unchanged — pnpm 11 and 12 share `lockfileVersion: '9.0'`.

### Expected CI state

The `release` smoke test **will fail on this PR**, with the same error that broke rc.95. That is intentional and is the evidence the test works — it fails precisely because the repo is back on pnpm 12 while nx is still 23.1.2.

It goes green once **#1182** (nx 23.2) lands: nx 23.2.0 bounds the old spelling to `<11.0.0` and adds a `configForm` branch emitting `--config.@scope:registry` for pnpm >= 11. Two ways to sequence this:

- merge #1182 first, then rebase this PR — it should go fully green, or
- merge this PR's test on pnpm 11 and flip to pnpm 12 with #1182

Happy to reshape into whichever you prefer; the sequencing is your call.

**Proved the test fails for the right reason and passes for the right reason**, by running the real target both ways:

`pnpm@11.25.0` — passes, with three real publishes:

```
$ pnpm nx run @aws/nx-plugin-e2e:smoke-test --name=release
Published to http://localhost:4873 with tag "release-e2e"   (x3)
✓ should publish every package with the release script
Test Files  1 passed | 33 skipped
```

`pnpm@12.2.1` — fails with the exact rc.95 error:

```
error: unexpected argument '--@aws:registry' found
× should publish every package with the release script
Error: nx release publish failed after 1 attempts
```

**Verified the anti-vacuous guard by sabotaging it.** Pointed the test at `0.0.0` (already on the registry) to simulate a publish nx would skip. Before the guard it passed green while publishing nothing — the exact false negative I was trying to avoid. With the guard:

```
AssertionError: @aws/nx-plugin@0.0.0 is already on the local registry — this test cannot prove a publish ran
```

**Verified the flag behaviour directly** against each pnpm major with a throwaway scoped package:

| pnpm | `--"@aws:registry=..."` | `--"config.@aws:registry=..."` |
|---|---|---|
| 11.22.0 | accepted | accepted |
| 11.25.0 | accepted | accepted |
| 12.2.1 | **`unexpected argument`** | accepted |

**Confirmed the CLI guard didn't break the release entrypoint** — running `scripts/release.ts` directly still executes the real flow (it resolved the next version as `1.0.0-rc.95` from the git tag and wrote the dist manifests), so the guard only suppresses execution on import.

Also confirmed `pnpm@12.2.1 install --frozen-lockfile` succeeds with no lockfile changes, biome lint is clean on both changed files, and `@aws/nx-plugin-e2e:build` passes. The pre-existing `tsc -p e2e/tsconfig.spec.json` errors are unchanged by this PR (identical count on clean main) and none are in the files touched here.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,7 @@ jobs:
           - audit
           - finch
           - trivy
+          - release
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -127,6 +127,7 @@ jobs:
           - audit
           - finch
           - trivy
+          - release
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
--- a/e2e/src/smoke-tests/release.spec.ts
+++ b/e2e/src/smoke-tests/release.spec.ts
@@ -1,0 +1,134 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { execSync } from 'node:child_process';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { afterAll, describe, expect, it } from 'vitest';
+// eslint-disable-next-line
+import { publishWithRetry } from '../../../scripts/release';
+import { MIGRATE_PACKAGES } from './migrate-versions';
+
+/**
+ * release smoke test — publishes the local build to the local registry through
+ * the release script's own `publishWithRetry`, and asserts all three packages
+ * land.
+ *
+ * The command `nx release publish` builds is derived from the active package
+ * manager's version, and a package manager that rejects the flag spelling nx
+ * picks fails every publish. Nothing else in the suite catches that: the other
+ * smoke tests publish with plain `npm publish`, and unit tests never invoke the
+ * executor. So the release lane's first signal used to be a failed release, on
+ * main, after the tag and GitHub release had already been created.
+ *
+ * Three details give this test teeth:
+ *
+ * - **It publishes a version the registry has never seen**, and asserts that
+ *   before publishing. nx short-circuits a package whose version already carries
+ *   the requested dist-tag and exits zero without running the publish command,
+ *   so reusing the `0.0.0` build `global-setup` published would pass whatever
+ *   the active package manager does.
+ * - **It publishes under a dedicated dist-tag.** `latest` keeps pointing at the
+ *   `0.0.0` build every other smoke test resolves.
+ * - **It runs the release script's own function**, not a copy of the command, so
+ *   a change to how the release publishes is covered by construction.
+ */
+const RELEASE_DIST_TAG = 'release-e2e';
+/**
+ * Version to publish under. Above every real release so it can never be
+ * mistaken for one, and unique per run so nx always performs a real publish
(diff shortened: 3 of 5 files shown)


cogwirrel (approved):

nx-plugin-for-aws:
Split into two PRs so the test can land independently of the pnpm 12 move:

- #1189 — `test(e2e)`: the release publish smoke test, on pnpm 11 so it lands green now
- #1190 — `chore(deps)`: the pnpm 12 move, which needs nx 23.2 (#1182) first

Keeping them together meant the test lane had to stay red until #1182 landed, which made this PR unmergeable and hid whether the test itself was sound.

PR #1189, opened 2026-09-03
test(e2e): cover the release publish against a local registry

nx-plugin-for-aws:
### Reason for this change

The `v1.0.0-rc.95` release tagged and cut a GitHub release, then failed to publish anything to npm:

```
pnpm publish ".../dist/packages/nx-plugin" --json --"@aws:registry=https://registry.npmjs.org/" --tag=latest --no-git-checks
error: unexpected argument '--@aws:registry' found
  tip: a similar argument exists: '--registry'
```

nx 23.1.2 gates that flag spelling on `>=9.15.7 <10.0.0 || >=10.5.0` — an unbounded upper range. pnpm 11 renamed the option to `--config.@scope:registry` and pnpm 12 dropped the old spelling, so moving the repo to pnpm 12 broke every publish.

**Nothing in CI caught it, and that missing coverage is the real defect.** The release publish had no test at all:

- the other smoke tests publish with plain `npm publish` in `global-setup`, which never exercises `nx release publish`
- unit tests never invoke the release-publish executor
- the command nx builds is *derived from the active pnpm version*, so it can only be validated by running it

The first signal was a failed release on main, **after** the git tag and GitHub release had already been created. That left a `v*` tag with nothing on npm, which then broke migrate shard 1 on every subsequent run (migrate resolves start versions from git tags).

### Description of changes

A `release` smoke test (`e2e/src/smoke-tests/release.spec.ts`) that publishes the local build to verdaccio and asserts all three packages land. Added to the `smoke_tests` matrix in `ci.yml` and `pr.yml`.

It calls the release script's own `publishWithRetry` rather than re-implementing the command, so the thing under test is the thing that actually releases. `scripts/release.ts` changes to support that:

- `publishWithRetry` is exported and takes optional overrides (`tag`, `additionalArgs`, `attempts`, `cwd`); defaults are unchanged, so the release path behaves exactly as before
- `main()` is guarded behind `import.meta.url === file://${process.argv[1]}`, matching the existing pattern in `scripts/stamp-migrations.ts`, so importing the module doesn't run a release

**Three things stop the test passing vacuously** — a release test that silently publishes nothing would be worse than no test:

1. **It publishes a version the registry has never served.** nx short-circuits a package whose version already carries the requested dist-tag, exits `0`, and never runs the publish command. `global-setup` has already published `0.0.0`, so reusing that version would pass green on *any* package manager. The test publishes `9.9.9-release-e2e.<pid>` and **asserts the precondition** before publishing.
2. **It publishes under a `release-e2e` dist-tag**, so `latest` keeps pointing at the `0.0.0` build every other smoke test resolves.
3. **It queries the registry afterwards** to confirm each package is really served at that version.

This PR is intentionally scoped to the test only and leaves the repo on pnpm 11, so it should be **fully green**. The pnpm 12 move is a separate PR — it depends on nx 23.2 (#1182) and would make this lane red until that lands.

Ran the real target both ways to prove the test discriminates.

`pnpm@11.25.0` (this branch) — passes, with three real publishes:

```
$ pnpm nx run @aws/nx-plugin-e2e:smoke-test --name=release
Published to http://localhost:4873 with tag "release-e2e"   (x3)
✓ should publish every package with the release script
Test Files  1 passed | 33 skipped
```

`pnpm@12.2.1` — fails with the exact rc.95 error:

```
error: unexpected argument '--@aws:registry' found
× should publish every package with the release script
Error: nx release publish failed after 1 attempts
```

**Verified the anti-vacuous guard by sabotaging it.** Pointed the test at `0.0.0` (already on the registry) to simulate a publish nx would skip. Before the guard it passed green while publishing nothing — the exact false negative worth avoiding. With the guard:

```
AssertionError: @aws/nx-plugin@0.0.0 is already on the local registry — this test cannot prove a publish ran
```

**Verified the flag behaviour directly** against each pnpm major with a throwaway scoped package:

| pnpm | `--"@aws:registry=..."` | `--"config.@aws:registry=..."` |
|---|---|---|
| 11.22.0 | accepted | accepted |
| 11.25.0 | accepted | accepted |
| 12.2.1 | **`unexpected argument`** | accepted |

**Confirmed the CLI guard didn't break the release entrypoint** — running `scripts/release.ts` directly still executes the real flow (it resolved the next version from the git tag and wrote the dist manifests), so the guard only suppresses execution on import. The subsequent `v1.0.0-rc.95` release on main published all three packages successfully with this same `release.ts` shape.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/e2e/src/smoke-tests/release.spec.ts
+++ b/e2e/src/smoke-tests/release.spec.ts
@@ -2,7 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-import { execSync } from 'node:child_process';
+import { execFileSync, execSync } from 'node:child_process';
 import { readFileSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { afterAll, describe, expect, it } from 'vitest';
@@ -10,36 +10,10 @@ import { afterAll, describe, expect, it } from 'vitest';
 import { publishWithRetry } from '../../../scripts/release';
 import { MIGRATE_PACKAGES } from './migrate-versions';
 
-/**
- * release smoke test — publishes the local build to the local registry through
- * the release script's own `publishWithRetry`, and asserts all three packages
- * land.
- *
- * The command `nx release publish` builds is derived from the active package
- * manager's version, and a package manager that rejects the flag spelling nx
- * picks fails every publish. Nothing else in the suite catches that: the other
- * smoke tests publish with plain `npm publish`, and unit tests never invoke the
- * executor. So the release lane's first signal used to be a failed release, on
- * main, after the tag and GitHub release had already been created.
- *
- * Three details give this test teeth:
- *
- * - **It publishes a version the registry has never seen**, and asserts that
- *   before publishing. nx short-circuits a package whose version already carries
- *   the requested dist-tag and exits zero without running the publish command,
- *   so reusing the `0.0.0` build `global-setup` published would pass whatever
- *   the active package manager does.
- * - **It publishes under a dedicated dist-tag.** `latest` keeps pointing at the
- *   `0.0.0` build every other smoke test resolves.
- * - **It runs the release script's own function**, not a copy of the command, so
- *   a change to how the release publishes is covered by construction.
- */
+/** Dist-tag to publish under, so `latest` keeps pointing at the 0.0.0 build. */
 const RELEASE_DIST_TAG = 'release-e2e';
-/**
- * Version to publish under. Above every real release so it can never be
- * mistaken for one, and unique per run so nx always performs a real publish
- * rather than re-tagging a version the registry already holds.
- */
+
+/** Unique per run, and above any version a real release could pick. */
 const releaseTestVersion = () => `999.9.9-release-e2e.${process.pid}`;
 
 const workspaceRoot = join(__dirname, '../../..');
@@ -50,6 +24,13 @@ const distManifest = (pkg: string) =>
     pkg.replace('@aws/', ''),
     'package.json',
   );
+
+const gitTags = (): string =>
+  execFileSync('git', ['tag', '-l'], {
+    cwd: workspaceRoot,
+    encoding: 'utf-8',
+    windowsHide: true,
+  });
 
 /** Versions the local registry currently serves for a package. */
 const versionsOnRegistry = (pkg: string, registry: string): string[] => {
@@ -65,7 +46,6 @@ const versionsOnRegistry = (pkg: string, registry: string): string[] => {
     // npm returns a bare string when only one version exists.
     return Array.isArray(versions) ? versions : [versions];
   } catch {
-    // Unpublished package — nothing on the registry yet.
     return [];
   }
 };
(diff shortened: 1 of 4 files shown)


e2e/src/smoke-tests/release.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+ */
+import { execSync } from 'node:child_process';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { afterAll, describe, expect, it } from 'vitest';
+// eslint-disable-next-line
+import { publishWithRetry } from '../../../scripts/release';
+import { MIGRATE_PACKAGES } from './migrate-versions';
+
+/**
```
cogwirrel: Too verbose, delete this whole comment and make others succinct and only include what’s absolutely necessary

e2e/src/smoke-tests/release.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    for (const pkg of MIGRATE_PACKAGES) {
+      const path = distManifest(pkg);
+      const contents = readFileSync(path, 'utf-8');
+      originalManifests.set(path, contents);
+      const manifest = JSON.parse(contents);
+      manifest.version = version;
+      writeFileSync(path, `${JSON.stringify(manifest, null, 2)}\n`);
+    }
+
+    await publishWithRetry({
```
cogwirrel, line 92: Make sure that this doesn’t create any git tags and GitHub release notes

e2e/src/smoke-tests/release.spec.ts
nx-plugin-for-aws: Done — deleted the file-level block and cut the rest back to what isn't evident from the code (dropped the `Three details give this test teeth` list, the `afterAll` and catch-branch comments, and trimmed the verbose JSDoc I'd added in `release.ts`). What's left is the four non-obvious bits: why the dist-tag isn't `latest`, why the version must be unique, why nx's skip-if-already-published behaviour matters, and why retries are capped at 1.

e2e/src/smoke-tests/release.spec.ts
nx-plugin-for-aws: Verified — it creates neither, and I've added an assertion so it stays that way.

Structurally, `nx release publish` is publish-only: tagging and changelog/GitHub-release live in the `nx release` step that `scripts/release.ts` runs *before* `publishWithRetry`, and the test calls only `publishWithRetry`. `nx/dist/src/command-line/release/publish.js` contains no `gitTag`/`createRelease` calls (12 such references in `release.js`, zero in `publish.js`). The `--tag` we pass is the **npm dist-tag**, not a git tag — easy to misread, which is what makes this worth checking.

Confirmed empirically too: snapshotted `git tag -l` and `gh release list` around a local run — 306 tags before and after, no `999*`/`release-e2e*` tag, and no new GitHub release.

Rather than rely on that, the test now snapshots tags before publishing and asserts they're unchanged at the end, so a future change to `publishWithRetry` can't start cutting releases from a test. I checked the assertion isn't vacuous by creating a tag mid-test — it fails as intended:

```
AssertionError: expected 'sabotage-tag-check\nv0.1.0\n…' to be 'v0.1.0\nv0.1.1\n…'
```

One thing worth flagging separately: the local registry is the only publish target, so even the npm side can't touch real npm — but `release.ts` run directly *does* write the dist manifests and would tag, which is why the CLI entrypoint stays behind the `import.meta.url` guard.

cogwirrel (approved):

PR #1196, opened 2026-09-04
feat(mcp-server): add a best-practices tool serving the cross-cutting guides

nx-plugin-for-aws:
> [!NOTE]
> **When squashing, keep only the first commit's body.** The branch carries three commits and GitHub concatenates all their bodies: `feat(mcp-server): add a best-practices tool …` (the real change), plus `docs: update translations` (translation workflow) and `chore: re-trigger CI` (empty commit to clear a wedged runner). Keep the PR title as the subject and drop the latter two bullets — the first body is written to be the release note.

### Reason for this change

Four cross-cutting guide pages are bundled into the published MCP package but reachable from nothing (`docs-integrity-07`). `guides/security`, `runtime-config`, `docker-bundling` and `local-development` are copied into the package by `packages/nx-plugin-mcp/project.json`, but `general-guidance` hard-coded `['workspace', 'typescript-project', 'python-project']` and no generator's `guidePages` names them. An agent driving the plugin therefore never saw the security controls the vended infrastructure carries, the runtime-config mechanism its projects discover each other through, container engine selection, or the `serve`/`dev` local development workflow — for the very generators it was invoking.

This PR also carries a small amount of related hardening and four docs corrections, described below.

#### A `best-practices` tool serving the cross-cutting guides

The pages above document behaviour that spans generators, so no single generator owns them and `generator-guide` will never return them. `best-practices` takes an enum of the available pages (`workspace`, `typescript-project`, `python-project`, `security`, `runtime-config`, `docker-bundling`, `local-development`), each described in the tool description, so an agent fetches only the pages its task needs rather than every page on every call. `general-guidance` keeps inlining the three pages every agent needs regardless of task and now points at `best-practices` for the rest; `TOOL_SELECTION_GUIDE` advertises it.

#### A guard against the two server entry points drifting apart

`packages/nx-plugin/src/mcp-server/server.ts` exposes `createServer()`, which registers every tool. The published `@aws/nx-plugin-mcp` binary does **not** call it — `packages/nx-plugin-mcp/src/index.ts` constructs its own `McpServer` and repeats the `add*Tool(server, generators)` calls by hand. The two lists are kept in sync only by whoever remembers to edit both, and nothing catches a divergence: `server.spec.ts` exercises `createServer()`, so a tool missing from the published entry point still passes every unit test; both files import the same registrars, so types and lint stay clean; and the package builds and publishes normally.

**To be clear about the blast radius: no released version is affected.** I checked `main` (`9d13812`) and both files register the same six tools, so no user has ever lost a tool. The hazard is latent, not shipped. I hit it in my own working tree while developing this PR — after registering the new tool in `server.ts` alone and rebuilding, driving `dist/packages/nx-plugin-mcp/bin/aws-nx-mcp.js` over stdio returned `MCP error -32602: Tool best-practices not found` for every call, with the whole test suite green. That was caught and fixed before the first push, so no commit on this branch ever carried the divergence either.

So this half is fragility-hardening rather than a user-visible bug fix, which is why the title leads with the tool. The new tool is registered in both entry points, and a new test in `server.spec.ts` scrapes the `add*Tool(server` calls out of `server.ts` and asserts each appears in the published entry point — it currently matches all seven and confirms all seven are wired.

I did **not** attempt the structural fix of having `index.ts` simply call `createServer()`. The blocker is generator resolution, not the tool list: `server.ts` goes through `listGenerators()`, whose `GENERATORS` is built from `path.resolve(import.meta.dirname, '..', '..')` in `utils/nx.ts`, whereas the published binary needs `buildGeneratorInfoList(path.resolve(__dirname, '..'))` relative to its own bundle and pre-filters hidden generators (`server.ts` filters at call time instead). Unifying that means parameterising the base directory and deciding where hidden-generator filtering belongs — a packaging-architecture decision I didn't want to make unilaterally inside a docs-and-metadata PR. Happy to do it here or as a follow-up; **@cogwirrel your call.**

#### Guide coverage tests (`docs-integrity-06`)

The only existing coverage (`connection-guide-coverage.spec.ts`) asserted page/metadata agreement for `connection` alone. Eight non-hidden generators resolve their guide page via the default kebab-cased id; all resolve correctly today, but renaming any of those pages would silently return an empty guide with no test failure. The new `guide-coverage.spec.ts` asserts that every non-hidden generator resolves at least one guide page that exists on disk (however it's resolved, so this covers those eight); that every page a generator declares in `guidePages` actually exists (a declared-but-missing page is silently dropped by `fetchGuideFrontmatters`); and that every `best-practices` page resolves.

#### Docs corrections

- **`docs-integrity-05`** — `snippets/required-prerequisites.mdx` (rendered on quick-start, every tutorial and the workspace guide) told readers to run `uv python install 3.14.0`. The minor is right, but generators deliberately pin bare `major.minor`, and migration `v1.0.0-rc.88/0001` exists specifically to strip exact patches from `.python-version` because `3.14.7` installs on Linux but not macOS, so `uv sync` fails there. The prerequisite instructed readers to create exactly the condition a shipped migration undoes.
- **`quickstart-06`** — `existing-project.mdx`'s `nx sync` hangs entry claimed `init` "pins the root `nx` devDependency to the version the plugin's own `@nx/*` packages resolve". It can't: devkit's `addDependenciesToPackageJson` drops any incoming version that isn't a strict upgrade (`removeLowerVersions`), so a workspace already on a newer `nx` keeps it and two `nx` versions coexist — precisely the deadlock the entry describes as the cause. The mitigation was documented as already applied when it isn't. The `init.ts` comment above that call asserted the same incorrect behaviour and is corrected too.
- **`quickstart-07`** — `workspace.mdx`'s `aws-nx-plugin.config.mts` example omitted `packageManager: { catalogs: true }`, which the preset writes into every workspace and which drives the `catalog:` protocol the same guide's Single Version Policy section describes. Anyone wanting to opt out of catalogs had no documented way to find the key.

`powers/` and `skills/` are regenerated from the edited snippet by the pre-commit hook, and the non-English locales by the translation workflow.

#### No migration

None is needed: nothing here changes generated output. The changes are MCP server code, tests, docs, and one source comment — there is no workspace state for a migration to carry forward.

- `pnpm nx run @aws/nx-plugin:lint --configuration=fix` — clean.
- `npx vitest run src/mcp-server` — the `guide-coverage`, `server` and `generator-info` specs pass (112 tests). One 120s timeout in the pre-existing `connection-guide-coverage` spec under load on my box; it passes on its own (37 tests).
- `pnpm nx build docs` — passed. CI covers it as well.
- **All three new guards proven to fail, then restored:** removing `addBestPracticesTool` from the published entry point fails the parity test; renaming a `best-practices` page fails `best-practices resolves the local-development page`; renaming a `guidePages` entry fails both `resolves at least one guide page` and `every guide page … declares exists on disk`.
- **Drove the built MCP server** (`dist/packages/nx-plugin-mcp/bin/aws-nx-mcp.js`) over stdio. `best-practices` returns all four newly-reachable pages (`security` 7,037 chars with `Security Controls` + `Shared Responsibility Model`; `runtime-config` 13,173; `docker-bundling` 18,729 with the trivy and finch guidance; `local-development` 3,474 with the `dev` target description) and 78,909 chars for all seven together. An invalid page is rejected by the enum rather than silently returning empty. `general-guidance` (44,755 chars) advertises the tool and names the pages it defers. `list-generators` returns 40,785 chars, and `generator-guide` returns non-empty, correct guides for all eight generators from `docs-integrity-06` (`ts#lambda-function` 18,527 / `ts#dcr-proxy` 28,597 / `ts#nx-plugin` 8,084 / `ts#dynamodb` 26,299 / `py#dynamodb` 39,988 / `py#rdb` 40,526 / `license` 16,947 / `agentcore-gateway` 27,583). The corrected `packageManager: { catalogs: true }`, `uv python install 3.14` and `nx` troubleshooting text all reach agents.

**`security.mdx` audit — negative result, no changes needed, no need to re-audit.** Since the page is now served to agents, I checked every claim it makes against what the generators actually produce. All of it is accurate: Checkov in the `build` target with `suppressRules` requiring a reason for each suppression; the `trivy` target scanning HIGH/CRITICAL; git-secrets pre-commit hooks; the vended custom authorizer stub returning `isAuthorized: false`; Cognito MFA required (`Mfa.REQUIRED` on CDK, `mfa_configuration` defaulting to `"ON"` on Terraform), the strong password policy, and deletion protection (`deletionProtection: true` / `"ACTIVE"`); CloudFront `REDIRECT_TO_HTTPS` with HSTS/CSP and a WAF with AWS managed rules; S3 `BlockPublicAccess.BLOCK_ALL`, `enforceSSL`, and KMS with `enableKeyRotation`; CMK-encrypted CloudWatch access log groups; Aurora `storageEncrypted` with a CMK plus generated Secrets Manager credentials with rotation; and the `grant*` methods on the vended constructs.

### Issue # (if applicable)

N/A — v1.0 bug bash findings `docs-integrity-05`, `docs-integrity-06`, `docs-integrity-07`, `quickstart-06`, `quickstart-07`.

### Checklist

----


docs/src/content/docs/en/snippets/required-prerequisites.mdx
```diff
@@ -5,5 +5,5 @@ title: Required Prerequisites
 - [Node >= 22](https://nodejs.org/en/download) (We recommend using something like [NVM](https://github.com/nvm-sh/nvm) to manage your node versions)
   - verify by running `node --version`
 - [UV >= 0.5.29](https://docs.astral.sh/uv/getting-started/installation/)
-  1. install Python 3.14 by running: `uv python install 3.14.0`
+  1. install Python 3.14 by running: `uv python install 3.14` (naming the bare `major.minor` lets uv resolve whichever patch your platform has a build for, matching the `.python-version` generators write)
```
cogwirrel: Remove the text in brackets

packages/nx-plugin/src/mcp-server/tools/general-guidance.ts
```diff
@@ -8,6 +8,24 @@ import { IAC_PROVIDERS } from '../../utils/iac-providers.js';
...
+ * single generator owns them and `generator-guide` would never return them.
+ */
+export const CROSS_CUTTING_GUIDE_PAGES = [
+  'workspace',
+  'typescript-project',
+  'python-project',
+  'security',
+  'runtime-config',
+  'docker-bundling',
+  'local-development',
```
cogwirrel: these should be behind another tool, eg `best-practices` which has an enum input of available pages.

packages/nx-plugin/src/mcp-server/guide-coverage.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+import { CROSS_CUTTING_GUIDE_PAGES } from './tools/general-guidance.js';
+
+/**
+ * The MCP tools serve guide content by name: `generator-guide` resolves the
+ * pages a generator declares in `generators.json`, and `general-guidance`
+ * resolves the cross-cutting pages no single generator owns. Neither path
+ * errors on a name that matches no file — the page is simply dropped — so
+ * without these tests a page rename silently empties an agent's guidance.
+ */
+describe('guide coverage', () => {
```
cogwirrel, line 16: unnecessary, remove

packages/nx-plugin/generators.json
```diff
@@ -14,7 +14,8 @@
       "factory": "./src/agentcore-gateway/generator",
       "schema": "./src/agentcore-gateway/schema.json",
       "description": "Generate an AgentCore Gateway project",
-      "metric": "g48"
+      "metric": "g48",
+      "guidePages": ["agentcore-gateway"]
```
cogwirrel: No need to explicitly name these, the "fallback" isn't a fallback, it's the default and guidePages are an override

nx-plugin-for-aws:
CI is green other than `Smoke Tests - react-website`, which is a pre-existing repo-wide failure unrelated to this PR — it fails identically on the sibling bug-bash PRs #1197 and #1198, neither of which touches the website generator.

The assertion is in `e2e/src/smoke-tests/react-website.spec.ts:131`:

```
```

The spec fetches the shadcn CLI fresh from npm (`pnpm dlx shadcn@4.19.0`, `TS_VERSIONS.shadcn`) and asserts on the import alias the CLI writes into the generated component. The CLI is now emitting a different alias shape than the assertion expects, so it breaks any PR that runs this job. Out of scope for this lane — flagging rather than fixing.

Build, Package and all 10 Unit Tests shards pass, including the new `guide-coverage.spec.ts`.

packages/nx-plugin/src/mcp-server/tools/general-guidance.ts
nx-plugin-for-aws: Done — added a `best-practices` tool (`packages/nx-plugin/src/mcp-server/tools/best-practices.ts`) with a `pages` enum of the available pages, each described in the tool description so the agent can pick what its task needs. `general-guidance` keeps inlining `workspace`/`typescript-project`/`python-project` and now points at `best-practices` for `security`, `runtime-config`, `docker-bundling` and `local-development`.

One thing worth flagging: `packages/nx-plugin-mcp/src/index.ts` builds its own `McpServer` rather than calling `createServer`, so registering the tool in `server.ts` alone left it absent from the published package — I hit `Tool best-practices not found` driving the built binary. Registered in both, and added a test in `server.spec.ts` that scrapes the `add*Tool(server` calls out of `server.ts` and asserts each appears in the published entry point, so they can't drift again.

packages/nx-plugin/generators.json
nx-plugin-for-aws: Understood — reverted, `generators.json` is back to its pre-PR state. I'd read the kebab id as a fallback rather than the default; the coverage test now targets the actual risk instead, asserting every non-hidden generator resolves a page that exists on disk however it's resolved, so a page rename fails the build whether the generator declares `guidePages` or relies on the default.

docs/src/content/docs/en/snippets/required-prerequisites.mdx
nx-plugin-for-aws: Done — the line is now just `install Python 3.14 by running: \`uv python install 3.14\``.

packages/nx-plugin/src/mcp-server/guide-coverage.spec.ts
nx-plugin-for-aws: Done — removed.

nx-plugin-for-aws:
Review comments addressed and pushed (`9d2c61b`) — PR title and description updated to match. CI: **70 pass, 1 fail**.

The one failure is `Smoke Tests - react-website`, a pre-existing repo-wide break unrelated to this PR. It fails with a byte-identical assertion on the sibling bug-bash PRs #1199 and #1200, neither of which touches the website generator:

```
```

`e2e/src/smoke-tests/react-website.spec.ts:131` fetches the shadcn CLI fresh from npm (`pnpm dlx shadcn@4.19.0`) and asserts on the import alias it writes into the generated component; the CLI now emits a different shape. Out of scope for this lane — flagging rather than fixing.

Build, Package and all 10 Unit Tests shards pass, including the new `guide-coverage.spec.ts` and the `server.spec.ts` entry-point parity test.

nx-plugin-for-aws:
Re-titled and restructured the description so the **published-entry-point defect leads** rather than sitting in the validation notes — it's a shipping bug independent of the bug bash and deserves review priority over the docs fixes.

Short version for reviewers: `packages/nx-plugin-mcp/src/index.ts` builds its own `McpServer` and repeats the `add*Tool(server, generators)` calls instead of calling `createServer()`. A tool registered only in `packages/nx-plugin/src/mcp-server/server.ts` is therefore **absent from what users install**, and nothing catches it — `server.spec.ts` tests `createServer()`, both files import the same registrars so types and lint are clean, and the package builds and publishes fine. Driving the built binary returned `MCP error -32602: Tool best-practices not found` for every call while the repo was fully green.

The guard scrapes the `add*Tool(server` calls out of `server.ts` and asserts each appears in the published entry point; it currently matches all 7 and confirms all 7 are wired.

On the structural alternative — having `index.ts` just call `createServer()` — the blocker is generator resolution, not the tool list: `server.ts` goes through `listGenerators()`, whose `GENERATORS` is built from `path.resolve(import.meta.dirname, '..', '..')` in `utils/nx.ts`, whereas the published binary needs `buildGeneratorInfoList(path.resolve(__dirname, '..'))` relative to its own bundle and pre-filters hidden generators. Unifying that means parameterising the base directory, which felt like a separate change. Happy to do it in this PR or a follow-up — your call.

Also worth stating plainly so it isn't repeated: the `security.mdx` audit in the description is a **negative result**. I checked every claim on that page against the generators (Checkov/`suppressRules`, trivy, git-secrets, the `isAuthorized: false` authorizer stub, Cognito MFA/password policy/deletion protection on both CDK and Terraform, CloudFront HSTS/CSP/WAF, S3 `BLOCK_ALL`/`enforceSSL`/KMS rotation, CMK-encrypted log groups, Aurora encryption with rotating Secrets Manager credentials, `grant*` methods) and all of it is accurate. No changes were needed and none were made.

cogwirrel (dismissed):

nx-plugin-for-aws:
Retitled to `feat(mcp-server): add a best-practices tool serving the cross-cutting guides` under the release-note rule, and **corrected the description** — I'd overstated the entry-point issue and want that on the record.

I went to confirm the blast radius before letting the title claim a fix, and the answer changes the framing. On `main` (`9d13812`) both `server.ts` and `packages/nx-plugin-mcp/src/index.ts` register the **same six tools**:

```
main (pre-PR)  server.ts: 6   index.ts: 6   in server.ts but NOT published: []
```

So **no released version is affected and no user has ever lost a tool.** The divergence existed only transiently in my own working tree while developing this PR — I registered the new tool in `server.ts`, rebuilt, drove the binary and got `Tool best-practices not found`. I fixed it before the first push, so no commit on this branch carried it either (verified across `e6c3357`/`38e66e7`/`HEAD`).

That makes it **fragility-hardening, not a shipping bug fix**. A `fix(mcp-server): register all tools in the published entry point` title would have promised a release note for something users never experienced — worse than underselling. The genuine user-facing change is the four cross-cutting guide pages (`security`, `runtime-config`, `docker-bundling`, `local-development`) going from bundled-but-unreachable to reachable, which is `docs-integrity-07` and a new capability, hence `feat`.

I've dropped the `> [!IMPORTANT]` callout accordingly — with the corrected facts it was compensating for the title rather than supporting it, and it asserted "silently absent from what users install" as present-tense fact when it never was. The mechanism, the missing-coverage explanation and the parity guard are all still documented under "A guard against the two server entry points drifting apart", now with the blast radius stated plainly.

Retained as you'd expect: the note on why I didn't attempt the structural `index.ts`→`createServer()` unification (generator-resolution mismatch, still **@cogwirrel's call**), the explicit `security.mdx` negative result, and a "No migration" section saying why none is needed — nothing here changes generated output.

Metadata-only edit via the REST API, so no force-push and the queued run from 11:33 is untouched.

nx-plugin-for-aws:
Drive-by from #1203, which had the same class of CI problem — this PR's run looks stalled rather than queued, and clearing it needs no push.

Run [33868489485](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33868489485) (created 11:33) has **32 jobs completed, all `success`, and 34 still `queued`** — with **no progress since 12:19**, over four hours ago. Twelve runs created after it have since completed.

The 34 stuck jobs carry a run-scoped CodeBuild label:

```
codebuild-nx-plugin-for-aws-runner-33868489485-1
```

Those are only claimable by a runner provisioned for that exact `<run-id>-<attempt>`, so they can't fall back to another executor. Once that runner is gone, nothing drains them.

This is a *partial* wedge — unlike #1203, where zero jobs ever started, here the runner served 32 jobs and then went away mid-run. Same remedy:

```bash
gh run cancel 33868489485       # the run must be terminal before rerun is accepted
gh run rerun --failed 33868489485
```

The retry gets a fresh `-2` label, CodeBuild provisions against it, and jobs start within minutes. On #1203 that took it from 12 permanently-stuck checks to 71/71 pass. Note `gh run rerun` on a *live* run reports `cannot be rerun; its workflow file may be broken` — misleading; the workflow is fine, it just refuses while the run is active.

Worth checking the label before assuming this, though: jobs labelled `macos-latest` sit in GitHub's shared pool and **do** drain on their own, so a long queue there is normal. Only the `codebuild-...-<run-id>-<attempt>` ones can wedge.

Since this PR is already approved, that rerun is probably all that stands between it and merge.

nx-plugin-for-aws:
@cogwirrel re-review requested — your approval was auto-dismissed by branch protection, not withdrawn.

To clear a wedged CI run I had to push to this branch, and `main` has *dismiss stale reviews* enabled, so the approval on `7518119` was dropped automatically (`reviewDecision` is now `REVIEW_REQUIRED`). Flagging the trade explicitly since it cost your sign-off:

**The push was worth it because the commit messages were wrong.** Both prior commits described the explicit `guidePages` entries as shipped work — the approach you asked me to revert. A squash merge takes its body from commit messages rather than the PR description, so this would have merged with a release note describing reverted work. The branch is now a single commit whose message describes what actually ships, and which states plainly that the entry-point guard covers a **latent** hazard — both entry-point lists agree on `main`, so no released version was ever affected.

**The reviewed content is byte-identical to what you approved.** Verified by comparing the patch against both bases:

```
git diff 9d13812 b729752  (approved)
git diff 80ba018 <head>   (now)
→ identical across packages/, docs/src/content/docs/en/, powers/, skills/
```

The only delta anywhere is 4 auto-generated translation files (`fr`/`ko`/`vi`) that the translation workflow rewrote on its own commit — zero English or source files touched.

So the diff you reviewed, the title, and the description are all unchanged; only the commit message and the base (rebased onto current `main`) differ. Nothing further is queued from my side, so a re-approval won't be invalidated by another push.

Still open and still yours to call: whether to do the structural `index.ts`→`createServer()` unification here or as a follow-up. The blocker is generator resolution rather than the tool list — `server.ts` goes via `listGenerators()` (built from `path.resolve(import.meta.dirname, '..', '..')`), whereas the published binary needs `buildGeneratorInfoList(path.resolve(__dirname, '..'))` relative to its own bundle and pre-filters hidden generators. Happy either way.

nx-plugin-for-aws:
**Note for whoever merges this — please trim the squash commit body.**

The branch carries three commits, and GitHub's squash UI concatenates all three bodies into the merge commit:

```
- feat(mcp-server): add a best-practices tool serving the cross-cutting guides
- docs: update translations
- chore: re-trigger CI
```

Only the first describes shipped work. The other two are infrastructure noise: `docs: update translations` is the translation workflow's own commit, and `chore: re-trigger CI` is an empty commit I pushed to clear a wedged CI run (jobs were queued against a runner label whose provisioning had died, and a new push was the only way to get a fresh label).

**Please keep the squash subject as the PR title** — `feat(mcp-server): add a best-practices tool serving the cross-cutting guides` — **and keep only the first commit's body**, dropping the other two bullets. That body is deliberately written to be the release note: it leads with the four cross-cutting guide pages going from bundled-but-unreachable to reachable, and states explicitly that the entry-point guard covers a *latent* hazard rather than a shipped bug (both entry-point tool lists agree on `main`, so no released version was ever affected).

Flagging it here because two PRs in this batch (#1211 and #1198) merged with release notes describing approaches they had reverted, precisely because the squash body came from commit messages rather than the PR description.

cogwirrel (approved):

PR #1197, opened 2026-09-04
docs(connection): grant the correct bedrock-agentcore action in the MCP Terraform snippets

nx-plugin-for-aws:
### Reason for this change

Five bug-bash findings across the two MCP connection guides and the two agent guides.

**`conn-agent-03` (major)** — the Terraform fragment in both MCP connection guides granted `bedrock-agentcore:InvokeAgent`, **an action that does not exist**. A Terraform user who copied it got a policy that grants nothing useful — no error at apply time, just `AccessDenied` at runtime when the agent tried to reach its connected MCP server. The CDK path in the same section grants the right thing, and the sibling A2A snippet already spelled it correctly, so only the Terraform readers were affected.

**`agents-01` (blocker, verified) / `agents-02` (major)** — the generated `<name>-serve` target sets only `PORT`, while `<name>-dev` sets `PORT` + `LOCAL_DEV=true`. That is the *whole* difference between them, and neither agent guide told the reader they differ — nor did either guide link to the existing [Local Development](https://awslabs.github.io/nx-plugin-for-aws/en/guides/local-development/) guide that explains `serve` vs `dev`. Without `LOCAL_DEV`, the generated `session.py` / `session.ts` takes the deployed branch and calls `get_agentcore_runtime_config()`, which raises `RUNTIME_CONFIG_APP_ID is not set`.

For `py#agent` this is especially bad: `get_session_manager()` is called lazily from inside `get_agent()`, which runs inside an already-started `JsonStreamingResponse`, so the exception fires after headers are flushed and is then masked by starlette's `Caught handled exception, but response already started.` The caller gets **HTTP 200 with a zero-byte body and no error at all** — the exact `curl` the guide told the reader to run appears to succeed while producing nothing. `ts#agent` has the same gap but tRPC surfaces the error to the client, hence the lower severity.

**`agents-09`** — the py-agent "Invoke the Local Server" `curl` hardcoded `8081` with no caveat, and the guide never said where to find the assigned port. Ports are auto-assigned per component from the workspace pool.

**`agents-10`** — the HTTP-only OpenAPI script / `-openapi` / `-generate-client` paragraph sat outside any protocol filter, so it described files and targets that don't exist for `a2a` and `ag-ui` agents. The `_openapi.py` path was also written as `<your-agent-name>_openapi.py`, which only matches when the agent name has no hyphens.

### Description of changes

**Docs only — no generator or template code changed, so there is no vended output to migrate.**

For `conn-agent-03`, both Terraform fragments now grant `bedrock-agentcore:InvokeAgentRuntime` and `bedrock-agentcore:InvokeAgentRuntimeWithWebSocketStream` on both the runtime ARN and `${arn}/*`, matching what the CDK path synthesizes and what the sibling `a2a-infrastructure.mdx` snippet already did. I also added a one-line note to each CDK fragment naming the actions `grantInvokeAccess` wires up, mirroring how the A2A and React snippets document theirs.

Verified against the real IAM action list rather than assumed: the AWS `InvokeAgentRuntime` API reference states the required permission is `bedrock-agentcore:InvokeAgentRuntime` (plus `InvokeAgentRuntimeForUser` when calling on behalf of a user id), and `aws-cdk-lib/aws-bedrockagentcore`'s `perms.js` defines exactly `RUNTIME_INVOKE_PERMS = ['bedrock-agentcore:InvokeAgentRuntime']` and `RUNTIME_INVOKE_USER_PERMS = ['bedrock-agentcore:InvokeAgentRuntimeForUser']`. There is no `InvokeAgent` action. I grepped the whole `docs/` tree and all of `packages/` — these two guides were the only occurrences, and there are none in any vended Terraform or CDK template.

For `agents-01`/`agents-02`, both agent guides' "Local Development" sections now note that a `<name>-serve` target is also generated, that it runs the agent against deployed infrastructure and therefore requires `RUNTIME_CONFIG_APP_ID`, and link to the **Local Development** guide for the full `dev` vs `serve` distinction rather than restating it. Both "Invoke the Local Server" sections now open by telling the reader to start the agent with `<name>-dev`.

Port guidance (`agents-09`) points at `metadata.ports` in `project.json` (and the `-dev` target's `--port` flag for Python), and frames `8081` as what a *first* HTTP agent gets in a fresh workspace rather than an absolute.

For `agents-10`, the OpenAPI paragraph is wrapped in `<OptionFilter when={{ protocol: 'http' }}>` and the script path corrected to `<your_agent_name>_openapi.py`, noting it is the agent name in `snake_case`.

### Review feedback

@cogwirrel pointed out that `guides/local-development.mdx` already documents the `dev` vs `serve` distinction, including the `RUNTIME_CONFIG_APP_ID` requirement — so the agent guides now link to that canonical guide instead of restating it, and two redundant `:::caution` blocks in the invocation sections were removed. Neither agent guide linked to it at all before, which is a good part of why the distinction was invisible to readers.

### A regression in the translated locales, caught and fixed

Worth calling out because it explains the commit history. Only English is authored here; the other eight locales are generated by the translation workflow. After the review change I reverted the non-English files so the workflow would regenerate them rather than hand-editing generated content — but that revert also undid the *already-correct* translated IAM fix in the connection guides, and the following translation run only regenerated the agent guides, leaving the bad `InvokeAgent` action in all eight locales. Fixed in `4b77eb2` by restoring the workflow's own output for those files. Final state verified by grep: **zero occurrences of the bare `bedrock-agentcore:InvokeAgent` action across all nine locales**, and no leftover references to files removed during the review.

### Deliberately not fixed — follow-up

**The empty-`200` itself remains.** `py#agent`'s `-serve` target returning HTTP 200 with a zero-byte body is a genuinely bad signal: the failure is invisible to the client because `get_session_manager()` resolves mid-stream, after response headers have been flushed. Hoisting session resolution ahead of `JsonStreamingResponse` in the HTTP `main.py`/`init.py` templates would turn it into a clean `500`. That is vended generator surface and wants its own change with tests, so it is recorded here rather than bundled in.

`-serve` reading deployed configuration is *intentional* — it is how you exercise real session storage and real AgentCore MCP transport locally, and the Local Development guide already describes it that way. Making it silently fall back to local behaviour would remove the only way to do that, and making it hard-fail would break the legitimate `RUNTIME_CONFIG_APP_ID=<id> nx run …:agent-serve` case. The bug this PR fixes is that the agent guides never said the targets differ or pointed at the guide explaining it.

Separately: the `docs` project has no `lint` target, so `nx run-many --target lint --all` never lints `.mdx`. Noted, not addressed here.

**Reproduced both agent findings before fixing anything**, in a fresh `--iac=cdk` workspace against the locally compiled plugin (`pnpm link`), replicating each target's environment exactly. No AWS deploy and no live Bedrock calls — dummy credentials throughout, with the Bedrock endpoint pointed at a dead local port for the control.

- `py#agent`, `-serve` environment (`PORT` only, `LOCAL_DEV` unset): `/ping` → `Healthy`, then the guide's documented `curl` → **`HTTP=200 bytes=0`**, body literally zero bytes on disk. The server log shows the exact two-stage failure — `session.py:21` → `get_agentcore_runtime_config()` → `RuntimeError: RUNTIME_CONFIG_APP_ID is not set`, masked by `RuntimeError: Caught handled exception, but response already started.`
- `py#agent`, `-dev` environment (same code, `LOCAL_DEV=true`): the session manager resolves and streaming starts — `tmp/agents/strands/py-agent-http-agent/session_<uuid>/` is written with `session.json` and messages. Only the deliberately unreachable model endpoint then fails, isolating `LOCAL_DEV` as the sole variable.
- `ts#agent`, `-serve` environment: `/ping` → `Healthy`, and invoking via the guide's own `.local` client factory returns a **visible** `TRPCClientError: RUNTIME_CONFIG_APP_ID is not set …`, confirming the severity split between the two generators is right.

**Confirmed the port and path claims** by generating four agents in one workspace and reading `project.json`: a first `py#agent` gets `8081`; a `ts#agent` alongside it gets `8082`; a second py HTTP agent gets `8083`; an `a2a` agent gets `9000`. The finding reported `8083` for a first agent, which turned out to be an artifact of that reporter's workspace already holding other projects — so the guide now frames the number as "a first agent in a fresh workspace" rather than restating either figure as absolute. For `agents-10`: a `--protocol=a2a` agent emits only `scripts/<name>/{agentcore,chat}.ts` with targets `a2a-chat`/`a2a-dev`/`a2a-serve` and no openapi anything, while an HTTP agent named `my-agent` emits `scripts/my_agent_openapi.py` — confirming both the protocol gap and the `snake_case` path.

**Doc validation:** CI is green — all 66 jobs in the latest run succeeded (71/71 checks). The `Build` job is the meaningful one for a docs change: it runs `nx build docs`, which exercises the starlight link validator and the build-time validator that checks every `<OptionFilter>` predicate against the generator's JSON schema, confirming `protocol: 'http'` is valid for `py#agent`. I also inspected the built HTML locally: both agent guides render with the new `OptionFilter` registered against the right predicate, and the connection pages render only `InvokeAgentRuntime` / `InvokeAgentRuntimeWithWebSocketStream` with no bare `InvokeAgent`.

`pnpm nx run @aws/nx-plugin:test` — 4230 passed / 21 failed. All failures are a pre-existing `@nx/js` environment error (`TypeError: Cannot read properties of null (reading 'version')` inside `libraryGeneratorInternal`), which cannot be reached by an `.mdx` change. **Confirmed pre-existing** by stashing my changes and re-running the affected suite: byte-identical 26 failures on the untouched tree. No snapshots changed.

### Issue # (if applicable)

N/A — bug bash findings `conn-agent-03`, `agents-01`, `agents-02`, `agents-09`, `agents-10`.

### Checklist

----


docs/src/content/docs/en/guides/py-agent.mdx
```diff
@@ -579,6 +589,10 @@ curl -N -X POST http://localhost:8081/invocations \
 The `-N` argument given to `curl` disables buffering the output stream, so you can see the streaming response in real-time.
 :::
 
+:::caution[Use `-dev`, not `-serve`]
+Use `<your-agent-name>-dev` for this, not `<your-agent-name>-serve`. `-serve` runs the deployed code path, so it needs `RUNTIME_CONFIG_APP_ID` set — see <Link path="guides/py-agent#-dev-vs--serve">`-dev` vs `-serve`</Link>. Without it, streaming has already begun by the time the agent tries to resolve its session storage, so the request returns an empty `200` with no error and the failure appears only in the server log.
+:::
```
cogwirrel: Remove this

docs/src/content/docs/en/guides/ts-agent.mdx
```diff
@@ -420,12 +426,18 @@ const client = MyAgentClient.local({ url: 'http://localhost:8081/ws' });
...
+
 :::tip[Quick Testing]
 Run with `tsx` as a quick way to test out your agent.
 
 <PackageManagerExecCommand commands={['tsx scripts/test.ts']} />
 :::
 
+:::caution[Use `-dev`, not `-serve`]
+Use `<your-agent-name>-dev` for this, not `<your-agent-name>-serve`. `-serve` runs the deployed code path, so it needs `RUNTIME_CONFIG_APP_ID` set — see <Link path="guides/ts-agent#-dev-vs--serve">`-dev` vs `-serve`</Link>. Without it, every invocation fails with `RUNTIME_CONFIG_APP_ID is not set`.
+:::
```
cogwirrel: Remove this

docs/src/content/docs/en/snippets/agent/dev-vs-serve.mdx
```diff
@@ -0,0 +1,27 @@
+---
```
cogwirrel: This is pointless, we should link to the local-development guide instead which talks about dev vs serve

nx-plugin-for-aws: Agreed — I'd missed that `guides/local-development.mdx` already covers `serve` vs `dev` including the `RUNTIME_CONFIG_APP_ID` requirement. Snippet deleted (and its 8 auto-generated translations removed; I reverted the translated guides to their pre-translation state so the workflow regenerates them from the updated English rather than me hand-editing generated locales).

Both agent guides' Local Development sections now carry a one-line pointer instead:

> A `<your-agent-name>-serve` target is also generated, which runs the agent against your deployed infrastructure and therefore requires `RUNTIME_CONFIG_APP_ID` to be set. See the Local Development guide for the difference between `dev` and `serve`.

Worth noting neither agent guide linked to that guide at all before, which is a good part of why the distinction was invisible.

docs/src/content/docs/en/guides/py-agent.mdx
nx-plugin-for-aws: Removed. The section now just opens by telling the reader to run `<your-agent-name>-dev` before the curl, which is enough — the `dev` vs `serve` detail belongs in the Local Development guide, linked from Local Development above.

docs/src/content/docs/en/guides/ts-agent.mdx
nx-plugin-for-aws: Removed, same as the py-agent one. The section now opens by telling the reader to run `<your-agent-name>-dev`, with the `dev` vs `serve` detail left to the Local Development guide.

cogwirrel (approved):

PR #1198, opened 2026-09-04
fix(py#project): prompt for the project type when not supplied

nx-plugin-for-aws:
### Reason for this change

`py#project` declares `type` as **required** but gave it neither an `x-prompt` nor a `$default`. It was the only required option across all non-hidden generators in that state, so the interactive run the guide documents (`<RunGenerator generator="py#project" />`) never asked which of `application` / `library` the reader wanted — they silently got `application`. The analogous `smithy#project` and `terraform#project` options both prompt. (`docs-integrity-08`)

This is also what the reported "`--projectType` is silently dropped" defect really was (`python-03`). Worth correcting the finding: **`projectType` appears nowhere in the docs** — I grepped `docs/src/content`, `docs/src/components` and `docs/src/i18n` — so it came from the finding's own prose framing rather than from any guide. `type` is the correct and only spelling: it is what the generator, `schema.d.ts`, every existing test and the sibling `smithy#project` / `terraform#project` generators use. So there was no doc/schema disagreement to reconcile, only a missing prompt — meaning `docs-integrity-08`'s fix *is* `python-03`'s fix.

Two related documentation defects on the FastAPI guide's "Accessing the Calling User" section:

- **`python-01`** — the guide's `identity.py` example imports `boto3`, which is not a dependency of the generated FastAPI project, and the guide never says to add it. The very next documented step (`typecheck`) fails with `unresolved-import`.
- **`python-02`** — the natural `add boto3` resolves latest and writes a floating `boto3>=1.43.88`, while `py#mcp-server` pins `boto3==1.43.83` (`packages/nx-plugin/src/utils/versions.ts`). Since uv treats the workspace as one resolution universe, the two constraints have no solution. `py#mcp-server`'s `uv sync` then fails **after** its files are written, leaving a half-generated workspace, and every `build` / `test` / `typecheck` fails until a human hand-edits the constraint.

### Description of changes

**`py#project` schema** — added an `x-prompt` (and `x-priority: important`) to the `type` option. Nx renders it as an autocomplete over the enum with `application` preselected from the existing `default`, so an interactive user is now asked instead of silently defaulted.

**No migration.** The change is additive: it only affects the choices offered at generation time, and the unchanged default leaves already-generated workspaces byte-identical. There is nothing in an existing workspace to carry forward.

**`fastapi.mdx`** — states `boto3` as a prerequisite for `identity.py` and gives the add command pinned to the version the generators use. The version renders from `PY_VERSIONS.boto3` via a direct import (the pattern `docker-bundling.mdx` already uses for `CONTAINER_VERSIONS.trivy`), so the documented command cannot drift from the pin.

**e2e smoke tests** — nine call sites passed `--projectType=application` to `py#project`, which has no such option. Nx drops unrecognised flags silently, so those runs were relying on the `application` default rather than the flag they named: the same failure mode as `python-03`, in our own suite. Corrected to `--type=application`. No behaviour change, but the suite no longer contradicts the schema.

This is also the clearest argument for rejecting an alias. Had `projectType` been aliased onto `type`, those nine sites would have started *working* — and the contradiction would have been masked permanently, leaving a suite that looked correct while still naming an option that does not exist.

The `python-project.mdx` guide is intentionally unchanged — the `type` option is covered by its generated parameters table.

Unit tests added to `packages/nx-plugin/src/py/project/generator.spec.ts` (new `type option` block): both enum values produce the matching `projectType`; `type` is prompted when absent (asserting choices and initial value); and not prompted when supplied.

- `pnpm nx run @aws/nx-plugin:test --skip-nx-cache -- --run py/project` — 45 passed on the rebased tree.
- `pnpm nx run @aws/nx-plugin:build` (full plugin suite) and `lint --fix` — green. No snapshots changed.
- `nx build docs` — passes with all internal links valid, and the built page renders the literal `boto3==1.43.83`, confirming the `PY_VERSIONS.boto3` import resolves through to output.

End-to-end in a scratch workspace against the locally compiled plugin:

- Interactive run (under a pty, no `--no-interactive`) prompts "What type of Python project would you like to create?" with `application` preselected. Previously the reader was never asked.
- `--type=library` → `project.json` `projectType: library`; `--type=application` → `application`.
- Replayed `python-02`'s repro using the newly documented pinned command: `py#api` → `run ws.my_api:add boto3==1.43.83` (writes an exact pin) → `py#project` → `py#mcp-server` all succeed, then `run-many --target build --all` passes fully (5 projects, 36 tasks). Previously `py#mcp-server` aborted on `uv sync` and 7 build tasks failed.
- Created the guide's `identity.py` verbatim and ran `ws.my_api:typecheck` → "All checks passed!", closing `python-01`'s failure.

Not deployed to AWS (no infrastructure change).

### Follow-ups deliberately out of scope

- **The exact pin `boto3: '==1.43.83'` in `versions.ts` is the underlying trap.** This PR fixes the documented path, but a user who types the natural `add boto3` still lands in the unsatisfiable state, with an error naming neither cause nor fix. A compatible-release constraint (`~=1.43`) would make the floating and generator-added constraints co-satisfiable. Out of scope here: it changes output for every Python generator.
- **`py#mcp-server`'s `uv sync` running after its files are written is inherent to Nx** — the install is in the generator's returned callback, which Nx invokes only after flushing the tree to disk. A pre-flight constraint-compatibility check before writing would be the real fix.

### Issue # (if applicable)

Closes `docs-integrity-08`, `python-03`, `python-02`, `python-01`.

### Checklist

----


docs/src/content/docs/en/guides/fastapi.mdx
```diff
@@ -191,7 +192,15 @@ When your API is protected by authentication, your route handlers often need to
...
 
 <OptionFilter when={{ auth: 'iam' }} description="Identity extraction for IAM-authenticated APIs">
-For `IAM` authentication, we look up the caller in Cognito using the sub extracted from the API Gateway request context. Create `identity.py` alongside `main.py`:
+For `IAM` authentication, we look up the caller in Cognito using the sub extracted from the API Gateway request context, which requires `boto3`. Add it to your API project:
+
+<NxCommands commands={[`run my-api:add boto3${PY_VERSIONS.boto3}`]} />
+
+:::caution[Pin `boto3` to the version above]
+Other generators add `boto3` pinned to exactly this version. Since your workspace shares a single UV virtual environment and lockfile, adding it unpinned (`add boto3`) resolves the latest release and writes a `boto3>=<latest>` constraint, which `uv` cannot satisfy alongside the pin. All Python targets in the workspace then fail to resolve dependencies.
+:::
```
cogwirrel: remove this caution

docs/src/content/docs/en/guides/python-project.mdx
```diff
@@ -23,6 +23,10 @@ You can generate a new Python project in two ways:
 
 <GeneratorParameters generator="py#project" />
 
+The `type` option selects whether you get an application (a deployable project, for example a lambda function or container image) or a library (reusable code, imported by other projects). It defaults to `application`, so pass `--type=library` for a library:
+
+<NxCommands commands={['g @aws/nx-plugin:py#project my_library --type=library']} />
```
cogwirrel: undo the changes to this file, this is covered by generator parameters

packages/nx-plugin/src/py/project/schema.json
```diff
@@ -32,7 +32,10 @@
       "type": "string",
       "description": "Whether the project is an application or library",
       "default": "application",
-      "enum": ["application", "library"]
+      "enum": ["application", "library"],
+      "aliases": ["projectType"],
```
cogwirrel: it should not accept an alias - remove the `projectType` references in docs.

nx-plugin-for-aws: Done — dropped the `aliases` entry, so `type` is the only accepted spelling. There were no `projectType` references anywhere in docs (grepped `docs/src/content`, `docs/src/components`, `docs/src/i18n`) — the finding's `projectType` came from the lane spec's prose framing rather than the guide itself, and the guide only ever rendered the option via the generated parameters table. Also removed the corresponding alias unit test.

docs/src/content/docs/en/guides/fastapi.mdx
nx-plugin-for-aws: Removed. The pinned `add` command stays, sourced from `PY_VERSIONS.boto3`.

docs/src/content/docs/en/guides/python-project.mdx
nx-plugin-for-aws: Reverted — `python-project.mdx` is back to its original state, and the `type` option is covered by the generator parameters table. The new `x-prompt` on the schema is what closes the finding, since the reader is now asked rather than silently defaulted to `application`.

nx-plugin-for-aws: Correction to my earlier reply: this was **not** actually reverted at the time I said it was. I used `git checkout -- <file>`, which restores from HEAD — and HEAD already contained my commit, so the file matched and `git status` looked clean while the change was still in the branch. The translated copies carried it too.

Now genuinely reverted against `origin/main` across all 9 locales, and confirmed absent from the PR's file list. Sorry for the incorrect status.

nx-plugin-for-aws:
Rebased onto `main` (now includes #1207, #1199, #1202).

While confirming the `macOS Smoke Tests - standalone (6/6)` failure wasn't mine, I found something worth fixing here: **9 e2e call sites passed `--projectType=application` to `py#project`**, which is not an option on that generator. Nx drops unrecognised flags silently, so those runs were relying on the `application` default rather than the flag they named — exactly the `python-03` failure mode, in our own test suite. Now corrected to `--type=application` across `generator-matrix.ts`, `infra-none`, `license-check`, `license-sync`, `local-dev`, `rdb` and `standalone`. No behaviour change (both resolved to `application`), but the suite no longer contradicts the schema.

On the macOS shard: it failed twice on this branch, so I checked the cause rather than assuming it was flaky. It is `EINVAL: invalid argument, open .../node_modules/@aws/nx-plugin/package.json` inside pnpm's `removeBinsOfDependency`, while generating **`ts#nx-plugin`** — a macOS pnpm/filesystem issue. `py#project` itself passes in that same job (`✓ should generate and build 'py#project' in isolation`), and this PR touches only `py#project`'s schema, its spec, and docs. It passes on #1204 and #1205.

cogwirrel (approved):

PR #1199, opened 2026-09-04
fix(agentcore-gateway): expose a non-optional gatewayUrl accessor and correct the DCR proxy guide

nx-plugin-for-aws:
### Reason for this change

Five bug-bash findings against the DCR proxy / AgentCore Gateway guides and one Terraform module input, three of which break a user who copies the documented code verbatim:

- **`gateway-01`** (major) — every CDK snippet in `ts-dcr-proxy.mdx` stores the App Client secret with a plain `new secretsmanager.Secret(...)`. That uses the AWS-managed key, which the `ts#infra` project's default `checkov` target flags as `CKV_AWS_149`. `checkov` is a `dependsOn` of `infra:build`, so `nx run-many --target build --all` **fails** on the copied snippet. `CKV_AWS_149` is not in the generated skip list.
- **`gateway-02`** (major) — `upstreamUrl: gateway.gateway.gatewayUrl` fails `tsc` with TS2322: the CDK `Gateway` L2 types `gatewayUrl` as `string | undefined`, but `DcrProxyProps.upstreamUrl` is `string`.
- **`gateway-03`** (major) — the Terraform tab sourced the gateway module from `src/app/agentcore-gateway/<name>`, a path the generator never creates (it writes `src/app/gateways/<name>`), so `terraform init` fails with `Unreadable module directory`.
- **`terraform-04`** (major) — the `ts-rdb` Terraform tab told the reader to pass `environment_variables` to the API module while the generated module declares `env`, so `terraform validate` rejects it with `Unsupported argument`. This is the documented path for loading the RDS CA bundle when connecting to Aurora without RDS Proxy.
- **`gateway-04`** (docs) — five snippets used a `':my-scope/common-constructs'` import specifier (leading colon instead of `@`), which resolves to nothing.

### Description of changes

**`gateway-02` fixed in the construct, not the doc.** The shared gateway construct already knew about the optional `gatewayUrl` and worked around it internally in `addGateway`, but never exposed the result to users, so the documented usage could not typecheck. Rather than papering over that in the guide with a non-null assertion, `AgentCoreGateway` now exposes a non-optional `gatewayUrl` accessor carrying the CloudFormation-attribute fallback. `addGateway` reads it (dropping its duplicated inline fallback and gaining `readonly gatewayUrl: string` in the shape it requires of its argument), the vended app-level gateway constructs use it for their runtime-config entry, and the guide's `upstreamUrl: gateway.gatewayUrl` now compiles as printed. The `connection` runtime-config patch was updated to match.

**A deterministic migration (`latest/gateway-url-accessor`) carries existing workspaces across**, since the core construct is vended `KeepExisting` and would otherwise pair new app constructs with a stale core one. The app-level pass is gated on the core construct actually carrying the accessor — a diverged or absent core construct leaves the app constructs untouched and is reported via `nextSteps`, rather than pointing them at an accessor that does not exist.

**`gateway-01` fixed in the doc.** The plain `Secret` is what fails; the generated `checkov.yml` skip list is correct not to suppress a real finding. All four CDK snippets now pass an `encryptionKey: new kms.Key(..., { enableKeyRotation: true })`, matching how every other vended construct in the plugin encrypts at rest.

The remaining three are doc-only and need no migration: the Terraform module path, the `env` input name (in `ts-rdb.mdx` plus the three `*-rdb-ssl-requirements` snippets that share the error — found by sweeping every `hcl` block's module arguments against the generated modules' declared variables; no others were wrong), and the import specifier.

**Not addressed:** finding `gateway-05` (the Cedar policy guide describing the deployed policy name) — the prose change was reverted per review.

**Unit tests.** `pnpm nx run @aws/nx-plugin:test` — full suite green. Added a `gateway-url-accessor` migration spec (9 cases: applies to the vended shape, rewrites the app-level constructs, **preserves user content written between two runs**, is idempotent, skips and reports a diverged construct, leaves app constructs alone when the core construct could not be migrated or is absent, and leaves a user-added accessor alone) plus a generator assertion for the new accessor. Both gating tests were confirmed to fail without the gate. The one `agentcore-gateway` snapshot change is the intended `MyGateway: this.gatewayUrl` line.

**Repo.** `pnpm nx run-many --target lint --all --fix` and `pnpm nx run-many --target build --all` both green, including `docs:build` and `generate-migrations` (the migration is registered as `latest-gateway-url-accessor`).

**CDK end-to-end** (fresh `--iac=cdk` workspace, locally compiled plugin, `agentcore-gateway --auth=cognito` + `ts#dcr-proxy` + `ts#infra` + `ts#react-website#auth`, with the guide's "Fronting an AgentCore Gateway" snippet pasted in verbatim):

- `nx run-many --target build --all` — **passes**, where the pre-fix snippet failed at `infra:checkov`.
- `checkov`: **188 passed / 0 failed** (`CKV_AWS_149` now PASSED).
- `tsc` clean on `upstreamUrl: gateway.gatewayUrl`; the synthesized template resolves it to `{"Fn::GetAtt": ["MyGatewayCF5984F0", "GatewayUrl"]}`.

**Terraform end-to-end** (fresh `--iac=terraform` workspace, modules wired exactly as the guides instruct): `terraform init` and `terraform validate` both **succeed**. Both regressions were also reproduced against the same workspace to confirm the fixes were the right ones — `environment_variables` gives `An argument named "environment_variables" is not expected here`, and `app/agentcore-gateway/my-gateway` gives `Unreadable module directory ... no such file or directory`.

Stopped at `synth` / `validate` — nothing was deployed to AWS.

### Issue # (if applicable)

N/A — from a v1.0 bug bash. Closes findings `gateway-01`, `gateway-02`, `gateway-03`, `gateway-04` and `terraform-04`.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/agentcore-gateway.mdx
+++ b/docs/src/content/docs/en/guides/agentcore-gateway.mdx
@@ -179,9 +179,7 @@ Refer to the [AWS documentation on AgentCore Gateway policies](https://docs.aws.
 
 ### Adding a policy
 
-To add a policy, create a new `.cedar` file alongside `permit-all.cedar`. Each `.cedar` file in `policies/` must contain exactly **one** `permit` or `forbid` statement and is deployed as a single `AWS::BedrockAgentCore::Policy` resource. Files containing multiple statements produce `unexpected token 'forbid'` errors at deploy time — split them into separate files.
-
-Policy names must be unique account-wide, so the deployed name is never the filename alone: CDK derives it from the construct path, and Terraform PascalCases the filename (`permit-all.cedar` → `PermitAll`) then appends the gateway's random suffix. The policy's **description** is what identifies its source file — `Policy loaded from permit-all.cedar` — so match on that when inspecting deployed policies.
+To add a policy, create a new `.cedar` file alongside `permit-all.cedar`. Each `.cedar` file in `policies/` must contain exactly **one** `permit` or `forbid` statement and is deployed as a single `AWS::BedrockAgentCore::Policy` resource. The policy resource's name is derived from the filename: `permit-all.cedar` becomes `PermitAll` (kebab/snake-case is converted to PascalCase). Files containing multiple statements produce `unexpected token 'forbid'` errors at deploy time — split them into separate files.
 
 For example, to permit only a specific agent role to invoke a particular tool, create:
 
@@ -334,7 +332,7 @@ The generated infrastructure creates policies with `IGNORE_ALL_FINDINGS`: AgentC
 
 One ordering constraint still applies: a policy referencing `AgentCore::Action::"<target>___<tool>"` only validates once the target has registered that tool with the Gateway, which is why the generated infrastructure creates policies after Gateway targets.
 
-If a policy fails to deploy, CloudFormation surfaces the rejection as an opaque `Resource stabilization failed` error — run `aws bedrock-agentcore-control list-policies --policy-engine-id <id>` to retrieve the validator's `statusReasons`, which contain the actual reason. Identify the offending policy by its `description` (`Policy loaded from <file>.cedar`) rather than its suffixed `name`.
+If a policy fails to deploy, CloudFormation surfaces the rejection as an opaque `Resource stabilization failed` error — run `aws bedrock-agentcore-control list-policies --policy-engine-id <id>` to retrieve the validator's `statusReasons`, which contain the actual reason.
 
 ### Example: forbid one tool while keeping a broader permit
 
--- a/docs/src/content/docs/en/guides/agentcore-gateway.mdx
+++ b/docs/src/content/docs/en/guides/agentcore-gateway.mdx
@@ -133,7 +133,7 @@ The generated infrastructure consumes an existing Cognito user pool and client �
 The generated construct requires an `identity` prop supplying the user pool and client:
 
 ```ts {7,9-11} title="packages/infra/src/stacks/application-stack.ts"
-import { MyGateway, UserIdentity } from ':my-scope/common-constructs';
+import { MyGateway, UserIdentity } from '@my-scope/common-constructs';
 
 export class ApplicationStack extends Stack {
   constructor(scope: Construct, id: string, props?: StackProps) {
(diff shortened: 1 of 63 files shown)


docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -179,7 +179,9 @@ Refer to the [AWS documentation on AgentCore Gateway policies](https://docs.aws.
...
+To add a policy, create a new `.cedar` file alongside `permit-all.cedar`. Each `.cedar` file in `policies/` must contain exactly **one** `permit` or `forbid` statement and is deployed as a single `AWS::BedrockAgentCore::Policy` resource. Files containing multiple statements produce `unexpected token 'forbid'` errors at deploy time — split them into separate files.
+
+Policy names must be unique account-wide, so the deployed name is never the filename alone: CDK derives it from the construct path, and Terraform PascalCases the filename (`permit-all.cedar` → `PermitAll`) then appends the gateway's random suffix. The policy's **description** is what identifies its source file — `Policy loaded from permit-all.cedar` — so match on that when inspecting deployed policies.
```
cogwirrel: revert this change

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -332,7 +334,7 @@ The generated infrastructure creates policies with `IGNORE_ALL_FINDINGS`: AgentC
 
 One ordering constraint still applies: a policy referencing `AgentCore::Action::"<target>___<tool>"` only validates once the target has registered that tool with the Gateway, which is why the generated infrastructure creates policies after Gateway targets.
 
-If a policy fails to deploy, CloudFormation surfaces the rejection as an opaque `Resource stabilization failed` error — run `aws bedrock-agentcore-control list-policies --policy-engine-id <id>` to retrieve the validator's `statusReasons`, which contain the actual reason.
+If a policy fails to deploy, CloudFormation surfaces the rejection as an opaque `Resource stabilization failed` error — run `aws bedrock-agentcore-control list-policies --policy-engine-id <id>` to retrieve the validator's `statusReasons`, which contain the actual reason. Identify the offending policy by its `description` (`Policy loaded from <file>.cedar`) rather than its suffixed `name`.
```
cogwirrel: revert this change

packages/nx-plugin/src/migrations/latest/gateway-url-accessor/migration.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+};
+
+export default async function migration(
+  tree: Tree,
+): Promise<MigrationReturnObject> {
+  const nextSteps: string[] = [];
+
+  await migrateCoreConstruct(tree, nextSteps);
+  await migrateAppConstructs(tree);
```
cogwirrel: app constructs should only migrate if we successfully migrated the core constructs

nx-plugin-for-aws: Good catch — fixed in 75a26a8. `migrateCoreConstruct` now returns whether the accessor ended up in place, and the app-level pass runs only when it did. Previously a diverged (or absent) core construct would still have had its app constructs rewritten to `this.gatewayUrl`, leaving the workspace failing to compile. Added two tests covering it (diverged core, and no core construct vended); both fail without the gate.

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Reverted in 75a26a8 — the original wording is back verbatim. (Finding `gateway-05` is therefore not addressed by this PR; the import specifier fix on this page still stands.)

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Reverted in 75a26a8 — the original sentence is back verbatim.

cogwirrel (approved):

PR #1201, opened 2026-09-04
fix(ts#rdb#trpc-connection): widen tRPC Context, and correct the database connection guide examples

nx-plugin-for-aws:
### Reason for this change

Bug bash findings on the `connection` generator's database guide family. What a user hits:

- **`conn-data-01` (blocker, verified)** — `trpc-rdb.mdx`'s first usage step, "Register the Plugin", does not compile. The generator writes `middleware/<db>.ts` exporting `I<Db>Context` but never widens `Context` in `init.ts`, so `t.procedure.concat(createMyDbPlugin())` resolves to tRPC's `TypeError<"Context mismatch">` (TS2345). Substituting `publicProcedure` gives the identical error, and the required `init.ts` edit is documented nowhere in the repo, so a reader is stuck. The Smithy connection generator already edits `context.ts` for the user via GritQL — tRPC was the inconsistent one.
- **`conn-data-02` (major)** — both Smithy connection guides import `<Op>OperationInput` / `<Op>OperationOutput` from the generated SSDK. Those symbols do not exist (TS2305); the real names are `<Op>ServerInput` / `<Op>ServerOutput`.
- **`conn-data-03` (major)** — `py-fast-api-rdb.mdx`'s `list_examples` route omits `.scalars()`, so `.all()` yields SQLAlchemy `Row` tuples and the route 500s inside `jsonable_encoder`.
- **`conn-data-04` (major)** — same root cause in `py-agent-rdb.mdx` / `py-mcp-server-rdb.mdx`: `.model_dump()` on `Row` raises `AttributeError`. Both pass `build`/`lint`/`typecheck`, so a v1.0 user only finds out at runtime.
- **`conn-data-05` (major)** — `ts-agent-dynamodb.mdx` uses a `func:` property neither Strands `tool()` overload accepts (TS2769).
- **`conn-data-08` (minor)** — `ts-mcp-server-dynamodb.mdx` uses the deprecated `server.tool()` overload.
- **`conn-data-09` + `databases-05` (docs)** — all six Python connection guides plus `py-rdb.mdx` import `from my_scope.my_db import ...`, but the generators emit a flat `<scope>_<name>` module (e.g. `ws_py_db`), so even a reader who correctly substitutes their own scope and name hits `ModuleNotFoundError`. Folded together since it is one fix.

### Description of changes

[...]

PR #1202, opened 2026-09-04
fix(docs): converge the graph-builder shortcuts with their step-by-step routes

nx-plugin-for-aws:
### Reason for this change

Closes bug bash findings **tutorial-03** (blocker) and **quickstart-05** (minor), which share one root cause.

Module 1 Task 2 of the Dungeon Adventure tutorial offers two routes, described as reaching "the same starting point": **Quick** (copy the commands from the embedded read-only graph) and **Step By Step** (run each generator from the prose). They did not produce the same workspace, and every later module assumed the step-by-step names:

- The preset passed `--name=inventory-server` / `--name=story-agent`, which the step-by-step route omits so the generators derive their own. That single divergence moves the source directory (`src/inventory-server/` vs `src/mcp-server/`), the construct class (`InventoryServer` vs `InventoryMcpServer`), the Nx target prefixes (`inventory-server-dev` vs `mcp-server-dev`, `story-agent-dev` vs `agent-dev`) and the generated Python connection client (`InventoryServerClientStrands` vs `InventoryMcpServerClientStrands`).
- `ts#website game-ui` produced construct `GameUi`, while Task 3's stack code references `GameUI`.
- The preset omitted `--allowSignup=true`, so the quick route generated `selfSignUpEnabled: false` — Module 4's closing "sign up for a new account" could not be followed, and failed silently at runtime rather than at compile time.
- The preset omitted `--cognitoDomain=game-ui`, giving domain prefix `dungeon-adventure-game-ui-<acct>` instead of `game-ui-<acct>`.

A reader who took the Quick route hit two `TS2724`s on Task 3's stack code (`InventoryMcpServer`, `GameUI`), a missing `packages/inventory/src/mcp-server/server.ts` in Module 2, and missing `agent-dev` / `mcp-server-inspect` targets in Modules 2 and 3.

The Quick Start guide's shortcut diverged the same way, in a milder form: it omitted `--cognitoDomain=my-demo` (deriving a different prefix) and ordered website before API, where the prose runs API first.

### Description of changes

The graph is a faithful picture of the architecture, but it cannot express everything the prose passes — the casing a generated construct class follows, an option belonging to a follow-up generator, or deliberately leaving `--name` off so the generator derives one. So the fix is at the emission layer rather than by renaming graph nodes (which would make the diagram disagree with the tutorial's own vocabulary):

- **`EmitOptions` takes `overrides`**, keyed by node name (`docs/src/lib/graph-builder/commands.ts`). A `NodeOverride` can pin the name passed to the generator (`generatorName`, or `null` to leave it off), the component name the connection commands then use (`componentName`), option values on the node's own generator, and option values on a follow-up generator.
- **A `Preset` declares `overrides`**, and the read-only `EmbeddedGraph` applies them. The interactive builder deliberately does not: there a preset is an editable starting point, so the graph the user sees and edits stays the single source of the commands.
- **The `dungeon-adventure` preset pins** `GameApi` / `GameUI` / `DungeonDb`, `cognitoDomain=game-ui`, `allowSignup=true`, and `generatorName: null` for the agent and MCP server so both routes get the generator-derived `agent` / `mcp-server` names. The `quick-start` preset pins `cognitoDomain=my-demo`.
- **Node and edge order in both presets** now follows the order their guide's own steps run in, so the copied commands read the same way round as the prose.
- **The copied commands no longer carry the explanatory comments.** The embedded graph's script is only ever copied, never displayed, so `toScript` is called unannotated and the clipboard gets clean, runnable commands. The interactive builder still shows and copies the annotated script, where the comments are the point.

No plugin code changed and no generated output changed, so **no migration is needed** — this is docs-site code plus preset data.

**End-to-end convergence, the whole point of the fix.** Rendered the fixed preset's script, ran it into a fresh workspace, ran the tutorial's step-by-step commands into a second fresh workspace, and diffed the two. Every divergence the finding reports is gone; each item from the verifier's evidence list checked individually:

| Check | Quick route | Step-by-step |
|---|---|---|
| `packages/inventory/src/` | `mcp-server/` | `mcp-server/` |
| MCP construct class | `InventoryMcpServer` | `InventoryMcpServer` |
| Website construct class | `GameUI` | `GameUI` |
| `inventory` MCP targets | `mcp-server-dev`, `mcp-server-inspect`, … | identical |
| `story` agent targets | `agent-dev`, `agent-chat`, … | identical |
| `selfSignUpEnabled` | `true` | `true` |
| `domainPrefix` | `game-ui-${account}` | `game-ui-${account}` |
| Python client | `InventoryMcpServerClientStrands` | `InventoryMcpServerClientStrands` |

The only remaining differences between the two workspaces are insertion-order artifacts of the two routes running the generators in a different sequence (key order in `tsconfig.base.json`, `nx.json`, `pnpm-workspace.yaml`, `app/index.ts`) plus one local dev port swap between `inventory` and `dungeon-db` (8000/8001), which no tutorial content references.

**Task 3's stack code compiles on the quick route.** Wrote Module 1 Task 3's `application-stack.ts` (importing `GameApi`, `GameUI`, `InventoryMcpServer`, `StoryAgent`, `UserIdentity`) into the quick-route workspace and ran `nx run @dungeon-adventure/infra:compile` — succeeds, so both `TS2724`s are resolved.

**Tests.** Added eight `emitCommands` tests in `graph-builder.spec.ts` covering each override field: an overridden project name, the kebab-cased reference a connection then uses, a follow-up targeted at the overridden project, `generatorName: null` leaving `--name` off, a derived component referenced by the name the generator records, pinned node options, pinned follow-up options, and a node with no override left untouched.

```
npx vitest run --root docs     # 132 passed
npx biome check                # clean
```

Did not deploy to AWS; stopped at compile.

### Issue # (if applicable)

N/A — from the v1.0 bug bash (findings `tutorial-03`, `quickstart-05`).

### Checklist

----


nx-plugin-for-aws:
CI note: every check passes except `Smoke Tests - react-website`, which is a pre-existing repo-wide failure unrelated to this PR.

It fails on the assertion at `e2e/src/smoke-tests/react-website.spec.ts:132`:

```
```

That test shells out to the upstream `shadcn` CLI, which fetches component source from shadcn's remote registry, so the generated import specifier is not something this repo controls. It reproduces identically on rerun, and fails the same way on unrelated PRs open right now (#1201, #1199) while passing on the last green `main` run — consistent with an upstream registry change rather than anything in this branch.

This PR changes only `docs/src` (docs-site code and preset data); no plugin code or generated output is touched, so it cannot affect what the `ts#website` generator produces.

docs/src/lib/graph-builder/preset-guide-parity.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+        // A positional argument is the name, which the prose always passes as
+        // a flag.
+        options.name = arg;
+      }
+    }
+    return { generator: generator.replace(/^@aws\/nx-plugin:/, ''), options };
+  });
+};
+
+describe.each([
```
cogwirrel: this is overkill, delete `preset-guide-parity.spec.ts`

nx-plugin-for-aws: Done, deleted in 15803f4.

Agreed — it regex-parsed each guide's MDX and carried a hand-maintained `SCHEMA_DEFAULTS` table, so it had a real chance of failing on unrelated prose edits while not actually proving much. The override behaviour is covered by the eight `emitCommands` tests in `graph-builder.spec.ts`, and the convergence itself was verified end-to-end by building both routes into separate workspaces and diffing them (results in the PR description).

cogwirrel (approved):

PR #1203, opened 2026-09-04
fix(ts#website): keep the shadcn tsconfig reference unique when the generator is re-run

nx-plugin-for-aws:
### Reason for this change

The one user-facing code change is the `ts#website` idempotency fix below; the rest are
corrections to the website, auth and docs-site guide pages. Nine bug-bash findings in total.

**Generator bug — the change this PR is titled for**

- **`website-04`** (minor) — re-running `ts#website` appended a *second* project reference to `packages/common/shadcn` in `tsconfig.app.json`. The generator's de-dupe matched the `tsconfig.json` path it writes, but `nx sync` has meanwhile rewritten the entry to the specific `tsconfig.lib.json` it resolves to, so the filter missed it. The workspace was then left out of sync and the next `nx build` hard-failed with `The workspace is out of sync` until `nx sync` collapsed the duplicate — and the extra reference regrew on every subsequent re-run.

**Docs bugs**

- **`website-01`** (major) — *every* CDK `application-stack.ts` snippet in the website and auth guides used a 2-arg `constructor(scope, id)`, but `ts#infra` generates an `application-stage.ts` that calls the stack with 3 args. Copying any snippet into the file it is explicitly `title=`d as broke the build with `TS2554` — surfacing in `application-stage.ts`, a file the user never touched.
- **`website-03`** (major) — the Runtime Configuration section read as though `useRuntimeConfig` ships with every `ts#website`, but `src/hooks/useRuntimeConfig.tsx` is only added by `ts#website#auth` or a `connection` generator. On a fresh website the documented import was `TS2307`, and the guide's own FileTree didn't list it either, so prose and FileTree contradicted each other.
- **`terraform-03`** (major) — the `.checkov.yaml` workaround the guide documented did not work. The vended `checkov` target passes `--config-file`, which suppresses checkov's auto-discovery entirely, so the file was ignored and the build stayed red. The guide also named the wrong target (`test` rather than `checkov`).
- **`website-02`** (docs) — the "Navigating Between Pages" example was not parseable TypeScript: `navigate({ to: '/products/$id', { params: { id }} })` has a keyless nested object literal. 6 TS errors from a 4-line snippet, 4 of them parse errors.
- **`website-08`** (docs) — the Generator Output FileTree omitted six generated files (including `__root.tsx` and `routeTree.gen.ts`, central to the TanStack Router setup) and annotated `src/hooks` as Cloudscape-only.
- **`website-07`** (docs) — the auth guide said threat protection is `AUDIT` mode and pointed at the wrong reference.
- **`website-05`** (docs) — `nx translate docs -- --all` and `-- --dry-run` both fail out of the box with `ERROR No target languages configured`, because `targetLanguages` ships empty. Only the `--languages` form works, yet all three were listed as runnable as soon as the generator has run.
- **`website-06`** (docs) — the FileTree omitted the generated `src/content.config.ts` and `.gitignore`.

### Description of changes

**`ts#website` (generated output)**

`website-04`: added `mergeTsProjectDirReference` to `ts-project-utils.ts`, which treats *any* reference resolving inside the referenced project's directory as already present and **leaves it in place**. That is the right layer: the churn is caused by `nx sync` legitimately rewriting the path, so the generator has to recognise both forms rather than the exact string it happened to write.

Leaving the existing entry alone (rather than filtering it out and re-appending) also keeps reference **order** stable. Re-appending is what makes a re-run reorder references once a second project is referenced — which itself leaves the workspace out of sync — so there is a regression test pinning the order.

**Docs** — English only (`docs/src/content/docs/en/**`); the translated copies are regenerated by the Translate Documentation workflow.

- `website-01`: all 7 snippets across both guides now keep `props?: StackProps` and pass `props` to `super`.
- `website-03`: fixed in the guide. It now states which generators add `useRuntimeConfig` and why — a website with no auth and no connections has nothing *in* its runtime configuration, so there is nothing for the hook to read until one has run — and the FileTree no longer lists it.
- `terraform-03`: fixed in the prose, not the target. The generator already vends a user-curated `packages/<infra>/checkov.yml` and the `terraform#project` guide already documents it correctly — the react-website guide was pointing at a second, ignored file.
- `website-02`: corrected to `navigate({ to: '/products/$id', params: { id } })`, with the pseudo-code `await ...` and the un-imported `<Button>` replaced so the block is genuinely copy-pasteable.
- `website-08` / `website-06`: FileTrees reconciled against actual generated output.
- `website-07`: reworded to "audit-only" enforcement (matching AWS's own terminology) and pointed at the adaptive-authentication reference.
- `website-05`: the `--languages` form is presented first as what works on a fresh project, with `--all` / `--dry-run` shown after the step that makes them work, plus a `:::caution` naming the exact error.

**No migration**, per review. Worth recording the one consequence so it's a deliberate call: the generator fix stops the duplicate reference being *created*, but a workspace that already re-ran `ts#website` on an affected version still has both entries committed, and hits `The workspace is out of sync` on its next `build`. `nx sync` collapses it, and `ts#sync` already runs as part of the vended `build`, so the failure is loud and names the command to run — materially different from a silent divergence.

Per the finding's `verifyNotes`, the related `project.json` `lint.dependsOn` reordering sub-claim could not be reproduced by a verifier and is **not** addressed here.

**Repo checks**

- `pnpm nx run @aws/nx-plugin:test` — 4265 tests / 241 files, all passing. The generator snapshots are byte-identical to `main`, so this PR changes nothing about what users are vended beyond the tsconfig reference itself.
- `pnpm nx run-many --target lint --all --fix` — clean.

**Tests added** (3)

- The shadcn reference stays at exactly one across a re-run, **and** across a re-run after `nx sync` has rewritten it to `tsconfig.lib.json` — the state a real workspace is actually in, and the case the old de-dupe missed.
- Reference **order** is unchanged by a re-run when a second project is also referenced.

**End-to-end, in real workspaces built from the locally compiled plugin**

- **CDK** (`ts#website` + `ts#website#auth` + `ts#infra`): reproduced `website-04` on the pre-fix build (duplicate reference; `nx build` failing `The workspace is out of sync`), then confirmed the fix — a re-run is now a clean no-op touching **zero files**, one reference remains, and `build` passes. Also confirmed `ts#website` alone leaves `src/hooks` absent, matching the corrected docs.
- Pasted **each** corrected CDK snippet into `packages/infra/src/stacks/application-stack.ts` in turn (auth+website, WAF-disabled + `suppressRules`, `S3_MANAGED` + `CKV_AWS_158`, custom domain + certificate, runtime config) and ran `nx run-many --target build --all` / `nx run infra:build` on each — all green, `Passed checks: 79, Failed checks: 0` from checkov every time.
- Typechecked the corrected navigate example as a real component against real `products` / `products/$id` routes.
- Confirmed `dev` starts and serves.
- **Terraform** (`--iac=terraform`): reproduced `terraform-03` exactly — 2 × `CKV_AWS_145` failures, and the guide's `packages/infra/src/.checkov.yaml` leaving them **unchanged** at 2. Then confirmed the newly documented `packages/infra/checkov.yml` suppression genuinely suppresses (`Failed checks: 0`), for both `CKV_AWS_145` and `CKV_AWS_7`. Full `nx run-many --target build --all` green with the documented fix in place.
- Diffed both corrected FileTrees against `find`-ed generated output, for shadcn and Cloudscape.
- Verified the astro-docs claims against a generated docs site: `translate -- --all` fails with `ERROR No target languages configured`, the documented `--languages jp,ko --dry-run` works, and `src/content.config.ts` / `.gitignore` are generated.

Not deployed to AWS, per instructions — stopped at `synth` / `checkov` / `validate`.

### Issue # (if applicable)

N/A — findings from a v1.0 bug bash (`website-01` … `website-08`, `terraform-03`).

### Checklist

----


nx-plugin-for-aws:
### CI status: 70/71 green

The one failing check — **Smoke Tests - react-website** — is **pre-existing on `main` and not caused by this PR**.

The assertion is in `e2e/src/smoke-tests/react-website.spec.ts:131`, which checks that a component added by the shadcn CLI resolves its alias to the package-local `#lib/utils` specifier:

```
```

That covers the shadcn CLI's own alias resolution against `components.json`. This PR touches neither `components.json` nor the shared shadcn package's `imports`/`exports` maps.

The same failure, with a byte-identical assertion message, occurs on #1198 (`fix/py-project-type-and-boto3-pin`) — a Python-only change that cannot affect shadcn alias resolution:

- this PR: https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33854346658/job/100966748376
- #1198: https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33846370554/job/100940370293

Everything this PR does actually touch is green, including the checks most likely to catch a regression from it:

- **all 10 Unit Test shards** — 4239 tests
- **Smoke Tests - idempotency** — directly exercises the `website-04` re-run fix
- **all 5 Smoke Tests - migrate shards** — exercise the two new migrations
- **Smoke Tests - dungeon-adventure**, **local-dev**, **create-workspace**, **existing-workspace**, **infra-none**
- **Build**, **Analyze**, **CodeQL**, **Translate Documentation**

Happy to rebase once the shadcn smoke test is fixed on `main`, if you'd prefer a fully green run before merging.

docs/src/content/docs/en/guides/astro-docs.mdx
```diff
@@ -108,7 +125,7 @@ Edit `scripts/translate.config.json` to change:
 | `modelId` | Bedrock model to use for translations. |
 | `awsRegion` | AWS region the Bedrock client is configured with. Can also be set via `AWS_REGION`. |
 | `concurrency` | Max number of concurrent agent invocations. |
-| `translationCommitMessage` | Commit message marker for translation commits (default `docs: update translations`). |
+| `translationCommitMessage` | Commit message marker for translation commits. Not in the generated config — add the key to override its `docs: update translations` default. |
```
cogwirrel: undo this

docs/src/content/docs/en/guides/react-website-auth.mdx
```diff
@@ -119,6 +119,8 @@ Set `advanced_security_mode` to `ENFORCED` in the `user_pool_add_ons` block in `
 </Fragment>
 </Infrastructure>
 
+Unlike the `enableWaf` and MFA options below, the enforcement mode is not exposed as a prop or variable, so this edit applies to every `UserIdentity` in your workspace.
```
cogwirrel: remove this

docs/src/content/docs/en/guides/react-website.mdx
```diff
@@ -401,12 +416,21 @@ module "my_website" {
...
+```yaml title="packages/infra/checkov.yml" {9}
+# Defines the checks to skip entirely. Remove if your requirements differ.
 skip-check:
-  - CKV_AWS_145
+  # Lambda function suppressions
+  - CKV_AWS_115 # Concurrent execution limit
+  - CKV_AWS_116 # Dead Letter Queue
+  - CKV_AWS_117 # Lambda functions in VPC
+  - CKV_AWS_173 # Encrypt Lambda environment variables
+  - CKV_AWS_272 # Code signing
```
cogwirrel: remove the lambda ones from this example. use the same `# ...`

docs/src/content/docs/en/guides/react-website.mdx
```diff
@@ -401,12 +416,21 @@ module "my_website" {
...
+  - CKV_AWS_145 # S3 buckets are not KMS encrypted with encryption = "S3_MANAGED"
 ```
+
+The `checkov` target passes `--config-file`, which turns off `checkov`'s own discovery of a `.checkov.yaml` — a config file dropped anywhere else is ignored.
```
cogwirrel: remove this line

packages/nx-plugin/src/ts/react-website/app/generator.ts
```diff
@@ -683,6 +676,19 @@ export async function tsReactWebsiteGenerator(
...
+  // config, so the `RuntimeConfigProvider` and `useRuntimeConfig` hook ship with
+  // all of them. The provider falls back to an empty config when no
+  // `runtime-config.json` is served, so a website with no backend still renders.
+  //
+  // Runs after the format above so the route tree stays excluded from the
+  // formatting this generator performs too.
+  await runtimeConfigGenerator(tree, {
+    project: fullyQualifiedName,
+    preferInstallDependencies: false,
+  });
```
cogwirrel: I don't think we need to add runtime config here - it's added by the connection generator / auth generator as needed. remove this.

packages/nx-plugin/src/migrations/latest/react-website-shadcn-tsconfig-reference/migration.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (seen) {
+      return false;
+    }
+    seen = true;
+    return true;
+  });
+};
+
+export default async function migration(
```
cogwirrel: no need for this migration, remove it.

packages/nx-plugin/src/migrations/latest/react-website-runtime-config/migration.ts
```diff
@@ -0,0 +1,79 @@
...
+ * `src/hooks/useRuntimeConfig.tsx` with every website, rather than only when an
+ * auth or connection generator pulled them in. This migration adds them to a
+ * website that doesn't have them yet, wiring the provider into `main.tsx` the
+ * same way the generator does.
+ *
+ * How to write a migration:
+ * - https://nx.dev/docs/kb/migration-generators
+ * - What `nextSteps` means: https://nx.dev/docs/reference/devkit/MigrationReturnObject
+ */
+export default async function migration(
```
cogwirrel: No need for this migration, remove it

packages/nx-plugin/src/ts/react-website/app/generator.ts
nx-plugin-for-aws: Done — reverted, and I've moved the fix to the docs instead.

For context on why I'd reached for the generator: the finding (`website-03`) was that the Runtime Configuration section reads as though `useRuntimeConfig` ships with every `ts#website`, so on a fresh website the documented import is a `TS2307`. The guide's own FileTree didn't list the hook either, so prose and FileTree contradicted each other.

Your framing resolves it better than mine did: a website with no auth and no connections has nothing *in* its runtime config, so vending a provider to read an empty object is dead weight. I'd been treating the hook's absence as the bug; it's really the docs implying it's unconditional.

So the guide now says plainly which generators add it and why, and the FileTree no longer lists it:

> `src/hooks/useRuntimeConfig.tsx` and `src/components/RuntimeConfig` are added to your website by the `ts#website#auth` generator and the `connection` generators, since those are what put values into the runtime configuration. A website with neither has no runtime configuration to read, so the hook is not vended until one has run.

Verified in a real workspace: `ts#website` alone leaves `src/hooks` absent, and a re-run is now a clean no-op touching zero files.

packages/nx-plugin/src/migrations/latest/react-website-runtime-config/migration.ts
nx-plugin-for-aws: Removed — it only existed to carry the runtime-config generator change, which I've also reverted per your other comment, so there's nothing left for it to migrate.

packages/nx-plugin/src/migrations/latest/react-website-shadcn-tsconfig-reference/migration.ts
nx-plugin-for-aws: Removed.

Flagging one consequence so it's a deliberate call rather than something that surfaces later: the generator fix stops the duplicate reference being *created*, but a workspace that already re-ran `ts#website` on an affected version still has both entries committed in its `tsconfig.app.json`. Those users hit `NX The workspace is out of sync` on their next `build`.

That said, `nx sync` collapses it — and `ts#sync` already runs as part of the vended `build`, so the failure is loud, self-describing and tells them exactly what to run. That's a materially different situation from a silent divergence, which is what CONTRIBUTING.md's "assume a migration is needed" is really guarding against. Happy to leave it out on that basis.

I've noted it in the PR description so the omission is recorded.

nx-plugin-for-aws:
Heads-up from the bug-bash lane on #1209 (tutorial + quick-start): we have both fixed the shadcn tsconfig reference, and our fixes are incompatible — worth agreeing on one before either merges.

**Same symptom, different target path.**

- This PR's `mergeTsProjectDirReference` treats any `<projectDir>/tsconfig*` as already-referenced and, when absent, writes **`<projectDir>/tsconfig.json`**. That stops the duplicate appearing on re-run.
- #1209 changes the generator to write **`<projectDir>/tsconfig.lib.json`** — the path `@nx/js:typescript-sync` actually resolves the dependency to.

The distinction matters because sync doesn't just tolerate `tsconfig.json`, it treats it as **stale** and rewrites it. So with the `tsconfig.json` form still on disk, a freshly generated workspace fails `nx sync:check` until a build has been run and its output committed. I verified that end to end: on `main` a fresh `ts#website --ux=shadcn` reports

```
packages/game-ui/tsconfig.app.json:
  - Missing references: packages/common/shadcn/tsconfig.lib.json
  - Stale references: packages/common/shadcn/tsconfig.json
```

and after the `tsconfig.lib.json` change, `nx sync:check` on raw generator output is exit 0 / "The workspace is up to date". That property is load-bearing for #1209 — its whole point is that a generated workspace lints and builds non-interactively without a sync step first (bug-bash findings `tutorial-01`, `tutorial-02`, `quickstart-01`, all major).

**Suggested split, if you agree:** #1209 keeps the `tsconfig.lib.json` change (it needs it, and it's already validated end to end there), and this PR drops its shadcn-reference change and migration. Your `mergeTsProjectDirReference` de-dupe then becomes unnecessary, because writing the resolved path means sync has nothing to rewrite and no second reference is ever created. Happy to do it the other way round if you'd rather own it — in that case please write `tsconfig.lib.json` rather than `tsconfig.json`, and I'll drop mine.

One more thing #1209 found in the same area, which affects you either way: the generator's old filter-and-re-append (`references.filter(ref => ref.path !== shadcnRef).concat(shadcnRef)`) moves the shadcn reference to the **end** on every re-run. Once any other reference is present, that swaps two entries and the `idempotency` smoke test never converges. #1209 replaces it with the existing `mergeTsReferences`, which preserves position. Flagging in case your version keeps the re-append.

Also noting for both of us: cogwirrel has asked several bug-bash PRs (including this one) to drop their migrations, so whichever of us keeps the generator change should probably expect the paired migration to go too.

cc @cogwirrel for a tiebreak if you'd prefer to just call it.

docs/src/content/docs/en/guides/astro-docs.mdx
nx-plugin-for-aws: Done — reverted to the original wording: `Commit message marker for translation commits (default `docs: update translations`).`

docs/src/content/docs/en/guides/react-website-auth.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/react-website.mdx
nx-plugin-for-aws: Done — the example is now just the `# ...` elision plus the one relevant check, matching the sibling `CKV_AWS_7` block:

```yaml title="packages/infra/checkov.yml" {3}
skip-check:
  # ...
  - CKV_AWS_145 # S3 buckets are not KMS encrypted when encryption is S3_MANAGED
```

docs/src/content/docs/en/guides/react-website.mdx
nx-plugin-for-aws: Removed. The corrected `checkov.yml` path in the sentence above it is what makes the workaround actually work, so the `--config-file` explanation wasn't carrying its weight.

nx-plugin-for-aws:
Following up on my note above — this is settled, and **you own it**. cogwirrel reviewed #1209 and told me `this is covered by another pr - check and undo` on the shadcn reference, plus `remove this migration`. Both are now reverted on my branch: #1209 no longer touches `ts/react-website/app/generator.ts` and its `shadcn-website-tsconfig-lib-reference` migration is deleted. So there's nothing left to conflict with, and no need to change your approach on my account.

One thing I found while mine was still in that is **worth folding into your fix**, because it's a separate bug in the same block and it will land on you now:

The existing code doesn't merge the reference, it removes and re-appends it:

```ts
references: [
  ...(tsconfig.references ?? []).filter(ref => ref.path !== sharedShadcnTsconfigRef),
  { path: sharedShadcnTsconfigRef },
]
```

So the shadcn reference **moves to the end of the list on every re-run**. With only that one reference present it's invisible. As soon as a second reference exists — e.g. after a `connection` generator adds the API's — the two entries swap places each run and the workspace never converges. That was the `Smoke Tests - idempotency` failure on #1209:

```
Error: Generators were not idempotent — re-running the matrix changed the workspace:
 M packages/website-no-router/tsconfig.app.json
 M packages/website/tsconfig.app.json

   "references": [
-      "path": "../common/shadcn/tsconfig.lib.json"
+      "path": "../my-api/tsconfig.lib.json"
-      "path": "../my-api/tsconfig.lib.json"
+      "path": "../common/shadcn/tsconfig.lib.json"
```

Your `mergeTsProjectDirReference` already returns `references` unchanged when the project is `alreadyReferenced`, so I think you're clean on the de-dupe. The order only churns via the re-append, so if you've replaced that branch you may already be fine — worth confirming with the idempotency smoke test, since it's the only thing that catches it. The existing `mergeTsReferences` in the same file preserves position and is what I ended up using.

Also flagging the constraint I hit, in case it informs which path you write: with a `tsconfig.json`-shaped reference on disk, `@nx/js:typescript-sync` reports it as **stale** and rewrites it to `tsconfig.lib.json`, so a freshly generated shadcn workspace fails `nx sync:check` until a build has been run and committed:

```
packages/game-ui/tsconfig.app.json:
  - Missing references: packages/common/shadcn/tsconfig.lib.json
  - Stale references: packages/common/shadcn/tsconfig.json
```

Writing `tsconfig.lib.json` directly makes `nx sync:check` exit 0 on raw generator output. Entirely your call whether that matters for `website-04` — just passing on what I measured so you don't have to rediscover it.

cogwirrel (approved):

nx-plugin-for-aws:
### All 71 checks green

Including the two most relevant to this change — **Smoke Tests - idempotency** and all 10 **Unit Tests** shards — plus **Smoke Tests - react-website**, which now passes on top of #1207.

### Note for anyone else stuck in the queue

This run sat at `queued` for **2h21m with not one job started**, while runs created *after* it completed normally. That turned out not to be queue position: the CodeBuild runner labels embed the run ID and attempt number —

```
codebuild-nx-plugin-for-aws-runner-33876633492-1
```

— so those jobs are only ever claimable by a runner provisioned for that exact `<run-id>-<attempt>`. When that provisioning doesn't happen, the jobs cannot fall back to any other executor and wait indefinitely. Nothing drains them, because no runner exists that is allowed to take them.

`gh run rerun` is rejected while the run is live (`cannot be rerun; its workflow file may be broken` — misleading, the workflow is fine). What cleared it:

```bash
gh run cancel <run-id>       # the run must be terminal first
gh run rerun --failed <run-id>
```

The retry gets a fresh label (`...-2`), CodeBuild provisions against it, and the jobs start within minutes — 13 completed and 39 in flight inside ten minutes here, after two hours of nothing.

Worth distinguishing from genuine starvation, which looks different: jobs *do* start, just slowly, and the oldest queued run is the one at the front. The tell for a wedge is a run that is **older than runs which have already finished**, with zero jobs ever started. No push is needed to fix it, so it costs nothing to try before assuming capacity.

PR #1204, opened 2026-09-04
fix(connection): make smithy extension traits take effect on the generated client

nx-plugin-for-aws:
### Reason for this change

Six findings from the v1.0 bug bash, all in the `conn-web` family.

**`conn-web-01`** (filed blocker, verified and reclassified major) — the `extensions.smithy` vended into the Smithy model by the react↔smithy connection declared its traits as bare `@specificationExtension`. Smithy therefore derives each OpenAPI key from the trait's shape id and emits **namespace-prefixed** keys (`x-ws-query`, `x-ws-mutation`, `x-ws-cursor`), while the client generator only reads `x-query` / `x-mutation` / `x-cursor` (`open-api/utils/codegen-data/types.ts` `VENDOR_EXTENSIONS`).

All four documented traits were therefore **silent no-ops** — no warning, the model builds, the client generates, and the shape is the *opposite* of what the guide shows. A user gets `mutationOptions` where the guide promises `queryOptions`, and no feedback loop to tell them why.

The remaining five are doc drift — the generators moved and the examples didn't. Each one is a hard compile error for anyone copying the guide:

- **`conn-web-02`** — the FastAPI/Smithy error-handling, best-practices and type-safety examples read `err.error.message` / `.reason` / `.traceId` / `.validationErrors` and use `case 502:`. No 5XX aggregate type is generated, so `case 502:` is rejected outright, and none of those field names exist on the payloads the generators actually emit.
- **`conn-web-03`** — `queryOptions()` / `queryKey()` shown with no arguments. The generator makes the input positional and **required** for any operation that accepts input at all, even when every field is optional — which is exactly the list endpoint the examples use. It cascades through the whole optimistic-update example, since `queryKey()` feeds `cancelQueries` / `getQueryData` / `setQueryData` / `invalidateQueries`.
- **`conn-web-04`** — reported as the Smithy "Grouping Operations" example reading the wrong accessor names. **Verified as not a bug** — see below.
- **`conn-web-06`** — the react-trpc "Type Safety" example calls `trpc.users.create.useMutation()`, the removed `@trpc/react-query` hooks API. The generator wires up the `@trpc/tanstack-react-query` options proxy.
- **`conn-web-07`** — every `agentId` in the react-agui guide (`'agent'`, `'story'`, `'research'`) mismatches the kebab-cased ids the generated hook registers (`'story-agent'`). `agentId` is typed as a plain `string`, so this fails **silently at runtime** — CopilotKit just never resolves the agent. No type error, no console error.

### Description of changes

**Generated output — `extensions.smithy` (`conn-web-01`)**

Added the explicit alias to each trait declaration, pinning the emitted key to the name the codegen reads:

```smithy
@specificationExtension(as: "x-query")    structure query {}
@specificationExtension(as: "x-mutation") structure mutation {}
@specificationExtension(as: "x-cursor")   structure cursor { ... }
```

This is the right layer: the codegen side already handles both the scalar and object forms Smithy produces (`getCursorOptions` accepts `false` / `{enabled:false}` / `'prop'` / `{inputToken:'prop'}`), and the unprefixed names are the documented contract shared with the FastAPI `openapi_extra` path. Renaming the keys the generator reads would instead break every FastAPI user.

**Generated output — AG-UI chat component (`conn-web-07`)**

`conn-web-07` is doc drift, but a silently-wrong string is worth more than a doc fix, so I paired it with a generated-side change. Each `connection` run now vends `src/components/<agent-name>-chat.tsx`:

```tsx
export const StoryAgentChat = (props: Omit<CopilotChatProps, 'agentId'>) => (
  <CopilotChat agentId={STORY_AGENT_ID} {...props} />
);
```

It binds `agentId` to the id the hook registers (exported as `<AGENT_NAME>_ID`) and forwards every other `CopilotChat` prop, wrapping the themed `CopilotChat` so it picks up the website's `ux` theme. The id is now a single source of truth and callers never handle it, so the class of defect the finding describes — a hand-written literal that silently fails to resolve — is gone rather than merely detectable.

**Docs**

Corrected the examples in all four guides to what the generators produce — every snippet below was pasted into a real generated workspace and typechecked (see validation).

- `queryOptions({})` / `queryKey({})` where the operation declares input (`conn-web-03`).
- Real error payload field names and only the statuses the accompanying backend actually models; dropped `case 502:` and the non-existent `*5XXResponse` / `*Response` type names in favour of the generated `*ResponseContent` / model names (`conn-web-02`).
- tRPC type-safety example switched to `useMutation(trpc.users.create.mutationOptions())` (`conn-web-06`).
- The AG-UI guide now renders the vended `<StoryAgentChat />` / `<ResearchAgentChat />` instead of passing `agentId` strings (`conn-web-07`).

Four further mismatches surfaced while verifying the above, each of which broke the corrected snippet until fixed, so they're in scope as part of making these examples compile:

- The guide's own Smithy error example put `InvalidRequestError` on `@httpError(400)`, which collides with the `ValidationException` the service already declares on 400 — only one payload reaches the client. Moved to `422`.
- The FastAPI example named a Pydantic model `ValidationError`, colliding with FastAPI's own model for the 422 it adds to every endpoint. Renamed to `ValidationErrorDetails`.
- The generated client camelCases Python fields, so `field_errors` is read as `fieldErrors`; the examples now read the camelCased name.
- Both guides passed `onError` to `useQuery`, removed in TanStack Query v5. Restructured to narrow on `error.status` at the render site. `useMutation` still takes `onError`, so those blocks are unchanged.

**`conn-web-04` — not a bug, no change**

I could not reproduce it. The finding reports that tag groups key on the full operation name, so the guide's `api.items.listItems` should be `api.items.groupedListItems` — but `groupedListItems` was an artefact of the verifier renaming the operations to avoid a collision with the untagged `ListItems` in the same model. Built with the guide's *exact* operation names and the generated proxy is:

```ts
public items = { createItem: {...}, listItems: {...} };
public users = { listUsers: {...} };
```

`groupOperationsByTag` (`open-api/utils/codegen-data.ts:650-668`) keys each member by the camelCased **full** operation name, which for `ListItems` *is* `listItems`. The collision the finding predicts doesn't arise either: `untaggedOperations` filters out anything tagged, so a tagged operation appears only in its group and never at top level.

So the guide was already correct. I left the snippet as-is and instead made the naming rule explicit in both guides, since that's the non-obvious part (FastAPI's counterpart keys on the endpoint function name, and qualifies with the tag only when a name is duplicated across tags — also verified and documented).

**Migration**

`latest/smithy-extension-trait-aliases` adds the `as:` alias to each trait in a Smithy model project's `extensions.smithy`. Smithy IDL isn't a language GritQL parses, so the edit is anchored to the exact `@trait` + `@specificationExtension` + `structure <name>` triple the generator vends (CONTRIBUTING's stated exception for formats GritQL can't parse). A declaration already carrying an argument, or diverged from the vended shape, is skipped and reported via `nextSteps`. It only rewrites what it recognises, and is idempotent.

The traits were *silent* no-ops, so an existing workspace keeps ignoring them after upgrading unless something rewrites the file — which is the argument for carrying it. I raised this on the thread offering to drop it; approved as-is, so it stays.

No migration for the AG-UI change, per review.

**`conn-web-01` end to end, before and after.** Fresh workspace (`ts#react-website` + `#auth` + `ts#api --framework=smithy` + `connection`) against the locally compiled plugin, with all four documented trait examples applied to the model.

Before — every key namespace-prefixed, and the model builds clean with no warning:

```
GET  /paged-items       {'x-ws-cursor': {'inputToken': 'nextToken'}}
POST /search-items      {'x-ws-query': {}}
GET  /start-processing   {'x-ws-mutation': {}}
GET  /unpaged-items     {'x-ws-cursor': {'enabled': False}}
```

After:

```
GET  /paged-items       {'x-cursor': {'inputToken': 'nextToken'}}
POST /search-items      {'x-query': {}}
GET  /start-processing   {'x-mutation': {}}
GET  /unpaged-items     {'x-cursor': {'enabled': False}}
```

And all four traits then behave exactly as the guide documents — confirmed in the regenerated `options-proxy.gen.ts`: `searchItems` flipped to `{queryKey, queryOptions, queryFilter}`, `startProcessing` to `{mutationKey, mutationOptions}`, `listPagedItems` gained `infiniteQueryOptions`, `listUnpagedItems` lost it.

**Every corrected example compiles.** Generated all four stacks the guides need into one workspace — smithy api + react website + connection, fastapi + connection, trpc + connection, and both a `ts#agent` and a `py#agent` with `--protocol=ag-ui` + connections — modelled the backends from each guide's own accompanying model, then pasted the corrected snippets in and ran `nx run @ws/web:compile` (tsc). Green.

I first reproduced each finding the same way, so the fixes are verified against the failures rather than assumed. E.g. `conn-web-06` reproduced as `TS2339: Property 'useMutation' does not exist on type 'DecorateMutationProcedure<...>'`, and `conn-web-03` as `TS2554: Expected 1-2 arguments, but got 0`. `conn-web-07` was confirmed by reading the ids back out of the generated hooks (`'story-agent'`, `'research-agent'` — neither `'story'` nor `'research'`).

**Generated-side changes verified in the workspace**, not just in tests: recompiled the plugin, deleted and re-ran both AG-UI connections, and confirmed the hooks emit `STORY_AGENT_ID` / `RESEARCH_AGENT_ID`, that `story-agent-chat.tsx` and `research-agent-chat.tsx` are vended bound to them, and that every guide example using `<StoryAgentChat />` typechecks.

**Tests.** 7 unit tests for the Smithy migration, covering the vended shape, user content surviving, a diverged shape being skipped and reported, and idempotency. Three generator assertions: that `extensions.smithy` carries all three aliases and no bare `@specificationExtension`, that the AG-UI hook exports and uses its id constant, and that the vended chat component binds `agentId` and forwards the rest.

```
pnpm nx run @aws/nx-plugin:test -u      # 4273 passed / 242 files
pnpm nx run-many --target lint --all --fix
pnpm nx run-many --target build --projects=@aws/nx-plugin   # includes lint + test
```

Two snapshots updated, both the AG-UI hook picking up the exported constant — reviewed and intended.

MDX validated by compiling all four guides through `@mdx-js/mdx`; the full `nx build docs` is left to CI.

Not deployed to AWS — no infrastructure changed.

### Issue # (if applicable)

N/A — bug bash findings `conn-web-01`, `-02`, `-03`, `-06`, `-07`. `conn-web-04` investigated and closed as not-a-bug (rationale above).

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/connection/react-agui.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-agui.mdx
@@ -46,16 +46,17 @@ The generator creates a **single shared** `AguiProvider` component, one hook per
 - src
   - components
     - AguiProvider.tsx Single `CopilotKitProvider` for every AG-UI agent. Created on the first `connection` run and updated on subsequent runs to register each new agent.
+    - \<agent-name>-chat.tsx A themed `<AgentName>Chat` bound to this agent's id. One file per `connection` run.
     - copilot
       - index.tsx Re-exports `CopilotChat`, `CopilotSidebar` and `CopilotPopup` with slot defaults that match your website's `ux` (Cloudscape, Shadcn, or no theme at all).
       - *ThemeComponents*.tsx Per-slot theme components (e.g. `CloudscapeAssistantMessage.tsx`, `ShadcnChatInput.tsx`). Only vended when `ux` is `cloudscape` or `shadcn`.
   - hooks
-    - useAgui\<AgentName>.tsx Registers one AG-UI agent. One file per `connection` run.
+    - useAgui\<AgentName>.tsx Registers one AG-UI agent and exports its id as `<AGENT_NAME>_ID`. One file per `connection` run.
     - useSigV4.tsx SigV4 signing (IAM only)
 
 </FileTree>
 
-Running `connection` a second time for a different agent **adds a new `useAgui<AgentName>.tsx` hook** and updates `AguiProvider.tsx` to register both hooks — any custom edits you've made to the provider are preserved. `main.tsx` keeps its single `<AguiProvider>` wrapper — you never end up with nested providers.
+Running `connection` a second time for a different agent **adds a new `useAgui<AgentName>.tsx` hook and `<agent-name>-chat.tsx` component** and updates `AguiProvider.tsx` to register both hooks — any custom edits you've made to the provider are preserved. `main.tsx` keeps its single `<AguiProvider>` wrapper — you never end up with nested providers.
 
 The following dependencies are added to the root `package.json`:
 
@@ -119,17 +120,14 @@ Both Session ID and Thread ID are provided by the browser. To restrict each user
 
 ### Adding a Chat Interface
 
-Instantiate CopilotKit components with `agentId` to select which agent to use. The id is the agent's name — the same one you chose when you ran the <Link path="guides/ts-agent">`ts#agent`</Link> or <Link path="guides/py-agent">`py#agent`</Link> generator — and you can also find it in the generated hook file (e.g. the key returned from `packages/web/src/hooks/useAgui<AgentName>.tsx`).
-
-Import the chat components from the generated `./components/copilot` module so the theme that matches your website's `ux` is applied automatically:
+The generator vends a `<AgentName>Chat` component per connected agent, already bound to that agent's id and themed to match your website's `ux`. Drop it anywhere inside the `<AguiProvider>` wrapper:
 
 ```tsx
-import { CopilotChat } from './components/copilot';
+import { StoryAgentChat } from './components/story-agent-chat';
 
 function ChatPage() {
   return (
-    <CopilotChat
-      agentId="agent"
+    <StoryAgentChat
       labels={{
         welcomeMessageText: 'How can I help you today?',
         chatInputPlaceholder: 'Ask me anything...',
@@ -139,15 +137,24 @@ function ChatPage() {
 }
 ```
 
+It forwards every `CopilotChat` prop except `agentId`, so anything you can pass to `<CopilotChat />` works here too.
+
 ### Connecting Multiple AG-UI Agents
 
-Run the `connection` generator once per agent. Every agent registered via the shared `AguiProvider` is visible from anywhere in the app — instantiate a CopilotKit component with a different `agentId` to route each chat to the agent you want:
+Run the `connection` generator once per agent. Each run vends that agent's own chat component, so routing a chat to a particular agent is a matter of which component you render:
 
 ```tsx
-import { CopilotChat } from './components/copilot';
+import { StoryAgentChat } from './components/story-agent-chat';
+import { ResearchAgentChat } from './components/research-agent-chat';
 
-<CopilotChat agentId="story" />    {/* talks to StoryAgent */}
-<CopilotChat agentId="research" /> {/* talks to ResearchAgent */}
+<StoryAgentChat />    {/* talks to StoryAgent */}
+<ResearchAgentChat /> {/* talks to ResearchAgent */}
+```
+
+If you need the raw id — to call CopilotKit's own hooks, say — each generated hook exports it:
+
+```tsx
+import { STORY_AGENT_ID } from './hooks/useAguiStoryAgent';
 ```
 
 ## Customising the Look and Feel
@@ -164,12 +171,13 @@ The generator reads `metadata.ux` from your React website project and vends a th
 | `shadcn` | Assistant messages render in a `bg-muted` bubble with a `Sparkles` avatar; user messages render right-aligned in a `bg-primary` bubble with a `User` avatar. The input is a rounded `Textarea` + pill-shaped send/stop `Button` (Enter submits, Shift+Enter newlines). Uses shadcn primitives from the shared `common-shadcn` package. |
 | `none` (or anything else) | No theme — the module just re-exports the default CopilotKit components. |
 
-Import the themed components from the **local theme module** (not `@copilotkit/react-core/v2` directly) so the theme is applied automatically:
+The vended `<AgentName>Chat` components are already themed. For a chat you wire up yourself, import from the **local theme module** (not `@copilotkit/react-core/v2` directly) so the theme is applied automatically:
 
 ```tsx
 import { CopilotChat } from './components/copilot';
+import { STORY_AGENT_ID } from './hooks/useAguiStoryAgent';
 
-<CopilotChat agentId="agent" />
+<CopilotChat agentId={STORY_AGENT_ID} />
 ```
 
 The theme is applied as slot defaults, so any slot you explicitly pass still wins — you keep full control whenever you need a one-off override.
@@ -188,8 +196,7 @@ For example, to drop in your own user-message renderer while keeping the rest of
 Per-chat overrides still work alongside the theme — anything you pass as a slot prop overrides the themed default:
 
 ```tsx
-<CopilotChat
-  agentId="agent"
+<StoryAgentChat
   // style the input and its children
   input={{
     textArea: 'text-blue-600',
@@ -208,25 +215,23 @@ Per-chat overrides still work alongside the theme — anything you pass as a slo
 Any slot can take a React component instead of a className, so you can replace the default entirely:
 
 ```tsx
-import { CopilotChat } from './components/copilot';
+import { StoryAgentChat } from './components/story-agent-chat';
 
-const MySendButton: React.FC<{ onClick: () => void }> = ({ onClick }) => (
+const MySendButton: React.FC<React.ButtonHTMLAttributes<HTMLButtonElement>> = ({
+  onClick,
+}) => (
   <button onClick={onClick} className="my-send-btn">
     Send
   </button>
 );
 
-<CopilotChat
-  agentId="agent"
-  input={{ sendButton: MySendButton }}
-/>;
+<StoryAgentChat input={{ sendButton: MySendButton }} />;
 ```
 
 Deeper overrides follow the same shape — e.g. replace just the copy button on assistant messages:
 
 ```tsx
-<CopilotChat
-  agentId="agent"
+<StoryAgentChat
   messageView={{
     assistantMessage: {
       copyButton: ({ onClick }) => <button onClick={onClick}>Copy</button>,
--- a/docs/src/content/docs/en/guides/connection/react-fastapi.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-fastapi.mdx
@@ -248,7 +248,7 @@ const createItem = useMutation({
   onSettled: () => {
     // This will run when the mutation completes (success or error)
     // Good place to invalidate queries that might be affected
-    queryClient.invalidateQueries({ queryKey: api.listItems.queryKey() });
+    queryClient.invalidateQueries({ queryKey: api.listItems.queryKey({}) });
   }
 });
 ```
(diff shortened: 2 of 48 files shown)


docs/src/content/docs/en/guides/connection/react-agui.mdx
```diff
@@ -119,17 +119,18 @@ Both Session ID and Thread ID are provided by the browser. To restrict each user
...
 ### Adding a Chat Interface
 
-Instantiate CopilotKit components with `agentId` to select which agent to use. The id is the agent's name — the same one you chose when you ran the <Link path="guides/ts-agent">`ts#agent`</Link> or <Link path="guides/py-agent">`py#agent`</Link> generator — and you can also find it in the generated hook file (e.g. the key returned from `packages/web/src/hooks/useAgui<AgentName>.tsx`).
+Instantiate CopilotKit components with `agentId` to select which agent to use. Each generated hook exports the id it registers as a `<AGENT_NAME>_AGENT_ID` constant — import that rather than writing the string by hand, since `agentId` is typed as a plain `string` and CopilotKit silently fails to resolve an id that doesn't match at runtime.
 
 Import the chat components from the generated `./components/copilot` module so the theme that matches your website's `ux` is applied automatically:
 
 ```tsx
 import { CopilotChat } from './components/copilot';
+import { STORY_AGENT_AGENT_ID } from './hooks/useAguiStoryAgent';
```
cogwirrel: The refactor is nice but let's go a step further.

Let's generate a `./components/story-agent-chat.tsx` which vends `<StoryAgentChat />` which wraps `<CopilotChat agentId={STORY_AGENT_ID} />` and forwards other props, and update docs accordingly.

docs/src/content/docs/en/guides/connection/react-agui.mdx
```diff
@@ -139,15 +140,21 @@ function ChatPage() {
 }
 ```
 
+:::note[Agent ids]
+The registered id is the kebab-cased name of the agent component, so an agent generated with `--name=StoryAgent` is registered as `story-agent`. The constant is derived from the same name, so `useAguiStoryAgent.tsx` exports `STORY_AGENT_AGENT_ID`.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
```diff
@@ -943,11 +951,15 @@ from pydantic import BaseModel
...
 
-class ValidationError(BaseModel):
+class ValidationErrorDetails(BaseModel):
     message: str
     field_errors: list[str]
 ```
 
+:::caution[Reserved model names]
+Avoid naming a model `ValidationError` — FastAPI ships its own `ValidationError` for the `422` response it adds to every endpoint. Two models with the same name collide in the OpenAPI spec, and the generated client gets only one of them.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
```diff
@@ -1088,6 +1089,10 @@ function ItemComponent() {
 }
 ```
 
+:::note[Query errors]
+TanStack Query v5 removed `onError` from `useQuery`, so query errors are handled by narrowing on `error.status` where you render. `useMutation` still takes `onError`.
+:::
```
cogwirrel: Remove this

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
```diff
@@ -1196,32 +1201,30 @@ function ItemList() {
...
         <li key={item.id}>{item.name}</li>
       ))}
     </ul>
   );
 }
 ```
 
+:::note[Passing an input]
+`queryOptions` and `queryKey` take the operation's input as their first argument whenever the operation declares any input at all, even when every parameter has a default. Pass `{}` for a list endpoint you're calling with no filters.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/connection/react-smithy.mdx
```diff
@@ -938,6 +976,10 @@ function ItemComponent() {
 }
 ```
 
+:::note[Query errors]
+TanStack Query v5 removed `onError` from `useQuery`, so query errors are handled by narrowing on `error.status` where you render. `useMutation` still takes `onError`.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/connection/react-smithy.mdx
```diff
@@ -1064,14 +1105,18 @@ function ItemList() {
...
         <li key={item.id}>{item.name}</li>
       ))}
     </ul>
   );
 }
 ```
 
+:::note[Passing an input]
+`queryOptions` and `queryKey` take the operation's input as their first argument whenever the operation declares any input at all, even when every field is optional. Pass `{}` for a list endpoint you're calling with no filters.
+:::
```
cogwirrel: remove this

packages/nx-plugin/src/migrations/latest/agui-hook-export-agent-id/migration.ts
```diff
@@ -0,0 +1,97 @@
...
+ * - https://nx.dev/docs/kb/migration-generators
+ * - What `nextSteps` means: https://nx.dev/docs/reference/devkit/MigrationReturnObject
+ *
+ * Guardrails:
+ * - Only touches a `src/hooks/useAgui<Name>.tsx` whose returned registry still
+ *   has the single string-literal-keyed shape the generator vends. A hook a user
+ *   has reshaped is left alone and reported via `nextSteps`.
+ * - Idempotent: a hook already exporting the constant no longer matches.
+ */
+export default async function migration(
```
cogwirrel: no need for a migration for this, remove

nx-plugin-for-aws: Removed — migration, spec, snapshot and the `migrations.json` entry are all gone.

Could I get your read on the other migration in this PR, `latest/smithy-extension-trait-aliases`? I want to apply the same rule rather than guess at the boundary.

My reason for thinking it might be different: the vended `extensions.smithy` declared its traits as bare `@specificationExtension`, so Smithy emitted `x-<namespace>-query` while the codegen only reads `x-query`. All three traits were **silent** no-ops — the model builds, the client generates, no warning anywhere. So unlike the `agentId` case (where a user hitting a broken chat would go and read the hook), a user here has no signal that anything is wrong, and an existing workspace keeps silently ignoring the traits after upgrading unless something rewrites the file.

Happy to drop it if you would rather this batch stay migration-free — just say and I will.

docs/src/content/docs/en/guides/connection/react-agui.mdx
nx-plugin-for-aws: Done. Each `connection` run now vends `src/components/<agent-name>-chat.tsx` exporting `<AgentName>Chat`, which binds `agentId` to the exported `<AGENT_NAME>_ID` and forwards the rest via `Omit<CopilotChatProps, 'agentId'>`. It wraps the themed `CopilotChat` from `./copilot`, so it picks up the `ux` theme automatically.

Also took your `STORY_AGENT_ID` naming over my `STORY_AGENT_AGENT_ID` — the latter stuttered.

Docs updated throughout: the chat-interface example is now `<StoryAgentChat labels={...} />`, the multi-agent example renders `<StoryAgentChat />` / `<ResearchAgentChat />` rather than passing ids, and the slot/Tailwind examples use the vended component too. The raw-id import is kept as a one-liner for anyone calling CopilotKit's own hooks. Generator test added asserting the binding and the prop forwarding.

docs/src/content/docs/en/guides/connection/react-agui.mdx
nx-plugin-for-aws: Removed. Moot now anyway — with the vended `<AgentName>Chat` component the reader never handles the id themselves.

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
nx-plugin-for-aws: Removed. The corrected code above it already shows the right shape.

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
nx-plugin-for-aws: Removed. The corrected code above it already shows the right shape.

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
nx-plugin-for-aws: Removed. The corrected code above it already shows the right shape.

docs/src/content/docs/en/guides/connection/react-smithy.mdx
nx-plugin-for-aws: Removed. The corrected code above it already shows the right shape.

docs/src/content/docs/en/guides/connection/react-smithy.mdx
nx-plugin-for-aws: Removed. The corrected code above it already shows the right shape.

cogwirrel (approved):

PR #1205, opened 2026-09-04
fix(ts#project): preserve target wiring added by other generators on re-run

nx-plugin-for-aws:
### Reason for this change

Closes bug bash finding **`ts-core-02`** (blocker, independently verified).

`ts#lambda-function` adds `bundle` to its host project's `build` and `assemble` `dependsOn` — this is what `typescript-infrastructure.mdx` relies on when it says `synth` depends on `^assemble` so "synthesizing builds the artifacts your stacks deploy, such as the Lambda bundles".

Re-running `ts#project` with the same name rewrote `project.json` from the base template and dropped `bundle` from both target lists. The `bundle` target definition itself survived, so nothing looked wrong; `assemble` simply stopped producing `dist/.../bundle`. The failure is silent — `nx run-many --target build --all` still passes from a wiped `dist` — until a clean `nx synth`, which fails with:

```
«CannotFindAsset» Cannot find asset at .../dist/packages/my-library/bundle/lambda/my-function
```

This violates the `CONTRIBUTING.md` §"Generator Idempotency" contract, which requires framework-owned config (project targets) to "update in place / merge, keyed so it overwrites rather than duplicates".

### Description of changes

**`utils/nx.ts` — new `mergeTarget` helper.** Merges the configuration a generator owns into a project's existing target:

- `dependsOn` is **unioned** — existing entries keep their order, desired ones are appended only when absent (deduplicated with the same deep equality `addDependencyToTargetIfNotPresent` uses, so the object dependency form is handled too). A re-run therefore neither drops an entry nor doubles one up.
- Every other key the caller passes wins, since those are the ones the generator owns.
- The result is passed through the existing `normalizeTargetKeyOrder`, so a target authored on the first run serializes identically to the same target merged into on a re-run.

**`ts#project` (`ts/lib/generator.ts`)** now merges rather than assigns `compile`, `build` and `assemble`. This is the right layer: `ts#project` is the generator that owns those three targets' *own* configuration, but it is not the only contributor to their `dependsOn` — `addArtifactDependencyToTargets` (used by `bundle`, `docker`, `synth`, `operations`, the AgentCore `<name>-package` targets, the RDB bundles) writes into the same lists from a dozen call sites. Fixing it in `ts#project` fixes it for all of them at once, rather than teaching each contributing generator to re-assert itself.

**No migration.** One was included initially, but per review it was removed: the window in which a user both re-ran `ts#project` over a project with an artifact target *and* has not since re-run the contributing generator is narrow enough that it is unlikely to have been hit in practice.

### Other project generators checked

Per the review request I checked whether the same replace-instead-of-merge pattern affects the other project generators that re-run over an existing project. **None of them do**, so this PR is scoped to `ts#project`:

- **`py#project`** — already guards the `compile`/`build`/`assemble` derivation behind `if (!projectConfiguration.targets.compile)`, so a re-run never touches those three. I verified by generating a Python project, adding a `py#lambda-function` to it, and re-running `py#project`: `build.dependsOn` keeps `bundle` and `assemble.dependsOn` keeps `bundle-x86`. The `project.json` **key order and JSON content are both unchanged**; the only difference is whitespace reflow (`updateProjectConfiguration` re-serializes arrays expanded, and `py#project` does not call `formatFilesInSubtree`). That is a cosmetic formatting issue with a materially different root cause, not this bug — noted here rather than fixed.
- **`ts#react-website`**, **`smithy#project`**, **`terraform#project`** — all guard their target authoring behind an existence check already.

Unit tests added:

- `ts/lib/generator.spec.ts` — per §"Testing expectations", a test that generates a project, runs `ts#lambda-function` against it, re-runs `ts#project` with the same name, and asserts `build.dependsOn` is exactly `['lint', 'compile', 'test', 'bundle']` and `assemble.dependsOn` exactly `['compile', 'bundle']` — so the `bundle` wiring survives **exactly once**: present, and not duplicated by the merge. A second test re-runs again and asserts `project.json` is byte-identical, so repeated re-runs can neither drop nor grow the list.
- `utils/nx.spec.ts` — seven cases for `mergeTarget`: authoring a fresh target, keeping another generator's dependency, adding a missing one without duplicating, deduplicating the object dependency form, letting the generator own every non-`dependsOn` key, identical key order whether authored or merged, and idempotence under repeated application.

Commands run:

- `pnpm nx run @aws/nx-plugin:test` — **241 files, 4272 tests, all passing**, with no snapshot changes (the generated `project.json` for a fresh project is unchanged; only the re-run path differs).
- `pnpm nx run-many --target lint --all --fix` — clean.

End-to-end in a real workspace, against the locally compiled plugin, reproducing the finding's exact repro steps:

1. `ts#project --name=my-library`, `ts#lambda-function --project=@ws/my-library --name=MyFunction`, `ts#infra --name=infra`, instantiated `MyLibraryMyFunction` in `application-stack.ts`, committed as the baseline.
2. Re-ran `ts#project --name=my-library` → **`git diff` is completely empty**. Before this change the same command emitted `UPDATE packages/my-library/project.json` and stripped both `bundle` entries.
3. `rm -rf dist .nx/cache && nx synth @ws/infra` → **succeeds** (previously `CannotFindAsset`).
4. `rm -rf dist .nx/cache && nx run-many --target build --all` → **succeeds**, checkov included.

Stopped at `synth` — nothing was deployed to AWS.

### Issue # (if applicable)

Bug bash finding `ts-core-02`.

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/ts/lib/generator.spec.ts
+++ b/packages/nx-plugin/src/ts/lib/generator.spec.ts
@@ -5,8 +5,8 @@
 import { readJson, readNxJson, type Tree, writeJson } from '@nx/devkit';
 import uniqBy from 'lodash.uniqby';
 import {
-  declaredNames,
   declareDependencies,
+  declaredNames,
 } from '../../utils/declared-dependencies.js';
 import { expectHasMetricTags } from '../../utils/metrics.spec.js';
 import {
@@ -15,6 +15,7 @@ import {
 } from '../../utils/shared-constructs.js';
 import { createTreeUsingTsSolutionSetup } from '../../utils/test.js';
 import { ownedDependencies } from '../../utils/version-upgrade-migration/owned-dependencies.js';
+import { tsLambdaFunctionGenerator } from '../lambda-function/generator.js';
 import {
   DEPENDENCIES,
   TS_LIB_GENERATOR_INFO,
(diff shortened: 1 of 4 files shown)


packages/nx-plugin/src/migrations/latest/restore-artifact-target-wiring/migration.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+const hasRewritableDependsOn = (
+  project: ProjectConfiguration,
+  target: string,
+): boolean => {
+  const dependsOn = project.targets?.[target]?.dependsOn;
+  return dependsOn === undefined || Array.isArray(dependsOn);
+};
+
+export default async function migration(
```
cogwirrel: no need for a migration here, unlikely user encountered this. remove the migration and associated files

nx-plugin-for-aws: Done — removed the migration and all its associated files (`migration.ts`, `migration.spec.ts`, `metadata.json`). The generated `migrations.json` entry drops out automatically since the manifest is assembled from folder discovery.

The fix itself is unchanged: `mergeTarget` in `utils/nx.ts` plus `ts#project` merging instead of assigning `compile`/`build`/`assemble`, with the test that adds a lambda function, re-runs `ts#project`, and asserts the `bundle` wiring survives exactly once.

cogwirrel (approved):

PR #1206, opened 2026-09-04
fix(ts#infra): stop the checkov gate passing on stale cached results

nx-plugin-for-aws:
### Reason for this change

Fixes 14 findings from the v1.0 bug bash covering the IaC guide pages. Three are bugs in generated output; the rest are guide examples a reader cannot follow as written.

**Generated output**

- **`ts-core-03`** (major) — *a security gate that doesn't gate.* The CDK `checkov` target declared `inputs: ["{workspaceRoot}/dist/{projectRoot}/cdk.out"]`. Nx resolves a fileset input against the files it tracks and `dist` is gitignored, so that input matched nothing and hashed to a constant. Once the scan had been cached every later run was a hit: the reporting agent added the guide's own `new Bucket(this, 'MyBucket')` and `nx build` reported success, while `--skip-nx-cache` on the identical `cdk.out` gave **6 failed HIGH checks** and the restored `checkov_report.json` recorded `resource_count: 0`.
- **`terraform-08`** (major) — `validate` is documented under "Common Targets" ahead of the bootstrap section, but on an application project it `dependsOn` `init`, which runs `scripts/init.ts` and hard-fails with a raw S3 `NoSuchBucket` trace until `bootstrap` has created the state bucket. `lint`, `test` and `checkov` all work pre-bootstrap, so the first target the guide invites the reader to try was the only unrunnable one.
- **`ts-core-10`** (docs, expanded to a generator fix on review) — `deploy` used express mode without `stageConfig` but waited for stabilization with it. Express mode returns before resources have stabilized, so a target that can name a production stage should not use it. The generator is now consistent: `--express` on `deploy-sandbox` only.

**Guide examples**

- **`terraform-02`** (major) — the guide's own "Implementing your Terraform Infrastructure" `main.tf` fails `nx build` with 7 Checkov failures with no note that it would. Per review, the example stays a bare bucket and the guide now says so.
- **`terraform-06`** (major) — the `prod` plan configuration wrote `prod.tfplan` from `-var-file=env/dev.tfvars`, so following the guide deploys dev-configured infrastructure into the prod state.
- **`ts-core-04`** (major) — the "Runtime Code" `bundle` target omitted `cwd`, so the documented rolldown target fails with `UNRESOLVED_ENTRY`.
- **`ts-core-05`** (major) — the `getAppConfig` snippet does not typecheck: 4 errors under the default strict tsconfig, so `compile` and `build` fail.
- **`terraform-05`**, **`ts-core-07`**, **`ts-core-08`**, **`ts-core-09`**, **`ts-core-11`**, **`ts-core-12`**, **`ts-core-13`** — doc/minor mismatches, detailed below. (**`terraform-07`** was reverted on review and is not addressed here.)

### Description of changes

**Generated output (3 changes, each with a deterministic migration)**

- `ts#infra`: the `checkov` target's `inputs` become `["default"]`, matching its `synth` sibling. `default` carries the project's own sources plus the transitive `dependentTasksOutputFiles` entry from `nx.json`, which Nx resolves against the upstream `synth` task — so the hash follows the template the scan actually reads. Fixed at the generator rather than by `"cache": false` because the scan *is* soundly cacheable; only its declared inputs were wrong.
- `ts#infra`: `deploy` no longer passes `--express`, so express mode is `deploy-sandbox`'s alone across both the `stageConfig` and default variants. `deploy` names whichever stage you give it, including production, and express mode returns before resources have stabilized — `--express` remains available per invocation. (Added on review; see the thread on `typescript-infrastructure.mdx`.)
- `terraform#project`: `validate` runs its own `terraform init -backend=false` ahead of `terraform validate` instead of depending on the backend-configured `init`, exactly mirroring what `test` already does. `TF_DATA_DIR` keeps that `.terraform` out of `src` so it never races the backend-configured targets, `TF_PLUGIN_CACHE_DIR` links providers rather than re-downloading them, and `outputs: []` keeps a cache hit to what it asserts (the data dir is symlinks into the shared cache, not an artifact). Fixed at the generator rather than documenting the ordering, because `validate` needs no credentials to do its job — the backend was incidental.

**Migrations** — `latest/checkov-scan-cache-inputs` and `latest/terraform-validate-without-backend`, both new and deterministic.

**Why these two need migrations.** Both change the *shape of a vended Nx target*, not source a user could reasonably reconcile themselves:

- The `checkov` one closes a **security hole in existing workspaces**. Every workspace generated before this fix has a gate that reports success on infrastructure it never scanned. Fixing only the generator leaves those users believing `build` checks their infrastructure when it does not — and the failure mode is silent, so nobody would think to look. There is no realistic prospect of a user hand-patching `inputs` on a target they have no reason to suspect.
- The `validate` one turns a target that **cannot run at all** on a fresh application project (raw S3 `NoSuchBucket`) into one that works. Reaching the fixed state by hand means knowing to swap a `dependsOn` for three commands plus two env vars and an `outputs: []` — generator-authored detail, not a user edit.

In both cases the target lives in `project.json`, which the generator writes once and then leaves alone, so without a migration a workspace generated last month silently diverges from one generated today — exactly the gap `CONTRIBUTING.md` says to close.

Each matches the exact shape the pre-fix generator produced before writing, skips and reports a customised target via `nextSteps` rather than clobbering it, and is a no-op on re-run and on a project generated after the fix. Neither changes deployed resources, so there is no deployment transition to exercise and no data-loss or in-place-update risk of the kind `CONTRIBUTING.md` warns about for IaC migrations.

For the express-mode change I amended the existing **unreleased** `latest/infra-deploy-express-mode` rather than adding a second migration that would undo it. `v1.0.0-rc.96` has shipped `--express` on `deploy`, so existing workspaces carry it and the migration has to *remove* it there while still *adding* it to `deploy-sandbox`. Folding both directions into the one entry means it expresses the end state and nobody upgrading across both gets add-then-remove churn. It keeps `--no-express` as a deliberate opt-out, reports a customised command instead of rewriting it (with the message matched to the direction of travel), and adds no `nextSteps` entry for a custom `deploy` that already waits — there is nothing for the user to do there.

**Guide pages**

- `terraform-project.mdx` — the `main.tf` example stays a bare bucket, with a note that it would fail the Checkov scan (per review — the earlier compliant-example approach was reverted); `prod.tfvars` uses the `aws_region` the root module declares (with a note naming the variables the generator declares); and the `prod` plan command uses `env/prod.tfvars`. The `validate` and `TF_PLUGIN_CACHE_DIR` prose changes were reverted on review — the `validate` generator fix stands on its own without prose, and `terraform-07` is dropped.
- `typescript-infrastructure.mdx` — the `stageConfig` FileTree matches the generated layout (`resolve-stage.ts`, the `src/infra/` nesting, the two READMEs), and the guide shows the flag being passed. The `deploy` express-mode prose is a single unconditional statement again, now that the generator behaves the same either way; the `deploy-ci` paragraph no longer claims it differs from `deploy` by never using express mode, since that is now only true of `deploy-sandbox`.
- `typescript-project.mdx` — the `bundle` target gains `cwd`, `inputs` and `dependsOn`, and the rolldown config gains `tsconfig`/`platform`, aligning it with `docker-bundling.mdx` and with what `ts#lambda-function` vends (the explanatory sentence was removed on review); the `hello.spec.ts` example is formatted the way the vended `biome.json` wants; the FileTrees list `README.md` and the root-level changes.
- `docker-bundling.mdx` — creating `.trivyignore` is documented as the prerequisite it is, alongside the target JSON that copies it; both Dockerfile examples apply the `apt-get upgrade` step the page's own tip says the generated Dockerfiles do (they do — this brings the examples in line).
- `runtime-config.mdx` — the `getAppConfig` snippets declare the shape they expect and cast, with a note that a cast is not a runtime check and Zod is the better choice for config you don't control.
- `ts-lambda-function.mdx` — the handler example matches the generated file (import order, `import type { Context }` / `export type { Context }`), and the FileTree covers `rolldown.config.ts`, the project `.gitignore` and the `project.json`/`package.json` edits.

English only; translations are generated.

**Unit tests** — 328 tests pass across `src/infra`, `src/terraform/project` and `src/migrations/latest` with no obsolete snapshots. 16 new tests across the two new migration specs (each covering: applies to the vended shape, migrates a library the same way, skips and reports a customised target, ignores projects another generator created, is idempotent, and leaves user IaC code untouched). Generator assertions and the `checkov` / `deploy` snapshot entries updated.

For the express-mode change specifically: a parameterised generator test pins `--express` to `deploy-sandbox` for both `stageConfig` values, the migration spec gains a case asserting the flag is stripped from a `deploy` that already carries it (the rc.96 shape), and its convergence test now perturbs a generated project in both directions — flag removed from the sandbox, flag added to `deploy` — and asserts it lands back exactly on today's targets.

```
pnpm nx run @aws/nx-plugin:test -- --run src/infra src/terraform/project src/migrations/latest
#  Test Files  21 passed (21) / Tests  328 passed (328)

pnpm nx run-many --target lint --projects=@aws/nx-plugin,docs --fix   # clean
```

**End to end, `--iac=terraform` workspace** (fresh workspace on the local build, `terraform#project` as both application and library):

- `nx validate tf-infra` — **passes on a fresh project with no bootstrap and no credentials** (previously an S3 `NoSuchBucket` trace); `nx validate tf-lib` passes too.
- The guide's corrected `prod.tfvars` + `project.json` prod configurations applied verbatim: `terraform plan -var-file=env/prod.tfvars` gives `Plan: 6 to add` with **no "Value for undeclared variable" warning**.
- `terraform init -backend=false && terraform validate` directly in `src/` — `Success!`.

**End to end, `--iac=cdk` workspace — `ts-core-03` proven both ways:**

1. Empty stack, `nx build @ws/infra` → success, cache populated.
2. Added the guide's `new Bucket(this, 'MyBucket')`, `nx build @ws/infra` again → **fails** with `Passed checks: 3, Failed checks: 6`, and the report on disk records `{'passed': 3, 'failed': 6, 'resource_count': 2}` — the real scan, not the restored empty one. Pre-fix this step was a `[local cache]` hit reporting success with `resource_count: 0`.

The unrelated upstream tasks still cache-hit in step 2, so this is a correctly *scoped* invalidation rather than caching being disabled.

The express-mode change was confirmed in a fresh workspace for both variants — `deploy` carries no `--express`, `deploy-sandbox` does, `deploy-ci` does not:

```
# default           deploy         -> cdk deploy --require-approval=never
#                   deploy-sandbox -> cdk deploy --require-approval=never "ws-infra-sandbox/*" --express
# with stageConfig  deploy         -> tsx .../infra-deploy.ts packages/staged2
#                   deploy-sandbox -> tsx .../infra-deploy.ts packages/staged2 "ws-staged2-sandbox/*" --express
```

Every corrected TypeScript snippet was typechecked in that workspace: the `runtime-config.mdx` `getAppConfig` snippet now compiles (`nx run @ws/my-library:compile` succeeds; it previously produced 4 errors), the `typescript-project.mdx` `bundle` target produces `dist/packages/snip-test/bundle/index.js` (previously `UNRESOLVED_ENTRY`), and the reformatted `hello.spec.ts` passes the `format` gate so `nx build` succeeds end to end. The `ts-lambda-function.mdx` handler example and FileTree were diffed against the generated `my-function.ts` and project contents.

No `terraform apply` and no `cdk deploy` were run.

**Docs** — CI covers the docs build (it OOMs locally on this box at the last of nine locales, after every `en` page has been generated).

### Issue # (if applicable)

N/A — bug bash findings `terraform-02`, `terraform-05`, `terraform-06`, `terraform-08`, `ts-core-03`, `ts-core-04`, `ts-core-05`, `ts-core-07`, `ts-core-08`, `ts-core-09`, `ts-core-10`, `ts-core-11`, `ts-core-12`, `ts-core-13`. `terraform-07` was reverted on review and is not addressed here.

### Checklist

----


nx-plugin-for-aws:
### `ts-core-03` — proof the security gate now actually gates

Since this is the finding with real security weight (a `checkov` gate wired into `build` that was reporting success on infrastructure it never scanned), here is the before/after in a real `--iac=cdk` workspace on the local build, using the guide's own `new Bucket(this, 'MyBucket')` example as the non-compliant resource.

**Why it was broken.** The target declared:

```json
"inputs": ["{workspaceRoot}/dist/{projectRoot}/cdk.out"]
```

Nx resolves a fileset input against the files it *tracks*, and `dist` is gitignored — so that input matched nothing and hashed to a constant. Every run after the first was therefore a cache hit regardless of what the stack contained.

**Before the fix** (reported by the bug bash, reproduced here):

```
$ pnpm nx run @ws/infra:checkov            # empty stack -> passes, populates cache
$ # add `new Bucket(this, 'MyBucket')` to packages/infra/src/stacks/application-stack.ts
$ pnpm nx run @ws/infra:checkov            # -> [local cache], "Successfully ran target checkov"
$ cat dist/packages/infra/checkov/checkov_report.json
{"passed": 0, "failed": 0, "skipped": 0, "parsing_errors": 0, "resource_count": 0, ...}
$ pnpm nx run @ws/infra:checkov --skip-nx-cache
Passed checks: 3, Failed checks: 6, Skipped checks: 0
```

A stale 0-resource report restored over the top of a stack with 6 failed HIGH checks, and `build` green.

**After the fix**, same two steps through `build` (the gate users actually rely on):

```
### STEP 1: empty stack, build (populates cache)
 NX   Successfully ran target build for project @ws/infra and 19 tasks it depends on
  Cache:             0/20 hit (0%)

### STEP 2: add the guide's non-compliant Bucket, build again
Passed checks: 3, Failed checks: 6, Skipped checks: 0
 NX   Running target build for project @ws/infra and 19 tasks it depends on failed
Failed tasks:
- @ws/infra:checkov
```

and the report on disk is the real scan, not a restored empty one:

```
{'passed': 3, 'failed': 6, 'resource_count': 2}
```

Note that the unrelated upstream tasks in step 2 still cache-hit (`@ws/my-library:compile [local cache]` etc.), so this is a correctly *scoped* invalidation rather than caching being disabled — `checkov` re-runs because the template it reads changed, and nothing else re-runs unnecessarily.

**Why `inputs: ["default"]` is the right layer.** It is exactly what the `synth` sibling declares. `default` carries the project's own sources plus the transitive `dependentTasksOutputFiles` entry from `nx.json`, which Nx resolves against the upstream `synth` task — so the hash follows the synthesized template the scan actually reads. I deliberately did **not** reach for `"cache": false`: the scan is soundly cacheable (its result is a function of the template, and its report genuinely is restorable from `outputs`), so disabling caching would have traded a correctness bug for a permanent build-time cost. Only the declared inputs were wrong. A unit test now pins the invariant directly:

```ts
expect(config.targets.checkov.inputs).toEqual(config.targets.synth.inputs);
```

**Carrying existing workspaces over.** `latest/checkov-scan-cache-inputs` is deterministic, and every workspace generated before this fix has the unsound gate — so shipping the migration is what actually closes the hole for existing users rather than only for new workspaces. It:

- keys off the recorded `ts#infra` generator metadata, so it never touches a `checkov` target another generator or the user owns (covered by a test that adds an unrelated project with identical pre-fix inputs and asserts it is left alone);
- matches the exact pre-fix `inputs` array before writing, and reports a customised target via `nextSteps` instead of clobbering it — the message tells the user what to check and why (`an input pointed only at dist matches no tracked file, so the scan can report a cached pass on infrastructure it never scanned`);
- is a no-op on re-run and on a project generated after the fix;
- leaves user infrastructure code untouched (asserted by writing a custom `application-stack.ts` between generation and migration and checking it byte-for-byte afterwards).

It changes no deployed resource — only Nx cache metadata — so there is no deployment transition to exercise, and no data-loss or in-place-update risk of the kind `CONTRIBUTING.md` warns about for IaC migrations.

Fixture-guard tests are included for both migrations (`should start from a fixture carrying the pre-fix inputs` / `should start from a fixture that depends on init`), so the suite fails loudly if a future generator change makes the "before" state unreachable and the assertions start passing vacuously.

---

For the record on the one red check: `Smoke Tests - react-website` is the repo-wide shadcn breakage being fixed in #1207 (which changes only `e2e/src/smoke-tests/react-website.spec.ts`). This PR's commit touches 17 files, none of them shadcn, react-website or e2e related — `git diff --name-only HEAD~1 HEAD` is limited to the six `en` guide pages, `ts#infra` / `terraform#project` and the two new migrations. Happy to rebase once #1207 lands.

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -107,10 +107,43 @@ You can start writing your Terraform infrastructure inside `src/main.tf`, for ex
...
++  #checkov:skip=CKV_AWS_144:Cross-region replication not required for this bucket
++  #checkov:skip=CKV_AWS_145:AES256 (S3-managed) encryption is sufficient for this bucket
++  #checkov:skip=CKV2_AWS_61:Lifecycle configuration not required for this bucket
++  #checkov:skip=CKV2_AWS_62:Event notifications not required for this bucket
 +  bucket = "my-unique-bucket-name"
 +}
++
++resource "aws_s3_bucket_versioning" "my_bucket" {
++  bucket = aws_s3_bucket.my_bucket.id
++  versioning_configuration {
```
cogwirrel: Undo the change, this is just an example, a bare s3 bucket is fine

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -107,10 +107,43 @@ You can start writing your Terraform infrastructure inside `src/main.tf`, for ex
...
++  bucket = aws_s3_bucket.my_bucket.id
++
++  block_public_acls       = true
++  block_public_policy     = true
++  ignore_public_acls      = true
++  restrict_public_buckets = true
++}
 ```
 
+A bare `aws_s3_bucket` fails the [Checkov](#security-testing) scan `build` runs, so the example declares the versioning, encryption and public-access-block resources the scan expects, and suppresses the remaining checks per resource with `#checkov:skip` comments stating why. Review each suppression against your own requirements — see [Security Testing](#security-testing) for both suppression mechanisms.
```
cogwirrel: Reword:

```
Note that the S3 bucket above would fail the Checkov security scan, which checks that the bucket has the appropriate security settings enabled.
```

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -134,8 +167,12 @@ To add new environments, create a new `src/env/<environment>.tfvars` file with t
 ```hcl
 # Production environment variables
 environment = "prod"
-region      = "us-west-2"
+aws_region  = "us-west-2"
 ```
+
+:::note[Variable names]
+Every name you set here must be declared in `src/variables.tf` — the scaffold declares `aws_region` and `environment`. Terraform only warns about a value for an undeclared variable, so a misspelled name is silently ignored rather than applied.
```
cogwirrel: Change to: ``Every name you set here must be declared in `src/variables.tf` — the generator declares `aws_region` and `environment`.``

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -247,6 +284,8 @@ You can validate your Terraform configuration using the `validate` target:
 
 <NxCommands commands={['validate tf-infra']} />
 
+Like `test`, this runs `terraform init -backend=false` first, so it installs the modules and providers validation needs without configuring the S3 backend. It therefore runs on a fresh project, before <Link path="guides/terraform-project#remote-state-bootstrap-application-projects-only">`bootstrap`</Link> and with no AWS credentials, as do `lint`, `test` and `checkov`.
```
cogwirrel: Remove

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -353,7 +392,7 @@ Set every variable your configuration requires in the `variables` block, otherwi
...
 
-Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. Note that the `test` target reads its path from `project.json`, so change it there too.
+Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. That only reaches the targets that go through the script, which is an application project's `init`. Every other target declares the path as an `env` entry in `project.json`, and Nx's `env` block wins over the ambient environment, so change it there too — on `test` and `validate` for either project type, and on `init` as well for a library project.
```
cogwirrel: undo

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
```diff
@@ -379,11 +388,11 @@ For more details, please refer to the [CDK bootstrapping documentation](https://
...
-| `deploy-sandbox` | Deploying your own sandbox stage during development. No stage argument needed. Uses express mode.        |
-| `deploy`         | Deploying any stage, by naming the stage or stacks you want.                                            |
-| `deploy-ci`      | Deploying from a CI/CD pipeline, using a pre-synthesized cloud assembly. Waits for full stabilization.   |
+| Target           | Use it for                                                                                                            |
+| ---------------- | --------------------------------------------------------------------------------------------------------------------- |
+| `deploy-sandbox` | Deploying your own sandbox stage during development. No stage argument needed. Uses express mode.                       |
+| `deploy`         | Deploying any stage, by naming the stage or stacks you want. Uses express mode unless you generate with `stageConfig`.  |
+| `deploy-ci`      | Deploying from a CI/CD pipeline, using a pre-synthesized cloud assembly. Waits for full stabilization.                  |
```
cogwirrel: Let's update the generator to be more consistent - `deploy` doesn't get express, `deploy-sandbox` does. Simplify the docs accordingly

docs/src/content/docs/en/guides/typescript-project.mdx
```diff
@@ -270,16 +275,21 @@ You can achieve this by adding a target such as the following to your `project.j
...
+        "command": "rolldown -c rolldown.config.ts",
+        "cwd": "{projectRoot}"
+      },
+      "dependsOn": ["compile"]
     },
   },
 }
 ```
 
+The `cwd` is what lets `rolldown` find its config: without it the command runs from the workspace root, where `rolldown.config.ts` does not exist.
```
cogwirrel: remove

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Done — reverted to the bare bucket.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Applied your wording verbatim.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Applied your wording verbatim.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Removed. The generator change behind it stays (`validate` now runs its own `terraform init -backend=false` instead of depending on the backend-configured `init`), so the target works on a fresh project — it just isn't called out in the prose any more.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Undone — restored the original sentence.

docs/src/content/docs/en/guides/typescript-project.mdx
nx-plugin-for-aws: Removed the sentence. The `cwd` (plus `inputs`/`dependsOn`) stays in the target itself, which is what fixes the `UNRESOLVED_ENTRY` the finding reported.

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
nx-plugin-for-aws: Done — made the generator consistent and simplified the docs to match, rather than documenting the inconsistency.

**Generator** (`src/infra/app/generator.ts`): the non-`stageConfig` `deploy` drops `--express`, so both variants now agree:

| Target | `stageConfig` | default |
| --- | --- | --- |
| `deploy` | waits | waits |
| `deploy-sandbox` | `--express` | `--express` |
| `deploy-ci` | waits | waits |

Express mode returns before resources have stabilized, so it belongs to the target that only ever deploys a sandbox. `deploy` names whichever stage you give it, including production, so waiting is the right default there — and `--express` is still available per invocation for a stage you're iterating on.

**Docs**: the table row is back to a single unconditional statement, and the "Deploying a Specific Stage" paragraph is back to one sentence — the `stageConfig`/non-`stageConfig` split I'd added is gone, since there's no longer a difference to explain. I also corrected the `deploy-ci` paragraph, which said it "differs from the `deploy` and `deploy-sandbox` targets" by never using express mode; that's now only true of `deploy-sandbox`, so it reads as sharing `deploy`'s stabilization behaviour and differing in taking no `--express` opt-out.

**Migration** — worth flagging, because this needs to *remove* a flag rather than add one. `v1.0.0-rc.96` has shipped `--express` on `deploy`, so existing workspaces have it. Rather than add a second migration that undoes a still-unreleased one, I amended `latest/infra-deploy-express-mode` (unreleased — added in f120924, no `version` recorded) to converge both targets on the wanted state: it adds the flag to `deploy-sandbox` where missing and strips it from `deploy` where present. That way one migration expresses the end state and there's no add-then-remove churn for anyone upgrading across both.

It keeps the existing guardrails and gains a couple:

- `--no-express` is still honoured as a deliberate opt-out (only relevant on `deploy-sandbox` now).
- A customised command is reported via `nextSteps` rather than rewritten, with the message matched to the direction of travel ("Add `--express`…" vs "Remove `--express`… this target can name any stage, including production").
- A custom `deploy` that already waits gets **no** `nextSteps` entry — there's nothing for the user to do, so it stays out of the list.

**Tests**: 11 passing on the migration spec, including a new one asserting `--express` is stripped from a `deploy` that already carries it (the rc.96 shape), and the convergence test now perturbs the generated project in *both* directions — flag removed from the sandbox, flag added to `deploy` — and asserts it lands back exactly on today's generated targets. On the generator side I added a parameterised test pinning the invariant for both `stageConfig` values:

```ts
expect(targets.deploy.options.command).not.toContain('--express');
expect(targets['deploy-sandbox'].options.command).toContain('--express');
expect(targets['deploy-ci'].options.command).not.toContain('--express');
```

Snapshots updated: the three `deploy` entries lose the flag, `deploy-sandbox` keeps it.

nx-plugin-for-aws:
All seven review comments are addressed — thanks. Summary of what changed since your pass:

**Applied your wording verbatim** (`terraform-project.mdx`): reverted to the bare S3 bucket with the Checkov note; `Every name you set here must be declared in src/variables.tf — the generator declares aws_region and environment.`; removed the `validate` paragraph; undid the `TF_PLUGIN_CACHE_DIR` rewrite (so `terraform-07` is dropped from this PR). Also removed the `cwd` sentence in `typescript-project.mdx`.

**The generator consistency change you asked for**: `deploy` no longer passes `--express`, so it's `deploy-sandbox`'s alone across both variants, and the prose is back to a single unconditional statement. Detail is on that thread — the one thing worth surfacing here is that `v1.0.0-rc.96` has already shipped `--express` on `deploy`, so I amended the still-unreleased `latest/infra-deploy-express-mode` to converge *both* directions (add to the sandbox, strip from `deploy`) rather than adding a second migration that undoes the first.

Two notes on scope, since both cut against the general direction of this batch:

**The retitle.** This is now `fix(ts#infra): stop the checkov gate passing on stale cached results` — the guide corrections are no longer in the title, and the scope moved to `ts#infra`, the generator whose vended target actually changes.

**The two new migrations.** I've kept `checkov-scan-cache-inputs` and `terraform-validate-without-backend`, and made the justification explicit in the description. Both alter the shape of a vended Nx target in `project.json`, which the generator writes once and then leaves alone — so unlike a change to vended source, there's no path by which an existing workspace picks the fix up. Specifically:

- The `checkov` one closes the hole in workspaces that already exist. Every workspace generated before this has a gate that reports success on infrastructure it never scanned, and the failure mode is silent — a user has no reason to suspect the target's `inputs`, so "they can patch it themselves" isn't realistic here.
- The `validate` one turns a target that can't run at all on a fresh application project (raw S3 `NoSuchBucket`) into one that works. Reaching the fixed state by hand means knowing to replace a `dependsOn` with three commands, two env vars and `outputs: []`.

Happy to drop either if you'd rather existing workspaces stay as they are, but the checkov one in particular seemed worth arguing for given it's a security gate.

**Validation**: 328 unit tests pass across `src/infra`, `src/terraform/project` and `src/migrations/latest` with no obsolete snapshots, and CI is **71/71 green** — including all 5 `migrate` shards, which replay the amended express-mode migration against real workspaces on recent releases, and `terraform`, which covers the `validate` change. The `ts-core-03` before/after proof is in [this comment](https://github.com/awslabs/nx-plugin-for-aws/pull/1206#issuecomment-5539307997): adding the guide's `new Bucket(this, 'MyBucket')` now **fails** `build` with 6 failed checks and a report recording 2 resources scanned, where before it was a `[local cache]` hit reporting success with `resource_count: 0`.

cogwirrel (approved):

PR #1208, opened 2026-09-04
feat(init): set up git-secrets credential scanning on existing workspaces

nx-plugin-for-aws:
### Reason for this change

Three findings from the v1.0 bug bash, all on the vended git-secrets pre-commit hook.

**`quickstart-02` (major)** — `workspace.mdx`'s "Suppressing False Positives" section documented two commands and **both failed silently**:

```sh
git secrets --add --allowed 'my-regex-pattern'
git secrets --add --allowed --literal 'my-literal+string'
```

Two independent causes:

1. `git secrets` is not on `PATH` in a generated workspace — only the vendored `.git-secrets/git-secrets` exists.
2. The vendored script's option parser accepts only the **short** forms and has no `*)` fallthrough. `--allowed` matches no arm, is consumed by the loop's trailing `shift`, and so is the pattern after it. `--add` then calls `add_config secrets.patterns ''` — empty value, wrong key — returns 1 and prints **nothing**.

A reader follows the guide, sees no error, believes the false positive is suppressed, and hits the same blocked commit again:

```
$ bash .git-secrets/git-secrets --add --allowed 'my-regex-pattern'; echo rc=$?
rc=1
$ git config --get-all secrets.allowed     # unchanged
$ git secrets --add --allowed 'x'
git: 'secrets' is not a git command.
```

**`ts-core-06` (major)** — `ts#infra --stageConfig` generates `stages.config.ts` containing `account: '123456789012'`, which the workspace's own hook flags, blocking `git commit` — while the guide's tip recommends committing that file, with no mention that scanning will stop you:

```
packages/common/infra-config/src/stages.config.ts:34:    //       account: '123456789012',
[ERROR] Matched one or more prohibited patterns
husky - pre-commit script failed (code 1)
```

**`quickstart-08` (minor)** — `workspace.mdx` and `security.mdx` both state that workspaces have commit-time credential scanning, but `nx add @aws/nx-plugin` (the `init` path) produced no `.husky/`, no `.git-secrets/`, no `.gitallowed`, no `husky` devDependency and no `prepare` script.

### Description of changes

**`quickstart-02` — fixed in the docs, not the script.** The vendored script is third-party code that `scripts/update-versions.ts` rewrites wholesale from the upstream tag on every git-secrets release, so a local edit to it would be silently reverted by the next bump. It is left byte-identical to upstream 1.3.0. The guide instead documents the flags this copy actually parses, and the vendored path it must be invoked by (it is deliberately not on `PATH` — vendoring exists so nobody has to install git-secrets globally):

```sh
# Allow a specific regex pattern (-a is the allowed flag)
bash .git-secrets/git-secrets --add -a -- 'my-regex-pattern'

# Allow a literal string, escaping special characters (-l is the literal flag)
bash .git-secrets/git-secrets --add -a -l -- 'my-literal+string'
```

A preset test pins the two `case` arms the guide depends on (`-a) ALLOWED=1`, `-l) LITERAL=1`), so a future upstream bump that changes them fails here rather than silently re-breaking the docs.

**`ts-core-06` — account ids stay flagged; the user opts out knowingly.** Rather than allowlisting them by default, the infrastructure guide's tip now says scanning will flag them and gives the entry to suppress it:

```text title=".gitallowed"
# Allow AWS Account IDs in the stage configuration
stages\.config\.ts:[0-9]+:.*[0-9]{12}
```

Scoped to the 12-digit shape rather than the bare filename: `.gitallowed` matches `<path>:<line>:<content>`, so a bare `stages.config.ts:` would allowlist the whole file and let a pasted access key through.

**`quickstart-08` — `init` now sets git-secrets up**, so the two guides' claim becomes true rather than being narrowed. An adopted workspace runs the same generators and gets the same `stages.config.ts`, so it wants the same protection, and it makes `init` and `preset` converge as the shared `applyWorkspaceInit` already intends. Extracted to `utils/git-secrets.ts`, shared by both, with a `--gitSecrets=false` opt-out mirroring the preset's.

Because `init` runs against a workspace someone already owns it never clobbers: an existing `.husky/pre-commit` or `prepare` script is left alone and the required content logged, and `.gitallowed` only gains missing entries.

**No migration.** Nothing in the generated output changes for an existing workspace — the vendored script is untouched and the `.gitallowed` default is unchanged — so there is nothing to carry a workspace across.

**Unit tests** — 58 pass across `src/init` and `src/preset`. 4 new `init` tests (hook set up; `--gitSecrets=false` opts out; a pre-existing `pre-commit`/`prepare` survives; user `.gitallowed` entries survive) plus the preset test pinning the parser arms the docs rely on. Lint clean. No snapshots changed.

**The documented commands, against the unmodified vendored script:**

| | result |
|---|---|
| `--add -a -- 'my-regex-pattern'` | `rc=0`, lands in `secrets.allowed` |
| `--add -a -l -- 'my-literal+string'` | `rc=0`, escaped to `my-literal\+string` |
| access key staged | still blocked |
| file containing the allowed pattern | commits |

**The `.gitallowed` snippet, through the real pre-commit hook** on a generated `stages.config.ts`: blocked without the entry, commits with it, and an access key added to that same file is still caught.

**`init` path** — a plain `create-nx-workspace` had no git-secrets; after `nx g @aws/nx-plugin:init` it has `.git-secrets/`, `.gitallowed`, `.husky/`, the `husky` dep and the `prepare` script; it blocks an access key on commit; and the documented suppression works there too. Both workspaces build, and `staged-infra:build` runs through `synth`. Not deployed to AWS.

The e2e `git-secrets` smoke test now covers the documented short-flag suppression, that a real secret is still caught afterwards, that the allowed pattern commits, and a second case asserting a generated `stages.config.ts` is blocked until the documented `.gitallowed` entry is added.

### Issue # (if applicable)

N/A — bug bash findings `quickstart-02`, `quickstart-08`, `ts-core-06`.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/typescript-infrastructure.mdx
+++ b/docs/src/content/docs/en/guides/typescript-infrastructure.mdx
@@ -272,8 +272,13 @@ Shared stages (under `shared.stages`) apply to any infra project in the workspac
 
 Project-specific stages (under `projects['packages/infra'].stages`) only apply to that project. When both exist for the same stage name, the project-specific entry takes priority.
 
-:::tip[Commit Stage Config]
-We recommend committing `stages.config.ts` so the team shares a single source of truth for stage credentials. If it contains personal profile names, you can add it to `.gitignore` and use a local override instead.
+:::note[Commit Stage Config]
+If you wish to commit `stages.config.ts` so the team shares a single source of truth for stage credentials, credential scanning may flag any AWS Account IDs specified in this file. You can suppress this in `.gitallowed` as follows:
+
+```text title=".gitallowed"
+# Allow AWS Account IDs in the stage configuration
+stages\.config\.ts:[0-9]+:.*[0-9]{12}
+```
 :::
 
 ### API Infrastructure
--- a/docs/src/content/docs/en/guides/workspace.mdx
+++ b/docs/src/content/docs/en/guides/workspace.mdx
@@ -2,7 +2,7 @@
 title: Workspace
 description: Reference documentation for workspaces
 ---
-import { FileTree } from '@astrojs/starlight/components';
+import { Aside, FileTree } from '@astrojs/starlight/components';
 import Link from '@components/link.astro';
 import CreateNxWorkspaceCommand from '@components/create-nx-workspace-command.astro';
 import PackageManagerShortCommand from '@components/package-manager-short-command.astro';
@@ -29,6 +29,7 @@ When you create a new workspace with `@aws/nx-plugin`, the preset generator sets
   - tsconfig.base.json Root TypeScript configuration
   - aws-nx-plugin.config.mts Nx Plugin for AWS configuration
   - .git-secrets/ Vendored git-secrets bash script for credential scanning
+  - .gitallowed Patterns git-secrets treats as false positives
   - .husky/ Git hooks
   - .mcp.json Nx Plugin for AWS MCP server configuration for Claude Code
   - .cursor/mcp.json ...and for Cursor
@@ -167,21 +168,26 @@ New workspaces are configured with [Biome](https://biomejs.dev/) for static anal
 
 ### Git Secrets
 
-New workspaces include [git-secrets](https://github.com/awslabs/git-secrets) pre-commit hooks that scan staged files for AWS credential patterns before each commit. This prevents accidentally committing access keys, secret keys, and other sensitive values.
+Workspaces are set up with [git-secrets](https://github.com/awslabs/git-secrets) pre-commit hooks that scan staged files for AWS credential patterns before each commit. This prevents accidentally committing access keys, secret keys, and other sensitive values.
+
+The script is vendored into the workspace at `.git-secrets/git-secrets` and run by the `.husky/pre-commit` hook, so there is nothing to install — but it is not on your `PATH`, so invoke it by path rather than as `git secrets`.
 
 #### Suppressing False Positives
 
 Patterns in git-secrets use [egrep-compatible regular expressions](https://github.com/awslabs/git-secrets#options-for-add). If git-secrets blocks a commit that does not contain real credentials:
 
 ```sh
-# Allow a specific regex pattern
-git secrets --add --allowed 'my-regex-pattern'
+# Allow a specific regex pattern (-a is the allowed flag)
+bash .git-secrets/git-secrets --add -a -- 'my-regex-pattern'
 
-# Allow a literal string (special characters are escaped)
-git secrets --add --allowed --literal 'my-literal+string'
+# Allow a literal string, escaping special characters (-l is the literal flag)
+bash .git-secrets/git-secrets --add -a -l -- 'my-literal+string'
+
+# List what is currently allowed
+git config --get-all secrets.allowed
 ```
 
-You can also create a `.gitallowed` file at the repository root with one egrep-compatible regex per line (shared with your team via version control):
+These are recorded in your local git config, so they apply only to your own clone. To share a suppression with your team, add it to the `.gitallowed` file at the repository root instead — one egrep-compatible regex per line, matched against `<path>:<line-number>:<line-contents>`:
 
 ```text title=".gitallowed"
 # Allow test fixtures
--- a/packages/nx-plugin/src/utils/git-secrets.ts
+++ b/packages/nx-plugin/src/utils/git-secrets.ts
@@ -1,0 +1,109 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { joinPathFragments, logger, type Tree, updateJson } from '@nx/devkit';
+import { readFileSync } from 'fs';
+import {
+  type DependencyDeclaration,
+  forDependencies,
+  type MustDeclare,
+} from './declared-dependencies.js';
+import { addDependenciesToPackageJson } from './dependencies.js';
+import { type ITsDepVersion, withVersions } from './versions.js';
+
+/** Dependencies a caller must declare to set up git-secrets. */
+export const GIT_SECRETS_DEPENDENCIES = [
+  { name: 'husky' },
+] as const satisfies readonly { name: ITsDepVersion }[];
+
+/** The vended script, which the husky pre-commit hook runs. */
+export const GIT_SECRETS_SCRIPT = '.git-secrets/git-secrets';
+
+const HUSKY_PRE_COMMIT = '.husky/pre-commit';
+
+/**
+ * Patterns allowed so a generated workspace commits cleanly, matched against
+ * `<path>:<line>:<content>`. Only the vended script, whose body contains the
+ * patterns it registers.
+ */
+export const GIT_SECRETS_ALLOWED = ['\\.git-secrets/git-secrets:'];
+
+/**
+ * Vend the git-secrets pre-commit hook: the script itself, the husky hook that
+ * runs it, and the allowlist.
+ *
+ * Re-runnable. The vendored script is always rewritten so an upgrade picks up a
+ * newer copy, while everything a user owns is preserved: `.gitallowed` only
+ * gains the entries a generated workspace needs, and an existing `pre-commit`
+ * hook or `prepare` script is left alone and reported rather than replaced.
+ */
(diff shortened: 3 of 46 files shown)


packages/nx-plugin/src/preset/git-secrets-files/git-secrets-dir/git-secrets
```diff
@@ -161,34 +161,61 @@ aws_provider() {
...
+  git-secrets --aws-provider [<credentials-file>]
+EOF
+}
+
+die() {
+  echo "$@" >&2
+  usage
+  exit 1
+}
+
```
cogwirrel: No! This is vendored. We can't edit it. Update the docs to use the short flags

packages/nx-plugin/src/migrations/latest/git-secrets-long-flags/migration.ts
```diff
@@ -0,0 +1,133 @@
...
+  const missing = GIT_SECRETS_ALLOWED.filter(
+    (pattern) => !existing.includes(pattern),
+  );
+  if (missing.length === 0) {
+    return;
+  }
+  tree.write('.gitallowed', `${[...existing, ...missing].join('\n')}\n`);
+};
+
+export default async function migration(
```
cogwirrel: remove this migration

docs/src/content/docs/en/guides/workspace.mdx
```diff
@@ -190,8 +198,14 @@ tests/fixtures/.*
 EXAMPLE[A-Z]{16}
 ```
 
+Your workspace's `.gitallowed` starts with the entries its own generated files need, including the account ids in <Link path="guides/typescript-infrastructure#stage-credential-configuration">`stages.config.ts`</Link>.
+
 For full details on managing patterns, see the [git-secrets documentation](https://github.com/awslabs/git-secrets#readme).
 
+<Aside type="tip">
+Opt out at creation time with `--no-gitSecrets`, or when adopting the plugin into an existing workspace with `nx g @aws/nx-plugin:init --gitSecrets=false`.
+</Aside>
```
cogwirrel: remove

docs/src/content/docs/en/guides/workspace.mdx
```diff
@@ -190,8 +198,14 @@ tests/fixtures/.*
 EXAMPLE[A-Z]{16}
 ```
 
+Your workspace's `.gitallowed` starts with the entries its own generated files need, including the account ids in <Link path="guides/typescript-infrastructure#stage-credential-configuration">`stages.config.ts`</Link>.
```
cogwirrel: remove this

packages/nx-plugin/src/utils/git-secrets.ts
```diff
@@ -0,0 +1,116 @@
...
+ * cleanly. Entries are matched against `<path>:<line>:<content>`.
+ *
+ * - the vended script itself, whose body contains the patterns it registers
+ * - `account` fields in `stages.config.ts`, which is where a workspace records
+ *   the account each stage deploys to and which the guide recommends committing.
+ *   Scoped to the account field so every other pattern is still caught there.
+ */
+export const GIT_SECRETS_ALLOWED = [
+  '\\.git-secrets/git-secrets:',
+  'stages\\.config\\.ts:[0-9]+:.*account.*[0-9]{4}-?[0-9]{4}-?[0-9]{4}',
```
cogwirrel: Don't add stage config by default. See comment in docs.

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
```diff
@@ -265,6 +265,8 @@ Project-specific stages (under `projects['packages/infra'].stages`) only apply t
 
 :::tip[Commit Stage Config]
 We recommend committing `stages.config.ts` so the team shares a single source of truth for stage credentials. If it contains personal profile names, you can add it to `.gitignore` and use a local override instead.
+
+Account ids in this file are exempted from the workspace's <Link path="guides/workspace#git-secrets">credential scanning</Link> hook by an entry in `.gitallowed`, so it commits as generated. An account id is not a secret; access keys and secret keys in this file are still caught.
 :::
```
cogwirrel: Don't exempt by default, change docs to:

```
:::note[Commit Stage Config]
If you wish to commit `stages.config.ts` so the team shares a single source of truth for stage credentials, credential scanning may flag any AWS Account IDs specified in this file. You can suppress this in `.gitallowed` as follows:

<example>
:::
```

packages/nx-plugin/src/preset/git-secrets-files/git-secrets-dir/git-secrets
nx-plugin-for-aws: Agreed, and thanks — reverted; the vendored script is byte-identical to upstream 1.3.0 again (`md5 d6ba3611…`, and `git diff origin/main` on that path is empty). I'd missed that `scripts/update-versions.ts` rewrites this file wholesale from the upstream tag, so my edit would have silently vanished on the next git-secrets release anyway.

The guide now documents the forms this copy actually parses:

```sh
bash .git-secrets/git-secrets --add -a -- 'my-regex-pattern'
bash .git-secrets/git-secrets --add -a -l -- 'my-literal+string'
```

Verified against the unmodified script: both return 0 and the patterns land in `secrets.allowed` (`my-regex-pattern`, `my-literal\+string`), a real access key is still blocked, and a file containing the allowed pattern now commits.

packages/nx-plugin/src/migrations/latest/git-secrets-long-flags/migration.ts
nx-plugin-for-aws: Removed — the whole `migrations/latest/git-secrets-long-flags/` folder is gone (migration, spec and the fixture). Follows naturally now the vendored script is untouched: there is nothing left to carry an existing workspace across.

packages/nx-plugin/src/utils/git-secrets.ts
nx-plugin-for-aws: Done — `GIT_SECRETS_ALLOWED` is back to just the vendored script entry, so account ids are flagged by default and the user opts out knowingly.

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
nx-plugin-for-aws: Used your wording verbatim, with the `<example>` filled in:

```text title=".gitallowed"
# Allow AWS Account IDs in the stage configuration
stages\.config\.ts:[0-9]+:.*[0-9]{12}
```

Verified through the real pre-commit hook on a staged generated `stages.config.ts`: blocked without the entry, commits with it, and an access key in that same file is still caught (`.gitallowed` matches `<path>:<line>:<content>`, so scoping it to the 12-digit shape rather than the bare filename keeps credentials there in scope). The e2e smoke test now asserts exactly that sequence.

docs/src/content/docs/en/guides/workspace.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/workspace.mdx
nx-plugin-for-aws: Removed.

cogwirrel (dismissed):

nx-plugin-for-aws:
**Note for whoever merges this:** if this lands after #1196 or #1215, expect conflicts **only in machine-generated locale files** — `zh/guides/workspace.mdx` (vs #1196) and `{es,fr,it,pt,vi}/guides/security.mdx` (vs #1215). The English sources shared with both PRs merge cleanly, so there is no content disagreement to adjudicate.

The branch itself carries **English-only** doc changes; the locale copies are re-added by the `translate-docs` bot on every push, so dropping them locally doesn't stick. Whichever side merges second is stale either way, and the next translate run regenerates all of them from the English source.

Which side to prefer if you do resolve by hand:

- `workspace.mdx` and `existing-project.mdx` — **this PR**, which owns the English edits.
- `security.mdx` — **#1215**. Both PRs touch the English source here (mine adds one sentence about `init` also setting up scanning; #1215 fixes locale link prefixes), but #1215's change is specifically to the locale files, so take theirs and let the bot pick my English sentence up on the next run.

cogwirrel (approved):

PR #1209, opened 2026-09-04
feat(ts#infra): declare workspace deps for generated cross-project imports

nx-plugin-for-aws:
### Reason for this change

A bug bash of the Dungeon Adventure tutorial and the quick-start surfaced eleven findings on the two highest-traffic onboarding pages. The substantive one — bug-bash findings `tutorial-01`, `tutorial-02` and `quickstart-01`, all major, all independently verified — is a single root cause in the generators.

The generators emit cross-project imports: `common/constructs` imports the tRPC API's router, `game-ui` imports its `AppRouter` type, `infra` imports `common-constructs`. But they leave declaring those imports to the `ts#sync` generator, which only runs as part of a task run. So a workspace that has never built is out of sync from the moment it is generated:

- `pnpm lint` fails biome's `noUndeclaredDependencies` at 4 sites, one of which (`packages/infra/src/main.ts:1`) is purely generated code — a reader who has written nothing still sees it. Verified: `pnpm lint` under a pseudo-TTY offers **no** sync prompt, because biome failing is not an Nx sync failure, so Nx never asks.
- `pnpm build` exits 1 with `The workspace is out of sync` wherever stdin is not a TTY.

### Description of changes

**Generators.** New `utils/local-project-dependency.ts` declares a generated cross-project import in the importing project's own `package.json`, at the point the import is emitted. Wired into the three sites that emit one:

- `utils/api-constructs` — the generated tRPC construct imports the API project's router (`trpc/backend` passes `projectRoot` for this).
- `trpc/react` — the generated client provider imports the API's `AppRouter` type.
- `infra/app` — `main.ts` and the application stack import `common-constructs`.

The specifier matches what `ts#sync` writes, so the two agree and re-running either is a no-op. An existing declaration, in `dependencies` or `devDependencies`, is never overwritten.

TypeScript project references are deliberately **not** written here — `@nx/js:typescript-sync` owns resolving which tsconfig a dependency maps to, and writing one alongside it appended a duplicate once sync had rewritten the first. That is what the `Smoke Tests - idempotency` failure on this PR was; it is fixed by leaving references to sync. Sync remains the safety net for imports the *user* adds.

**Docs.** Per review, the sync guidance stays minimal — the guides mention the prompt rather than teaching `nx sync` as a required step:

- `tutorial-05` — the `ts#mcp-server` FileTree listed a `Dockerfile`, and the `py#agent` prose claimed an "extra Dockerfile" and that the artifact "uploads your agent Docker image to ECR". Neither generator produces a Dockerfile on the tutorial's default `infra=agentcore`; both package a zip via `AgentRuntimeArtifact.fromCodeAsset`. All three statements corrected.
- `tutorial-04` — the `ts#api` FileTree listed a root `vitest.workspace.ts` (it is `vitest.config.mts`), and the `echo.ts` snippet imported from `../schema/echo.js` (it is `../schema/index.js`).
- `tutorial-06` — the `useGameApi.tsx` snippet updated to the generated hook signature.
- `tutorial-10` — Module 4 Task 1 names `http://localhost:4200/` where it tells you to open it, instead of three tasks later.
- `tutorial-01` — "You **will** be prompted" softened to "You **may** be prompted", since a generated workspace now has nothing to sync.
- `tutorial-02` — Module 2 gains an `nx sync` step, because the `dungeon-db` imports the reader writes there are their own and genuinely need it.
- `quickstart-01` and Module 3 — a short recovery tip pointing at the prompt or `nx sync`.

**e2e.** `tutorial-07`: Module 2's diff "before" sides were captured pre-lint, so biome's import reordering meant they no longer matched the reader's disk. The dungeon-adventure smoke test's Module 1 stage now runs the documented `lint` before `build` — which is what made the templates drift — and the three affected `.old.template` files are updated to the post-lint state. `2/agents/story-agent.ts.template` (the *after* side) is reordered to match, so the rendered diff shows only Module 2's three real changes rather than the whole import block churning.

**No migration.** Running `nx sync` in an existing workspace produces exactly the same result as the new generator-time declarations, and any existing workspace has necessarily already synced in order to have built at all.

### What this does *not* fix

**A non-interactive run still fails once the reader adds their own cross-project import.** This was the verified core of `tutorial-01` / `tutorial-02` / `quickstart-01`, and the reduced scope leaves it open, so recording it plainly rather than letting it disappear:

- Nx `process.exit(1)`s in its non-TTY branch (`tasks-runner/run-command.js`, `if (!process.stdout.isTTY)`) *before* it would prompt. To be precise about the severity: that branch does emit the out-of-sync error **with** actionable guidance, including "Run `nx sync` (no flags) to sync now", so the reader is not left without a next step — what is TTY-gated is the prompt and the chance to continue in the same invocation. I verified `sync.applyChanges: true` in `nx.json` does **not** help, because the exit happens before that branch is reached.
- After this PR a freshly generated workspace has nothing to sync, so the pages are fine as written — I confirmed `nx sync:check`, `pnpm lint` and `pnpm build` all pass non-interactively on raw Module 1 generator output.
- The residual exposure is Module 2 onward, where the reader's own imports need a sync. Module 2 names `nx sync` as a step, so following the page verbatim works; a reader who skips it and has no TTY gets an exit 1 plus the diagnostic above, and has to re-run after syncing.

If that's worth closing later, the smallest lever is on the Nx side (a non-interactive `--sync` opt-in) rather than in these guides.

### Descoped after review

- The **shadcn tsconfig reference** change (`tsconfig.json` → `tsconfig.lib.json`) and its migration are reverted — #1203 owns that fix (`website-04`), now merged. This branch is rebased onto it and does not touch `ts/react-website/` or add any migration.
- Notes for `tutorial-08` (`agent-chat` requires a TTY) and `tutorial-09` (the CWD-coupled local session path) were removed at review request; both remain open as findings.
- `quickstart-04` (re-running `ts#infra` reverts the reader's stack) is another lane's fix.

**On the `Smoke Tests - idempotency` failure seen earlier on this PR:** fixed, and worth recording what it actually was. The shadcn reference is removed and re-appended on each run, so it migrates to the end of the list. With one reference that is invisible; the project reference my helper *also* wrote supplied the second, and the two then swapped places every run:

```
 "references": [
-      "path": "../common/shadcn/tsconfig.json"
+      "path": "../my-api/tsconfig.lib.json"
-      "path": "../my-api/tsconfig.lib.json"
+      "path": "../common/shadcn/tsconfig.json"
```

Rather than work around the re-append, the helper no longer writes project references at all — that is `@nx/js:typescript-sync`'s job, and it was the wrong layer for this change. `lint` still passes on raw generator output, which is what the findings were about; references were never needed for that. I verified the swap is gone and confirmed `main`'s generator reproduces the underlying re-append on its own, so the root cause is `website-04` / #1203 (now merged), independent of this PR.

**Reproduced the reported failure first**, from a fresh `pnpm create @aws/nx-workspace` running Module 1's generators verbatim plus Task 3's stack: `pnpm lint` exit 1 with 4 × `noUndeclaredDependencies` across 3 projects; `pnpm build` non-TTY exit 1, out of sync.

**Worked the tutorial end to end, non-interactively, in a fresh workspace** built with the locally compiled plugin — all 14 Module 1 generators, then Module 2 Tasks 1–3 from the e2e templates. Stopped before AWS deployment.

Module 1, on raw generator output with **no sync step run at all**:

- `pnpm lint` → **exit 0**. This is the fix: previously 4 × `noUndeclaredDependencies` across 3 projects.
- `pnpm build` → exit 0, 58 tasks
- `find -name 'Dockerfile*'` → **zero results**, confirming `tutorial-05`.

Module 2 (`nx sync` → `lint` → `build`): exit 0 at each step, 58 tasks. Every documented command succeeded without a prompt.

**FileTrees and snippets verified against real generated output** — the `vitest.config.mts` name, the `schema/index.js` import, and the `useGameApi.tsx` / `story-agent.ts` snippets byte-for-byte. Confirmed the `.old.template` diffs are import-ordering only by running biome over the generated files.

**Tests.** 6 new unit tests for `addLocalProjectDependency`: manifest declaration, no-op on re-run, preserves a user-set version, preserves a `devDependencies` placement, leaves the consumer tsconfig untouched, and a missing dependency project.

```
pnpm nx run @aws/nx-plugin:test --skip-nx-cache
  Test Files  246 passed (246)
       Tests  4340 passed (4340)
```

No snapshot changes. `pnpm nx run-many --target lint --fix` is clean. `nx build docs` OOMs in this container, so I verified the MDX by hand: every component used is imported in each page edited, and every `<Link path>` is locale-agnostic.

**Two earlier CI failures were mine and are fixed:**

- *`Smoke Tests - dungeon-adventure`* (Linux and Windows) — an `AssertionError` on `2/agents/story-agent.ts.old.template`. Adding the documented `lint` to the e2e Module 1 stage reordered that construct's imports too, which I had missed.
- *`Translate Documentation`* — `translation has 17 code block(s) but the source has 16`, across all eight locales. `scripts/translate.ts` `restoreCodeBlocks` requires source and translation to have equal fenced-block counts, so **removing** a fenced block from an already-translated page fails every locale. Now passing. Every page on this branch is back to parity with its translations, and the translated files are reset to `main` so the workflow regenerates them from the final English.

### Issue # (if applicable)

N/A — from the v1.0 bug bash. Closes findings `tutorial-01`, `tutorial-02`, `tutorial-04`, `tutorial-05`, `tutorial-06`, `tutorial-07`, `tutorial-10`, `quickstart-01`. `tutorial-08`, `tutorial-09` and `quickstart-04` are descoped and remain open.

### Checklist

----


nx-plugin-for-aws:
Diagnosis of the failing `Translate Documentation` job, from the [run 33863854159 log](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/33863854159) (I am on the docs-link-validator lane, #1215, and checked whether my translate-script changes were implicated — they are not; my changes are on `buildSystemPrompt` and a post-write `<Link path>` locale strip, neither of which is in play here).

All 8 failures are the same file in every locale:

```
[translate] ERROR 8 translation(s) failed after retries:
  get_started/tutorials/dungeon-game/1.mdx → jp   (and ko, es, pt, fr, it, zh, vi)
```

and the underlying cause is the pre-existing `restoreCodeBlocks` guard, which throws when the code-block count differs:

```
[translate] translation attempt 1/4 for get_started/tutorials/dungeon-game/1.mdx → jp failed
  (translation has 17 code block(s) but the source has 16); retrying in 2000ms
```

It exhausts all 4 attempts for every locale, so it is systematic rather than flaky. The agent is consistently emitting one extra fenced block for this page — the summaries in the log mention adding an Aside/tip around the lint-error section (`Added the "The workspace is out of sync" tip after the lint error tip`), which is the part this PR edits. Worth checking whether the new Aside in `1.mdx` has an unbalanced or nested fence, since `splitOnFences` counts fence lines and a stray ``` inside an Aside would make the translated file legitimately parse as 17.

That guard is deliberate (a mismatched count means the document lost its shape, so it refuses to stitch), so the fix is on the source page rather than the script.

nx-plugin-for-aws:
Hi — flagging an overlap so we don't both ship a fix for the same bug.

We're independently fixing the same shadcn tsconfig reference issue in the same file. #1203 (`fix/website-auth-docs-guides`) covers it as bug-bash finding `website-04`; this PR changes the same lines in `ts/react-website/app/generator.ts`, plus adds `migrations/latest/shadcn-website-tsconfig-lib-reference/`.

**I think the idempotency failure on this PR is caused by the shape of the fix, and is worth a look before deciding who keeps it.**

The failure is:

```
 M packages/website/tsconfig.app.json
-      "path": "../common/shadcn/tsconfig.lib.json"
+      "path": "../my-api/tsconfig.lib.json"
```

That's the two references *swapping order*, not a duplicate. The generator still does `filter`-then-append:

```ts
references: [
  ...(tsconfig.references ?? []).filter((ref) => ref.path !== sharedShadcnTsconfigRef),
  { path: sharedShadcnTsconfigRef },
]
```

Appending unconditionally moves the shadcn reference to the **end** on every run. With only shadcn present that's invisible; once a second project's reference exists (a website connected to an API), the re-run reorders them and `nx sync` sees a change. Pointing the ref at `tsconfig.lib.json` makes the `filter` match, so it stops *duplicating* — but the entry is still removed and re-appended, so the order still churns.

#1203 takes a different angle: rather than predicting which filename `nx sync` will resolve to, it treats *any* reference resolving inside the target project's directory as already-present and leaves it **in place**:

```ts
const alreadyReferenced = references.some(
  (ref) => ref.path === projectDirRef || ref.path.startsWith(`${projectDirRef}/tsconfig`),
);
return alreadyReferenced ? references : [...references, { path: `${projectDirRef}/tsconfig.json` }];
```

Because nothing is removed and re-appended, order is stable, and it holds whether the on-disk ref is the `tsconfig.json` the generator wrote or the `tsconfig.lib.json` `nx sync` rewrote it to.

I've added a regression test on #1203 for exactly the shape failing here — shadcn ref first, another project's ref second, re-run, assert order unchanged — and it passes.

**Suggested split:** #1203 keeps the `ts#website` tsconfig fix (it's that PR's assigned finding and already has the tests), and this PR drops its `generator.ts` hunk so it can stay focused on the `ts#infra` sync fix and the tutorial/quick-start docs. Happy to do it the other way round if you'd prefer — the important thing is one of us drops it. Worth noting cogwirrel has asked several lanes to drop migrations from this batch, so `shadcn-website-tsconfig-lib-reference` may be going anyway.

Let me know which you'd prefer and I'll adjust.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
```diff
@@ -151,21 +152,18 @@ export type AppRouter = typeof appRouter;
...
+import { EchoInputSchema, EchoOutputSchema } from '../schema/index.js';
 
 export const echo = publicProcedure
   .input(EchoInputSchema)
   .output(EchoOutputSchema)
   .query((opts) => ({ message: opts.input.message }));
 ```
 
-This file is the implementation of the `echo` method and as you can see is strongly typed by declaring its input and output data structures.
+This file is the implementation of the `echo` method and as you can see is strongly typed by declaring its input and output data structures. The schemas are imported from `../schema/index.js`, the barrel file that re-exports every schema module.
```
cogwirrel: revert this line

docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
```diff
@@ -903,9 +901,11 @@ export class StoryAgent extends Construct implements IGrantable, IConnectable {
...
 
-You may notice an extra `Dockerfile`, that references the Docker image from the `story` project, allowing us to co-locate the Dockerfile and agent source code.
+:::note[No container image required]
+`AgentRuntimeArtifact.fromCodeAsset` uploads a zip of `dist/packages/story/package/story-agent`, so no container engine or ECR repository is involved. Pass `--infra=agentcore-ecr` to either the `py#agent` or `ts#mcp-server` generator instead if you would rather deploy a container image — that option vends a `Dockerfile` alongside the source.
+:::
```
cogwirrel: remove this note

docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
```diff
@@ -1551,31 +1553,34 @@ If for whatever reason you ever wanted to clear your cache (stored in the `.nx`
 
 </Drawer>
 
-Using the command line, run the following command to fix any lint issues first:
+Using the command line, first make sure the workspace is in sync with the imports your code has:
```
cogwirrel: Undo the changes from here downwards on 1.mdx. Given the generators set up everything so it's already in sync, it's fine to then say when building you may be prompted with the following: ...

docs/src/content/docs/en/get_started/tutorials/dungeon-game/2.mdx
```diff
@@ -107,6 +107,17 @@ To implement the API methods, make the following changes within `packages/game-a
...
+:::note[`LOCAL_SESSION_STORAGE_DIR` is relative to the project root]
+`queryActions` reads the transcript from S3 when deployed, and from the local
+directory the Story Agent's `session.py` writes to when `LOCAL_DEV=true`. That
+path is written relative to `packages/game-api`, matching the `dev` target's
+`cwd: {projectRoot}` — the same relative literal `session.py` uses from
+`packages/story`, so both resolve to `tmp/agents/strands/story-agent` at the
+workspace root. Run the API through its `dev` or `serve` target so the CWD is
+right; invoking `tsx src/local-server.ts` from the workspace root instead
+resolves to a directory that doesn't exist and returns an empty transcript.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/get_started/tutorials/dungeon-game/2.mdx
```diff
@@ -170,11 +181,19 @@ The `DungeonDb` construct registers the deployed table name in <Link path="guide
...
 There's no need to deploy to AWS to try out our API — the `dev` target runs the Game API against [DynamoDB Local](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html). Because we connected the Game API to the `DungeonDb` project in Module 1, this target also starts DynamoDB Local automatically.
 
-First, fix any lint issues:
+The procedures and MCP server we just wrote import `@dungeon-adventure/dungeon-db`, a new cross-project import, so start by syncing the workspace:
+
+<NxCommands commands={["sync"]} />
+
+<Aside type="tip" title="Why this step is needed">
+`nx sync` declares `@dungeon-adventure/dungeon-db` in the `game-api` and `inventory` `package.json` files and adds the matching TypeScript project references. Without it, `lint` fails the `noUndeclaredDependencies` rule on both projects, and `build` fails outright wherever Nx has no interactive terminal to offer the sync on (CI, an AI coding agent, an IDE task runner).
+</Aside>
```
cogwirrel, line 179: Change to: `nx sync` declares `@dungeon-adventure/dungeon-db` in the `game-api` and `inventory` `package.json` files and adds the matching TypeScript project references for you, saving you from needing to add this by hand when declaring dependencies between projects.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/3.mdx
```diff
@@ -47,6 +47,10 @@ If you encounter any lint errors, run the following command to automatically fix
 <PackageManagerShortCommand commands={["lint"]} />
 </Aside>
 
+<Aside type="tip" title="The workspace is out of sync">
+If the build reports `The workspace is out of sync`, run <code>nx sync</code> and build again — see <Link path="get_started/tutorials/dungeon-game/1">Module 1 Task 4</Link>.
+</Aside>
```
cogwirrel, line 52: use backticks not `<code>`. remove: ` — see <Link path="get_started/tutorials/dungeon-game/1">Module 1 Task 4</Link>.`

docs/src/content/docs/en/get_started/tutorials/dungeon-game/3.mdx
```diff
@@ -59,6 +63,29 @@ Then, in a second terminal, start the chat:
 
 Your first message should tell the agent your hero's name and the genre (for example: `My name is Alice. Start my zombie adventure.`) and the story will stream back.
 
+<Aside type="caution" title="`agent-chat` needs an interactive terminal">
```
cogwirrel: remove all this

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -142,6 +142,10 @@ export class ApplicationStack extends Stack {
 ```
 
 This is all the CDK we need to write to deploy our full stack application.
+
+:::caution[This file is yours from here on]
+`application-stack.ts` is the only file this guide asks you to author, so commit it before running any further generators — that way `git` can always recover it. See the <Link path="guides/typescript-infrastructure">ts#infra generator guide</Link> for more on working in this project.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -224,6 +228,14 @@ If you encounter any lint errors, you can run the following command to automatic
 <PackageManagerShortCommand commands={["lint"]} />
 :::
 
+:::tip[The workspace is out of sync]
+If the build reports `The workspace is out of sync`, run the following command and then build again:
```
cogwirrel: If the build reports `The workspace is out of sync`, accept the prompt to sync the workspace or run the following sync command directly before retrying the build:

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -224,6 +228,14 @@ If you encounter any lint errors, you can run the following command to automatic
...
+
+<NxCommands commands={["sync"]} />
+
+`nx sync` writes the TypeScript project references and `package.json` dependency entries your imports require. The generators write both for the code they vend, so this is only needed once you have added a cross-project import yourself. `nx build` also offers to sync, but only when it has an interactive terminal to prompt on — in CI, from an AI coding agent, or from an IDE task runner it fails instead, so running `nx sync` explicitly is the portable habit. Use `nx sync:check` in CI to verify without writing anything.
```
cogwirrel: change to: ``This writes the TypeScript project references and `package.json` dependency entries your imports require.``

packages/nx-plugin/src/ts/react-website/app/generator.ts
```diff
@@ -629,16 +632,14 @@ export async function tsReactWebsiteGenerator(
...
         ux === 'shadcn'
-          ? [
-              ...(tsconfig.references ?? []).filter(
-                (ref) => ref.path !== sharedShadcnTsconfigRef,
-              ),
-              {
-                path: sharedShadcnTsconfigRef,
-              },
-            ]
+          ? mergeTsReferences(tsconfig.references, [
```
cogwirrel: this is covered by another pr - check and undo

packages/nx-plugin/src/migrations/latest/shadcn-website-tsconfig-lib-reference/migration.ts
```diff
@@ -0,0 +1,103 @@
...
+ * `@nx/js:typescript-sync` resolves a dependency to its `tsconfig.lib.json`, so
+ * a reference to the solution `tsconfig.json` reads to it as stale and is
+ * rewritten on every task run — leaving `nx sync:check` failing until a build
+ * has been run and its changes committed. Generators now write the resolved
+ * path directly.
+ *
+ * Only the exact path the generator produced is replaced; anything else in the
+ * reference list is left alone.
+ */
+export default async function migration(
```
cogwirrel: remove this migration

docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
nx-plugin-for-aws: Reverted — the sentence is gone. Kept the snippet itself corrected: the generated `procedures/echo.ts` imports from `../schema/index.js`, not `../schema/echo.js` (bug-bash finding `tutorial-04`).

docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
nx-plugin-for-aws: Removed. The prose above it still drops the "uploads your agent Docker image to ECR" claim, since on the tutorial default of `infra=agentcore` the construct packages a zip via `AgentRuntimeArtifact.fromCodeAsset` — I confirmed zero Dockerfiles are generated anywhere in a full Module 1 workspace.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/1.mdx
nx-plugin-for-aws: Done — Task 4 is back to `lint` then `build`, with the original prompt transcript restored and only "You will be prompted" softened to "You may be prompted". Also reverted my edit to the "Fixing Import Errors" aside further up the page, which was part of the same sync framing.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/2.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/2.mdx
nx-plugin-for-aws: Applied verbatim.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/3.mdx
nx-plugin-for-aws: Both done — backticks instead of `<code>`, and the `<Link>` back-reference dropped.

docs/src/content/docs/en/get_started/tutorials/dungeon-game/3.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Applied verbatim.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Applied verbatim.

packages/nx-plugin/src/ts/react-website/app/generator.ts
nx-plugin-for-aws: Undone — reverted `generator.ts` to `main`, and dropped the two regression tests I had added alongside it. Confirmed #1203 is the other PR (its `mergeTsProjectDirReference` covers the same reference), and I have commented there so they know they now solely own it.

One thing worth passing on, since it was only visible once both changes were in play: the surrounding code removes the shadcn reference and re-appends it on every run, so it migrates to the end of the list. With only the shadcn reference present that is invisible, but as soon as a second reference exists the two swap places and the `idempotency` smoke test never converges — that was the `Smoke Tests - idempotency` failure on this PR. Flagged on #1203 so it can be fixed there.

packages/nx-plugin/src/migrations/latest/shadcn-website-tsconfig-lib-reference/migration.ts
nx-plugin-for-aws: Removed, along with its spec.

nx-plugin-for-aws:
All 12 review comments are applied — replied on each thread individually. Summary of what changed: the shadcn tsconfig change and its migration are gone (#1203 owns that), and the sync guidance is back to mentioning the prompt rather than teaching `nx sync` as a step. The generators still declare each cross-project import they emit, so the workspace is in sync from generation.

One thing I want to flag explicitly rather than let it drop quietly, since it was the verified core of three of this lane's findings (`tutorial-01`, `tutorial-02`, `quickstart-01`, all major): **the sync prompt only exists when stdin is a TTY.** In a non-interactive run, Nx `process.exit(1)`s before it would prompt, so "accept the prompt to sync" isn't available. I verified this on the current Nx (`tasks-runner/run-command.js`, the `if (!process.stdout.isTTY)` branch — it errors and exits before reaching the `applyChanges` branch), and confirmed that `sync.applyChanges: true` in `nx.json` does **not** help for the same reason.

Where that still bites, after this PR:

- **Module 2** — the reader's own `dungeon-db` imports genuinely need a sync. Interactively they get the prompt; in CI, from an AI coding agent, or from most IDE task runners both `lint` and `build` exit 1. Module 2 does still name `nx sync` as a step, so a reader who follows the page verbatim is fine — this is only about the recovery path if they skip it.
- **Quick-start / Module 1 / Module 3** — these now only mention the prompt. Post-fix that should be moot, because the generators leave nothing to sync; I confirmed `nx sync:check` is exit 0 / "The workspace is up to date" on raw Module 1 generator output, and that `pnpm lint` and `pnpm build` both pass non-interactively with no sync step at all. So the remaining exposure is limited to a reader who adds their own cross-project import and then hits it without a TTY.

Not proposing any further change here — just recording it so the reduced scope is a decision rather than an oversight. If it's worth closing off later, the smallest lever is on the Nx side (a non-interactive `--sync` opt-in) rather than in these guides.

Separately: `macOS Smoke Tests - standalone (6/6)` is failing on `pnpm install --no-frozen-lockfile` inside `ts#nx-plugin` with `Failed to start verdaccio` / exit 130 (SIGINT). It's on a path this PR doesn't touch and passes on sibling PRs, so I'm treating it as infrastructure and will rerun it.

nx-plugin-for-aws:
Thanks — I see the `ts/react-website/app/generator.ts` hunk and the `shadcn-website-tsconfig-lib-reference` migration are both gone now, so the overlap is resolved: **#1203 owns the shadcn tsconfig reference fix** (`website-04`) and this PR stays on the `ts#infra` sync fix and the tutorial/quick-start docs. No further action needed from my side.

For the record, #1203's version keeps the existing reference **in place** rather than filtering and re-appending it, so reference order is stable on re-run — that's the specific shape that was tripping the idempotency matrix here. It's covered by a regression test using the same two-reference layout (shadcn first, another project second, re-run, assert order unchanged).

cogwirrel (dismissed):

nx-plugin-for-aws:
Two follow-ups, plus a correction to something I said earlier.

**1. The tsconfig reference bug is pre-existing, not caused by this PR — and this PR no longer touches that file at all.**

I said earlier that the re-append was the mechanism and implied my change triggered it. That was half right and worth correcting properly, because I went and measured it rather than reasoning about it. I compiled the plugin from `origin/main`'s `packages/nx-plugin/src` — none of this branch's changes present — and re-ran `ts#website --ux=shadcn` in an already-synced workspace:

```
before: ['../common/shadcn/tsconfig.lib.json', '../game-api/tsconfig.lib.json']
after:  ['../common/shadcn/tsconfig.lib.json', '../game-api/tsconfig.lib.json',
         '../common/shadcn/tsconfig.json']
```

`main` alone produces the duplicate. The mechanism is the filter's *predicate*, not the re-append: the generator filters on `tsconfig.json`, but `nx sync` has already rewritten that reference to `tsconfig.lib.json`, so the filter misses and a second reference is appended. Any workspace that has synced once reproduces it, with or without this PR. That's `website-04`, which #1203 fixes with `mergeTsProjectDirReference`; I've written the measurement up on that thread.

So: not fixed here, and correctly so — `git diff origin/main -- packages/nx-plugin/src/ts/react-website/` on this branch is empty.

**2. Correction on the non-interactive failure mode — I overstated it.**

I said a non-TTY run fails "with no prompt and no instruction to recover". The second half is wrong. Nx's non-TTY branch (`run-command.js`, `if (!process.stdout.isTTY)`) emits the out-of-sync error *with* its fix guidance before exiting 1 — I have it in my own repro logs:

```
To sync the workspace:
- Run `nx sync` (no flags) to sync now.
- Run `nx sync:check` to preview the changes without modifying any files.
- Set `sync.applyChanges` to `true` in your `nx.json` to sync automatically when running tasks in interactive environments.
```

So the reader does get an actionable next step. What's TTY-gated is the prompt and the chance to continue within the same invocation — you have to sync and re-run. (The third bullet is misleading in a non-TTY context: I verified `sync.applyChanges: true` does **not** help, because the exit happens before that branch is reached.) The PR body now states it at that severity rather than the stronger one.

**Housekeeping:** the branch is squashed to a single commit, since a squash-merge takes its body from commit messages rather than the PR description, and the earlier commits described the shadcn/migration work that has since been dropped. The squashed tree is byte-identical to the approved state (`git diff a9b3e9f HEAD` is empty) — only the history is collapsed.

nx-plugin-for-aws:
Rebased onto current `main` (picks up #1203) and pushed as a single commit. Ready for re-review — the branch is final: title, commit subject and body all match what ships.

**The `Smoke Tests - idempotency` failure was mine, and it is fixed properly rather than papered over.** I had assumed it was purely #1203's re-append bug; that was only half right, so I measured it instead of arguing it. The re-append moves the shadcn reference to the end of the list on every run — invisible with one reference. The second reference was supplied by **my** helper, which also wrote a TypeScript project reference, and the two then swapped every run:

```
 "references": [
-      "path": "../common/shadcn/tsconfig.json"
+      "path": "../my-api/tsconfig.lib.json"
-      "path": "../my-api/tsconfig.lib.json"
+      "path": "../common/shadcn/tsconfig.json"
```

(Worth noting the failing idempotency case never runs `nx sync`, which is how I pinned the second reference on my code rather than on sync.)

The fix is to stop writing project references at all — that's `@nx/js:typescript-sync`'s job, and it was the wrong layer for this change. `lint` was the actual failure in the findings, and that only needs the `package.json` entry; references were never required for it. So the helper is now manifest-only, which is both smaller and correctly scoped. Verified: the swap is gone, and `pnpm lint` still exits 0 on raw Module 1 generator output with all three manifest deps declared.

I also confirmed `main`'s generator reproduces the underlying re-append entirely on its own, with none of this branch's code compiled in — so that root cause is `website-04` / #1203, independent of this PR.

**Answering the outstanding questions:**

1. **The `react-website` hunk and the migration are gone.** `git diff origin/main --name-only -- packages/nx-plugin/src/ts/react-website/ packages/nx-plugin/src/migrations/` is empty. #1203's `mergeTsProjectDirReference` is present on my base.
2. **Title and commit message are corrected and identical.** Both now read `feat(ts#infra): declare workspace deps for generated cross-project imports` — no claim about "without a manual sync", since the generator change delivers the `lint` fix but not a sync-free non-interactive `build` for the reader's own imports. Squashed to one commit so a squash-merge body describes only what ships.
3. **The residual gap is stated at the right severity in the body.** I had overstated it: nx's non-TTY path *does* emit the out-of-sync error together with its fix guidance ("Run `nx sync` (no flags) to sync now") before exiting 1. So a CI run fails with an actionable diagnostic — what's TTY-gated is the prompt and the chance to continue in the same invocation. Corrected.

Validation on the rebased base: `246 test files / 4340 tests passed`, no snapshot changes, `lint --fix` clean, and the Module 1 + Module 2 walkthrough re-verified non-interactively end to end.

nx-plugin-for-aws:
All **71 checks green, zero failures** — including `Smoke Tests - idempotency`, which is the confirmation that the reference-swap fix works rather than just my local check.

Branch is final and ready for re-review: title and commit subject are identical, single commit, rebased on current `main`.

cogwirrel (approved):

PR #1210, opened 2026-09-04
feat(ts#rdb): scan RDB container images under CDK

nx-plugin-for-aws:
### Reason for this change

Bug-bash findings across the RDB guides and the React/RDB connection pages. The one change to generated output:

**`databases-03` — the Image Scanning section was unreachable for CDK users.** Both RDB guides render the shared trivy snippet, promising a `trivy` target and a `.trivyignore`. In a CDK workspace neither existed for any RDB project (nor a `docker` target), despite all of them shipping Dockerfiles: `addDockerScanTarget` sat inside `if (iac === 'terraform')`. `pnpm trivy` reported "No tasks were run", so the migration image a CDK workspace deploys was never scanned.

The rest are documentation-only: the `dev` target's dependency on the Nx Daemon was undocumented on three of the four connection guides (`conn-web-05`), examples that do not typecheck (`conn-web-08`, `conn-web-09`, `conn-web-10`), a dead in-page anchor (`databases-04`), FileTrees that do not match generator output (`conn-web-11`, `conn-data-06`, `databases-07`), an undocumented generated line (`conn-data-07`), and an orphaned snippet (`conn-data-10`).

### Description of changes

**Image scanning for CDK RDB projects.** The RDB generators were the outlier: `ts#agent` and `ts#mcp-server` already add `docker` + `trivy` under both providers, and CDK builds the very same context (`DockerImageCode.fromImageAsset`). Both `ts#rdb` and `py#rdb` now wire up the build and its scan regardless of provider.

Turning the scan on surfaced two HIGH findings per RDB image. Both are **fixed at the source rather than suppressed** — the ignore list I first proposed is gone:

- **TypeScript:** the image installs the prisma CLI to run migrations, and the CLI's own ranges resolve `deepmerge-ts` below the CVE-2026-40345 fix and `mysql2` below the GHSA-3f6p-5ww8-9rcr fix. The install now overrides both via `npm pkg set`, exactly as the agent and MCP image builds already do for `minimatch`. The versions live in `TS_VERSIONS` and are declared through `PRISMA_IMAGE_DEPENDENCIES`, so the weekly sync keeps them current.
- **Python:** the findings were pip's *vendored* copies of `msgpack` and `setuptools` (under `site-packages/pip/_vendor/`) — not OS packages, so `microdnf upgrade` correctly never touched them, and pip is already at its latest so upgrading it does not help. Nothing installs packages at runtime, so both images drop pip.

**`conn-web-05` is documented, not worked around.** `nx watch` needs the daemon, so the `dev` targets built on it do too. Previously only the FastAPI guide mentioned this, and it framed the consequence as merely losing auto-regeneration; the Smithy, py-agent and AG-UI guides said nothing. All four now carry the same note stating the `dev` target requires the daemon (wording per review), and the Nx troubleshooting page matches.

Other doc fixes: guarded `data`/`isPending` and `users.data` in react-trpc; replaced `item.error.message` with `JSON.stringify(item.error.error)` in react-fastapi/react-smithy (the error is a `{ status, error }` union, which those guides document further down); typed the AG-UI slot component as `ButtonHTMLAttributes<HTMLButtonElement>`; pointed the ts-rdb anchor at `connection/trpc-rdb#using-the-middleware`; corrected the tRPC hooks FileTree (one file exporting both hooks) and the ts-rdb FileTree; annotated the conditional Dockerfile in four connection guides. For `conn-data-10` the orphaned `py-lambda-rdb-ssl-requirements` snippet is wired into `py-fast-api-rdb.mdx` rather than deleted — its content is accurate and its TypeScript counterparts are already rendered from the equivalent pages, so FastAPI readers connecting without RDS Proxy were the only ones left without guidance.

**No migration**, per review. Existing workspaces keep the shape they already deploy; new ones get the scan from today's generators.

`pnpm nx run @aws/nx-plugin:test -u`, `pnpm nx run-many --target lint --all --fix` and `pnpm nx run-many --target build --projects=@aws/nx-plugin` all pass. Snapshot changes are limited to the two RDB Dockerfiles. Added cases parameterise the RDB generator specs over both IaC providers to assert the image targets exist for each, assert they are absent with `--infra none`, and assert each Dockerfile carries its fix.

End-to-end in a scratch CDK workspace (`pnpm create @aws/nx-workspace`, linked to the local build) with `ts#rdb`, `py#rdb`, `ts#react-website`, `py#fast-api`, `ts#trpc-api` and an AG-UI `ts#agent`, all connected:

- **Image scanning, with real containers:** both RDB projects now get `docker`, `trivy` and `.trivyignore` under CDK. Built the TypeScript and Python migration images and ran the pinned Trivy image against each. Before the fixes Trivy **exits 1** (two HIGH findings per image — which is how I found them); after, it **exits 0** with an empty ignore file. Confirmed the overridden `deepmerge-ts@8.0.0` and `mysql2@3.22.0` actually land with the schema engine intact and `prisma --version` working, and that the Python migration handler still imports and runs without pip (`alembic`, `sqlalchemy`, `asyncpg`, `boto3` and `migration_handler.handler` all load). This also matters because the `Smoke Tests - trivy` lane scans every image the generator matrix produces.
- **Local database workflow:** started the Postgres container via `nx run <db>:dev` and applied a migration with `nx run <db>:prisma migrate dev`, which succeeded against the live container.
- **Examples:** pasted every corrected example into the workspace and confirmed `typecheck` passes. Each was first confirmed to fail with exactly the error the finding reported (TS18048, TS2339 on `UseQueryResult`/`message`, TS2322 on the slot) before the fix.
- **`conn-web-05`:** confirmed with `NX_DAEMON=false` that the watch task exits non-zero and takes `dev` down with it, i.e. the daemon really is required — which is what the note now states.

Not deployed to AWS. Ports in 14670-14699 and distinct container names were used to avoid collisions with other scratch workspaces on this box.

### Issue # (if applicable)

N/A — from a v1.0 bug bash.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/connection/react-agui.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-agui.mdx
@@ -205,12 +205,14 @@ Per-chat overrides still work alongside the theme — anything you pass as a slo
 
 ### Replacing a slot with a custom component
 
-Any slot can take a React component instead of a className, so you can replace the default entirely:
+Any slot can take a React component instead of a className, so you can replace the default entirely. Type your component against the props the slot declares — `sendButton` renders a `<button>`, so it receives `ButtonHTMLAttributes`:
 
 ```tsx
 import { CopilotChat } from './components/copilot';
 
-const MySendButton: React.FC<{ onClick: () => void }> = ({ onClick }) => (
+const MySendButton: React.FC<React.ButtonHTMLAttributes<HTMLButtonElement>> = ({
+  onClick,
+}) => (
   <button onClick={onClick} className="my-send-btn">
     Send
   </button>
@@ -247,6 +249,10 @@ The connection generator automatically configures `dev` integration:
 
 :::tip[Hot Reloading]
 The website and connected agent hot-reload together, enabling you to quickly iterate on both sides without deploying to AWS.
+:::
+
+:::note[The dev target requires the Nx Daemon]
+The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
 :::
 
 ## More Information
--- a/docs/src/content/docs/en/guides/connection/react-fastapi.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-fastapi.mdx
@@ -125,8 +125,8 @@ For more fine-grained control, you can use the `watch-generate:<ApiName>-client`
 />
 :::
 
-:::warning[File Watcher Dependency]
-`watch-generate:<ApiName>-client` relies on the [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) command, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be running. Therefore if you have disabled the daemon, the client won't automatically regenerate when making changes to your FastAPI.
+:::note[The dev target requires the Nx Daemon]
+The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
 :::
 
 ### Using the API Hook
@@ -146,12 +146,14 @@ function MyComponent() {
   const api = useMyApi();
   const item = useQuery(api.getItem.queryOptions({ itemId: 'some-id' }));
 
-  if (item.isLoading) return <div>Loading...</div>;
-  if (item.isError) return <div>Error: {item.error.message}</div>;
+  if (item.isPending) return <div>Loading...</div>;
+  if (item.isError) return <div>Error: {JSON.stringify(item.error.error)}</div>;
 
   return <div>Item: {item.data.name}</div>;
 }
 ```
+
+The error is a discriminated union of `{ status, error }` objects rather than an `Error`, so there is no top-level `message` to render. Narrow on `status` to read a specific response body — see [Error Handling](#error-handling) below.
 
 <Drawer title="Using the API client directly" trigger="Click here for an example using the vanilla client directly.">
 ```tsx {5,13}
@@ -222,7 +224,7 @@ function CreateItemForm() {
 
       {createItem.isError && (
         <div className="error">
-          Error: {createItem.error.message}
+          Error: {JSON.stringify(createItem.error.error)}
         </div>
       )}
     </form>
@@ -370,12 +372,12 @@ function ItemList() {
       }),
   });
 
-  if (items.isLoading) {
+  if (items.isPending) {
     return <LoadingSpinner />;
   }
 
   if (items.isError) {
-    return <ErrorMessage message={items.error.message} />;
+    return <ErrorMessage message={JSON.stringify(items.error.error)} />;
   }
 
   return (
--- a/docs/src/content/docs/en/guides/connection/react-py-agent.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-py-agent.mdx
@@ -183,6 +183,10 @@ The connection generator automatically configures `dev` integration:
 The website and connected agent will hot-reload, enabling you to quickly iterate on both together without deploying to AWS.
 :::
 
+:::note[The dev target requires the Nx Daemon]
+The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
+:::
+
 ## More Information
 
 For more information, please refer to:
--- a/docs/src/content/docs/en/guides/connection/react-smithy.mdx
+++ b/docs/src/content/docs/en/guides/connection/react-smithy.mdx
@@ -126,6 +126,10 @@ For more fine-grained control, you can use the `watch-generate:<ApiName>-client`
 />
 :::
 
+:::note[The dev target requires the Nx Daemon]
+The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
+:::
+
 ### Using the API Hook
 
 The generator provides a `use<ApiName>` hook which you can use to call your API with TanStack Query.
@@ -143,12 +147,14 @@ function MyComponent() {
   const api = useMyApi();
   const item = useQuery(api.getItem.queryOptions({ itemId: 'some-id' }));
 
-  if (item.isLoading) return <div>Loading...</div>;
-  if (item.isError) return <div>Error: {item.error.message}</div>;
+  if (item.isPending) return <div>Loading...</div>;
+  if (item.isError) return <div>Error: {JSON.stringify(item.error.error)}</div>;
 
   return <div>Item: {item.data.name}</div>;
 }
 ```
+
+The error is a discriminated union of `{ status, error }` objects rather than an `Error`, so there is no top-level `message` to render. Narrow on `status` to read a specific response body — see [Error Handling](#error-handling) below.
 
 <Drawer title="Using the API client directly" trigger="Click here for an example using the vanilla client directly.">
 ```tsx {5,13}
@@ -219,7 +225,7 @@ function CreateItemForm() {
 
       {createItem.isError && (
         <div className="error">
-          Error: {createItem.error.message}
+          Error: {JSON.stringify(createItem.error.error)}
         </div>
       )}
     </form>
@@ -331,12 +337,12 @@ function ItemList() {
       }),
   });
 
-  if (items.isLoading) {
+  if (items.isPending) {
     return <LoadingSpinner />;
   }
 
   if (items.isError) {
-    return <ErrorMessage message={items.error.message} />;
+    return <ErrorMessage message={JSON.stringify(items.error.error)} />;
   }
 
   return (
--- a/docs/src/content/docs/en/guides/connection/ts-mcp-server-rdb.mdx
+++ b/docs/src/content/docs/en/guides/connection/ts-mcp-server-rdb.mdx
@@ -36,13 +36,13 @@ Select your MCP server project as the source and your relational database projec
 
 ## Generator Output
 
-The generator modifies two files in your MCP server's source directory:
+The generator modifies the following files in your MCP server's source directory:
 
 <FileTree>
 
 - packages/my-service/src/my-mcp
   - server.ts Prisma client fetched and available to all tools registered inside `createServer`
-  - Dockerfile RDS CA bundle installed for SSL connections to Aurora
+  - Dockerfile RDS CA bundle installed for SSL connections to Aurora (only when the MCP server's `infra` is `agentcore-ecr`)
 
 </FileTree>
 
@@ -52,11 +52,12 @@ Additionally, the `<mcp-server-name>-dev` target is updated to depend on the dat
 
 The Prisma client is fetched inside `createServer` and available to all tools and resources registered there:
 
-```ts title="packages/my-service/src/my-mcp/server.ts" {1,4}
+```ts title="packages/my-service/src/my-mcp/server.ts" {1,4,5}
 import { getPrisma as getMyDb } from '@my-scope/my-db';
 
 export const createServer = async () => {
   const myDb = await getMyDb();
+  myDb.$on('error', console.error);
   const server = new McpServer({ name: 'my-service', version: '1.0.0' });
   // register tools/resources that use myDb
   return server;
--- a/packages/nx-plugin/src/ts/rdb/generator.ts
+++ b/packages/nx-plugin/src/ts/rdb/generator.ts
@@ -32,6 +32,7 @@ import {
   IMAGE_BUILD_CACHE,
   NODE_IMAGE_DEPENDENCIES,
   nodeImageVersions,
+  PRISMA_IMAGE_DEPENDENCIES,
 } from '../../utils/docker.js';
 import { formatFilesInSubtree } from '../../utils/format.js';
 import { FS_DEPENDENCIES, FsCommands } from '../../utils/fs.js';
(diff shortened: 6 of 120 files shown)


packages/nx-plugin/src/utils/watch-target.ts
```diff
@@ -0,0 +1,39 @@
...
+ * A continuous target which re-runs another target whenever the projects it
+ * derives from change.
+ *
+ * The executor is used rather than a `nx watch` command because `nx watch`
+ * exits 1 without the Nx Daemon, and a `continuous` task exiting non-zero fails
+ * the whole target that depends on it — taking the `dev` server down with it.
+ * The executor idles instead, so `dev` still comes up and only regeneration on
+ * change is lost.
+ */
+export const watchTarget = ({
```
cogwirrel: no. watch needs the daemon. no wrapper.

packages/nx-plugin/src/ts/rdb/generator.ts
```diff
@@ -341,29 +342,31 @@ export const tsRdbGenerator = async (
...
-    }
+    addDockerScanTarget(
+      tree,
+      {
+        project: projectConfig,
+        containerEngine,
+        trivyTargetName: 'trivy',
+        dockerTargetName: 'docker',
+        imageTags: [migrationDockerImageTag],
+        ignoredFindings: TS_RDB_TRIVY_IGNORES,
```
cogwirrel: is there no way we can address the cause? trivy won't error for unpatched?

packages/nx-plugin/src/py/rdb/utils.ts
```diff
@@ -4,6 +4,27 @@
...
 
+/**
+ * Findings the migration and create-db-user images' scan suppresses.
+ *
+ * Both are reached only through build-time tooling the Lambda Python base image
+ * ships, not through the handler's own dependencies, and `microdnf upgrade` in
+ * the vended `Dockerfile` does not move them — they are Python packages rather
+ * than OS ones. Remove an entry once a base image ships a version past it.
+ */
+export const PY_RDB_TRIVY_IGNORES = [
```
cogwirrel: same here, any way we can address this?

packages/nx-plugin/src/migrations/latest/rdb-image-scan-targets/migration.ts
```diff
@@ -0,0 +1,164 @@
...
+  const [scope, name] = projectName.startsWith('@')
+    ? projectName.slice(1).split('/')
+    : projectName.split('.');
+  return `${scope.replace(/_/g, '-')}-${name.replace(/_/g, '-')}-${tagSuffix}:latest`;
+};
+
+const divergedNextStep = (projectName: string, bundleTarget: string): string =>
+  `${projectName}: no \`${bundleTarget}\` target that stages the image build context, so the \`docker\` and \`trivy\` targets were not added. Add them by hand if you want the migration image scanned — see the Image Scanning section of the generator's guide.`;
+
+export default async function migration(
```
cogwirrel: remove this migration

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
```diff
@@ -50,6 +50,10 @@ root.render(
 
 <RunGenerator generator="connection" />
 
+:::note[Naming your FastAPI project]
+`--targetProject` takes the FastAPI's **Nx project name**, which for `py#fast-api --name=MyPyApi` is `<scope>.my_py_api` (the unqualified `my_py_api` also works). That is not the same as the hyphenated Python distribution name (`<scope>-my-py-api`) that `uv` prints while installing. Run <NxCommands commands={['show projects']} /> to list the names the generator accepts.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/connection/ts-mcp-server-rdb.mdx
```diff
@@ -52,17 +52,20 @@ Additionally, the `<mcp-server-name>-dev` target is updated to depend on the dat
...
 export const createServer = async () => {
   const myDb = await getMyDb();
+  myDb.$on('error', console.error);
   const server = new McpServer({ name: 'my-service', version: '1.0.0' });
   // register tools/resources that use myDb
   return server;
 };
 ```
 
+`getPrisma()` creates the client with `log: ['warn', 'error']`, and the `$on('error', ...)` subscription routes those error events to the console so database failures are visible in the MCP server's logs.
```
cogwirrel: remove

packages/nx-plugin/src/ts/rdb/generator.ts
nx-plugin-for-aws: Good call — both are addressable, and the ignore list is gone.

On `--ignore-unfixed`: the scan already passes it, so unfixed findings were never the issue. Both of these are reported as `fixed`, reached through the prisma CLI's own dependency tree (the image installs the CLI to run migrations), whose ranges resolve them below the fix:

- `deepmerge-ts` 7.1.5 -> 8.0.0 (CVE-2026-40345), via `@prisma/config`
- `mysql2` 3.15.3 -> 3.22.0 (GHSA-3f6p-5ww8-9rcr), a direct dep of `prisma`

So the fix is to override them at the point the Dockerfile installs the CLI, exactly as the agent/MCP image builds already do for `minimatch`. Verified in a real image: without it Trivy **exits 1**, with it **exits 0**, and both fixed versions land with the schema engine intact and `prisma --version` working. The versions live in `TS_VERSIONS` so the weekly sync keeps them current.

packages/nx-plugin/src/py/rdb/utils.ts
nx-plugin-for-aws: Also addressed at the cause — this ignore list is gone too, and my original note about the base image was wrong.

Tracing the two findings: they are **pip's vendored copies**, at `/var/lang/lib/python3.14/site-packages/pip/_vendor/{msgpack,cachecontrol,...}` — not OS packages (so `microdnf upgrade` correctly does not touch them) and not our bundle. pip is already at its latest (26.2.1), so upgrading it does not help; the vulnerable copies are what it ships.

But nothing installs packages at runtime — the handler's dependencies are baked into the image by the bundle step — so pip is dead weight. Both Dockerfiles now drop it. Verified in a real image: Trivy **exits 0** afterwards, and the migration handler still imports and runs (`alembic`, `sqlalchemy`, `asyncpg`, `boto3` and `migration_handler.handler` all load fine).

packages/nx-plugin/src/utils/watch-target.ts
nx-plugin-for-aws: Understood — wrapper removed entirely, along with the executor, the shared `watchTarget` helper and its migration. The vended watch targets are back to the plain `nx watch` command.

`conn-web-05` is now fixed as documentation only. The point worth keeping is that the FastAPI guide understated the symptom: it said the client "won't automatically regenerate", when in fact the watch task is `continuous`, so its non-zero exit fails the whole `dev` target and the website is never served at all (curl → 000). The four guides and the Nx troubleshooting page now say that plainly — the daemon is required for `nx dev <website>`, and it is also off by default under `CI=true`.

packages/nx-plugin/src/migrations/latest/rdb-image-scan-targets/migration.ts
nx-plugin-for-aws: Removed — migration, spec and its folder are all gone. Same for the `watch-executor` migration that came with the wrapper, so this PR now ships no migrations at all.

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/connection/ts-mcp-server-rdb.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/connection/react-agui.mdx
```diff
@@ -249,6 +251,10 @@ The connection generator automatically configures `dev` integration:
 The website and connected agent hot-reload together, enabling you to quickly iterate on both sides without deploying to AWS.
 :::
 
+:::caution[The dev target requires the Nx Daemon]
+If your website also has a connection whose `dev` target watches for changes (a FastAPI, Smithy or Python agent connection), that watching uses [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) and requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon). With the daemon disabled the watch target exits immediately and takes the whole `dev` target down with it, so **nothing is served**. Note the daemon is also off by default when `CI=true`. Re-enable it (`NX_DAEMON=true`) or run the underlying servers directly.
+:::
```
cogwirrel: Change to:

:::note[The dev target requires the Nx Daemon]
The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
:::

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
```diff
@@ -125,8 +125,10 @@ For more fine-grained control, you can use the `watch-generate:<ApiName>-client`
 />
 :::
 
-:::warning[File Watcher Dependency]
-`watch-generate:<ApiName>-client` relies on the [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) command, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be running. Therefore if you have disabled the daemon, the client won't automatically regenerate when making changes to your FastAPI.
+:::caution[The dev target requires the Nx Daemon]
+`watch-generate:<ApiName>-client` uses [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching), which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon). With the daemon disabled it exits immediately, and because `dev` depends on it, the whole `dev` target fails — **your website is not served at all**, rather than merely losing regeneration on change. Note the daemon is also off by default when `CI=true`.
+
+If you have disabled the daemon, either re-enable it (`NX_DAEMON=true`) or run the underlying servers directly, e.g. `nx run <WebsiteProject>:"generate:<ApiName>-client"` once, then the website's and API's own serve commands.
 :::
```
cogwirrel, line 130: change to:

:::note[The dev target requires the Nx Daemon]
The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
:::

docs/src/content/docs/en/guides/connection/react-py-agent.mdx
```diff
@@ -183,6 +183,10 @@ The connection generator automatically configures `dev` integration:
 The website and connected agent will hot-reload, enabling you to quickly iterate on both together without deploying to AWS.
 :::
 
+:::caution[The dev target requires the Nx Daemon]
```
cogwirrel: change to:

:::note[The dev target requires the Nx Daemon]
The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
:::

docs/src/content/docs/en/guides/connection/react-smithy.mdx
```diff
@@ -126,6 +126,12 @@ For more fine-grained control, you can use the `watch-generate:<ApiName>-client`
 />
 :::
 
+:::caution[The dev target requires the Nx Daemon]
```
cogwirrel: change to:

:::note[The dev target requires the Nx Daemon]
The `dev` target hot-reloads local dev servers. Many of these rely on [`nx watch`](https://nx.dev/docs/guides/tasks--caching/workspace-watching) to achieve this, which requires the [Nx Daemon](https://nx.dev/docs/concepts/nx-daemon) to be enabled.
:::

docs/src/content/docs/en/guides/connection/react-agui.mdx
nx-plugin-for-aws: Applied verbatim.

docs/src/content/docs/en/guides/connection/react-fastapi.mdx
nx-plugin-for-aws: Applied verbatim. Also dropped the separate note I had added higher up this page, per your other comment.

docs/src/content/docs/en/guides/connection/react-py-agent.mdx
nx-plugin-for-aws: Applied verbatim.

docs/src/content/docs/en/guides/connection/react-smithy.mdx
nx-plugin-for-aws: Applied verbatim. I also matched the same framing on the Nx troubleshooting page, which had the old "hot-reloading will not function" wording.

packages/nx-plugin/src/py/rdb/utils.ts
nx-plugin-for-aws: Tested both alternatives you'd expect to work, and neither does — but the images are clean now, so the ignore list is gone either way.

The findings are **pip's vendored copies**, at `site-packages/pip/_vendor/`. `pip/_vendor/vendor.txt` in the newest base image pins exactly the vulnerable versions:

```
msgpack==1.1.2
setuptools==70.3.0
```

What I tried:

1. **`pip install --upgrade pip setuptools msgpack`** — Trivy still **exits 1**. It installs clean top-level `msgpack 1.2.2` / `setuptools 84.0.0`, but `pip/_vendor/` is untouched, so the vulnerable copies remain and are still what gets reported.
2. **A newer base image tag** — `3.14` already resolves to the newest stable (`3.14.2026.09.01.20`); the only newer tags are `-experimental`. That tag vendors the same two versions, so it does not help. (Scanning the bare tag reports a *different* HIGH, in `openssl-fips-provider`, which our existing `microdnf upgrade` already clears — so the OS-package half of the Dockerfile is doing its job.)

Also confirming `--ignore-unfixed` was never the mechanism: both are reported as `fixed`, so they were never being kept alive by that flag.

So the fix is to remove pip. Nothing installs packages at runtime — the handler's dependencies are baked in by the bundle step — so both Dockerfiles now drop it. Verified in a real image: Trivy **exits 0**, and the migration handler still imports and runs (`alembic`, `sqlalchemy`, `asyncpg`, `boto3` and `migration_handler.handler` all load).

cogwirrel (dismissed):

cogwirrel (approved):

PR #1211, opened 2026-09-04
fix(terraform#project): preserve user-authored files on generator re-run

nx-plugin-for-aws:
### Reason for this change

Four generators silently destroyed the code their own guides tell the reader to write, whenever the generator was re-run with the same name. Each exited 0 with no prompt, no warning and no backup, and `build`/`lint`/`test`/`typecheck` all still passed afterwards — so the regression was completely invisible.

- **`terraform-01`** (blocker) — `terraform#project` overwrote `src/main.tf`, `src/variables.tf`, `src/outputs.tf` and `src/env/dev.tfvars`. `checkov.yml` and `project.json` were already protected, which made the inconsistency worse: the two files the generator guarded are the ones the user is *least* likely to have edited.
- **`ts-core-01`** (blocker) — `ts#infra` overwrote `src/main.ts` and `src/stacks/application-stack.ts`, deleting every stack the user declared and every extra stage.
- **`databases-01`** (major) — `ts#dynamodb` reset `src/entities/index.ts` and `config.json`, unregistering the user's entities and dropping any GSI they added.
- **`databases-02`** (major) — `py#dynamodb` reset `<name>/entities/__init__.py` and `config.json`. Higher impact than the TS twin: an unimported PynamoDB model never registers its discriminator, so the documented cross-entity `BaseModel.gsi1_index.query(...)` dispatch silently stops returning those types.

The contract being violated is `CONTRIBUTING.md` §"Generator Idempotency"'s own absolute rule — *"Never destroy user intent"* — not an inferred expectation. `ts#rdb` / `py#rdb` already complied and were used as the reference.

### Description of changes

Every generator in the lane now emits its templates through a **single `generateFiles` call with `OverwriteStrategy.KeepExisting`**. No template files were moved or edited: `files/` is byte-identical to `main` in all four generators, so the entire behavioural change is the strategy plus one guard.

**`terraform#project`** — `Overwrite` → `KeepExisting`. One-word change.

**`ts#infra`** — `Overwrite` → `KeepExisting`, plus the real fix: `tree.delete(joinPathFragments(libraryRoot, 'src'))` is now guarded behind `isNewProject`. That delete was the destructive step all along — it removed the whole `src` directory on every run, so no overwrite strategy could have saved the user's stacks. The CDK app layout only needs to replace the library `src` that `tsProjectGenerator` scaffolds, which happens exactly once. (Thanks @cogwirrel — the first revision split templates into framework/user halves to decide what could safely converge; guarding the delete makes that classification unnecessary, and the result is both simpler and safer.)

**`ts#dynamodb` / `py#dynamodb`** — `KeepExisting` on the entity and client templates, and the two files the verifier called out are **merged rather than kept wholesale**, because plain KeepExisting would leave a re-run unable to register anything and would silently strand the generated entity:

- **Entity barrels** are merged via a new `addNamedExport` helper (TS) and the existing `addPythonReExport` (Python) — the same approach `ts#rdb`/`py#rdb` already use for their barrels. The generated entity's export is re-added only if absent; the user's own exports are untouched. This matters most in Python, where the barrel import is what registers a model's discriminator.
- **`config.json`** is merged by a new `writeDynamoDBConfig` util: `tableConfig.globalSecondaryIndexes` is preserved (defaults seed a brand-new project only, and an intentionally-emptied list is respected), while `runtimeConfigKey` and `localDev` are rewritten from the generator's options so changing e.g. `--tableName` still takes effect. This mirrors `ts#rdb`, whose existing comment already documents `config.json` as framework-owned.

`addPythonReExport` gained a "create the file if absent" branch, matching `addStarExport`, since that barrel is no longer vended as a template.

**No migration**, deliberately. This changes *when* a generator writes, not *what* it writes — a workspace that never re-ran a generator is already byte-identical to one generated today. For the workspaces this bug actually hurt, the destroyed content is gone from the tree and survives only in the user's git history: a codemod cannot recover it, and anything attempting to would either no-op or overwrite whatever the user has since rewritten by hand, reintroducing the very failure being fixed. The two snapshot deltas are the barrels converging on the shape `ts#rdb`/`py#rdb` already produce (a retained `// Export your library code here` stub, and the Nx docstring on the Python package `__init__.py`) — cosmetic, and covered by the "formatting or stylistic changes to vended templates" exception.

**Tests.** The pre-existing idempotency tests gave false assurance: they compared generator output to generator output without ever writing user content between the runs (`terraform/project/generator.spec.ts:823` was the cited example). Every test added here writes **distinctive user content between the two runs and asserts it survives**:

- `terraform#project` — authors a bucket, a variable, an output, `dev.tfvars`, a second `prod.tfvars` and a hand-added `networking.tf`; plus a library-project variant, curated `checkov.yml`, and edits to `bootstrap/main.tf` and `scripts/init.ts`.
- `ts#infra` — authors a `Bucket` in the stack, a beta stage in `main.ts`, a stack of the user's own and custom Checkov skips; and a dedicated test that a file added under `src/` survives, which is what pins the guarded delete.
- `ts#dynamodb` / `py#dynamodb` — follow the guides' Data Modelling and GSI steps verbatim (new entity + curated barrel + `gsi3` + an edited generated entity); assert the barrel re-registers a *removed* generated export without discarding the user's, and that `localDev` still converges under a changed `--tableName` while the GSI list is carried through.

New unit tests cover `addNamedExport` (5 cases) and `writeDynamoDBConfig` (4 cases, incl. the empty-GSI-list edge case). Every one of these passed unchanged under the simplified single-`KeepExisting` design — none of them depended on the framework/user split.

```
pnpm nx run-many --target lint --projects=@aws/nx-plugin --fix   # green
pnpm nx run-many --target build --projects=@aws/nx-plugin        # green (runs the full test target)
```

111 tests across `terraform#project` + `ts#infra`, 54 across the DynamoDB pair, no snapshot writes. I also re-ran every consumer of the two shared helpers I touched — `src/ts/rdb`, `src/py/rdb`, `src/py/agent/**`, `src/utils/agent-connection` (394 tests) — and the full `src/migrations` + `src/utils/version-upgrade-migration` suite (832 tests).

**End-to-end**, against the locally compiled plugin linked into scratch workspaces (stopped at `synth`/`validate`, nothing deployed):

- *Terraform workspace* — generated `terraform#project`, appended a marked bucket/variable/output/tfvar, md5summed, re-ran with identical args. The re-run printed **no `UPDATE` lines at all** and all eight checksums verified `OK`. `terraform init -backend=false && terraform validate` succeeded and `terraform fmt -check` was clean. (`nx validate` needs a bootstrapped remote-state bucket, i.e. an AWS deploy, so I validated the HCL directly.)
- *CDK workspace* — generated `ts#infra` + `ts#dynamodb` + `py#dynamodb`, authored six distinctively marked edits across the stack, `main.ts`, both entity barrels, two new entity files and `gsi3` in both `config.json`s, then re-ran all three with identical args. All six markers survived; `gsi3` survived in both configs; `localDev.tableName` still converged; the only write was a `config.json` re-serialisation. `nx run-many --target build --all` then passed end to end including `synth` and `checkov` — the first checkov run failed *only* on the deliberately-bare `MyPreciousBucket` I had authored, which is itself proof the preserved code reached the synthesized template; hardening that bucket turned the build green.

I did not run `nx build docs` locally (it OOMs this box); CI covers it. The English guide notes from the first revision have been removed per review, so this PR now touches no docs at all.

### Issue # (if applicable)

Closes bug-bash findings `terraform-01`, `ts-core-01`, `databases-01`, `databases-02`.

### Checklist

----


docs/src/content/docs/en/guides/py-dynamodb.mdx
```diff
@@ -65,6 +65,10 @@ The generated project uses [PynamoDB](https://pynamodb.readthedocs.io/) for enti
 
 Add or update entity files under `<name>/entities/`, using the generated example entity as a starting point:
 
+:::note[Re-running the generator]
+Your entity modules and the `<name>/entities/__init__.py` barrel are yours, so re-running the generator for an existing project leaves them as you wrote them — the generated `BaseModel` and `ExampleModel` imports are re-registered in the barrel if they are missing, without touching your own imports. In `config.json` the `tableConfig.globalSecondaryIndexes` list is preserved, while `runtimeConfigKey` and `localDev` are refreshed from the generator's options.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -111,6 +111,10 @@ You can start writing your Terraform infrastructure inside `src/main.tf`, for ex
 +}
 ```
 
+:::note[Re-running the generator]
+Everything under `src` — `main.tf`, `variables.tf`, `outputs.tf`, `providers.tf` and the `env/*.tfvars` files — is yours, so re-running the generator for an existing project leaves it as you wrote it. `checkov.yml` and `project.json` are preserved too. The `bootstrap` configuration and the `scripts` helpers that back the nx targets are regenerated, so a re-run picks up fixes to them.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/ts-dynamodb.mdx
```diff
@@ -63,6 +63,10 @@ The local development scripts are shared across all DynamoDB projects (both Type
 
 The generated project uses [ElectroDB](https://electrodb.dev/) for type-safe entity modelling on a single DynamoDB table, following [DynamoDB's single-table design](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/data-modeling-foundations.html). Add or update entity files under `src/entities/`, using the generated example entity as a starting point.
 
+:::note[Re-running the generator]
+Your entity files and the `src/entities/index.ts` barrel are yours, so re-running the generator for an existing project leaves them as you wrote them — the generated example entity's export is re-registered in the barrel if it is missing, without touching your own exports. In `config.json` the `tableConfig.globalSecondaryIndexes` list is preserved, while `runtimeConfigKey` and `localDev` are refreshed from the generator's options.
+:::
```
cogwirrel: remove this

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
```diff
@@ -85,6 +85,12 @@ export class ApplicationStack extends Stack {
 }
 ```
 
+:::note[Re-running the generator]
+Everything under `src` — your stacks, your stages and `main.ts` — is yours, so re-running the generator for an existing project leaves it as you wrote it, as does `checkov.yml`. `cdk.json` is regenerated, so a re-run picks up fixes to it.
+
+Because `main.ts` is preserved, re-running with `--stageConfig` on a project generated without it adds the `common/infra-config` and `common/scripts` packages and repoints the deploy targets, but does not rewrite `main.ts` to read from the config. Follow the `stageConfig` example below to wire `resolveStage` in yourself.
+:::
```
cogwirrel: remove this

packages/nx-plugin/src/infra/app/generator.ts
```diff
@@ -140,23 +141,45 @@ export async function tsInfraGenerator(
...
-      pkgMgrCmd: getPackageManagerDisplayCommands().exec,
-      dir: lib.dir,
-      stageConfig,
-      ...schema,
-      ...esmVars(tree),
+      overwriteStrategy: OverwriteStrategy.KeepExisting,
     },
+  );
+
+  // `cdk.json` and the README are framework-owned, so they converge on a re-run.
```
cogwirrel: no. all files here should be generated with keepexisting - undo the framework/user split. the fix is to guard the delete of src as you've done above.

nx-plugin-for-aws: Done — collapsed to a single `generateFiles` with `KeepExisting` and reverted the framework/user split, so `files/app/` is byte-identical to main again. You're right that the guarded `src` delete was the actual fix: the `Overwrite` strategy was never what destroyed the stack, deleting the whole directory on every run was.

Applied the same simplification to `terraform#project`, which I'd split the same way — it's now one blanket `KeepExisting` too, with no template moves. `ts#dynamodb` and `py#dynamodb` were already a single blanket call.

I swapped the `cdk.json` convergence test for one asserting the guarded delete directly (a file the user adds under `src/` survives a re-run). All the tests that write user content between the two runs still pass unchanged — 111 tests across terraform + infra, 54 across the DynamoDB pair.

docs/src/content/docs/en/guides/py-dynamodb.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/ts-dynamodb.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/typescript-infrastructure.mdx
nx-plugin-for-aws: Removed.

cogwirrel (approved):

PR #1212, opened 2026-09-04
feat(agentcore-gateway,agentcore-harness): add lint/format/typecheck targets

nx-plugin-for-aws:
### Reason for this change

**The user-facing change:** `agentcore-gateway` and `agentcore-harness` projects got no `build`, `lint` or `typecheck` target and no `tsconfig.json` (finding `gateway-06`). Their vended TypeScript — `local-dev.ts` (~130 lines, edited by every connection generator) and `scripts/chat.ts` — was invisible to `nx run-many --target build --all` and to the Biome sweep, so `pnpm build` gave a false all-green and a type error in code the user is expected to run only surfaced at `nx dev` runtime.

Two smaller generator defects ride along:

- **`agents-08`** — `ts#agent` was not byte-idempotent: a re-run rewrote `project.json`, moving the `<name>-package` target's `dependsOn` from before `options` to after it. Identical content, key order only, but an unexplained tracked-file diff on every invocation.
- **`conn-web-12`** — `targetComponent` rejects the agent name the guides tell you to supply. After `ts#agent --name=StoryAgent`, `--targetComponent=StoryAgent` fails with `Component 'StoryAgent' not found ... Available components: story-agent`. **See the note below — this one is not settled yet.**

The PR also carries documentation corrections for the same page group (`agents-05`, `agents-06`, `agents-07`, `gateway-07`, `gateway-08`, `conn-agent-01`, `conn-agent-02`, `conn-agent-04`, `conn-agent-05`). Those are guide fixes rather than release-note material, so they are deliberately out of the title; the substantive ones are that the `ts#mcp-server` tool/resource examples failed the workspace's own `lint` and `compile` when pasted, and that the TypeScript connection guides called `create(sessionId)` when the generated client's `create()` is zero-arg (`TS2554` if copied) — the Python guides get that right, so it was TS-only drift.

### Description of changes

**Generated output:**

- `addScriptProjectTargets` (`utils/script-project-targets.ts`) gives a script-only TypeScript project the same gates a `ts#project` gets: a `noEmit` `tsconfig.json` (written only when absent, so a user's own config survives) plus `typecheck` (`tsc --noEmit`), `lint`, `format` and a `build` that runs them. Wired into `agentcore-gateway` and `agentcore-harness`. `compile` is deliberately absent — these projects run through `tsx`, so there is no artifact to produce.

  Per review, the lint/format targets are **not** redefined here: `biomeTargets` and `registerBiomeNamedInput` are extracted from `ts/lib/biome.ts` and called from both `configureBiomeLint` and `addScriptProjectTargets`, so there is one definition and the two cannot drift. Only the tsconfig handling is specific to the script-project case. Net −70 lines.

  The tsconfig is deliberately **not** registered as a workspace project reference: a `noEmit` project produces no declarations to consume, and `@nx/js/typescript` disables an inferred `typecheck` for any project reachable from one — I hit that failure mode first, hence the plain `nx:run-commands` target.

- The packaging, docker and trivy targets now go through the existing `normalizeTargetKeyOrder` helper rather than authoring key order by hand. Fixing it in the shared helpers (`agent-core-packaging.ts`, `docker.ts`) also makes `ts#mcp-server`, `py#agent` and `py#mcp-server` byte-idempotent, which had the same latent defect.

**Docs** (English only; translations are bot-generated): filetrees reconciled against real generated output; the tRPC/`client.ts` claim moved inside the `protocol: 'http'` filter; the MCP examples reformatted to the vended Biome style with `fetchSomeData` defined; `create(sessionId)` → `create()` with the prose rewritten to describe the actual `AsyncLocalStorage` (`enterSessionContext` / `runWithSessionId`) plus `agentcore-fetch.ts` header mechanism, mirroring how the Python guides phrase it; and a new snippet noting that `ts#infra` must be generated before there is an `application-stack.ts` to edit.

> **`conn-web-12` is not addressed by this PR.** Per review the `connection/schema.json` description edit is reverted and the `targetComponent` casing paragraph is removed from the three react connection guides, so the `findComponentInMetadata` fallback has been dropped too rather than shipping a behaviour change with no documentation. `packages/nx-plugin/src/connection/` is now byte-identical to `main`. The finding — `--targetComponent=StoryAgent` failing when the recorded component name is `story-agent` — remains **open** and should be tracked separately.

**Migration** — added `latest/script-project-quality-gates` (deterministic) for the `gateway-06` change, since it alters generated *targets* rather than application code a user would have hand-patched. It skips a project already carrying `typecheck`, and skips-and-reports via `nextSteps` a project reworked into a compiled `ts#project` (detected by `tsconfig.lib.json`) rather than layering a `noEmit` config over one that emits. 7 unit tests, including one that writes user content between two runs and asserts it survives. Happy to drop it if you'd rather not carry it.

**Unit tests** — added tests asserting the new gate targets and tsconfig shape, that the noEmit tsconfig is not added as a project reference, and idempotency tests that write user content between the two runs (`local-dev.ts`, a customised `tsconfig.json`, `agent.ts`) and assert it survives — the existing tests passed only because they compared generator output to generator output. Also a parameterised test asserting the packaging target's `dependsOn` precedes `options`, so the ordering regression cannot return silently.

```
pnpm nx run @aws/nx-plugin:test -u                           # 659 passed across the affected specs
pnpm nx run-many --target lint --all --fix                   # clean
pnpm nx run-many --target build --projects=@aws/nx-plugin    # clean
```

The one snapshot change (`agentcore-harness`) is exactly the new gate targets plus the new `tsconfig.json` — the intended change to what users get. The refactor produced **no** further snapshot drift, confirming it is behaviour-preserving. Four other snapshot files showed pure obsolete-snapshot prunes from running `-u` with a filter; I reverted those as they are not part of this change.

**End-to-end**, in scratch workspaces against the locally compiled plugin:

- Generated the projects each page targets across the documented permutations — `ts#agent` with `protocol=http|a2a|ag-ui` and `infra=agentcore|agentcore-ecr`, `ts#mcp-server`, `agentcore-gateway`, `agentcore-harness`, `ts#dcr-proxy`, `py#agent`, `py#mcp-server` — and verified every corrected FileTree against `find` output.
- Pasted both corrected MCP examples in, registered them in `server.ts`, and confirmed `lint` and `compile` pass (they failed both before).
- `gateway-06`: both projects now appear in `nx show projects --with-target build|lint|typecheck`; `nx run-many --target build --all` runs their new targets; a deliberate type error is caught in **both** files (`local-dev.ts` → `TS2322`; `scripts/chat.ts` → `TS2322`) and passes again once reverted; `nx sync:check` is clean, confirming the noEmit tsconfig needs no project reference. **Re-verified after the refactor**: byte-identical generated `project.json`, and the type error is still caught.
- Idempotency: re-ran `ts#agent` (both infra modes), `ts#mcp-server`, `agentcore-gateway` and `agentcore-harness` against committed workspaces — `git diff` is now empty in every case, where `ts#agent` previously produced a 2-insertion/2-deletion `project.json` diff.

Not deployed to AWS and no live Bedrock calls.

### Issue # (if applicable)

N/A — bug bash findings.

### Checklist

----


docs/src/content/docs/en/guides/connection/agentcore-gateway-agent.mdx
```diff
@@ -37,6 +38,8 @@ A TypeScript `http` agent serves tRPC over WebSocket, and AgentCore Gateway does
 
 Select the Gateway project as the source and the agent project as the target. If the agent project contains multiple components, specify `targetComponent` to disambiguate.
 
+<Snippet name="connection/project-names" />
```
cogwirrel: no, remove this snippet and all references you've added.

docs/src/content/docs/en/guides/agentcore-gateway.mdx
```diff
@@ -42,9 +42,13 @@ The generator creates a new project at `packages/<name>/`, plus a CDK construct
       - permit-all.cedar Default Cedar policy that permits authenticated callers
       - README.md Reference for writing Cedar policies
     - local-dev.ts Local gateway for local development — aggregates attached MCP servers (`mcp`) or proxies attached agents (`http`)
-    - project.json Adds the `serve` and `dev` targets
+    - package.json Declares the dependencies `local-dev.ts` imports
+    - tsconfig.json TypeScript configuration used by the `typecheck` target
+    - project.json Adds the `serve`, `dev`, `build`, `lint`, `format` and `typecheck` targets
 </FileTree>
 
+`local-dev.ts` is checked by the workspace's build sweep: `build` runs `lint` and `typecheck`, the latter type-checking the project with `tsc --noEmit`. Nothing is compiled — the script runs directly through [`tsx`](https://tsx.is/).
```
cogwirrel: remove

docs/src/content/docs/en/guides/agentcore-harness.mdx
```diff
@@ -32,10 +32,13 @@ The generator creates a standalone project at `packages/<name>/`. Because AWS ru
   - packages/\<name>/
     - src/PROMPT.md The Harness system prompt
     - scripts/chat.ts Multi-turn chat client for the deployed Harness
-    - project.json Adds the `chat` target
+    - tsconfig.json TypeScript configuration used by the `typecheck` target
+    - project.json Adds the `chat`, `build`, `lint`, `format` and `typecheck` targets
     - README.md Chat and customization instructions
 </FileTree>
 
+`scripts/chat.ts` is checked by the workspace's build sweep: `build` runs `lint` and `typecheck`, the latter type-checking the project with `tsc --noEmit`. Nothing is compiled — the script runs directly through [`tsx`](https://tsx.is/).
```
cogwirrel: remove

packages/nx-plugin/src/connection/schema.json
```diff
@@ -22,11 +22,11 @@
...
     "sourceComponent": {
       "type": "string",
-      "description": "The source component to connect from (component name, path relative to source project root, or generator id). Use '.' to explicitly select the project as the source."
+      "description": "The source component to connect from (component name in any casing, path relative to source project root, or generator id). Use '.' to explicitly select the project as the source."
     },
     "targetComponent": {
       "type": "string",
-      "description": "The target component to connect to (component name, path relative to target project root, or generator id). Use '.' to explicitly select the project as the target."
+      "description": "The target component to connect to (component name in any casing, path relative to target project root, or generator id). Use '.' to explicitly select the project as the target."
     },
```
cogwirrel: undo

packages/nx-plugin/src/utils/script-project-gates.ts
```diff
@@ -0,0 +1,140 @@
...
+ * `ts#project` gets.
+ */
+const SCRIPT_TSCONFIG_FILE_NAME = 'tsconfig.json';
+
+/**
+ * The lint and format targets a `ts#project` gets, so a script project is
+ * covered by the same Biome sweep. Falls back to a no-op when the workspace
+ * carries no `biome.json`, matching `configureBiomeLint`.
+ */
+const biomeTargets = (
```
cogwirrel: reuse/refactor from ts project generator / utils

nx-plugin-for-aws: Done — extracted `biomeTargets` and `registerBiomeNamedInput` from `ts/lib/biome.ts` and `addScriptProjectGates` now calls them, so there's one definition of the lint/format targets rather than a mirrored copy. `configureBiomeLint` is refactored onto the same helpers, so `ts#project` and the script projects can't drift.

The tsconfig handling stays in `script-project-gates.ts` since it's the part that genuinely differs (`noEmit`, and deliberately not registered as a workspace project reference — `@nx/js/typescript` disables an inferred `typecheck` for any project reachable from a `noEmit` one).

Net -70 lines. Re-verified after the refactor: identical generated `project.json`, no snapshot drift, 665 tests pass, and `typecheck` still catches a deliberate type error in `local-dev.ts` (`TS2322`) and passes once reverted.

docs/src/content/docs/en/guides/connection/agentcore-gateway-agent.mdx
nx-plugin-for-aws: Removed — snippet file deleted along with all nine references and the `Snippet` imports that were only there for it.

docs/src/content/docs/en/guides/agentcore-gateway.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/agentcore-harness.mdx
nx-plugin-for-aws: Removed.

packages/nx-plugin/src/connection/schema.json
nx-plugin-for-aws: Reverted — both descriptions are back to the original wording.

One question on the behaviour behind it, so I don't leave it half-done. The bug bash finding was that `--targetComponent` rejects the name the guides tell you to pass: after `ts#agent --name=StoryAgent`, running the connection generator with `--targetComponent=StoryAgent` fails with `Component 'StoryAgent' not found ... Available components: story-agent`, because the recorded component name is kebab-cased. The guides say to supply "the same one you chose when you ran the ts#agent or py#agent generator".

This PR also has a one-line change in `findComponentInMetadata` that falls back to matching the kebab-cased reference (exact name still wins first, so it can't redirect a reference that already matches). Would you like:

1. that generator fallback kept, so `StoryAgent` is accepted; or
2. the fallback dropped too, and the guides changed to tell readers to pass the kebab-cased id?

Happy either way — just don't want to guess. If you'd rather this finding wasn't addressed in this PR at all, say so and I'll strip the `findComponentInMetadata` change and its tests.

docs/src/content/docs/en/guides/connection/react-agui.mdx
```diff
@@ -33,6 +33,8 @@ Before using this generator, ensure you have:
 
 You will be prompted to select your React website as the source project and the project containing your AG-UI Agent as the target project. If your target project contains multiple components (such as multiple agents or other component types), you will be prompted to specify a `targetComponent` to disambiguate.
 
+`targetComponent` accepts the `--name` you gave the agent generator in any casing (`StoryAgent`, `story-agent`), the component's path relative to the target project root, or its generator id.
```
cogwirrel: remove this line and any other repeats in other guides

packages/nx-plugin/src/agentcore-gateway/generator.ts
```diff
@@ -143,6 +144,9 @@ export const agentcoreGatewayGenerator = async (
   project.targets ??= {};
   project.targets['serve'] ??= localGatewayTarget(port);
   project.targets['dev'] ??= localGatewayTarget(port);
+  // local-dev.ts is TypeScript the connection generators edit and the user is
+  // invited to tune, so it goes through the workspace's build sweep.
+  addScriptProjectGates(tree, { project });
```
cogwirrel: rename to addScriptProjectTargets

packages/nx-plugin/src/agentcore-harness/generator.spec.ts
```diff
@@ -44,6 +44,47 @@ const CHAT_TARGET = {
   },
 };
 
+/**
+ * The quality gates `scripts/chat.ts` is covered by, so the vended TypeScript
+ * the user runs is checked by the workspace's build sweep.
+ */
+const GATE_TARGETS = {
```
cogwirrel: use a snapshot instead of defining this duplication

packages/nx-plugin/src/agentcore-gateway/generator.ts
nx-plugin-for-aws: Renamed to `addScriptProjectTargets`, and renamed the file to `utils/script-project-targets.ts` to match (a `-gates` file exporting `addScriptProjectTargets` read oddly). Both call sites and the migration updated.

packages/nx-plugin/src/agentcore-harness/generator.spec.ts
nx-plugin-for-aws: Done — dropped the hand-written target definitions. The existing `snapshotTreeDir` test already covers the whole project directory including `project.json`, so the gates' exact shape now comes from that snapshot; I checked the generated snapshot rather than blessing it blind, and it does contain `build` (`dependsOn: [lint, typecheck]`), `lint` (`dependsOn: [format]`), `format` and `typecheck`.

The behavioural assertions kept alongside it are the ones a snapshot can't make on its own: the exact set of target names (so a leftover `invoke` target still fails), that `build` pulls the gates in, and that `chat` is untouched on a re-run and on an `infra: none` → `agentcore` upgrade. No obsolete snapshots left behind.

docs/src/content/docs/en/guides/connection/react-agui.mdx
nx-plugin-for-aws: Removed — that line appeared in three guides (`react-agui`, `react-ts-agent`, `react-py-agent`), and the translation sync had already propagated it to `fr` and `it`, so it's gone from all nine files. I audited every locale afterwards to confirm nothing references it.

That line was the documentation half of the `targetComponent` casing finding, so with it gone I've also dropped the `findComponentInMetadata` fallback and its tests rather than leaving the behaviour change undocumented — the finding is recorded as still open instead of half-applied. Answered on the schema thread too.

packages/nx-plugin/src/connection/schema.json
nx-plugin-for-aws: Reverted, and I've now taken the whole `targetComponent` casing change back out — `packages/nx-plugin/src/connection/` is byte-identical to `main` again (generator, schema and spec). Removing the doc paragraph you flagged on `react-agui.mdx` left the `findComponentInMetadata` fallback as an undocumented behaviour change, so dropping both was the coherent option rather than shipping half of it.

That leaves the underlying finding open: after `ts#agent --name=StoryAgent`, `--targetComponent=StoryAgent` still fails with `Component 'StoryAgent' not found ... Available components: story-agent`. I've noted it as **not addressed by this PR** in the description so it gets tracked separately rather than being assumed fixed. Happy to pick it up in its own PR if you'd like it changed either way.

cogwirrel (dismissed):

nx-plugin-for-aws:
All eight review comments are addressed and CI is green (71/71). Summary of this round:

- `addScriptProjectGates` → **`addScriptProjectTargets`**, and the file renamed to `utils/script-project-targets.ts` to match.
- Biome targets now come from **`ts/lib/biome.ts`** (extracted `biomeTargets` / `registerBiomeNamedInput`, shared with `configureBiomeLint`) instead of being redefined — net −70 lines, and no snapshot drift, confirming it's behaviour-preserving.
- Harness spec duplication replaced by the existing `project.json` snapshot; I verified the snapshot actually contains the wired targets rather than blessing it blind.
- `connection/project-names` snippet and the `targetComponent` casing paragraph removed everywhere, including the `fr`/`it` translations the sync had already propagated them to (a stale `fr` copy was what broke the docs build — fixed and audited across all 1503 mdx files).
- `packages/nx-plugin/src/connection/` is byte-identical to `main`; the `targetComponent` casing finding is recorded as **not addressed here**.
- Squashed to one commit so the squash-merge body matches what ships.

**One question on your title edit** — you renamed it to "add lint/format/**compile** targets", but the target I've added is named `typecheck` (`tsc --noEmit -p tsconfig.json`), not `compile`. I chose `typecheck` because these projects run through `tsx` and emit nothing, so there's no artifact for a `compile` to produce — whereas in a `ts#project`, `compile` is `tsc --build tsconfig.lib.json` and does emit.

Would you like it renamed to `compile` for consistency with the other generators, or should the title say `typecheck`? Happy to do either — just flagging so the release note and the vended target name agree.

nx-plugin-for-aws:
Nudging once on the title question above, since it's the only thing outstanding and it lands in the release note.

Evidence for `typecheck` being the right name, from this repo's own convention:

- `ts#project` registers **both** names via the `@nx/js/typescript` plugin — `typecheck.targetName: 'typecheck'` and `build.targetName: 'compile'` — so they are distinct roles, not synonyms.
- `ts#project`'s `compile` is `tsc --build tsconfig.lib.json` and declares `outputs: ['{workspaceRoot}/dist/{projectRoot}/tsc']` — it **emits**.
- `py#project`'s `typecheck` is `uv run ty check` with **no** `outputs` — it only checks.

A Gateway/Harness runs through `tsx` and emits nothing, so the target I add is `tsc --noEmit -p tsconfig.json` with no `outputs`. By the above, that is a `typecheck`, and naming it `compile` would be the one `compile` in the workspace that produces no artifact.

So my suggestion is to **leave the target as `typecheck` and adjust the title** to `add lint/format/typecheck targets`. You set the title, so it's yours to change — happy to do it myself if you'd prefer, but I'd rather not push to the branch and dismiss your next approval over a title edit.

If you'd actually rather the target were named `compile` for cross-generator consistency, say so and I'll rename it and re-verify — just note it'd then be an emit-less `compile`.

Everything else is settled: 71/71 green, all eight comments addressed, and the two findings I couldn't close (`conn-web-12`, `gateway-07`) are recorded as still open rather than assumed fixed.

cogwirrel (approved):

PR #1213, opened 2026-09-04
fix(ts#api): preserve user API code when the generator is re-run

nx-plugin-for-aws:
### Reason for this change

Closes the `ts-api` lane of the v1.0 bug bash. Findings addressed: **ts-api-02, 03, 04, 05, 06, 07, 08, 09, 10, 11**. (`ts-api-12` was dropped per review — see below.)

The headline issue (**ts-api-02**, major) is that `ts#api` is not idempotent. Re-running it for an existing API name reports `UPDATE .../src/router.ts` and `UPDATE .../src/procedures/echo.ts` and resets both to the scaffold — the exact two files the tRPC guide instructs the reader to edit ("If you add a new procedure, make sure you register it by adding it to the router in `src/router.ts`"). The reporting agent hit this for real mid-lane and lost a registered nested router, breaking `compile` of a downstream caller. This is the same family as the confirmed `ts#infra` / `terraform#project` blockers.

The rest are guide examples that don't work as written. A reader following them hits:

- **ts-api-03** (major) — the Cognito identity middleware isn't filtered by `infra`, so `http-lambda` readers get `APIGatewayProxyEvent` where the generator emits `APIGatewayProxyEventV2WithJWTAuthorizer` (tRPC `Context mismatch`, TS2345) *and* the wrong claims path (`authorizer.claims` vs `authorizer.jwt.claims`, so it would always throw `FORBIDDEN`). The IAM example directly above it *is* correctly tabbed.
- **ts-api-04** (major) — every identity example opens with `@aws-sdk/client-cognito-identity-provider`, which no generated API depends on and neither guide mentions. TS2307 on the first line.
- **ts-api-05** (minor) — the examples use a value import for `aws-lambda` types, failing biome's `noUndeclaredDependencies`. `lint --configuration=fix` does **not** repair it.
- **ts-api-06** (major) — the Smithy walkthrough adds a required `getIdentity` to `ServiceContext` but only patches `handler.ts`, breaking the generated `local-server.ts` — contradicting the guide's own "Manual Context Required" caution.
- **ts-api-07** (major) — `withOverrides` and `api.integrations.$router` are documented adjacently but are mutually exclusive on a scaffolded API.
- **ts-api-08** (docs) — Smithy operations are PascalCase but CDK integration keys are camelCase, undocumented.
- **ts-api-09 / 10 / 11** (docs) — a FileTree listing a `model/package.json` that is never generated, two bare relative links that 404, and a FileTree omitting `src/index.ts` (the entrypoint the guide's own client example imports from).

### Description of changes

**Idempotency (`ts-api-02`)** — `ts#api` now generates its source with `OverwriteStrategy.KeepExisting`, so a re-run leaves every existing file alone. That covers the two files the reported loss was in (`router.ts`, `procedures/echo.ts`) and everything else the guides instruct edits to. The same change is applied to the Smithy backend.

The placeholder barrel `ts#project` scaffolds is deleted only when the project did not already exist, so the API still vends its own `src/index.ts` on a first run; on a re-run nothing is deleted and `KeepExisting` preserves what is there. The Smithy generator had the same `tree.delete` unguarded, so it gets the same guard.

Only `ts#api` is touched here — the other generators in this family are a separate PR (#1211).

**Guides** — root cause was on the docs side for all of these; the generated output was right:

- `ts-api-03`: the Cognito example is now tabbed REST/HTTP like the IAM one, with `APIGatewayProxyEventV2WithJWTAuthorizer` and `authorizer.jwt.claims` on the HTTP tab. Username reads `claims.username ?? claims['cognito:username']`, since an ID token carries the latter.
- `ts-api-04`: `@aws-sdk/client-cognito-identity-provider` is documented as a prerequisite via `<InstallCommand>`, with the version rendered from `TS_VERSIONS` at build time so the documented pin can't go stale — the pattern `docker-bundling.mdx` uses for `CONTAINER_VERSIONS.trivy`. Neither generator declares the package: it only appears because a reader followed an optional walkthrough, so it's the user's dependency rather than something `ts#api` vends and owns. `<InstallCommand>` gained a `projectDir` passthrough (`buildInstallCommand` already accepted it) so the Smithy command targets `packages/<api>/backend` rather than npm/bun's wrong default.
- `ts-api-05`: `import type` throughout.
- `ts-api-06`: the walkthrough now includes the `local-server.ts` edit with a stub identity for local dev.
- `ts-api-07`: one sentence noting `$router` is unavailable once every operation is overridden. The types are correct — this is deliberate, since an all-overridden API routes nothing to the router — so the fix is documentation, not a type change.
- `ts-api-08`: a note in the Smithy guide's Integrations section with an example on each of the CDK and Terraform tabs (both metadata generators key off the same `dotNotationName`).
- `ts-api-09` / `ts-api-11`: FileTrees corrected against `find` output on real generated projects.
- `ts-api-10`: `<Link path="/guides/...">` as the rest of the docs use.

**Migration** — none added. The generated output is unchanged (the snapshots are untouched); the only behaviour change is what a *re-run* overwrites, which no existing workspace's on-disk state depends on. The `TS_VERSIONS` addition is read only by the guides to render the install command. Everything else is docs.

**`ts-api-12` intentionally not addressed** — the `auth`-has-no-`none` caution was removed per review, so that finding is left open.

**Unit tests** — `pnpm nx run @aws/nx-plugin:test`: **4324 passed / 245 files, no failures and no snapshot changes**. New tests, since per `CONTRIBUTING.md` the existing idempotency tests pass only because they compare generator output to generator output:

- `trpc` and `smithy`: `should preserve user-authored implementation files when re-run` — writes user content into every generated source file between the two runs (including a registered nested router and a new procedure, reproducing the reported loss) and asserts it survives byte-for-byte.
- `trpc`: `should vend its own entrypoint over the ts#project placeholder` — asserts the API's own `src/index.ts` is what lands.

**End-to-end**, in a scratch workspace on the locally compiled plugin, generating the permutations each example targets (`--framework=trpc|smithy`, `--infra=rest-lambda|http-lambda`, `--auth=iam|cognito`) plus a `shared`/`custom` API and `ts#infra`:

- **Idempotency reproduced and fixed**: wrote a nested `users` router, a new procedure and `// MY CUSTOM CODE` into a generated API, re-ran the generator, confirmed all survived. Re-ran the Smithy generator over the completed identity walkthrough and confirmed all eight user files unchanged by `md5sum -c`.
- **Every corrected example pasted in verbatim**, then `build` (which runs `compile`, `lint` and `format`) green on all six projects — so `cog-http` now compiles where `ts-api-03` blocked it, and lint passes where `ts-api-05` failed.
- **`local-server.ts` verified specifically**: reverting just the new `getIdentity` line reproduces the exact `TS2345` from the finding, confirming the edit is what closes `ts-api-06`. With it, `tsx src/local-server.ts` serves and `GET /echo?message=hello` returns `{"message":"local says hello"}`.
- **`ts-api-07`/`08` confirmed against the real types**: `integrations.echo` and `withOperationOptions({ echo })` compile for Smithy's `operation Echo` (and the generated `metadata.gen.ts` shows `echo`); `$router` resolves with no overrides, and a `@ts-expect-error` on the all-overridden case passes — proving `$router` really is dropped, as the snippet now says. `synth` and `checkov` both green.

Stopped at `synth` — nothing deployed to AWS.

Also: `pnpm nx run-many --target lint --all --fix` clean, `pnpm nx run-many --target build --all` green, and `pnpm nx run docs:test` 124 passed (covers link integrity, so the `ts-api-10` fix is checked). The docs site build is left to CI.

The end-to-end run above exercised the original template-split implementation; review simplified that to a single blanket `KeepExisting`, which the unit tests re-verify — both `preserve user-authored implementation files` tests still pass unchanged.

### Issue # (if applicable)

N/A — bug bash findings ts-api-02 … ts-api-12.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/trpc.mdx
+++ b/docs/src/content/docs/en/guides/trpc.mdx
@@ -16,6 +16,8 @@ import NxCommands from '@components/nx-commands.astro';
 import Infrastructure from '@components/infrastructure.astro';
 import Snippet from '@components/snippet.astro';
 import OptionFilter from '@components/option-filter.astro';
+import InstallCommand from '@components/install-command.astro';
+import { TS_VERSIONS } from '../../../../../../packages/nx-plugin/src/utils/versions';
 
 [tRPC](https://trpc.io/) is a framework for building APIs in TypeScript with end-to-end type safety. Using tRPC, updates to API operation inputs and outputs are immediately reflected in client code and are visible in your IDE without the need to rebuild your project.
 
@@ -49,15 +51,18 @@ The generator will create the following project structure in the `<directory>/<a
 
 <FileTree>
   - src
+    - index.ts Package entrypoint re-exporting the router, context, client and schema
     - init.ts Backend tRPC initialisation
     - handler.ts Lambda handler entrypoint
     - router.ts tRPC router definition
     - schema Schema definitions using Zod
+      - index.ts Barrel re-exporting every schema
       - echo.ts Example definitions for the input and output of the "echo" procedure
       - z-async-iterable.ts Zod helper for subscriptions (REST API only)
     - procedures Procedures (or operations) exposed by your API
       - echo.ts Example procedure
     - middleware
+      - index.ts Barrel re-exporting the middleware, and the procedure context type
       - error.ts Middleware for error handling
       - logger.ts middleware for configuring AWS Powertools for Lambda logging
       - tracer.ts middleware for configuring AWS Powertools for Lambda tracing
@@ -65,9 +70,15 @@ The generator will create the following project structure in the `<directory>/<a
     - local-server.ts tRPC standalone adapter entrypoint for local development server
     - client
       - index.ts Type-safe client for machine-to-machine API calls
+  - rolldown.config.ts Bundle configuration for the Lambda deployment package
   - tsconfig.json TypeScript configuration
+  - tsconfig.lib.json TypeScript configuration for the library sources
+  - tsconfig.spec.json TypeScript configuration for the tests
+  - vitest.config.mts Vitest configuration
   - package.json Project manifest defining the project's package name and dependencies
   - project.json Project configuration and build targets
+  - README.md Project readme
+  - .gitignore Ignores the project's build output
 
 </FileTree>
 
@@ -297,6 +308,10 @@ As an example, let's implement some middlware to extract some details about the 
 <OptionFilter when={{ auth: 'iam' }} description="Identity middleware example for IAM-authenticated APIs">
 This example walks through identity middleware for `IAM` authentication. We look up the caller in Cognito using the sub extracted from the API Gateway event.
 
+The lookup uses the Cognito Identity Provider client, which is not a dependency of a generated tRPC API. Install it into your API project first:
+
+<InstallCommand pkg={`@aws-sdk/client-cognito-identity-provider@${TS_VERSIONS['@aws-sdk/client-cognito-identity-provider']}`} project="@my-scope/my-api" />
+
 First, we define what we'll add to the context:
 
 ```ts
@@ -334,8 +349,8 @@ In our case, we want to extract details about the calling Cognito user. We'll do
 ```ts
 import { CognitoIdentityProvider } from '@aws-sdk/client-cognito-identity-provider';
 import { initTRPC, TRPCError } from '@trpc/server';
-import { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
-import { APIGatewayProxyEvent } from 'aws-lambda';
+import type { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
+import type { APIGatewayProxyEvent } from 'aws-lambda';
 
 export interface IIdentityContext {
   identity?: {
@@ -345,12 +360,17 @@ export interface IIdentityContext {
 }
 
 export const createIdentityPlugin = () => {
-  const t = initTRPC.context<IIdentityContext & CreateAWSLambdaContextOptions<APIGatewayProxyEvent>>().create();
+  const t = initTRPC
+    .context<
+      IIdentityContext & CreateAWSLambdaContextOptions<APIGatewayProxyEvent>
+    >()
+    .create();
 
   const cognito = new CognitoIdentityProvider();
 
   return t.procedure.use(async (opts) => {
-    const cognitoAuthenticationProvider = opts.ctx.event.requestContext?.identity?.cognitoAuthenticationProvider;
+    const cognitoAuthenticationProvider =
+      opts.ctx.event.requestContext?.identity?.cognitoAuthenticationProvider;
 
     let sub: string | undefined = undefined;
     if (cognitoAuthenticationProvider) {
@@ -397,8 +417,8 @@ export const createIdentityPlugin = () => {
 ```ts
 import { CognitoIdentityProvider } from '@aws-sdk/client-cognito-identity-provider';
 import { initTRPC, TRPCError } from '@trpc/server';
-import { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
-import { APIGatewayProxyEventV2WithIAMAuthorizer } from 'aws-lambda';
+import type { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
+import type { APIGatewayProxyEventV2WithIAMAuthorizer } from 'aws-lambda';
 
 export interface IIdentityContext {
   identity?: {
@@ -408,7 +428,12 @@ export interface IIdentityContext {
 }
 
 export const createIdentityPlugin = () => {
-  const t = initTRPC.context<IIdentityContext & CreateAWSLambdaContextOptions<APIGatewayProxyEventV2WithIAMAuthorizer>>().create();
+  const t = initTRPC
+    .context<
+      IIdentityContext &
+        CreateAWSLambdaContextOptions<APIGatewayProxyEventV2WithIAMAuthorizer>
+    >()
+    .create();
 
   const cognito = new CognitoIdentityProvider();
 
@@ -481,12 +506,16 @@ export interface IIdentityContext {
 
 Note that we define an additional _optional_ property on the context. tRPC manages ensuring that this is defined in procedures which have correctly configured this middleware.
 
-Next, the middleware itself:
+Next, the middleware itself. The event type and the location of the claims differ between a REST API and an HTTP API, so the implementation depends on your selected `infra`:
+
+<Tabs syncKey="http-rest">
+<TabItem label="REST API" _filter={{ infra: 'rest-lambda' }}>
+A REST API's Cognito User Pools authorizer places the claims at `event.requestContext.authorizer.claims`:
 
 ```ts
 import { initTRPC, TRPCError } from '@trpc/server';
-import { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
-import { APIGatewayProxyEvent } from 'aws-lambda';
+import type { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
+import type { APIGatewayProxyEvent } from 'aws-lambda';
 
 export interface IIdentityContext {
   identity?: {
@@ -497,7 +526,9 @@ export interface IIdentityContext {
 
 export const createIdentityPlugin = () => {
   const t = initTRPC
-    .context<IIdentityContext & CreateAWSLambdaContextOptions<APIGatewayProxyEvent>>()
+    .context<
+      IIdentityContext & CreateAWSLambdaContextOptions<APIGatewayProxyEvent>
+    >()
     .create();
 
   return t.procedure.use(async (opts) => {
@@ -506,7 +537,7 @@ export const createIdentityPlugin = () => {
       | undefined;
 
     const sub = claims?.sub;
-    const username = claims?.username;
+    const username = claims?.username ?? claims?.['cognito:username'];
 
     if (!sub || !username) {
       throw new TRPCError({
@@ -527,6 +558,59 @@ export const createIdentityPlugin = () => {
   });
 };
 ```
+</TabItem>
+<TabItem label="HTTP API" _filter={{ infra: 'http-lambda' }}>
+An HTTP API's JWT authorizer delivers a payload-v2 event whose claims sit one level deeper, at `event.requestContext.authorizer.jwt.claims`. The context must be typed on `APIGatewayProxyEventV2WithJWTAuthorizer` to match the one the generated `publicProcedure` uses — otherwise `.concat()` fails with tRPC's `Context mismatch` error:
+
+```ts
+import { initTRPC, TRPCError } from '@trpc/server';
+import type { CreateAWSLambdaContextOptions } from '@trpc/server/adapters/aws-lambda';
+import type { APIGatewayProxyEventV2WithJWTAuthorizer } from 'aws-lambda';
+
+export interface IIdentityContext {
+  identity?: {
+    sub: string;
+    username: string;
+  };
+}
+
+export const createIdentityPlugin = () => {
+  const t = initTRPC
+    .context<
+      IIdentityContext &
+        CreateAWSLambdaContextOptions<APIGatewayProxyEventV2WithJWTAuthorizer>
+    >()
+    .create();
+
+  return t.procedure.use(async (opts) => {
+    const claims = opts.ctx.event.requestContext?.authorizer?.jwt?.claims as
+      | Record<string, string>
+      | undefined;
+
+    const sub = claims?.sub;
+    const username = claims?.username ?? claims?.['cognito:username'];
+
+    if (!sub || !username) {
+      throw new TRPCError({
+        code: 'FORBIDDEN',
+        message: 'Unable to determine calling user',
+      });
+    }
+
+    return await opts.next({
+      ctx: {
+        ...opts.ctx,
+        identity: {
+          sub,
+          username,
+        },
+      },
+    });
+  });
+};
+```
+</TabItem>
+</Tabs>
 
 You can then mix the plugin into any procedure that needs the caller's identity:
 
--- a/docs/src/content/docs/en/guides/ts-smithy-api.mdx
+++ b/docs/src/content/docs/en/guides/ts-smithy-api.mdx
@@ -18,6 +18,8 @@ import PackageManagerShortCommand from '@components/package-manager-short-comman
 import Infrastructure from '@components/infrastructure.astro';
 import Snippet from '@components/snippet.astro';
 import OptionFilter from '@components/option-filter.astro';
+import InstallCommand from '@components/install-command.astro';
+import { TS_VERSIONS } from '../../../../../../packages/nx-plugin/src/utils/versions';
 
 [Smithy](https://smithy.io/) is a protocol-agnostic interface definition language for authoring APIs in a model driven fashion.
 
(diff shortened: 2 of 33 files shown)


nx-plugin-for-aws:
Heads-up on overlap with #1216 (cc same lane): both PRs append new tests to `packages/nx-plugin/src/trpc/backend/generator.spec.ts` immediately after `should be idempotent when re-run with same options`, so whichever merges second will hit a textual conflict there. Both sides are pure additions, so the resolution is to keep both blocks — no semantic overlap.

No template conflict: this PR only moves files under `trpc/backend/files/**` (into `files/framework/` and `files/user/`), while #1216 changes the `utils/api-constructs/files/cdk/app/apis/**` construct templates. #1216 does update the `trpc`/`smithy` snapshots; this PR changes no generated output, so its snapshots are untouched and the two sets of snapshot changes don't collide.

Happy to rebase on whichever lands first.

docs/src/content/docs/en/guides/trpc.mdx
```diff
@@ -37,40 +39,59 @@ You can generate a new tRPC API in two ways:
 
 :::tip[API Type]
 Select `rest-lambda` (default) as your `infra` if you would like to use [tRPC Subscriptions](https://trpc.io/docs/server/subscriptions) to stream responses.
+
+Select `none` if you want the tRPC backend and its local `serve` target with no infrastructure as code — a router you call from another project in your workspace, or one you deploy yourself. You can escalate later by re-running the generator with the same `name` and an `infra` option, which adds the infrastructure and leaves your procedures alone.
```
cogwirrel: remove

docs/src/content/docs/en/guides/trpc.mdx
```diff
@@ -37,40 +39,59 @@ You can generate a new tRPC API in two ways:
...
 
 :::tip[Integration Pattern]
 The `integrationPattern` option defaults to `isolated`, which creates one Lambda per tRPC procedure. Select `shared` if you would prefer a single shared Lambda handler for the whole API, with optional per-procedure overrides.
 :::
 
+:::caution[Authentication is mandatory]
+`auth` has no `none` option — every deployed API is protected. `iam` (the default) and `cognito` are handled by API Gateway for you; `custom` puts a Lambda Authorizer you own in front of the API, and is the escape hatch for anything else, including a public API. The generated `src/authorizer.ts` **denies all requests by default**, so implement it to allow whatever traffic you intend before deploying.
+
+`infra: none` generates no API Gateway API and so no authorizer at all, but it also gives you nothing to call over the network.
+:::
```
cogwirrel: remove

docs/src/content/docs/en/guides/trpc.mdx
```diff
@@ -37,40 +39,59 @@ You can generate a new tRPC API in two ways:
...
+  - tsconfig.spec.json TypeScript configuration for the tests
+  - vitest.config.mts Vitest configuration
   - package.json Project manifest defining the project's package name and dependencies
   - project.json Project configuration and build targets
+  - README.md Project readme
+  - .gitignore Ignores the project's build output
 
 </FileTree>
 
+`src/index.ts` is the package's public entrypoint — it is what makes `appRouter`, the `Context` type, the [client](#invoking-your-trpc-api) and your schemas importable from `@my-scope/my-api` by other projects in your workspace.
```
cogwirrel: remove

docs/src/content/docs/en/guides/ts-smithy-api.mdx
```diff
@@ -801,6 +827,32 @@ output "lambda_function_name" {
...
+  integrations: MyApi.defaultIntegrations(this)
+    // `Echo` in the model, `echo` here
+    .withOperationOptions({ echo: { timeout: Duration.seconds(60) } })
+    .build(),
+});
+
+api.integrations.echo.handler.addToRolePolicy(...);
+```
+
+Using `Echo` is a type error rather than a silent failure, and TypeScript suggests the camelCase form in the message.
```
cogwirrel: remove this line

docs/src/content/docs/en/guides/ts-smithy-api.mdx
```diff
@@ -801,6 +827,32 @@ output "lambda_function_name" {
...
+:::
+</Fragment>
+<Fragment slot="terraform">
+:::note[Operation names are camelCase in Terraform]
+The generated module's operation-keyed outputs and its `operations.json` use the **camelCase** form of the Smithy operation name, so a Smithy operation `Echo` is keyed `echo`.
+:::
```
cogwirrel, line 869: add an example

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -66,6 +66,24 @@ const api = new MyApi(this, 'MyApi', {
...
+    .withOverrides({ sayHello: { integration: new HttpIntegration('https://example.com') } })
+    .build(),
+});
+
+// TS2339: Property '$router' does not exist on type ...
+api.integrations.$router.handler.addEnvironment('LOG_LEVEL', 'DEBUG');
+```
+
+Leave at least one operation on the default integration if you need to configure the router — with a single-operation API that means not overriding it at all.
+:::
```
cogwirrel: remove this

packages/nx-plugin/src/trpc/backend/generator.ts
```diff
@@ -274,16 +282,31 @@ export async function tsTrpcApiGenerator(
 
   updateProjectConfiguration(tree, projectConfig.name, projectConfig);
 
+  // The Lambda entrypoint, the context type and the client are derived from the
+  // `infra` and `auth` options, so they converge on the requested shape — this is
+  // what lets an API escalate from `--infra none` to a deployed one.
   generateFiles(
```
cogwirrel: No: all generated files should be keep existing. there is no framework vs user - undo that

packages/nx-plugin/src/trpc/backend/generator.ts
```diff
@@ -145,6 +150,9 @@ export async function tsTrpcApiGenerator(
   );
   const backendRoot = projectConfig.root;
 
+  // Clear the placeholder barrel from ts#project so the API's own is written.
+  deletePlaceholderIndex(tree, joinPathFragments(backendRoot, 'src'));
```
cogwirrel: too complex - we can just delete the placeholder index if the project didn't exist before. if it does already exist and we didn't call tsProjectGenerator, don't delete anything - keepexisting will preserve.

packages/nx-plugin/src/trpc/backend/generator.ts
nx-plugin-for-aws: Undone — reverted to a single `generateFiles` call over `files/` with `OverwriteStrategy.KeepExisting`, and moved the templates back to their original `files/src/` layout so `files/framework` and `files/user` are gone. The template tree is now byte-identical to `main`; the only generator change is `Overwrite` → `KeepExisting`. Same for the Smithy generator.

One consequence worth flagging: I had a `should converge the framework-owned files on the requested options when re-run` test asserting that `middleware/index.ts` and `client/index.ts` pick up a changed `infra`/`auth` on a re-run. That behaviour is gone by design now, so I removed the test rather than leave it asserting something untrue. Net effect is that re-running with different `infra`/`auth` no longer moves the context type or client on an existing project — flagging in case you'd rather that were called out in the guide.

packages/nx-plugin/src/trpc/backend/generator.ts
nx-plugin-for-aws: Simplified as suggested — now just:

```ts
if (!projectExists) {
  // Clear the placeholder barrel from ts#project so the API's own is written.
  tree.delete(joinPathFragments(backendRoot, 'src', 'index.ts'));
}
```

using the `projectExists` flag already computed above. Dropped the `deletePlaceholderIndex` / `TS_PROJECT_PLACEHOLDER_INDEX` helper and its content comparison entirely, so `ts/lib/generator.ts` is back to identical with `main` (it had no other callers). Applied the same guard in the Smithy generator, which had the same unguarded `tree.delete`.

Trimmed the placeholder test to just assert the API's own barrel is vended, since the `ts#project`-first case it used to cover was only meaningful with the content matching.

docs/src/content/docs/en/guides/trpc.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/trpc.mdx
nx-plugin-for-aws: Removed the whole `Authentication is mandatory` caution (this was finding ts-api-12, which reported that the guide never states `auth` has no `none` option — happy to leave that undocumented).

docs/src/content/docs/en/guides/trpc.mdx
nx-plugin-for-aws: Removed. The FileTree above still lists `src/index.ts`, which was the substance of ts-api-11.

docs/src/content/docs/en/guides/ts-smithy-api.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/ts-smithy-api.mdx
nx-plugin-for-aws: Added an example to the Terraform note, mirroring the CDK one above it so both sides show the camelCase key in context:

```hcl
# `Echo` in the model, `echo` here
resource "aws_iam_role_policy" "echo_permissions" {
  name = "echo-additional-permissions"
  role = module.my_api.lambda_execution_role_names["echo"]

  policy = jsonencode({
    Version = "2012-10-17"
    Statement = [
      {
        Effect   = "Allow"
        Action   = ["s3:GetObject"]
        Resource = "arn:aws:s3:::my-bucket/*"
      }
    ]
  })
}
```

docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
nx-plugin-for-aws: Removed the caution block, replaced with a single sentence right after the `$router` example:

> Note that `$router` is no longer available if you override every operation via `withOverrides`, since no operation is left using the default router integration.

Happy to drop that line too if you'd rather the constraint went unmentioned — it was finding ts-api-07, where the two adjacent examples couldn't be combined (TS2339) with nothing in the snippet saying so.

nx-plugin-for-aws:
Updated overlap position with #1216 after the review changes (my earlier note is now out of date on the template point).

My change set is down to five files:
```
packages/nx-plugin/src/smithy/ts/api/generator.{ts,spec.ts}
packages/nx-plugin/src/trpc/backend/generator.{ts,spec.ts}
packages/nx-plugin/src/utils/versions.ts
```
The template split is undone, so `trpc/backend/files/**` and `ts/lib/generator.ts` are now byte-identical to `main` — no template overlap with #1216 at all.

The single shared file is `trpc/backend/generator.spec.ts`, where both PRs append a test after `should be idempotent when re-run with same options`. Both are pure additions, so whoever merges second keeps both blocks. #1216 also updates the `trpc`/`smithy` snapshots; this PR changes no generated output so its snapshots are untouched, and those don't collide.

Happy to rebase on #1216 once it lands.

packages/nx-plugin/src/smithy/ts/api/generator.ts
```diff
@@ -72,6 +72,9 @@ export const DEPENDENCIES = declareDependencies<TsSmithyApiMetadata>()({
     { name: '@aws-lambda-powertools/tracer' },
     { name: '@aws-lambda-powertools/metrics' },
     { name: '@aws-sdk/client-appconfigdata' },
+    // Not installed: the guide's identity resolver adds it, and declaring it
+    // keeps that pin current in a workspace that followed the guide.
+    { name: '@aws-sdk/client-cognito-identity-provider', versionOnly: true },
```
cogwirrel: no! remove

packages/nx-plugin/src/trpc/backend/generator.ts
```diff
@@ -73,6 +73,9 @@ export const DEPENDENCIES = declareDependencies<TsTrpcApiMetadata>()({
     { name: '@trpc/client' },
     { name: 'aws4fetch' },
     { name: '@aws-sdk/credential-providers' },
+    // Not installed: the guide's identity middleware adds it, and declaring it
+    // keeps that pin current in a workspace that followed the guide.
+    { name: '@aws-sdk/client-cognito-identity-provider', versionOnly: true },
```
cogwirrel: no!!!!

nx-plugin-for-aws: Removed, and the same in the Smithy generator — the point lands: the declaration describes what the generator vends and therefore owns, and a package that only shows up because someone followed an optional guide walkthrough is the user's dependency, not `ts#api`'s. Declaring it `versionOnly` had the sync managing a pin for something that may not be in the workspace at all.

`ts-api-04` (the finding behind this — the identity examples need `@aws-sdk/client-cognito-identity-provider` and neither guide said so, so they fail with TS2307 on line 1) stays fixed purely in the docs. Both guides state it as a prerequisite with an `<InstallCommand>`, and the version renders from `TS_VERSIONS` at build time so the documented pin can't go stale:

```mdx
import { TS_VERSIONS } from '../../../../../../packages/nx-plugin/src/utils/versions';

<InstallCommand pkg={`@aws-sdk/client-cognito-identity-provider@${TS_VERSIONS['@aws-sdk/client-cognito-identity-provider']}`} project="@my-scope/my-api" />
```

which is the pattern `docker-bundling.mdx` already uses for `CONTAINER_VERSIONS.trivy`. The `TS_VERSIONS` entry is kept for that rendering only — nothing declares it, and I confirmed the pin-ownership test is one-directional (declarations must be vended, not every pin declared), so an undeclared docs-only pin is fine.

packages/nx-plugin/src/smithy/ts/api/generator.ts
nx-plugin-for-aws: Removed here too — see the [reply on the tRPC one](https://github.com/awslabs/nx-plugin-for-aws/pull/1213#discussion_r3935672744) for the detail. Neither API generator declares `@aws-sdk/client-cognito-identity-provider` now; the guides state it as a prerequisite instead.

cogwirrel (dismissed):

cogwirrel (dismissed):

nx-plugin-for-aws:
Rebased onto `main` (now carrying #1216) and all review feedback applied. Ready for another look — summary of where it landed:

**Generator** (`trpc/backend/generator.ts`, `smithy/ts/api/generator.ts` — 5 files total):
- Single `generateFiles` call with blanket `OverwriteStrategy.KeepExisting`; the framework/user template split is gone and `files/**` is byte-identical to `main`.
- Placeholder barrel deleted only `if (!projectExists)`, using the flag already computed above. The `deletePlaceholderIndex` helper and its content-matching are gone, so `ts/lib/generator.ts` is untouched.
- Both `@aws-sdk/client-cognito-identity-provider` `versionOnly` declarations removed. `ts-api-04` is fixed purely in the docs — an `<InstallCommand>` prerequisite with the version rendered from `TS_VERSIONS` at build time.

**Conflict resolution** — #1216 and this PR both appended to `trpc/backend/generator.spec.ts`. Kept both: #1216's paired collision tests (`extends the aliased core construct rather than itself` / `leaves the import unaliased for a name that does not collide`) and this PR's `should preserve user-authored implementation files when re-run`. They assert different properties; all pass.

Also dropped three files and the translated locales that an earlier squash had inadvertently pulled in — the branch no longer touches #1203's `ts#website` fix or any other lane's work. Change set is now 3 English guide pages, `install-command.astro`, and the 5 plugin files.

`pnpm exec vitest --run`: **4337 passed / 245 files**, no failures, no snapshot changes or obsolete snapshots. Collapsed to one commit whose message describes the final design, so the squash body won't describe an approach that was reverted.

cogwirrel (approved):

PR #1214, opened 2026-09-04
fix(connection): insert imports/exports below header comments

nx-plugin-for-aws:
### Reason for this change

Bug bash findings for the tooling / meta-generator guides. The headline fix is `tooling-02`:

`ts#nx-migration` writes `"nx-migrations": {"migrations": "./migrations.json"}` into the plugin's `package.json`, but `ts#nx-plugin`'s `package` target only copied `generators.json` and `executors.json` into `dist`. The manifest was therefore absent from the published tarball while `package.json` still pointed at it, so `nx migrate <plugin>@latest` — the entire premise of the guide's "Running Migrations" section — found no migrations on a published plugin. The migration *code* was packaged; only the manifest was missing.

Also covers `tooling-01`, `tooling-03`, `tooling-04`, `tooling-05`, `tooling-06`, `tooling-07`, `tooling-08`, `tooling-10`, `tooling-11`, `tooling-12`.

### Description of changes

**Packaging (`tooling-02`)** — the `package` target's manifest assets are now one `nxPluginManifestAssets()` helper that includes `migrations.json`. `ts#nx-migration` also calls `addNxPluginManifestAssets`, so a plugin generated before this change picks the asset up when it gains its first migration. It's anchored to the vended `generators.json` asset, so a `package` target the user has rewritten is left alone.

**Keeping the license header first (`tooling-01`, `tooling-10`)** — fixed at the two inserters. `license#sync`'s behaviour is unchanged; the only edit to that file is exporting `parseFile`.

- `ts#nx-generator` inserted its star export at offset 0 of the plugin's `src/index.ts`, above the license header. `addStarExport` now inserts below a leading comment, reusing the sync generator's own `parseFile` to split hashbang/header/body against the comment syntax configured for the file's extension (`license.source.header.commentSyntax`, falling back to the built-in table) rather than assuming C-style comments.
- `tooling-10` was biome's import sorting hoisting a sorted-first import above the header in the generated `mcp-server/http.ts`. The template shipped its imports unsorted, so `biome check --write` reordered them and carried the header down with the import it was attached to. The template now ships them sorted, so there is nothing to hoist. Verified against the vended `biome.json`: the header stays first.

**Re-run safety (`tooling-03`)** — `ts#nx-plugin` deletes only the `ts#mcp-server` sample files it created *in that run*, so a re-run preserves a customised `server.ts` (the file the guide tells you to extend) and its license header.

**Orphan sample tool (`tooling-06`)** — the generator deleted the old `add.ts`/`sample-guidance.ts`, but the `ts#mcp-server` template now ships `tools/divide.ts`, which `server.ts` never registers. Every generated plugin carried that orphan; it's now removed.

**Docs:**

- `tooling-04` — `nx-generator.mdx` imported `@aws/nx-plugin/sdk/open-api.js`; the `.js` resolves to `open-api.js.js`, failing `tsc` with TS2307. Now extensionless.
- `tooling-05` — generator-testing guidance and the scaffolded `generator.spec.ts` template use `createTreeUsingTsSolutionSetup()`, matching `nx-migration.mdx`.
- `tooling-06`/`tooling-07` — both stale FileTrees corrected.
- `tooling-08` — the skip command rendered as `pnpm LICENSE_DEPENDENCY_CHECK=skip lint`, which pnpm rejects with `Command "LICENSE_DEPENDENCY_CHECK=skip" not found`. Now uses `NxCommands`' `env` prop so the assignment precedes the package manager (3 occurrences).
- `tooling-11` — the publish walkthrough now says to remove `"private": true` and uses the real dist path (`dist/tools/nx-plugin/package`) for the guide's own `--directory=tools` recommendation.
- `tooling-12` — documented where `npmCollector`/`pythonCollector` are imported from, and the exceptions schema including the previously-undocumented `version` and `spdx` fields.

**`tooling-09` — finding not upheld, no change.** I'd originally replaced the interactive "Would you like to sync?" prompt with non-interactive remediation output, having only observed the non-TTY path. Re-checked under a real pseudo-terminal (`script -qec`) and the prompt does appear exactly as the guide had it — nx gates it on `process.stdout.isTTY && !isCI()` alone, and `sync.applyChanges` only *skips* it. The guide was right, so it's unchanged apart from the malformed `packages/<my-project>LICENSE` path, which was the one real defect.

**No migration**, per review feedback on this batch.

`pnpm nx run @aws/nx-plugin:test -u` (4304 passed, 244 files), `pnpm nx run-many --target lint --all --fix` and `pnpm nx build @aws/nx-plugin`, all green. Added tests:

- `addStarExport`: inserts below a block comment, a run of line comments, and a hashbang + comment; plus a case using a configured non-C-style (`##`) syntax so the config-driven path is covered.
- `ts#nx-plugin`: a customised `server.ts` written *between* runs survives — the existing tests passed only because they compared generator output to generator output; `divide.ts` absent; `migrations.json` in package assets.
- `ts#nx-migration`: the asset is added to an existing `package` target without disturbing the others.

Snapshot changes reviewed: the `divide.ts` entry removed, and the `http.ts` import order updated. Nothing else moved.

**End-to-end**, in a scratch workspace on the locally compiled plugin using the guide's recommended combination (`ts#nx-plugin` + `license` + `ts#nx-generator` + `ts#nx-migration`):

- `nx run <plugin>:package` → `migrations.json` present in `dist/tools/nx-plugin/package/`, alongside `generators.json`.
- The star export lands below the header immediately after `ts#nx-generator`; `lint` + `nx sync` run repeatedly settle with one header per file and no further changes.
- Full `ts#nx-generator` round trip: scaffolded a generator, then ran it — loaded as ESM and wrote its output.
- Typechecked the SDK import both ways: extensionless compiles; the `.js` form reproduces TS2307.
- Re-ran `ts#nx-plugin` over a customised `server.ts` — preserved byte-for-byte, a clean no-op.

Added the packaged-manifest assertion to the `nx-plugin` e2e smoke test. Did not deploy to AWS.

### Issue # (if applicable)

N/A — from the v1.0 bug bash.

### Checklist

----


nx-plugin-for-aws:
Rebased onto `main` (picks up #1207).

One note on the migration in this PR, since I see migrations being dropped elsewhere in this batch as something users would have already fixed themselves. I think this one is a genuine exception, and I'd rather flag it than have it slip through unnoticed:

A workspace affected by this bug is stuck at **two** headers, and that is not a state a user *can* have already addressed. The finding is precisely that `lint` and `license#sync` never converged: `biome check --write` hoists the sorted-first import above the header, the sync no longer recognises a header that isn't first in the file, so it prepends another — then reports "All files are up to date". A user who hand-deleted the duplicate would watch it regrow on the next `lint` + `sync`. That's the difference from the fixes where the user's own edit is durable.

Concretely, `nx sync` in **this repo** removed a real duplicated header from `packages/nx-plugin/src/terraform/project/schema.d.ts` — so the doubled state exists in the wild, on this codebase, today.

The migration itself is just `licenseSyncGenerator(tree)` (which now collapses duplicates) plus backfilling the `migrations.json` package asset; it's a no-op on a workspace that isn't affected, and it's idempotent.

That said, this is your call and I'll drop it without further argument if you'd prefer. The substantive fixes stand either way:
- position-independent header detection in `license#sync`
- `addStarExport` inserting below a leading comment
- the `package` target including `migrations.json`
- `ts#nx-plugin` preserving a customised `server.ts` on re-run

docs/src/content/docs/en/guides/license.mdx
```diff
@@ -58,14 +57,17 @@ License headers are out of sync in the following source files:
...
+- Run `nx sync:check` to preview the changes without modifying any files.
+- Set `sync.applyChanges` to `true` in your `nx.json` to sync automatically when running tasks in interactive environments.
 ```
 
-Select `Yes` to sync the changes.
+Run the sync to apply the changes:
+
+<NxCommands commands={['sync']} />
+
+Setting `sync.applyChanges` to `true` in your `nx.json` makes Nx offer to sync for you whenever a task runs in an interactive terminal, instead of failing the task.
```
cogwirrel: check again in interactive terminal - this is the flow we want to document. selecting yes is the action.

packages/nx-plugin/src/utils/ast.ts
```diff
@@ -189,6 +189,37 @@ export const addSingleImport = async (
...
+    }
+    i++;
+  } else {
+    while (lines[i]?.trimStart().startsWith('//')) {
+      i++;
+    }
+  }
+  if (i === 0) {
+    return `${statement}\n${contents}`;
+  }
```
cogwirrel: This should use the default comment syntax and any user's custom comment syntax from config, not just hardcoded to c-style comments

packages/nx-plugin/src/migrations/latest/deduplicate-license-headers/migration.ts
```diff
@@ -0,0 +1,41 @@
...
+ * reconciles them, since it now recognises the header wherever it sits and moves
+ * it back to the top.
+ *
+ * Separately, a plugin's `package` target copied only `generators.json` and
+ * `executors.json` into `dist`, so a plugin whose `package.json` points at
+ * `migrations.json` published a manifest path absent from the tarball and
+ * `nx migrate <plugin>@latest` found no migrations. The missing asset is added to
+ * any project whose `package` target carries the ones we vend.
+ */
+export default async function migration(
```
cogwirrel: remove this

packages/nx-plugin/src/license/sync/generator.ts
```diff
@@ -374,6 +374,39 @@ const parseFile = (
...
+ * inserting a statement at the top pushes it down a line. Recognising the header
+ * wherever it sits — rather than only at the top — lets the sync move it back
+ * instead of prepending a second copy, and collapses copies an earlier sync
+ * added.
+ *
+ * Lines are compared exactly, so only a copy rendered at column 0 as the header
+ * itself is matches. An indented comment carrying the same text — the header
+ * quoted inside code, say — is left alone.
+ */
+const removeRenderedHeaders = (content: string, header: string): string => {
```
cogwirrel: No, undo changes to license sync generator

nx-plugin-for-aws: Reverted — `license/sync/generator.ts` and its spec are back to pristine, so `removeRenderedHeaders` is gone.

Fixed both duplication paths at the inserters instead:

- `ts#nx-generator`: `addStarExport` now inserts below the leading comment, so it never pushes the header off the top.
- `tooling-10` (biome hoisting a sorted-first import above the header in the generated `mcp-server/http.ts`): the root cause was the template shipping its imports unsorted, so `biome check --write` reordered them and carried the header down with the import it was attached to. The template now ships them pre-sorted, so there's nothing to hoist. Verified against the vended `biome.json`: header stays first, one header per file.

packages/nx-plugin/src/utils/ast.ts
nx-plugin-for-aws: Good catch — fixed. The helper now resolves the comment syntax for the file's extension from `license.source.header.commentSyntax`, falling back to the built-in `LANGUAGE_COMMENT_SYNTAX` table the sync itself uses, rather than assuming `/*`…`*/` and `//`. Block and line handling both come from the resolved syntax, and a file whose extension has no known syntax is left alone.

Added a test covering a configured non-C-style syntax (`##` on a `.xyz` file) so the config path is exercised, not just the C-style default.

packages/nx-plugin/src/migrations/latest/deduplicate-license-headers/migration.ts
nx-plugin-for-aws: Removed — the migration directory and its spec are deleted, and the `migrations.json` entry is gone (that file is generated and gitignored, so it regenerates without the entry).

It follows anyway now: with the sync-generator change reverted and both duplication paths fixed at the inserters, there's no doubled-header state left for a migration to collapse.

docs/src/content/docs/en/guides/license.mdx
nx-plugin-for-aws: You're right — I'd only ever seen the non-TTY path, so I mis-read this as the docs being wrong. Re-checked under a real pseudo-terminal (`script -qec`) in a generated workspace and the prompt does appear, verbatim:

```
? Would you like to sync the identified changes to get your workspace up to date?
You can skip this prompt by setting the `sync.applyChanges` option to `true` in your `nx.json`.
Yes, sync the changes and run the tasks
No, run the tasks without syncing the changes
```

nx gates it on `process.stdout.isTTY && !isCI()` alone (`run-command.js`); `sync.applyChanges` isn't required — it only *skips* the prompt. So the original text was correct and I've restored it, including "Select `Yes` to sync the changes" as the action.

I kept two things: a note that the prompt is interactive-only and that CI fails the task with `nx sync` instructions instead, and the fix to the malformed `packages/<my-project>LICENSE` path (missing separator), which was the one real defect here.

docs/src/content/docs/en/guides/license.mdx
```diff
@@ -58,15 +57,18 @@ License headers are out of sync in the following source files:
...
+You can skip this prompt by setting the `sync.applyChanges` option to `true` in your `nx.json`.
 Yes, sync the changes and run the tasks
 No, run the tasks without syncing the changes
 ```
 
 Select `Yes` to sync the changes.
 
+:::note[Non-interactive terminals]
+The prompt only appears in an interactive terminal. In CI — or any non-TTY environment — the task fails instead, and Nx prints instructions to run `nx sync` (or `nx sync:check` to preview). Set `sync.applyChanges` to `true` in your `nx.json` to skip the prompt and always sync automatically.
+:::
```
cogwirrel: remove - the below note covers this

docs/src/content/docs/en/guides/license.mdx
```diff
@@ -79,6 +81,8 @@ The license sync generator performs three main tasks:
 
 When the sync generator is run, it will ensure that all source code files in your workspace (based on your configuration) contain the appropriate license header. The header is written as the first block comment or consecutive series of line comments in the file (besides the shebang/hashbang if present in a file).
 
+Each file ends up with exactly one header. A header the formatter's import sorting has moved further down the file is recognised where it sits and moved back to the top rather than joined by a second copy, so repeated `lint` and `nx sync` runs converge.
```
cogwirrel: remove this

docs/src/content/docs/en/guides/nx-generator.mdx
```diff
@@ -562,7 +565,7 @@ describe('your generator', () => {
 
 Key points for testing generators:
 
-- Use `createTreeWithEmptyWorkspace()` to create a virtual file system
+- Use `createTreeUsingTsSolutionSetup()` from `@aws/nx-plugin/sdk/utils/test` to create a virtual file system. It seeds the workspace markers our preset writes (`tsconfig.base.json`, `pnpm-workspace.yaml`, `biome.json`, a `type: module` root `package.json`), which any generator composing an `@aws/nx-plugin` generator needs. `createTreeWithEmptyWorkspace()` from `@nx/devkit/testing` gives you a bare tree without them
```
cogwirrel: remove: `createTreeWithEmptyWorkspace()` from `@nx/devkit/testing` gives you a bare tree without them

docs/src/content/docs/en/guides/nx-generator.mdx
```diff
@@ -40,13 +40,16 @@ The generator will create the following project files within the given `project`
...
+  - generators.json Created or updated to define your generators
+  - project.json Updated to record the generator against the project
+  - package.json Created if absent, and pointed at `generators.json`
 </FileTree>
 
-:::warning[Project Modification]
-This generator will update the selected `project` to use CommonJS, as Nx Generators only support CommonJS at present ([refer to this GitHub issue for ESM support](https://github.com/nrwl/nx/issues/15682)).
+:::note[Module Format]
+The generated generator matches the module format of the project you add it to. In the default ESM workspace, the scaffolded files use `import.meta.dirname` and explicit `.js` import specifiers; in a CommonJS workspace they use `__dirname` and extensionless specifiers.
 :::
```
cogwirrel: remove this note altogether

docs/src/content/docs/en/guides/ts-nx-plugin.mdx
```diff
@@ -33,24 +33,36 @@ The generator will create the following project structure:
...
+  - tsconfig.lib.json TypeScript configuration for your plugin's source
+  - tsconfig.spec.json TypeScript configuration for your plugin's tests
+  - vitest.config.mts Vitest configuration for your plugin's tests
+  - README.md Documentation for your plugin, served by the `generator-guide` tool
+  - LICENSE Written by the <Link path="/guides/license">`license`</Link> generator's sync, when configured
 </FileTree>
 
+:::note[Module Format]
+The plugin matches the module format of your workspace. In the default ESM workspace, its `package.json` is written as `type: module` and the generated code uses `import.meta.dirname` and explicit `.js` import specifiers; in a CommonJS workspace it uses `__dirname` and extensionless specifiers.
+:::
```
cogwirrel: remove this note altogether

docs/src/content/docs/en/guides/ts-nx-plugin.mdx
```diff
@@ -114,22 +126,24 @@ For more details about writing MCP Servers, refer to the <Link path="/guides/ts-
 
 ### Preparing for Publication
 
-1. **Update Package Information**: Edit your plugin's [`package.json`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json) with proper name, version, description, and repository information.
+1. **Update Package Information**: Edit your plugin's [`package.json`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json) with proper name, version, description, and repository information. Generated projects are `"private": true` so they aren't published by accident — remove that field, since `npm publish` refuses a private package with `EPRIVATE`.
```
cogwirrel: change last sentence to:

Also remove `"private": true`.

e2e/src/smoke-tests/license-sync.spec.ts
```diff
@@ -80,4 +80,57 @@ describe('smoke test - license-sync', () => {
     // Ensure the original content is still present
     expect(pyContent).toContain('greeting = "hello world"');
   });
+
+  it('should converge on one license header per file across lint and sync', async () => {
```
cogwirrel: remove this

packages/nx-plugin/src/utils/ast.ts
```diff
@@ -190,90 +194,87 @@ export const addSingleImport = async (
...
+    while (lines[i]?.trimStart().startsWith(lineComment)) {
+      i++;
+    }
   }
-
-  // Prepend new export to file
-  tree.write(filePath, `export * from '${from}';\n${contents}`);
+  if (i === 0) {
+    return `${statement}\n${contents}`;
+  }
```
cogwirrel: double check and possibly reuse code from license sync generator which is doing the same sort of thing and is well tested

nx-plugin-for-aws: Good call — now reuses it rather than reimplementing. `insertBelowLeadingComments` calls the sync generator's `parseFile` to split hashbang/header/body against the resolved `commentSyntax`, then re-joins with the statement after the header.

To be clear about the other thread: the only change to `license/sync/generator.ts` is making `parseFile` exported — a one-word diff, no behaviour change. The `removeRenderedHeaders` addition is fully reverted.

docs/src/content/docs/en/guides/license.mdx
nx-plugin-for-aws: Removed — the note below covers it.

docs/src/content/docs/en/guides/license.mdx
nx-plugin-for-aws: Removed.

docs/src/content/docs/en/guides/nx-generator.mdx
nx-plugin-for-aws: Removed that clause.

docs/src/content/docs/en/guides/nx-generator.mdx
nx-plugin-for-aws: Removed the note.

docs/src/content/docs/en/guides/ts-nx-plugin.mdx
nx-plugin-for-aws: Removed the note.

docs/src/content/docs/en/guides/ts-nx-plugin.mdx
nx-plugin-for-aws: Done — used your wording verbatim.

e2e/src/smoke-tests/license-sync.spec.ts
nx-plugin-for-aws: Removed — `license-sync.spec.ts` is back to its original state.

packages/nx-plugin/src/utils/ast.ts
nx-plugin-for-aws: One refinement worth flagging from reusing `parseFile`: it returns the first leading comment whatever it is, and some vended `index.ts` files open with a placeholder (`// Export your library code here`). Skipping that unconditionally reordered exports in `ts#dynamodb` and the agent-connection generators — caught by their snapshots.

So the leading comment is now matched against the configured header content (the first line of `header.content.lines`, or of the `filePath` it points at) and only skipped when it really is the license header. Any other comment keeps the previous behaviour, with the export above it. Added a test for that case, and the snapshot churn is back down to just the three files this PR intends to change.

nx-plugin-for-aws:
Closing out one loose end from my own earlier report, since it was noted as unverified.

I'd flagged "16 obsolete `ts#dynamodb` snapshots on `main`" as a possible pre-existing issue. **It isn't — there is nothing wrong with `main`.** The obsolete count was manufactured by my own bug, and I can now show the mechanism.

`snapshotTreeDir` walks a directory and asserts one snapshot per file inside a single test:

```ts
tree.children(dir).forEach((subDir) => snapshotTreeDir(tree, joinPathFragments(dir, subDir)));
```

A throwing `expect` aborts that `forEach`, so every file after the mismatch is never asserted — and vitest reports those untouched keys as obsolete. One mismatch fabricates a pile of "obsolete" entries that don't exist on their own.

Demonstrated on pristine `main` (80ba018), which passes 28/28 with zero obsolete:

- corrupt the **first** snapshot key of one test → `1 failed, 6 obsolete`
- corrupt the first key of **both** affected tests → `2 failed, 8 obsolete`

The count scales with how early the mismatch lands, which is how CI reached 16 — my `parseFile` change altered a file early in the shared `common/` tree, orphaning everything after it in both tests. Worth noting my initial cascade guess was wrong in its specifics too: `my-table/src/index.ts` is the *last* key in that test, so that key alone couldn't have orphaned anything.

`main` is clean, this PR is clean (245 files / 4343 tests, no obsolete), and there is no follow-up needed for anyone else. Sorry for the noise on that one.

docs/src/content/docs/en/guides/license.mdx
```diff
@@ -58,15 +57,17 @@ License headers are out of sync in the following source files:
...
 ? Would you like to sync the identified changes to get your workspace up to date?
 Yes, sync the changes and run the tasks
 No, run the tasks without syncing the changes
 ```
 
 Select `Yes` to sync the changes.
 
+:::note[Non-interactive terminals]
+The prompt only appears in an interactive terminal. In CI — or any non-TTY environment — the task fails instead, and Nx prints instructions to run `nx sync` (or `nx sync:check` to preview). Set `sync.applyChanges` to `true` in your `nx.json` to skip the prompt and always sync automatically.
+:::
```
cogwirrel: remove this whole note

packages/nx-plugin/src/utils/ast.spec.ts
```diff
@@ -180,6 +184,109 @@ describe('ast utils', () => {
...
+      // The placeholder ts#project vends is the file's own comment, so the
+      // export goes above it as it always did.
+      tree.write('index.ts', `// Export your library code here\n`);
+
+      await addStarExport(tree, 'index.ts', './module');
+
+      expect(tree.read('index.ts', 'utf-8')).toBe(
+        `export * from './module';\n// Export your library code here\n`,
+      );
+    });
```
cogwirrel: Remove this behaviour. Whether or not the first comment is a license header, we want to add the import below it.

packages/nx-plugin/src/utils/ast.ts
```diff
@@ -189,6 +194,59 @@ export const addSingleImport = async (
...
+const insertBelowLeadingComments = async (
+  tree: Tree,
+  filePath: string,
+  contents: string,
+  statement: string,
+): Promise<string> => {
+  const headerConfig = (await readLicenseConfig(tree))?.source?.header;
+  if (!headerConfig) {
+    return `${statement}\n${contents}`;
+  }
```
cogwirrel: Remove this - if user hasn't set up the license generator we should still add the import below the first comment/shebang, based on our default comment syntax

packages/nx-plugin/src/utils/ast.ts
```diff
@@ -189,6 +194,59 @@ export const addSingleImport = async (
...
+
+  const { hashbang, header, body } = parseFile(contents, syntax);
+  // The header's first line of content is enough to tell it from an unrelated
+  // comment, and holds whichever format the extension is configured with.
+  const [firstHeaderLine] =
+    'lines' in headerConfig.content
+      ? headerConfig.content.lines
+      : (tree.read(headerConfig.content.filePath, 'utf-8')?.split('\n') ?? []);
+  const isLicenseHeader =
+    !!header && !!firstHeaderLine && header.includes(firstHeaderLine);
```
cogwirrel: Per the above comment, we must add the import after all header comments, license or not, so don't try to guess what's a license.

packages/nx-plugin/src/utils/ast.ts
nx-plugin-for-aws: Removed. `insertBelowLeadingComments` no longer reads the license config at all — it resolves the default comment syntax for the file's extension from `LANGUAGE_COMMENT_SYNTAX` and inserts below the shebang and leading comment block unconditionally.

This is better than what I had: a workspace without the license generator configured previously fell through to prepending *above* the comment, which is exactly the case you're pointing at. It's also synchronous again, since there's nothing async left to await.

packages/nx-plugin/src/utils/ast.ts
nx-plugin-for-aws: Agreed, and done — the `firstHeaderLine` comparison and `isLicenseHeader` flag are both gone. No guessing: the leading comment block is skipped whatever it contains.

I kept `parseFile` for *finding* the block, which is the part worth reusing, and dropped everything that tried to classify it. The function is down to resolving the syntax, one `parseFile` call, and a re-join.

packages/nx-plugin/src/utils/ast.spec.ts
nx-plugin-for-aws: Removed that test, and replaced it with the inverse assertion — a leading `// Export your library code here` now keeps the export *below* it, with no license config in the test at all. That's the behaviour you're describing, and having it asserted stops the old "insert above an unrecognised comment" path creeping back.

docs/src/content/docs/en/guides/license.mdx
nx-plugin-for-aws: Removed the whole note.

While there I also restored `This will result in an error in CI.`, which I'd dropped as part of the same mistaken `tooling-09` edit — it's accurate, since the task does fail in CI. The workflow block now differs from `main` by exactly one character: the missing `/` in `packages/<my-project>LICENSE`.

e2e/src/smoke-tests/license-sync.spec.ts
nx-plugin-for-aws: Confirming this one is done — `e2e/src/smoke-tests/license-sync.spec.ts` is byte-identical to `main` (empty diff). The convergence loop I'd added there is gone.

packages/nx-plugin/src/utils/ast.ts
nx-plugin-for-aws: Correcting my previous reply — I over-applied this and have put part of it back, so it's clear what did and didn't go.

**Removed:** the header *identification*. No `header.content` read, no `firstHeaderLine` comparison, no `isLicenseHeader` flag. The leading comment block is skipped whatever it says.

**Kept:** the config read, but only to resolve **comment syntax**, mirroring the precedence `license/sync/generator.ts` already uses:

```ts
const syntax =
  (configuredSyntax ?? {})[extension] ?? LANGUAGE_COMMENT_SYNTAX[extension];
```

My first pass dropped that too and relied on `LANGUAGE_COMMENT_SYNTAX` alone. That regressed the case `commentSyntax` exists for: an extension the built-in table doesn't cover (say `.xyz`) has no syntax to match, so the comment block isn't found and the export lands *above* it — the very thing this change is meant to prevent. User config first, defaults second, and if neither covers the extension the statement goes at the top (no throw — this isn't the sync path).

Added a test for exactly that: `.xyz` configured via `commentSyntax` only, asserting the export lands below the comment. I verified it's a real guard by removing the config precedence and watching it fail, rather than assuming.

packages/nx-plugin/src/utils/ast.ts
```diff
@@ -217,8 +263,15 @@ export const addStarExport = async (
     return;
   }
 
-  // Prepend new export to file
-  tree.write(filePath, `export * from '${from}';\n${contents}`);
+  tree.write(
+    filePath,
+    await insertBelowLeadingComments(
```
cogwirrel, line 293: We should do this for everything that is prepending code to a file, not just typescript exports.

nx-plugin-for-aws: Done — all four remaining prepending helpers now route through `insertBelowLeadingComments`:

| helper | before | now |
|---|---|---|
| `addDestructuredImport` | handled `#!` only, else prepended above comments | inserts below shebang + comment |
| `addPythonDestructuredImport` | `from X import Y\n${contents}` | same, via `py` syntax (`#`) |
| `addSingleImport` | `import X\n${contents}` | inserts below |
| `addNamedExport` | `export { … }\n${contents}` | inserts below |

Only the *write a new statement* branch changed in each; the merge-into-existing-import fast paths are untouched. Added a test per helper asserting the statement lands below a leading block comment, one for a shebang + comment, and one for Python (`#` comments, which only work because the syntax is resolved per extension). Verified they're real guards by reverting `addSingleImport` and watching its test fail.

**One thing worth flagging, because it changed the design.** Widening this surfaced a reentrancy bug: `license/config.ts`'s `ensureDependencyCheckBlock` calls `addDestructuredImport` **on `aws-nx-plugin.config.mts` itself**. With the syntax lookup calling `readLicenseConfig`, that evaluated the config file mid-edit — while its own imports were half-written — throwing `ReferenceError: DEFAULT_LICENSE_ALLOWLIST is not defined` across 29 tests in `license/generator.spec.ts` and `terraform/project/generator.spec.ts`.

So the configured `commentSyntax` is now read **as text from the tree** rather than by evaluating the module. Same precedence you asked for (config first, then `LANGUAGE_COMMENT_SYNTAX`), no evaluation, and the helper is synchronous again. Anything the text read doesn't recognise falls through to the built-in table, so it degrades to today's behaviour rather than breaking. The `.xyz` test still passes and still fails if I remove the config read, so the path is genuinely covered.

nx-plugin-for-aws: Replaced the text-based config read with a try/catch around the real load — agreed the regex parsing was brittle, and it's gone:

```ts
let syntax: CommentSyntax | undefined;
try {
  syntax = (await readLicenseConfig(tree))?.source?.header?.commentSyntax?.[extension];
} catch {
  // Reading the config evaluates it, which throws on a config that doesn't
  // load — including this file itself while a generator is mid-edit.
}
syntax ??= LANGUAGE_COMMENT_SYNTAX[extension];
```

Config first, defaults second, and an unknown extension inserts at the top rather than throwing.

The try/catch isn't just defensive, it's load-bearing: `license/config.ts`'s `ensureDependencyCheckBlock` calls `addDestructuredImport` **on `aws-nx-plugin.config.mts` itself**, so the load runs against a config whose imports are half-written and throws `ReferenceError: DEFAULT_LICENSE_ALLOWLIST is not defined`. That took out 29 tests when the read was unguarded. Catching it covers both that and a genuinely broken user config, so I noted the mid-edit case in the comment since it's the non-obvious one.

Added a test with a config that fails to evaluate (imports a nonexistent module), asserting the import still lands below the comment. Verified it's a real guard — removing the try/catch makes it fail with `Cannot find module './does-not-exist.js'` rather than silently passing.

cogwirrel (dismissed):

cogwirrel (approved):

PR #1215, opened 2026-09-04
fix(ts#astro-docs): validate internal doc links and stop translations breaking them

nx-plugin-for-aws:
### Reason for this change

Closes **docs-integrity-03**, **docs-integrity-04**, **docs-integrity-01**, **docs-integrity-02** and **quickstart-03**.

`pnpm nx build docs` printed *"All internal links are valid"* while **236 dead internal targets** shipped. The gate v1.0 relies on to catch broken links was effectively a no-op, because nothing validated `<Link path>` — the prop that nearly every internal cross-reference in these docs goes through.

Two consequences a reader hits:

- **docs-integrity-04**: the translate script's prompt told the agent to *"Rewrite link paths that embed the source locale (`/en/foo` becomes `/jp/foo`)"*. But `<Link>` resolves `path` via `getRelativeLocaleUrl(currentLocale, path)`, so it prepends the locale itself — the injected one got prepended again. **545 `path` values across 178 files** in the 8 generated locales rendered as `/nx-plugin-for-aws/jp/jp/guides/...` and 404'd. Worst hit were the most cross-linked pages (`guides/terraform-project`, `guides/typescript-project`, `guides/react-website-auth`), 40-50 pages per locale.
- **docs-integrity-01 / quickstart-03 / docs-integrity-02**: `existing-project.mdx` linked `guides/py-project` (the slug is `python-project`), and the pdk-migration FAQ used `path="/guides/trpc.mdx"` — the `.mdx` is emitted verbatim into the href.

### Description of changes

**The finding's suggested fix does not work, so this takes a different layer.** `docs-integrity-03` proposed `components: [['Link', 'path']]` on `starlightLinksValidator`. I tried it and measured the result: it is both wrong and lossy, because that option assumes the prop holds a **final href**, whereas `path` is a locale-agnostic slug the component still has to resolve:

- It reports every *correct* link as invalid — `path="/guides/trpc"` is compared against the built route `/nx-plugin-for-aws/en/guides/trpc`, so the true positives drown in false ones (the deliberately-broken `/guides/nope-one` and the valid `/guides/trpc` were reported identically).
- It only reads **block-level** JSX (`mdxJsxFlowElement`), so it silently skips inline `<Link>`s. Counting by AST node type, **all 432** `<Link>` usages in the `en` docs are inline (`mdxJsxTextElement`) and **0** are block-level — even a `<Link>` alone on its own line is wrapped in a paragraph by MDX, so it parses as inline. The option would therefore have validated **none** of them, while still reporting the false positives above.

So validation lives in a small remark plugin, `docs/src/plugins/remark-link-validator.ts`, which checks the prop on its own terms — against the pages that exist in the same locale, on both JSX node types, over the whole tree.

Both of those choices are load-bearing, and measured rather than defensive. Counting `<Link>` nodes by AST type and by JSX-ancestor depth in the `en` docs: **all 432 are inline** (`mdxJsxTextElement`, so a visitor handling only `mdxJsxFlowElement` sees none), and **97 of them are nested inside other JSX**, up to three deep — `Aside`, `Drawer > details`, `Drawer > details > Aside`. A walk that recurses only into JSX children rather than the whole tree therefore misses 97 links, not a handful. My first version of this plugin did exactly that and passed a deliberately-broken page, which is what prompted the count. It fails the build on a `path` with no page in that locale, and separately on a `path` carrying a locale segment, with a message naming the fix. Fenced code is skipped, so the `<Link>` example in `guides/astro-docs.mdx` stays documentation.

**Root cause for docs-integrity-04 is the translation pipeline**, fixed in both copies of the script (`scripts/translate.ts` and the `ts#astro-docs` template it is vended from):

- the prompt instruction that caused it is gone, replaced by an explicit statement that a `path` is locale-agnostic and must be copied verbatim;
- `stripLocaleFromLinkPaths` removes a leading locale the agent adds anyway, so the fix does not depend on prompt adherence — the same belt-and-braces shape as the existing `restoreCodeBlocks` / `assertFrontmatterIsParseable` checks.

**Past output repaired mechanically.** A script stripped the injected prefix from all 545 `path` values across the 178 affected files; the diffs touch only the `path` prop. `PR-BRIEF.md` says English docs only *because* translations are generated — this lane is the deliberate exception the task called out, since it fixes the generator of those translations and repairs its prior output. The two `en` link fixes are applied across all 9 locales because the defect was copied into each.

**The validator genuinely fails the build** — checked in both directions, since a passing build can't distinguish a fixed validator from a still-silent one:

- inline `<Link path="guides/does-not-exist">` → `✗ Build failed`, `[remark-link-validator] en/guides/zz-probe.mdx: ... points at 'guides/does-not-exist', which is not a page in the 'en' locale.`
- `<Link path="/jp/guides/trpc">` in a `jp` page → `✗ Build failed`, `... starts with the locale 'jp'. ... renders as '/jp/jp/guides/trpc' and 404s. Drop the locale: path="guides/trpc".`

Both probes were then removed. Notably the *first* implementation of the plugin passed this deliberately-broken page — it used a JSX-subtree walk that never reaches an inline `<Link>` inside a paragraph — which is exactly the failure mode this check exists to catch.

**Dead-target count to zero.** An independent scan of every `<Link path>` across all 9 locales against each locale's page set: **545 locale-prefixed → 0**, and **0** remaining dead targets in every locale (was 236 distinct in the built site). A cheap parse+validate pass over all **1494** docs pages using the same mdast stack the build uses: all clean; re-adding a broken link makes it fail, confirming the pass is discriminating rather than vacuous.

```
pnpm nx run @aws/nx-plugin:test --skip-nx-cache   # 242/242 files, 4274/4274 tests
pnpm nx run-many --target lint --all --fix        # clean
pnpm nx run-many --target build --projects=@aws/nx-plugin   # green
```

No snapshots changed. **Tests added:** 1 asserting the generator vends a translate script that keeps `<Link path>` locale-agnostic.

### Scoped out (recorded as a follow-up, not done here)

While diagnosing this I checked whether one structural gate would cover the other translate-pipeline defects seen in this batch. Measured against the real failure shapes: parsing the agent's output with the mdast stack the docs build already uses **catches** a truncated/duplicated closing tag (#1199's failure mode) and **subsumes** the existing `assertImportsAreUnique` regex with better diagnostics, but does **not** catch an unterminated fence (#1209's), since that is valid MDX — so it complements the existing block-count check rather than replacing it. Not added here: it is out of this lane and changes vended output. Left as a follow-up.

**Migration:** none. One was included initially, but removed at review — the fix that matters is the generator no longer producing locale-prefixed `path` values, and the repaired locale files ship in this PR.

### Issue # (if applicable)

N/A — from the v1.0 bug bash.

### Checklist

----


nx-plugin-for-aws:
Note for the translate-pipeline defect cluster (locale prefixes here; `restoreCodeBlocks` fence miscount in #1209; a truncated duplicate `</CardGrid>` appended to the `it`/`jp` gateway guides in #1199) — I was asked whether the guard pattern this PR adds generalises to the malformed-MDX cases.

**Partly, and the useful part is cheap.** The guards in this script are all textual (`splitOnFences`, regex). Adding one *parse* assertion — the same mdast stack the docs build uses (`remark-parse` + `remark-mdx` + `remark-frontmatter`), which is already a direct dependency — catches #1199 outright. Measured against the actual failure shapes:

| case | parse |
|---|---|
| valid baseline | parses |
| **#1199** duplicate trailing `</CardGrid>` | **throws** — ``Unexpected closing slash `/` in tag, expected an open tag first`` |
| **#1199** truncated duplicate (partial tag) | **throws** — `Unexpected end of file after name, expected …` |
| unclosed `<CardGrid>` | **throws** — ``Expected a closing tag for `<CardGrid>` (4:1-4:11)`` |
| duplicate `import` | **throws** — `Could not parse import/exports with acorn` |
| **#1209** extra unterminated fence | **parses** ✗ |

Two things follow:

1. A parse assertion would have turned #1199 from a failed docs build into a caught-and-retried translation, and it **subsumes `assertImportsAreUnique`** — that guard exists to pre-empt exactly the acorn error the parser reports directly, with a line/column. It is strictly stronger than the regex it would replace.
2. It would **not** have caught #1209. An unterminated fence is valid MDX (it is just a very long code block), so only the existing `restoreCodeBlocks` count check sees it — which is what fired there. So the two guards are complementary, not redundant.

The general point: every guard here validates *text*, so each new failure mode needs a new bespoke check. One structural check — does the output still parse as MDX? — covers the whole class of "the agent emitted something that is not valid MDX" in one place, which is the shape of the cluster.

I have deliberately **not** added this to this PR (out of lane, and it changes vended output so it would need its own migration). Flagging it as the fix for the cluster issue rather than smuggling it in here.

docs/src/content/docs/en/guides/astro-docs.mdx
```diff
@@ -122,6 +122,11 @@ import Link from '@components/link.astro';
 <Link path="guides/getting-started">Read the getting-started guide</Link>
 ```
 
+`path` must be locale-agnostic — the component prepends the reader's locale
+itself, so a `path` that already names one (`/jp/guides/getting-started`)
+resolves to `/jp/jp/guides/getting-started` and 404s. The translation script
+strips a locale the translating agent adds anyway.
```
cogwirrel: remove

packages/nx-plugin/src/migrations/latest/translate-script-locale-agnostic-link-paths/migration.ts
```diff
@@ -0,0 +1,120 @@
...
+const readReplacement = (name: string): string =>
+  readFileSync(
+    join(import.meta.dirname, 'files', `${name}.ts.fixture`),
+    'utf-8',
+  ).trimEnd();
+
+const editedNextStep = (projectName: string): string =>
+  `${projectName}: its scripts/translate.ts has been customised, so it was left as it is. Translations must keep every \`<Link path>\` locale-agnostic — a path naming a locale renders a doubled segment and 404s — see https://awslabs.github.io/nx-plugin-for-aws/guides/astro-docs/ for the version the generator vends.`;
+
+export default async function migration(
```
cogwirrel: no need for a migration for this, remove

nx-plugin-for-aws: Removed — the whole `migrations/latest/translate-script-locale-agnostic-link-paths/` directory (migration, spec, grit patterns, fixtures). `migrations.json` is generated from the source tree rather than checked in, so its entry drops out on the next `generate-migrations`.

The fix itself is unaffected: `remark-link-validator.ts`, the corrected prompt and `stripLocaleFromLinkPaths` in both `scripts/translate.ts` and the `ts#astro-docs` template all stay, as do the 545 repaired `<Link path>` values. Re-ran the scan after deleting: still 0 locale-prefixed and 0 dead `<Link path>` across all 9 locales.

docs/src/content/docs/en/guides/astro-docs.mdx
nx-plugin-for-aws: Removed. Also dropped the eight translated copies of it — the CI translate job had already picked the paragraph up into `es`/`fr`/`it`/`jp`/`ko`/`pt`/`vi`/`zh`, so removing only the English version would have left the locales carrying a paragraph the source no longer has. All 9 copies of `guides/astro-docs.mdx` are back to their previous state.

cogwirrel (dismissed):

cogwirrel (approved):

PR #1216, opened 2026-09-04
fix(ts#api): stop a project name colliding with the core construct it extends

nx-plugin-for-aws:
### Reason for this change

Closes bug bash finding **`ts-api-01`** (blocker, verified).

Naming a project after the core CDK construct its infrastructure extends generated a **self-referencing class that cannot compile**. `ts#api --name=http-api --framework=trpc --infra=http-lambda` emitted `packages/common/constructs/src/app/apis/http-api.ts` containing:

```ts
import { HttpApi } from '../../core/api/http-api.js';

export class HttpApi<
  TIntegrations extends ApiIntegrations<Operations, HttpApiIntegration>,
> extends HttpApi<Operations, TIntegrations> {
```

That is 5 TS errors (`TS2395`, `TS2440`, `TS2506`, `TS2339`), which fails `@ws/common-constructs:compile` and with it `build` for the **whole workspace** — `common-constructs` is a dependency of every API project's build. There is no warning at generate time, and recovering by hand means unpicking the project from `apis/index.ts`, `package.json`, `project.json`, `tsconfig.base.json` and `tsconfig.json`.

### Description of changes

Each app-level CDK construct template aliases its core import to `Core<Symbol>` **only when the chosen name would actually shadow it**. In the ordinary case the generated code keeps the plain import and is byte-identical to today, so the blast radius is limited to the workspaces that were broken.

The collision check is computed in the template from the same class the template extends (`apiNameClassName === 'HttpApi'`, `nameClassName === 'DynamoDBTable'`, …), so it needs no generator changes, no new shared helper and no additions to `templateOptions`.

**Why a check rather than a reserved-name denylist.** The reserved name isn't a property of the generator — it depends on which core construct the chosen *infrastructure* extends:

- `ts#api --name=http-api --infra=http-lambda` → collides (extends `HttpApi`)
- `ts#api --name=rest-api --infra=rest-lambda` → collides (extends `RestApi`)
- `py#api --name=rest-api --framework=fastapi` → **collides** (9 errors), because py#api defaults to REST
- `py#api --name=http-api` → **compiles fine**, same generator, same name, different infra

So `http-api` is fatal in one configuration and perfectly legal in another. A denylist would have to be keyed on (generator × framework × infra) and kept in sync with every new core construct, and would reject names that are fine. Deriving the reserved name from the class the template extends handles that variability by construction.

Templates covered — the same latent shape existed beyond the two API cases, so all of them now check:

| Template | Colliding name |
| --- | --- |
| `api-constructs` http | `HttpApi` |
| `api-constructs` rest | `RestApi` |
| `website-constructs` | `StaticWebsite` / `StaticWebsiteProps` |
| `dynamodb-constructs` | `DynamoDBTable` / `DynamoDBTableProps` |
| `rdb-constructs` | `AuroraDatabase` / `…Engines` / `…Props` |
| `agent-core-constructs` gateway | `AgentCoreGateway` / `…Props` |

`lambda-functions` needed no change — it extends CDK's own `Function`, not a vended core construct. **Terraform is not affected**: the HCL templates interpolate the project name only into strings and paths (`website_name`, S3 keys, `name_prefix`), and the `module` labels are fixed literals (`module "http_api"`), so no user-chosen name can shadow a module reference.

No migration. With conditional aliasing there is nothing to carry: a workspace whose name doesn't collide is unchanged by this PR, and one that does collide never compiled in the first place.

**Diff size:** 9 files, +203/-24, and **no snapshots move** — the earlier revision of this PR aliased unconditionally, which was 23 files / +867/-81 with 12 snapshot files churning. That churn was the blast radius @cogwirrel objected to, and conditional aliasing removes all of it.

Each affected generator has a **pair** of tests, and the second of each pair is what demonstrates the small blast radius:

- **colliding name → aliased**: `expect(construct).toContain('HttpApi as CoreHttpApi')`, `toContain('extends CoreHttpApi<')`, plus `not.toContain('extends HttpApi<')` and no unaliased core import left.
- **ordinary name → untouched**: `expect(construct).toContain('extends HttpApi<')` and `expect(construct).not.toContain('CoreHttpApi')` — i.e. the vended output for a normal project is exactly as before.

Applied to `http-api`/`http-lambda` and `rest-api`/`rest-lambda` in `trpc/backend` and again in `py/fast-api` (covering the infra-dependent variability above), and to the non-API case in `ts/react-website` with a website named `static-website`.

```
pnpm exec vitest run src/trpc/backend/generator.spec.ts src/py/fast-api/generator.spec.ts \
  src/ts/react-website/app/generator.spec.ts src/agentcore-gateway/generator.spec.ts
# 230 passed, 0 snapshots written
```

That no snapshot needed updating — and that `agentcore-gateway/generator.spec.ts` needed no assertion changes — is itself the evidence that ordinary output is unchanged.

End-to-end in a scratch workspace against the locally compiled plugin, both directions in the same workspace:

```
pnpm nx g @aws/nx-plugin:ts#api --name=http-api --framework=trpc --infra=http-lambda --auth=cognito
#   -> import { HttpApi as CoreHttpApi } ... extends CoreHttpApi<Operations, TIntegrations>
pnpm nx g @aws/nx-plugin:ts#api --name=orders   --framework=trpc --infra=http-lambda --auth=cognito
#   -> import { HttpApi } ...              extends HttpApi<Operations, TIntegrations>
pnpm nx sync && pnpm install
pnpm nx run @ws/common-constructs:compile   # was 5 errors on http-api -> now clean
```

Nothing was deployed to AWS.

### Known limitation (pre-existing, not introduced here)

For a website named `static-website` specifically, one *separate* pre-existing error remains after aliasing: `common/constructs/src/index.ts` star-re-exports both `./app/index.js` and `./core/index.js`, so an app class sharing a core class's name is an ambiguous re-export (`TS2308` on `StaticWebsite` and `StaticWebsiteProps`). I confirmed this error is present before this change too, so it's a distinct barrel-shadowing issue in the `common/constructs` export surface rather than a regression, and fixing it means changing how those barrels re-export. Left out of scope here and flagged for follow-up. The API cases are unaffected, since `core/api/*` isn't re-exported through `core/index.ts` — which is why `http-api`/`rest-api` compile cleanly end to end.

### Issue # (if applicable)

N/A — from the v1.0 bug bash (`ts-api-01`).

### Checklist

----

### DIFF
--- a/packages/nx-plugin/src/py/fast-api/generator.spec.ts
+++ b/packages/nx-plugin/src/py/fast-api/generator.spec.ts
@@ -573,4 +573,54 @@ describe('fastapi project generator', () => {
 
     expect(tree.exists('packages/common/constructs')).toBeTruthy();
   });
+
+  // The reserved name follows the infrastructure, not the framework: a REST API
+  // named `rest-api` collides where an HTTP one named `http-api` does.
+  describe.each([
+    { name: 'http-api', infra: 'http-lambda' as const, coreClass: 'HttpApi' },
+    { name: 'rest-api', infra: 'rest-lambda' as const, coreClass: 'RestApi' },
+  ])(
+    'a project named $name, colliding with the core $coreClass',
+    ({ name, infra, coreClass }) => {
+      it('extends the aliased core construct rather than itself', async () => {
+        await pyFastApiProjectGenerator(tree, {
+          name,
+          directory: 'apps',
+          infra,
+          auth: 'iam',
+          iac: 'cdk',
+        });
+
+        const construct =
+          tree.read(
+            `packages/common/constructs/src/app/apis/${name}.ts`,
+            'utf-8',
+          ) ?? '';
+
+        expect(construct).toContain(`${coreClass} as Core${coreClass}`);
+        expect(construct).toContain(`extends Core${coreClass}<`);
+        expect(construct).toContain(`export class ${coreClass}<`);
+        expect(construct).not.toContain(`extends ${coreClass}<`);
+      });
+
+      it('leaves the import unaliased for a name that does not collide', async () => {
+        await pyFastApiProjectGenerator(tree, {
+          name: 'test-api',
+          directory: 'apps',
+          infra,
+          auth: 'iam',
+          iac: 'cdk',
+        });
+
+        const construct =
+          tree.read(
+            'packages/common/constructs/src/app/apis/test-api.ts',
+            'utf-8',
+          ) ?? '';
+
+        expect(construct).toContain(`extends ${coreClass}<`);
+        expect(construct).not.toContain(`Core${coreClass}`);
+      });
+    },
+  );
 });
--- a/packages/nx-plugin/src/trpc/backend/generator.spec.ts
+++ b/packages/nx-plugin/src/trpc/backend/generator.spec.ts
@@ -1271,4 +1271,74 @@ describe('trpc backend generator', () => {
     const projectConfig = JSON.parse(secondProjectJson);
     expect(projectConfig.metadata.ports).toEqual([2022]);
   });
+
+  // A project named after the core construct its infrastructure extends used to
+  // emit `export class HttpApi ... extends HttpApi` with `HttpApi` imported too.
+  describe.each([
+    {
+      name: 'http-api',
+      infra: 'http-lambda' as const,
+      coreClass: 'HttpApi',
+      corePath: '../../core/api/http-api.js',
+    },
+    {
+      name: 'rest-api',
+      infra: 'rest-lambda' as const,
+      coreClass: 'RestApi',
+      corePath: '../../core/api/rest-api.js',
+    },
+  ])(
+    'a project named $name, colliding with the core $coreClass',
+    ({ name, infra, coreClass, corePath }) => {
+      it('extends the aliased core construct rather than itself', async () => {
+        await tsTrpcApiGenerator(tree, {
+          name,
+          directory: 'apps',
+          infra,
+          integrationPattern: 'isolated',
+          auth: 'iam',
+          iac: 'cdk',
+        });
+
+        const construct =
+          tree.read(
+            `packages/common/constructs/src/app/apis/${name}.ts`,
+            'utf-8',
+          ) ?? '';
+
+        // The core class is imported under an alias, so the app class of the
+        // same name neither collides with nor extends itself.
+        expect(construct).toContain(`${coreClass} as Core${coreClass}`);
+        expect(construct).toContain(`from '${corePath}'`);
+        expect(construct).toContain(`extends Core${coreClass}<`);
+        expect(construct).toContain(`export class ${coreClass}<`);
+        expect(construct).not.toMatch(
+          new RegExp(`import \\{[^}]*\\b${coreClass}\\s*\\}`),
+        );
+        expect(construct).not.toContain(`extends ${coreClass}<`);
+      });
+
+      it('leaves the import unaliased for a name that does not collide', async () => {
+        await tsTrpcApiGenerator(tree, {
+          name: 'test-api',
+          directory: 'apps',
+          infra,
+          integrationPattern: 'isolated',
+          auth: 'iam',
+          iac: 'cdk',
+        });
+
+        const construct =
+          tree.read(
+            'packages/common/constructs/src/app/apis/test-api.ts',
+            'utf-8',
+          ) ?? '';
+
+        // Aliasing is applied only where it is needed, so the vended code for an
+        // ordinary name is unchanged.
+        expect(construct).toContain(`extends ${coreClass}<`);
+        expect(construct).not.toContain(`Core${coreClass}`);
+      });
+    },
+  );
 });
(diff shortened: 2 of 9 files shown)


nx-plugin-for-aws:
Rebased onto `main` (picks up #1207) — clean, no conflicts. Also confirmed no overlap with #1213: that PR touches the trpc *backend* templates (`files/user/**`, `files/framework/**`) and the API guides, whereas this one only touches the vended app-level CDK construct templates under `utils/*-constructs/files/cdk/app/**`. The one file we both touch is `trpc/backend/generator.spec.ts`, where my additions are appended at the end.

### On the migration — happy to drop it if you'd rather

I've noticed migrations being turned down elsewhere in this batch on the grounds that users would already have dealt with the change in their own workspaces, so let me put the case for this one briefly and then defer to you.

The reasoning that usually applies — "the workspace still built, so nothing to carry" — inverts here: a user who hit this had a workspace that **did not compile at all** (`common-constructs:compile` fails, taking `build` for every project with it). So they cannot have simply lived with it. They either renamed the project away from the colliding name, or hand-patched the import themselves — and if they hand-patched it, they picked *some* alias, which won't be the `Core<Symbol>` convention this PR standardises on. The migration's job is exactly that reconciliation: converge a hand-aliased or partially-aliased construct onto today's generated shape, which is the "would it be in the state today's generators produce" bar rather than the "does it still build" bar.

It also isn't a no-op for the uncollided majority, which is the part I'd weigh most: aliasing is applied *uniformly*, so every existing `common/constructs` app construct diverges from today's output, colliding name or not. Without the migration, `nx migrate` leaves every workspace's constructs on the old unaliased shape.

That said, it's a cosmetic divergence for anyone who never hit the collision, and I recognise that's a weaker case than #1206's. If you'd prefer it gone, say the word and I'll drop the migration and keep the aliasing sweep plus the tests.

### Why aliasing rather than a reserved-name denylist

Flagging the reasoning explicitly since it's the obvious alternative. A denylist can't express this bug, because **the reserved name varies by framework and infra** — it depends on which core construct the chosen infrastructure extends, not on the generator:

- `ts#api --name=http-api --infra=http-lambda` → collides (extends `HttpApi`)
- `ts#api --name=rest-api --infra=rest-lambda` → collides (extends `RestApi`)
- `py#api --name=rest-api --framework=fastapi` → **collides** (9 errors), because py#api defaults to REST
- `py#api --name=http-api` → **compiles fine**, same generator, same name, different infra

So `http-api` is legal in one generator and fatal in another, and a user has no way to reason about which. A denylist would have to be keyed on (generator × framework × infra) and kept in sync with every new core construct, and it would reject names that are perfectly fine. Aliasing removes the collision structurally for every name, present and future, and needs no maintenance. That's also why I swept the whole class rather than just the two API cases — the same latent shape existed for `static-website`, `aurora-database`, `dynamo-db-table` and `agent-core-gateway`.

Terraform was checked and needs no equivalent change: the HCL templates only interpolate the project name into strings and paths, and the `module` labels are fixed literals.

cogwirrel (commented):
This is way too big a blast radius for a small change. Instead, let's just do a small change to the templates only (no migrations) which imports as Corexxx only when it would otherwise clash.

nx-plugin-for-aws:
@cogwirrel thanks — reworked exactly as asked, templates only.

**Aliasing is now conditional.** Each template emits `import { X as CoreX }` only when the chosen name would actually shadow the core class; otherwise it keeps the plain import and the vended file is byte-identical to today. The check is computed inline in the template from the same value it extends (`apiNameClassName === 'HttpApi'`, `nameClassName === 'DynamoDBTable'`, …), so there's no generator change, no new helper and nothing added to `templateOptions`.

**The migration is gone** (`migrations/latest/alias-core-construct-imports/` and all its files). It follows from conditional aliasing anyway: a non-colliding workspace is unchanged by this PR, and a colliding one never compiled.

**Blast radius, before → after:**

| | files | diff | snapshots |
| --- | --- | --- | --- |
| unconditional (what you reviewed) | 23 | +867/-81 | 12 churning |
| conditional (now) | **9** | **+203/-24** | **0** |

The snapshot count is the part I'd point at: nothing moved, and `agentcore-gateway/generator.spec.ts` needed no assertion changes back, which is the mechanical proof that ordinary output is untouched.

**Tests are now paired per generator** — one for each direction:

```ts
// colliding name -> aliased
expect(construct).toContain('HttpApi as CoreHttpApi');
expect(construct).toContain('extends CoreHttpApi<');
expect(construct).not.toContain('extends HttpApi<');

// ordinary name -> unchanged
expect(construct).toContain('extends HttpApi<');
expect(construct).not.toContain('CoreHttpApi');
```

across `http-api`/`http-lambda` and `rest-api`/`rest-lambda` in both `trpc/backend` and `py/fast-api`, plus the non-API case (`static-website`) in `ts/react-website`. 230 tests pass, 0 snapshots written. Verified end-to-end too: generating `http-api` and `orders` into the same workspace yields the aliased and plain forms respectively, and `common-constructs:compile` goes from 5 errors to clean.

One thing I kept from the analysis because it's why the *check* is right rather than just smaller: the reserved name follows the infra, not the generator — `py#api --name=rest-api` fails (py#api defaults to REST) while `py#api --name=http-api` compiles fine. Deriving it from the class the template extends handles that; a fixed denylist couldn't.

Also noted in the description: for a website named `static-website` there's a *separate* pre-existing `TS2308` from `common/constructs/src/index.ts` star-re-exporting both `./app` and `./core`. It reproduces without this change, so I've left it out of scope rather than widen the diff again — happy to raise it separately if you'd like it fixed.

cogwirrel (approved):

PR #1223, opened 2026-09-05
feat(ts#react-website): move the shared shadcn package onto the cn package

nx-plugin-for-aws:
### Reason for this change

The shadcn registry now serves its components importing `cn` from the [`cn`](https://www.npmjs.com/package/cn) npm package and declares it as a registry dependency, rather than importing a local `@/lib/utils` helper:

```json
{ "name": "skeleton", "dependencies": ["cn"], "type": "registry:ui" }
```

So a component added by `shadcn add` in the shared shadcn package no longer resolves the package's own `#lib/utils`, and every generated website fails with `TS2307: Cannot find module 'cn'`. This is the change #1222 deferred by leaving the refreshed templates out.

### Description of changes

- The vended component templates import `cn` from `cn`, as the registry now serves them.
- `cn` replaces `clsx` and `tailwind-merge` in the shared shadcn package's dependencies — it is a zero-dependency drop-in for `twMerge(clsx(...))` that also re-exports `clsx`, `twMerge` and `ClassValue`.
- `src/lib/utils.ts` becomes `export { cn } from 'cn';` — the ["Add a cn helper"](https://ui.shadcn.com/docs/installation/manual) step of shadcn's own install guide, verbatim ("Components import `cn` directly from the `cn` package", re-exported from `lib/utils.ts` "so your own imports come from one place"). The CLI does not need the file — it resolves the `utils` alias from `components.json` alone, and `shadcn add` succeeds with no `src/lib` at all — but the [monorepo guide](https://ui.shadcn.com/docs/monorepo) keeps it and has cross-workspace consumers import `cn` through `<pkg>/lib/utils`, which is exactly what the vended CopilotKit components do.
- A `shadcn-cn-package` migration carries an existing workspace across: it rewrites the component imports, replaces the generated helper, declares `cn` and drops `clsx`/`tailwind-merge`. It only touches a package whose `components.json` aliases still match the generated shape, only rewrites an import whose sole binding is `cn`, only replaces `src/lib/utils.ts` while it still holds the generated `twMerge(clsx(...))` helper, and only drops a package once nothing under `src` imports it.

- `pnpm nx test nx-plugin -u` — 4477 passing, snapshots updated, 8 new migration tests covering the happy path, the customised-helper and diverged-alias skips, a retained `clsx`, and idempotency
- `pnpm nx run-many --target build --all`
- `pnpm lint`
- The smoke tests build a generated `ux=shadcn` website end to end

### Issue # (if applicable)

N/A

### Checklist

----


packages/nx-plugin/src/migrations/latest/shadcn-cn-package/migration.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  updateJson(tree, packageJsonPath, (json) => {
+    for (const name of DROPPED_PACKAGES) {
+      if (!stillImported.has(name)) {
+        delete json.dependencies?.[name];
+        delete json.devDependencies?.[name];
+      }
+    }
+    return json;
+  });
```
cogwirrel: use the removeDependencies method so that this will eg remove it from catalogs etc

nx-plugin-for-aws: Done in 5bed10b9.

There was no catalog-aware removal to call — only devkit's `removeDependenciesFromPackageJson`, which just deletes the manifest keys — so I've wrapped it in `utils/dependencies.ts` the same way `addDependenciesToPackageJson` is, and the migration now calls that.

The wrapper only drops the catalog entry once no manifest in the workspace still declares the package (scanned via `visitNotIgnoredFiles`), so removing a dep from one project can't strand another project's `catalog:` reference. It covers all three catalog homes, matching `writeCatalogVersions`: `pnpm-workspace.yaml`, `.yarnrc.yml` and bun's `package.json#catalog`.

Six tests on the helper (one per package manager, the still-referenced case, a project-manifest removal, idempotency) plus one on the migration asserting the `clsx`/`tailwind-merge` catalog entries go while `class-variance-authority` stays.

Two existing call sites still use devkit's directly and I left them alone deliberately:
- `v1.0.0-rc.94/0002-shx-fs-commands` — already released, so its behaviour is pinned.
- `ts/react-website/app/generator.ts` — it strips the `react`/`react-dom` that `@nx/react` seeds into the root manifest, and it runs before the website manifest declares them, so catalog pruning there would drop an entry that is about to be referenced.

Happy to fold either in if you'd rather they were consistent.

nx-plugin-for-aws: Switched to `getProjects` in b8549fd4 — no tree walk now, just the root manifest plus each project's, deduped against the path being written.

Worth noting the tradeoff it comes with: a `package.json` belonging to no Nx project is no longer seen, so its `catalog:` reference wouldn't hold an entry back. I've said so on `workspaceManifests` and I explicitly keep `packageJsonPath` in the set, which covers the one case that would otherwise bite — a caller passing a name in only one of `dependencies`/`devDependencies` while the target manifest still declares it in the other.

The two tests that stand in for sibling manifests now register real projects with `addProjectConfiguration`, so they still exercise the lookup rather than passing vacuously.

cogwirrel (approved):

PR #1228, opened 2026-09-06
docs: fix quick start issues and improve graphs

nx-plugin-for-aws:
### Reason for this change

I ran the Quick Start Guide verbatim in two fresh workspaces — one `--iac=cdk`, one `--iac=terraform` — from `pnpm create @aws/nx-workspace` all the way through deploying to AWS, signing in through Cognito, calling the deployed API from a browser, and tearing everything down. Most of the guide is accurate, but five things did not hold.

### Description of changes

**The diagram's "Copy commands" button always emitted the CDK infra generator.** `quick-start.mdx` passed `iac="cdk"` to `<EmbeddedGraph>`, and `iac` is a plain prop with no link to the `syncKey="iac"` tab state the rest of the page uses. A reader on the Terraform tab copied `ts#infra infra` and ended up with a CDK project, so Step 4's Terraform sample could not apply. The diagram now sits inside `<Infrastructure>` tabs, one `<EmbeddedGraph>` per provider. Verified in a browser: the Terraform tab's button now copies `terraform#project infra --type=application --no-interactive`, and the second graph sizes correctly when its tab is revealed.

**Terraform printed neither the CloudFront domain nor the user pool id.** Terraform only prints root module outputs, and the scaffolded `packages/infra/src/outputs.tf` declares just `aws_region` and `environment`. After `pnpm nx apply infra` the only way to find the distribution domain was `terraform state show`. The Step 4 Terraform sample now re-exports `website_url` and `user_pool_id`.

**Step 6 gains an optional bootstrap-teardown section.** Terraform has a `bootstrap-destroy` target; CDK has no equivalent, so it documents deleting the `CDKToolkit` stack per region and then the asset bucket — which the bootstrap template marks `DeletionPolicy: Retain`, and whose object versions `aws s3 rb --force` cannot clear. Every command was run against a throwaway `eu-west-2` bootstrap and verified.

**The embedded graph now shows which IaC provider it scaffolds.** A mark in the bottom right of the diagram, in each provider's own branding: the AWS CDK product icon in Squid Ink (swapped for the white variant in dark mode) and Terraform's in HashiCorp purple. It sits outside the scaled extent so it stays legible however far the diagram shrinks.

**NPM 10 cannot complete Step 1.** Node 22 — the documented minimum — ships NPM 10, and `npm create @aws/nx-workspace -- my-project` aborts inside create-nx-workspace's `npm install` with `npm error Cannot read properties of null (reading 'edgesOut')` (arborist `#loadPeerSet`), leaving a directory with `nx.json` and `package.json` but no `node_modules` and no `packages/`. Reproduced against the published `@aws/create-nx-workspace@1.0.0-rc.98` with a clean `HOME` pointed at npmjs; the same directory installs cleanly under `npm@11`. CI already installs `npm@11.6.1`, so this path is never exercised. The prerequisite now says NPM >= 11.

Finally, the closing tutorial tip is now a **Next Steps** section with the Dungeon Adventure tutorial first, then the workspace guide and the existing-project guide.

Only `docs/.../en/` is edited; the `powers/` and `skills/` changes are the pre-commit hook regenerating them from the prerequisites snippet.

Both workspaces were created with locally built packages published to Verdaccio, and every command was run exactly as the guide renders it for the pnpm tab.

Verified working, unchanged:

- Step 1 — workspace creation, `cd my-project`, and (driven through a real pty) both documented prompts: "Which provider would you like to manage your infrastructure?" and "Which container engine would you like to use?"
- Step 2 — all six generator invocations, including the documented output paths
- Step 3 — `pnpm dev` serving the website on `localhost:4200` with the tRPC dev server on 2022; the drawer's `before` block is byte-identical to the generated `index.tsx`, and the `after` block compiles and renders "Hello from the API!" in a browser. `Spinner`, `useDemoApi`, the `echo` query and `schema/echo.ts` all match.
- Step 4 — the CDK sample compiles, lints, synths and passes checkov; the Terraform sample is `terraform fmt`-clean and `terraform validate`-clean. The "workspace is out of sync" note reproduces exactly and `pnpm nx sync` resolves it.
- Bootstrap/deploy — `nx bootstrap infra` on both, the tip's `nx run infra:bootstrap --args="aws://<account-id>/us-east-1"` forwarding correctly, then `deploy-sandbox infra` (CDK, us-west-2) and `apply infra` (Terraform, us-east-1)
- Step 5 — `load-runtime-config demo-website` on both providers, `serve demo-website` on 4200, and the `admin-create-user` snippet as written. First sign-in prompts for a new password then offers exactly "SMS" and "Authenticator app", matching the guide's wording. Completed TOTP sign-in on both stacks and confirmed the website calls the **deployed** IAM-auth API successfully, both locally and from CloudFront — which also exercises the Terraform sample's `execute-api:Invoke` policy.
- Step 6 — CDK `destroy-sandbox infra`, and Terraform `destroy infra` once deletion protection was off

All AWS resources created during this testing have been deleted.

`pnpm nx run docs:build` passes with all internal links valid.

### Not changed here

Two problems I found are code, not docs, and want their own change:

1. **The vended Terraform modules' boto3 provisioners do not pin a region.** My first `apply` failed after 10 minutes with `User pool us-east-1_xxxxxxxxx does not exist` for a pool that existed. `core/user-identity/add-callback-url` (and the other `uv run --with boto3` provisioners) call `boto3.client(...)` with no region, and botocore ignores `AWS_REGION` — it reads `AWS_DEFAULT_REGION` or `~/.aws/config` — while the Terraform AWS provider honours `AWS_REGION`. Anyone whose `AWS_REGION` differs from their `~/.aws/config` region hits a wrong-region client, and anyone who sets only `AWS_REGION` with no config file gets no region at all. The provisioners should take the provider's region through their `environment` block.
2. **`EmbeddedGraph`'s `packageManager` is also a fixed prop**, defaulting to `pnpm`, so the copy button emits pnpm commands regardless of the package-manager tab the reader selected. Harmless (the commands still work) but inconsistent; wiring both `iac` and `packageManager` to the synced tab state would remove the need for the tabbed diagram in this PR.

One more bug found while verifying the rendered page, not fixed here:

3. **A `<Drawer>` nested inside `<Infrastructure>` tabs renders a dead trigger.** The `<template id="drawer-template">` comes through empty, so `createPortal` gets a null portal and throws before the click handler is attached — the trigger renders and does nothing, silently. The container is already marked `data-drawer-initialized`, so nothing retries it. I hit this putting the CDK teardown commands in a drawer and inlined them instead.

Two smaller observations, no change made:

- The CDK "Bootstrap Regions" tip may be unnecessary: running `pnpm nx bootstrap infra` with the target region set to us-west-2 bootstrapped **both** us-west-2 and us-east-1 in one pass, since `cdk bootstrap` enumerates every environment in the app. Both reported "no changes" in my already-bootstrapped account, so I could not prove it would *create* the us-east-1 stack, and I left the tip alone rather than document unproven behaviour.
- The generated `packages/infra/src/main.tf` emits a provider deprecation warning on every plan: `data.aws_region.current.id` should be `data.aws_region.current.region`.

### Issue # (if applicable)

N/A

### Checklist

----


docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -266,9 +273,22 @@ module "demo_website" {
...
+}
+
+output "user_pool_id" {
+  value = module.user_identity.user_pool_id
+}
 ```
 
 This is all the Terraform we need to write to deploy our full stack application.
+
+:::note[Root Module Outputs]
```
cogwirrel: Remove this note, just the code change above is enough

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -361,7 +381,7 @@ The first time you sign in as this user you are prompted to choose a new passwor
 
 </Drawer>
 
-Your website is also served from CloudFront. Open its distribution domain name (which is printed to your terminal after deployment) to use the version running entirely on AWS.
```
cogwirrel, line 364: Undo this line change

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -376,14 +396,27 @@ When you have finished, delete the resources you deployed to avoid ongoing costs
 </Fragment>
 </Infrastructure>
 
+The Cognito user pool is created with deletion protection enabled, so it survives the commands above. Turn deletion protection off in the [Amazon Cognito console](https://console.aws.amazon.com/cognito/) before removing it.
```
cogwirrel: Leave this unchanged and instead fix the identity terraform module to retain the user pool so terraform destroys cleanly

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -376,14 +396,27 @@ When you have finished, delete the resources you deployed to avoid ongoing costs
...
 Congratulations! 🎉 You have successfully built and deployed a full-stack application using `@aws/nx-plugin`!
 
-:::tip[Next: Full-Stack Tutorial]
-For a more in-depth tutorial for building a full-stack application, check out the <Link path="get_started/tutorials/dungeon-game/overview">Dungeon Adventure Tutorial</Link>.
-:::
+## Next Steps
+
+- <Link path="get_started/tutorials/dungeon-game/overview">Dungeon Adventure Tutorial</Link> — build a full-stack AI dungeon adventure game, for a more in-depth walkthrough than this guide.
+- <Link path="guides/workspace">Workspaces</Link> — how your workspace is laid out, and the configuration the generators share.
+- <Link path="get_started/existing-project">Add to an Existing Project</Link> — adopt the plugin in an Nx workspace or monorepo you already have.
```
cogwirrel: “adopt the plugin in your existing codebase”

docs/src/content/docs/en/snippets/recommended-prerequisites.mdx
```diff
@@ -1,8 +1,9 @@
 ---
 title: Recommended Prerequisites
 ---
-- [PNPM >= 11](https://pnpm.io/installation#using-npm) (you can also use [Yarn >= 4](https://yarnpkg.com/getting-started/install), [Bun >= 1](https://bun.sh/docs/installation), or [NPM >= 10](https://nodejs.org/en/learn/getting-started/an-introduction-to-the-npm-package-manager) if you prefer)
+- [PNPM >= 11](https://pnpm.io/installation#using-npm) (you can also use [Yarn >= 4](https://yarnpkg.com/getting-started/install), [Bun >= 1](https://bun.sh/docs/installation), or [NPM >= 11](https://nodejs.org/en/learn/getting-started/an-introduction-to-the-npm-package-manager) if you prefer)
   - verify by running `pnpm --version`, `yarn --version`, `bun --version` or `npm --version`
+  - Node 22 ships with NPM 10, which cannot resolve the plugin's peer dependencies. Run `npm install -g npm@11` before creating a workspace with NPM.
```
cogwirrel: Remove this bullet point

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Done — note removed, the outputs in the sample stand on their own.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Done — reverted to the original wording.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/snippets/recommended-prerequisites.mdx
nx-plugin-for-aws: Done — bullet removed. The NPM floor in the line above is still bumped to 11, since 10 cannot complete Step 1 at all.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Note reverted to the original wording.

On fixing the module instead: I can't find a way to make Terraform retain the pool *and* exit cleanly, and I'd rather check which trade-off you want than guess.

What I verified:

- `aws_cognito_user_pool` has no retain/skip-destroy argument. Full attribute list from `terraform providers schema -json` (AWS provider 6.63.0): `alias_attributes, arn, auto_verified_attributes, creation_date, custom_domain, deletion_protection, domain, email_verification_message, email_verification_subject, endpoint, estimated_number_of_users, id, last_modified_date, mfa_configuration, name, region, sms_authentication_message, sms_verification_message, tags, tags_all, user_pool_tier, username_attributes`.
- Terraform core has no retain-on-destroy either — `lifecycle` offers only `create_before_destroy`, `prevent_destroy`, `ignore_changes`, `replace_triggered_by`, `precondition`, `postcondition`. `prevent_destroy = true` makes `destroy` fail rather than skip, so it trades one error for another.

So the options I can see:

1. **Mirror CDK exactly** — create the pool through an `aws_cloudformation_stack` with `DeletionPolicy: Retain` (there's precedent in the `metrics` module). `terraform destroy` then deletes the stack, CloudFormation retains the pool, destroy exits 0, and the existing note is true for both providers. But it means moving the pool, client, domain and WAF association into CloudFormation and re-plumbing every output — a big change for this module.
2. **Let Terraform delete it** — add a destroy-time provisioner that flips `deletion_protection` to `INACTIVE` just before the delete. Small, self-contained change, `destroy` exits 0, but the pool is *removed*, so the note would need a Terraform-specific line saying so (roughly what my first revision did).

I'd lean towards 2 unless matching CDK's retain semantics matters to you — it's a far smaller change, and "destroy removes everything" is arguably the friendlier Quick Start outcome. Happy to do either in a follow-up; say which and I'll pick it up.

Worth flagging in the meantime: with the note reverted and the module unchanged, the Terraform path of Step 6 is still inaccurate — `pnpm nx destroy infra` exits non-zero with `InvalidParameterException: The user pool cannot be deleted because deletion protection is activated` and leaves behind everything it hadn't reached yet (reproduced twice against a real deployment).

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -380,10 +396,44 @@ When you have finished, delete the resources you deployed to avoid ongoing costs
...
+aws s3 rb s3://$BUCKET
+```
+
+Each `delete-objects` reports an error rather than doing nothing when there is nothing of that kind left to remove, which is safe to ignore.
+</Fragment>
+<Fragment slot="terraform">
+<NxCommands commands={['bootstrap-destroy infra']} />
+
+:::caution[Your Terraform State Is Deleted Too]
+This empties and deletes the S3 bucket holding your Terraform state, so run it only once nothing you care about is still deployed from that state.
```
cogwirrel: “once you have torn down all deployed infrastructure”

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -380,10 +396,44 @@ When you have finished, delete the resources you deployed to avoid ongoing costs
 The Cognito user pool is created with deletion protection enabled, so it is not removed automatically. Delete it (and the IAM role Cognito uses to send SMS messages) manually if you no longer need them.
 :::
 
+### Remove the Bootstrap Resources
+
+The resources you bootstrapped are shared by every project deploying to that AWS account and region, and cost next to nothing to leave in place. Remove them only if nothing else deploys there.
```
cogwirrel: Too much. Remove above heading and just say “To remove bootstrap resources, follow below:”

docs/src/content/docs/en/get_started/quick-start.mdx
```diff
@@ -380,10 +396,44 @@ When you have finished, delete the resources you deployed to avoid ongoing costs
...
+
+<Infrastructure>
+<Fragment slot="cdk">
+Delete the bootstrap stack in every region you bootstrapped, including `us-east-1`. This removes the bootstrap roles and the container image repository:
+
+```sh
+aws cloudformation delete-stack --stack-name CDKToolkit --region <region>
+```
+
+The asset bucket is retained on purpose, so delete it separately. It is versioned, which `aws s3 rb --force` cannot handle, so remove every version and delete marker first:
```
cogwirrel: Remove “it is versioned” onwards to the end of the fragment

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Done.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Done — heading and paragraph replaced with that one line.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Done — the fragment now ends at "The asset bucket is retained on purpose, so delete it separately."

Flagging one consequence: that leaves no command for the bucket. Worth knowing the obvious one doesn't work — the bootstrap bucket is versioned, so `aws s3 rb --force` fails with `BucketNotEmpty: ... You must delete all versions in the bucket` (I hit this on a real bootstrap bucket). Emptying it from the S3 console does handle versions.

Happy to add a short pointer to the console if you'd like something actionable there, or leave it as is.

docs/src/content/docs/en/get_started/quick-start.mdx
nx-plugin-for-aws: Followed up on whether Terraform has a built-in skip-destroy, since that would be the clean fix here. It doesn't — verified against Terraform 1.16.1 and AWS provider 6.63.0:

- `lifecycle { skip_destroy = true }` → `An argument named "skip_destroy" is not expected here`. Not a core meta-argument.
- `lifecycle { prevent_destroy = true }` → errors the destroy rather than skipping the resource, so it swaps one failure for another.
- `removed { from = ... lifecycle { destroy = false } }` → *is* valid, but it only applies to a resource you have deleted from configuration (forget-don't-destroy). It's not a destroy-time retain.
- `terraform destroy -exclude=...` → not a flag in 1.16.1 (`flag provided but not defined: -exclude`); only `-target` exists, and it's per-invocation anyway.

`skip_destroy` does exist as a **provider-level** argument — 19 AWS resources expose it (`aws_cloudwatch_log_group`, `aws_ecs_task_definition`, `aws_db_parameter_group`, …) — but `aws_cognito_user_pool` is not one of them. So the tidy fix is an upstream provider ask, not something we can configure today.

That leaves the two options from above. Still leaning towards the destroy-time provisioner that flips `deletion_protection` to `INACTIVE` (small, self-contained, `destroy` exits 0, pool removed) unless you specifically want to match CDK's retain semantics, which needs the CloudFormation wrapper. Let me know and I'll do it in a follow-up.

cogwirrel (approved):

PR #1232, opened 2026-09-07
docs: correct API generator in shared snippet, pnpm allowBuilds and RuntimeConfig usage

nx-plugin-for-aws:
### Reason for this change

Three unrelated documentation errors found while sanity-checking the docs ahead of v1.0. Each one breaks a reader who follows the page exactly as written.

**1. The shared API integrations snippet sends FastAPI and Smithy readers to the wrong generator.**

`docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx` is included by three guides:

- `guides/trpc.mdx:828` (frontmatter `generator: ts#api`)
- `guides/ts-smithy-api.mdx:873` (frontmatter `generator: ts#api`)
- `guides/fastapi.mdx:635` (frontmatter `generator: py#api`)

...but its Terraform "change the integration pattern" step hardcoded `generator="ts#api"`. A reader on the FastAPI page following those Steps generates a brand new **TypeScript** API instead of re-running the Python one. Both `ts#api` and `py#api` expose `integrationPattern` with the same `["isolated","shared"]` enum, so there is a correct command for each — it just was not being shown.

**2. The `pnpm-workspace.yaml` fix in `existing-project` does not work.**

The note documented `allowBuilds` as a YAML sequence:

```yaml
allowBuilds:
  - nx
```

pnpm 11 reads `allowBuilds` as a **map**. This is confirmed three ways in-repo: `packages/nx-plugin/src/utils/pnpm-workspace.ts:12` types it `Record<string, boolean>`; `packages/nx-plugin/src/utils/init.ts:133-138` writes `Object.fromEntries(deps.map(d => [d, true]))`; and this repo's own `pnpm-workspace.yaml` uses the map form (`nx: true`). With the list form `nx` is not allow-listed, so the reader still hits the `ERR_PNPM_IGNORED_BUILDS` failure that the note exists to prevent.

**3. The PDK migration walkthrough contains a snippet that cannot compile.**

`snippets/pdk-migration/example/02-migrate-website.mdx` assigned to a `config` property:

```ts
RuntimeConfig.ensure(this).config = { ...RuntimeConfig.ensure(this).config, ... };
```

`RuntimeConfig` (`packages/nx-plugin/src/utils/files/common/constructs/src/core/runtime-config.ts.template`) exposes only `ensure`, `of`, `set(namespace, key, value)` and `get(namespace)` over a `private readonly _namespaces` map. There is no `config` property, so this is a type error. The shape was also wrong: the vended `UserIdentity` construct nests exactly these four fields under one key (`utils/identity-constructs/files/cdk/core/user-identity.ts.template:122`), and consumers destructure `const { cognitoProps } = useRuntimeConfig()` (`ts/react-website/cognito-auth/files/app/components/CognitoAuth/index.tsx.template:14`).

### Description of changes

- `snippets/api/type-safe-api-integrations.mdx` — the integration-pattern step no longer names a specific generator. It now reads "Re-run the generator that created your API with the other integration pattern (eg `--integrationPattern=shared`)", which is correct on all three including guides. This also removes the last generator-specific reference from the snippet, bringing it in line with every other shared `api/` snippet (`api-choice-note`, `api-architecture`, `access-logging`, …), none of which name a generator. The now-unused `RunGenerator` import is dropped.
- `get_started/existing-project.mdx` — `- nx` → `nx: true`.
- `snippets/pdk-migration/example/02-migrate-website.mdx` — use `RuntimeConfig.ensure(this).set('connection', 'cognitoProps', { … })`.

English-only, per the contributing guide; translations are regenerated from these sources.

- `pnpm nx run docs:build` — passes. 1540 pages built, and the `starlight-links-validator` reports **"All internal links are valid."**
- Confirmed `RunGenerator`, `Tabs` and `TabItem` have no remaining usage in the snippet after the change, so the removed imports leave nothing dangling.
- `pnpm nx run @aws/nx-plugin:test --skip-nx-cache` — 252 files / 4493 tests pass. Run with the cache skipped deliberately: docs are not declared inputs to that target, so it otherwise reports a cache hit and would not actually re-execute.
- Confirmed no test snapshot embeds this snippet's prose (no match for `Re-run the API generator` anywhere in the tree), so there are no snapshots to update.

### Issue # (if applicable)

N/A

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
+++ b/docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
@@ -1,11 +1,10 @@
 ---
 title: Type-Safe API Integrations
 ---
-import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
+import { Steps } from '@astrojs/starlight/components';
 import Infrastructure from '@components/infrastructure.astro';
 import Drawer from '@components/drawer.astro';
 import OptionFilter from '@components/option-filter.astro';
-import RunGenerator from '@components/run-generator.astro';
 
 The REST/HTTP API CDK constructs are configured to provide a type-safe interface for defining integrations for each of your operations.
 
@@ -424,15 +423,7 @@ Unlike CDK, the integration pattern is baked into the generated module. To chang
 
 <Steps>
   1. Delete the previously generated API module in `packages/common/terraform/src/app/apis`
-  2. Re-run the generator that created your API with the other integration pattern:
-      <Tabs syncKey="apiLanguage">
-        <TabItem label="TypeScript">
-          <RunGenerator generator="ts#api" requiredParameters={{ integrationPattern: 'shared' }} />
-        </TabItem>
-        <TabItem label="Python">
-          <RunGenerator generator="py#api" requiredParameters={{ integrationPattern: 'shared' }} />
-        </TabItem>
-      </Tabs>
+  2. Re-run the generator that created your API with the other integration pattern (eg `--integrationPattern=shared`)
 </Steps>
 
 With the `isolated` pattern, the module reads the operations from a generated file:
--- a/docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
+++ b/docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
@@ -1,7 +1,7 @@
 ---
 title: Type-Safe API Integrations
 ---
-import { Steps } from '@astrojs/starlight/components';
+import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
 import Infrastructure from '@components/infrastructure.astro';
 import Drawer from '@components/drawer.astro';
 import OptionFilter from '@components/option-filter.astro';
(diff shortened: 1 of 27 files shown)


docs/src/content/docs/en/snippets/api/type-safe-api-integrations.mdx
```diff
@@ -424,8 +424,15 @@ Unlike CDK, the integration pattern is baked into the generated module. To chang
...
-      <RunGenerator generator="ts#api" requiredParameters={{ integrationPattern: 'shared' }} />
+  2. Re-run the generator that created your API with the other integration pattern:
+      <Tabs syncKey="apiLanguage">
+        <TabItem label="TypeScript">
+          <RunGenerator generator="ts#api" requiredParameters={{ integrationPattern: 'shared' }} />
+        </TabItem>
+        <TabItem label="Python">
+          <RunGenerator generator="py#api" requiredParameters={{ integrationPattern: 'shared' }} />
+        </TabItem>
+      </Tabs>
```
cogwirrel: No tabs, just change 2 to:

```
2. Re-run the generator that created your API with the other integration pattern (eg `--integrationPattern=shared`)
```

nx-plugin-for-aws: Done in e6fe9ba — step 2 is now exactly that line, no tabs. Also dropped the `RunGenerator` import, which became unused in this snippet as a result.

`docs:build` passes (1540 pages, links validator clean). The translation workflow will re-run and replace the tabbed version it generated for the other 8 locales.

cogwirrel (approved):

PR #1261, opened 2026-09-08
ci(workflows): track the last validated commit in a ref, and release from a docs push

nx-plugin-for-aws:
### Reason for this change

Two problems with the docs-only skip on `main`, both surfaced by real runs.

**1. A stale API response made a docs push run the whole suite.** [Run 34202837561](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/34202837561) (`e757cb8`, docs only) compared against `6c3e900` — seven commits back — and ran all 57 smoke legs. It should have compared against `76161e5`. The base came from `GET /actions/workflows/{id}/runs?branch=main&status=success`, and that endpoint is eventually consistent: called twice minutes apart it returned month-old runs once and current runs the next time. At 08:07:48, when `Detect Changes` ran, it returned neither `d4ef5cd`'s run (green at 07:11:17) nor `76161e5`'s (green at 07:18:15), so the script fell back to a much older green ancestor. The direction is safe — an older base only widens the diff — but it defeats the optimisation exactly when it should apply.

**2. A docs push could swallow a pending release.** Push something releasable, then push docs on top:

- the releasable commit's own run declines to publish, because `latest_commit == github.sha` is false once the branch has moved on;
- the docs push skipped the release job outright, since #1239 gated it on `docs_only`.

So the release waited for the next push touching anything outside `docs/`. Nothing has actually been lost — every commit since `v1.0.0` is `docs`, `ci` or `chore`, so `skip_release` was true throughout — but with batched pushes this sequence is routine, not an edge case.

### Description of changes

**The last validated commit lives in a ref, not in the API.** A new `mark_validated` job pushes `refs/ci/last-validated` at the end of a green push run on `main`, and `scripts/docs-only-change.mts` reads it with `git ls-remote`. No API call, nothing to be stale, and the script loses its token, its workflow-id lookup and its `fetch` — the `changes` job drops `actions: read` again.

- The ref only ever moves forwards: if it points at a commit the current one does not descend from, the job leaves it alone, so a run finishing out of order cannot rewind it onto a commit whose successors are unvalidated.
- A cancelled or failed run leaves the ref where it was, so its commits fold into the next push's diff — the property #1239 added, now on a dependable signal.
- `mark_validated` needs `changes`, `package`, `build`, `unit_tests` and `smoke_tests_complete`, so "validated" means the whole suite passed, or was legitimately skipped for a docs only change.
- Before the ref exists — the first run after this merges — there is no base, so everything runs and the ref is created. `workflow_dispatch` neither reads nor moves it.

**A docs push releases whatever is still unreleased.** The `Should skip release` check moves into the `changes` job, which is early enough for the release job's own `if` to use it: the job is now skipped only when the change is docs only **and** there is nothing to release. The release job stops recomputing the same thing (its duplicate step and unused `skip_release` output are gone) and its publish step reads `needs.changes.outputs.skip_release`. The publish step's `latest_commit == github.sha` guard is untouched.

All in a throwaway fork (`nx-plugin-for-aws/nxpaws-ci-skip-test`) carrying these workflows with `runs-on` and the long running step bodies stubbed, but `changes`, `smoke_tests_complete` and `mark_validated` verbatim, plus a ruleset requiring `Build` and `Smoke Tests Complete`, and tags so `git describe` behaves as it does here.

Pushes to `main`:

| Scenario | Compared against | Result |
| --- | --- | --- |
| First push, no ref yet | nothing | `docs-only=false`, full suite, **ref created** |
| Docs push after a green run | the ref | `docs-only=true`, 6 smoke jobs skipped, ref advanced |
| Releasable (`feat`) push | the ref | `docs-only=false`, 57 legs, `Release` ran |
| **Docs push while that release is still pending** | the ref | `docs-only=true` (smoke skipped) **and `Release` ran** — the fix |
| Docs push with nothing pending | the ref | `docs-only=true`, **`Release` skipped** — the behaviour #1239 asked for, preserved |
| Docs push on top of a **cancelled code** push | the ref, not the cancelled commit | `docs-only=false`, 57 legs, `Release` ran |
| Docs push on top of a **cancelled docs** push | two docs commits back | `docs-only=true`, still skipped — not over-conservative |
| `workflow_dispatch` | nothing | `docs-only=false`, 57 legs, `mark_validated` skipped |

Pull requests, unchanged by this PR and re-run against it:

| Scenario | Result |
| --- | --- |
| Docs only | `docs-only=true`, 6 jobs skipped, gate green, `CLEAN` |
| Docs + code | `docs-only=false`, all 54 legs |
| Code commit → run cancelled → docs commit | `docs-only=false`, all 54 legs |
| One smoke leg failing | `Smoke Tests Complete` **failed**, PR `BLOCKED` |

The ref's forward-only guard was also exercised directly in a scratch repository: first run creates it, an older commit finishing late is refused ("does not descend from"), a descendant advances it.

### Issue # (if applicable)

Follows up #1235 and #1239.

### Checklist

----

### DIFF
--- a/.github/actions/docs-only-change/action.yml
+++ b/.github/actions/docs-only-change/action.yml
@@ -22,6 +22,5 @@ runs:
         # push has the script resolve its own base - see the script.
         BASE_SHA: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || '' }}
         HEAD_SHA: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-        GITHUB_TOKEN: ${{ github.token }}
       # Node strips the types, so this job needs no dependencies installed.
       run: node ./scripts/docs-only-change.mts
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,15 +35,20 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      actions: read
     outputs:
       docs_only: ${{ steps.docs_only.outputs.docs-only }}
+      skip_release: ${{ steps.should_skip_release.outputs.skip_release }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0
       - id: docs_only
         uses: ./.github/actions/docs-only-change
+      # Read here rather than in the release job, which needs the answer to
+      # decide whether to run at all.
+      - name: Should skip release
+        id: should_skip_release
+        run: echo "skip_release=$(git log $(git describe --tags --abbrev=0)..HEAD --oneline | grep -qE 'feat\(?.*\)?:|fix\(?.*\)?:|revert\(?.*\)?:|perf\(?.*\)?:|deps\(?.*\)?:' && echo false || echo true)" >> $GITHUB_OUTPUT
   package:
     name: Package
     runs-on: codebuild-nx-plugin-for-aws-runner-${{ github.run_id }}-${{ github.run_attempt }}-large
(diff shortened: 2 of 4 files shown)


nx-plugin-for-aws:
Closing in favour of a simpler approach: reverting the docs-only logic from `ci.yml` entirely, so pushes to `main` always run the full suite, and keeping the skip only for pull requests where the diff is unambiguously docs only. That removes the last-green tracking, the marker ref and the release catch-up along with it — none of which are needed once the push path stops skipping.

PR #1263, opened 2026-09-08
fix(terraform#project): give the provider plugin cache a single writer

nx-plugin-for-aws:
### Reason for this change

The `terraform-deploy` smoke test fails intermittently in `nx apply infra` with a provider checksum error, most recently on `main` in [run 35295062284](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/35295062284) and earlier in [run 34202837561](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/34202837561):

```
Error: Required plugins are not installed
  - registry.terraform.io/hashicorp/local: the cached package for hashicorp/local 2.9.0
    (in .terraform/providers) does not match any of the checksums recorded in the dependency lock file
```

`plan` depends on both `init` and `validate`, so Nx schedules them concurrently, and `test` can join them. All three run their own `terraform init` against one per-project plugin cache. Two `terraform init` runs filling the cache at once each compute a different hash for the same provider, because the hash covers a directory the other is still writing. HashiCorp documents the plugin cache as not concurrency-safe. The generator's existing comment anticipated this, but isolated projects from each other rather than a project's own targets.

### Description of changes

**`terraform#project`: a single writer for the provider cache.** A new `terraform-init` target runs `terraform init -backend=false` into its own `TF_DATA_DIR` (`dist/<project>/terraform-init`) and is the only target that fills the plugin cache. `init`, `test` and `validate` `dependsOn` it, so by the time they run the cache is warm and they only link out of it. `test` and `validate` no longer run their own `terraform init`; they read the `TF_DATA_DIR` `terraform-init` prepared. The target is deliberately uncached: a cache hit on another machine would report success over an empty cache and hand the race straight back.

Shipped with the `terraform-init-target` migration, which adds the target and the `dependsOn` entries to existing workspaces, and skips and reports a `project.json` whose targets have diverged.

**Deploy smoke tests.** `applyInfra` (from #1272) retries only the CloudWatch Logs delivery-source `ConflictException`; this PR adds a 30s pause before the re-run, since the service asks the caller to try again later and an immediate re-run lands while the delivery configuration is still being updated.

**Docs.** The Terraform guide's plugin-cache paragraph is shortened; the `TF_PLUGIN_CACHE_DIR` override note was dropped because `test` and `validate` no longer set the variable themselves.

**Not in this PR.** An earlier revision also added a `time_sleep` before the AgentCore Harness is created, to prevent the `Role validation failed` IAM eventual-consistency error. That was dropped ([discussion](https://github.com/awslabs/nx-plugin-for-aws/pull/1263#issuecomment-5724678612)): provider v6.64.0 retries that error on Create and #1270 vends that version, so the plugin would be carrying a template change plus a migration for something fixed upstream.

**Unit tests.** `src/migrations`, `src/agentcore-harness` and `src/terraform/project`: 77 files, 868 tests pass after the split. New coverage: the `terraform-init` target shape and `dependsOn` wiring in `terraform/project/generator.spec.ts`, and 10 tests for the `terraform-init-target` migration covering convergence on what the generator writes today, idempotency, and that a diverged `project.json` is skipped and reported. `scripts/generate-migrations.ts` assembles the manifest without the removed migration.

**Real deployment, in a workspace built from an earlier revision of this branch** (`us-west-2`, plugin published to a local registry, Terraform 1.15.7 as CI uses). That revision still included the Harness wait, but the plugin-cache behaviour is unchanged since:

- Cold cache (`.terraform` and `dist` deleted): `nx plan infra` ran `terraform-init` for both projects first, then `init` and `validate` concurrently, then `plan`, with no checksum error. `run-many --target build --all` and `nx destroy` likewise.
- Migration transition against live resources, as [CONTRIBUTING requires for IaC migrations](https://github.com/awslabs/nx-plugin-for-aws/blob/main/CONTRIBUTING.md#migrations-that-change-infrastructure): reverted the workspace to the pre-fix shape, applied it, ran `nx migrate --run-migrations`, then diffed. Both `project.json` files came out byte-identical to freshly generated ones. `lint` passes on the migrated files.
- Re-ran the generator: no changes to the files this PR touches.
- The full `terraform-deploy` smoke test, run locally against AWS, hit the delivery-source `ConflictException` on the first apply; the retry converged the remaining resources and the test went on to exercise the deployed stack.

### Issue # (if applicable)

N/A. Merge before #1270 so its `main` run does not hit the race again, and so `update-versions` can regenerate #1270 against the new `main`.

### Checklist

----

### DIFF
--- a/docs/src/content/docs/en/guides/terraform-project.mdx
+++ b/docs/src/content/docs/en/guides/terraform-project.mdx
@@ -357,9 +357,15 @@ run "plan_is_valid" {
 
 Set every variable your configuration requires in the `variables` block, otherwise the run fails with "has a required variable ... with no set value".
 
-Every target that runs `terraform init` reuses a provider cache under `.terraform/plugin-cache` in your workspace root, so providers are downloaded once rather than on every run. Each project gets its own directory there: two `terraform init` runs filling one cache at the same time can each compute a different hash for the same provider, which terraform then rejects against your `.terraform.lock.hcl`. See the [Terraform documentation](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) for more information.
+Every target that runs `terraform init` reuses a provider cache under `.terraform/plugin-cache` in your workspace root, so providers are downloaded once rather than on every run. Each project gets its own directory there. See the [Terraform documentation](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) for more information.
 
-Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. Note that the `test` target reads its path from `project.json`, so change it there too.
+The `install-providers` target is what fills that cache, and every target that runs `terraform init` depends on it:
+
+<NxCommands commands={['install-providers tf-infra']} />
+
+You will not normally run it yourself. It exists because two `terraform init` runs filling one cache at the same time can each compute a different hash for the same provider, which terraform then rejects against your `.terraform.lock.hcl` with "the cached package ... does not match any of the checksums recorded in the dependency lock file". Downloading the providers once, ahead of the targets that need them, leaves those targets only reading from a cache that is already warm.
+
+Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. Note that the `install-providers`, `test` and `validate` targets read their path from `project.json`, so change it there too.
 
 <OptionFilter when={{ type: 'application' }} description="Application-only targets: plan / apply / destroy / etc.">
 ### Application-Only Targets
--- a/docs/src/content/docs/en/guides/terraform-project.mdx
+++ b/docs/src/content/docs/en/guides/terraform-project.mdx
@@ -359,14 +359,6 @@ Set every variable your configuration requires in the `variables` block, otherwi
 
 Every target that runs `terraform init` reuses a provider cache under `.terraform/plugin-cache` in your workspace root, so providers are downloaded once rather than on every run. Each project gets its own directory there. See the [Terraform documentation](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) for more information.
 
-The `install-providers` target is what fills that cache, and every target that runs `terraform init` depends on it:
-
-<NxCommands commands={['install-providers tf-infra']} />
-
-You will not normally run it yourself. It exists because two `terraform init` runs filling one cache at the same time can each compute a different hash for the same provider, which terraform then rejects against your `.terraform.lock.hcl` with "the cached package ... does not match any of the checksums recorded in the dependency lock file". Downloading the providers once, ahead of the targets that need them, leaves those targets only reading from a cache that is already warm.
-
-Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. Note that the `install-providers`, `test` and `validate` targets read their path from `project.json`, so change it there too.
-
 <OptionFilter when={{ type: 'application' }} description="Application-only targets: plan / apply / destroy / etc.">
 ### Application-Only Targets
 
--- a/packages/nx-plugin/src/agentcore-harness/generator.spec.ts
+++ b/packages/nx-plugin/src/agentcore-harness/generator.spec.ts
@@ -545,6 +545,24 @@ describe('agentcore-harness generator', () => {
       expect(tree.exists(CDK_HARNESSES_INDEX_PATH)).toBe(false);
     });
 
+    it('waits for the execution role to propagate before creating the Harness', () => {
+      // CreateHarness validates the execution role by assuming it, and IAM is
+      // eventually consistent: depending on the policy alone, the validation can
+      // run first and the Harness lands in CREATE_FAILED with "Role validation
+      // failed".
+      expect(tf).toMatch(
+        /resource "time_sleep" "execution_role_propagation" \{\n {2}count = var\.create_execution_role \? 1 : 0/,
+      );
+      expect(tf).toContain(
+        'depends_on = [aws_iam_role.execution_role, aws_iam_role_policy.execution_role]',
+      );
+      expect(tf).toContain(
+        'depends_on = [time_sleep.execution_role_propagation]',
+      );
+      // The wait's provider is pinned like every other the module uses.
+      expect(tf).toContain('source  = "hashicorp/time"');
+    });
+
     it('declares exactly the retained input variables', () => {
       // Set equality, so a variable added or removed here is deliberate. The
       // system prompt stays a file read rather than a variable.
--- a/packages/nx-plugin/src/migrations/latest/terraform-harness-role-propagation/migration.ts
+++ b/packages/nx-plugin/src/migrations/latest/terraform-harness-role-propagation/migration.ts
@@ -1,0 +1,162 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+  joinPathFragments,
+  type MigrationReturnObject,
+  type Tree,
+} from '@nx/devkit';
+import {
+  applyGritQL,
+  GRIT_INSERT_PLACEHOLDER,
+  insertViaGritQL,
+  matchGritQL,
+} from '../../../utils/ast.js';
+import { formatFilesInSubtree } from '../../../utils/format.js';
+import {
+  PACKAGES_DIR,
+  SHARED_TERRAFORM_DIR,
+} from '../../../utils/shared-constructs-constants.js';
+import { TERRAFORM_VERSIONS } from '../../../utils/versions.js';
+
+/**
+ * Wait for the AgentCore Harness execution role to propagate before creating the
+ * Harness.
+ *
+ * IAM is eventually consistent, and CreateHarness validates the execution role by
+ * assuming it. Depending on the role's policy alone, the validation can run before
+ * the role has propagated and the Harness lands in CREATE_FAILED with "Role
+ * validation failed", failing the apply.
+ *
+ * These files are generated with `KeepExisting`, so an upgraded workspace keeps
+ * the older module until this runs. Modules that have diverged from the generated
+ * shape are left untouched and reported via `nextSteps`.
+ */
+
+const TERRAFORM_HARNESS_APP_DIR = `${PACKAGES_DIR}/${SHARED_TERRAFORM_DIR}/src/app/harnesses`;
+
+const hcl = (pattern: string) => `language hcl\n${pattern}`;
+
+const divergedMessage = (filePath: string) =>
+  `${filePath}: has diverged from the generated shape - left untouched. To wait for the execution role to propagate before the Harness is created, compare it against the agentcore-harness generator's Terraform template and add the time_sleep.execution_role_propagation resource the Harness depends on.`;
+
+/**
+ * Indented as `terraform fmt` writes a `required_providers` entry, since `.tf`
+ * files are not reformatted after a migration and the vended `format` target
+ * fails on one that isn't.
+ */
+const TIME_PROVIDER_TEXT = `    time = {
+      source  = "hashicorp/time"
+      version = "${TERRAFORM_VERSIONS.time}"
+    }`;
+
+const TIME_SLEEP_TEXT = `# IAM is eventually consistent, and CreateHarness validates the execution role
+# by assuming it. Without this wait the validation intermittently runs before
+# the role and its policy have propagated and the Harness lands in CREATE_FAILED
+# with "Role validation failed".
+resource "time_sleep" "execution_role_propagation" {
+  count = var.create_execution_role ? 1 : 0
+
+  depends_on = [aws_iam_role.execution_role, aws_iam_role_policy.execution_role]
+
+  create_duration = "30s"
+
+  triggers = {
+    assume_role_policy = aws_iam_role.execution_role[0].assume_role_policy
+    policy             = aws_iam_role_policy.execution_role[0].policy
+  }
+}`;
+
+/**
+ * Every shape the migration rewrites, so a module missing any of them is
+ * reported rather than half-migrated. The role and its policy are indexed, which
+ * is the shape the earlier `terraform-harness-vpc-role-and-memory-grant`
+ * migration leaves behind.
+ */
+const REQUIRED_SHAPES = [
+  '`required_providers { $_ }`',
+  '`resource "aws_iam_role" "execution_role" { $_ }`',
+  '`resource "aws_iam_role_policy" "execution_role" { $_ }`',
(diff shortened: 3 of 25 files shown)


docs/src/content/docs/en/guides/terraform-project.mdx
```diff
@@ -357,9 +357,15 @@ run "plan_is_valid" {
...
 
-Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. Note that the `test` target reads its path from `project.json`, so change it there too.
+The `install-providers` target is what fills that cache, and every target that runs `terraform init` depends on it:
+
+<NxCommands commands={['install-providers tf-infra']} />
+
+You will not normally run it yourself. It exists because two `terraform init` runs filling one cache at the same time can each compute a different hash for the same provider, which terraform then rejects against your `.terraform.lock.hcl` with "the cached package ... does not match any of the checksums recorded in the dependency lock file". Downloading the providers once, ahead of the targets that need them, leaves those targets only reading from a cache that is already warm.
+
+Set `TF_PLUGIN_CACHE_DIR` in your environment to point the vended `init` script at a cache you manage yourself — a volume shared between workspaces, say. Note that the `install-providers`, `test` and `validate` targets read their path from `project.json`, so change it there too.
```
cogwirrel: remove lines 362 to 368

packages/nx-plugin/src/terraform/project/generator.ts
cogwirrel: Let's remove these redundant commands from `test` and `validate`, and rename `install-providers` to `terraform-init`

packages/nx-plugin/src/terraform/project/generator.ts
```diff
@@ -302,6 +307,36 @@ export async function terraformProjectGenerator(
...
+    // `validate` — so without one writer ahead of them they race over the cache
+    // and terraform rejects the resulting hash against the lock file. They each
+    // depend on this, so by the time they run the cache is warm and they only
+    // link out of it.
+    //
+    // `-backend=false` installs the providers without configuring the S3
+    // backend, so this needs no credentials and runs on a fresh workspace before
+    // `bootstrap`. Deliberately uncached: a cache hit on another machine would
+    // report success over an empty cache and hand the race back to the targets
+    // that depend on it.
```
cogwirrel: Remove this comment

packages/nx-plugin/src/terraform/project/generator.ts
nx-plugin-for-aws: Done — renamed to `terraform-init`, and `test` and `validate` now run against the `TF_DATA_DIR` it prepared rather than initialising their own:

```jsonc
"terraform-init": {
  "options": {
    "commands": [{ "command": "shx mkdir -p <cache>", "forwardAllArgs": false }, "terraform init -backend=false"],
    "env": { "TF_DATA_DIR": "../../../dist/{projectRoot}/terraform-init", "TF_PLUGIN_CACHE_DIR": "<cache>" }
  }
},
"test":     { "options": { "command": "terraform test",     "env": { "TF_DATA_DIR": "../../../dist/{projectRoot}/terraform-init" } }, "dependsOn": ["terraform-init"] },
"validate": { "options": { "command": "terraform validate", "env": { "TF_DATA_DIR": "../../../dist/{projectRoot}/terraform-init" } }, "dependsOn": ["terraform-init"] }
```

So one `terraform init` per project instead of three, and the separate `terraform-test`/`terraform-validate` data dirs are gone. `parallel: false` went with the multi-command list, and `TF_PLUGIN_CACHE_DIR` too, since neither target installs anything now.

Verified in a real workspace: `test` and `validate` run concurrently against the shared data dir from a cold cache, with a real `.tftest.hcl` present (`Success! 1 passed, 0 failed` alongside `Success! The configuration is valid`), plus cold `build`/`plan`/`apply`/`destroy`. The migration rewrites both targets and is only allowed to touch the exact shape the generator produced — a target that has taken on another command, or been pointed at a data dir of its own, is reported via `nextSteps` instead.

packages/nx-plugin/src/terraform/project/generator.ts
nx-plugin-for-aws: Removed. The race explanation was already on `pluginCacheDir` above, so it was duplicated there.

What is left on the target is four lines, and only the part that has no other home now that `test` and `validate` no longer carry it — why the init is `-backend=false`, and why the target is uncached. Happy to drop those too if you would rather it carried nothing.

docs/src/content/docs/en/guides/terraform-project.mdx
nx-plugin-for-aws: Removed 362–368, so the section is back to the single paragraph about the shared cache.

Note that takes the pre-existing `TF_PLUGIN_CACHE_DIR` override paragraph with it (it was line 368). That reads right to me — the target is an implementation detail and the override note pointed at target names that have now changed — but say the word if you meant 362–366 and I will put it back.

shsrams:
Reviewing this against the failure it fixes. The plugin-cache half is exactly what `main` needs right now and I'd like to land it quickly; the harness role-propagation half needs a decision first. So I'd suggest splitting them.

### Land now: the provider plugin-cache fix

`main`'s `terraform-deploy` failed on precisely this in [run 35295062284](https://github.com/awslabs/nx-plugin-for-aws/actions/runs/35295062284) (attempt 1), on `@e2e-test/infra:plan:dev`:

```
Error: Required plugins are not installed
  - registry.terraform.io/hashicorp/local: the cached package for hashicorp/local 2.9.0
    (in .terraform/providers) does not match any of the checksums recorded in the dependency lock file
```

The log shows `validate` and `init:dev` starting three lines apart, which matches your diagnosis. The `terraform-init` single-writer target and its migration are the fix, and nothing about them depends on the discussion below.

### Please reconsider: the harness role-propagation wait

Provider v6.64.0 (2026-09-09) added an upstream retry for the same error:

> `resource/aws_bedrockagentcore_harness`: Retry `` Role validation failed for '...' `` IAM eventual consistency errors on Create. Because this error is returned while waiting for a newly-created harness to stabilize, the failed harness is deleted and creation is restarted.

\#1270 bumps the vended provider from 6.63.0 to 6.64.0, so that retry is about to reach users.

To be fair to the change you made, the two are not equivalent:

- `time_sleep` **prevents** the failure, at the cost of 30s on every harness deploy.
- The provider **recovers** from it, by deleting the failed harness and recreating it — the apply succeeds, but with an extra creation cycle and CloudTrail events reflecting it.

My inclination is still to drop it, for maintenance reasons rather than behavioural ones:

1. It is a vended template change plus a migration existing workspaces must run. If we ship it and later conclude the provider retry is enough, removing it costs a *second* migration.
2. Once 6.64.0 lands, this is no longer a hard failure, so the plugin would be carrying a workaround for something fixed upstream.

If you disagree, say so — you have deployment evidence here that I don't, and "prevent rather than recover" is a defensible position. I'd just rather decide before it ships than after.

Dropping it removes roughly 360 lines: the `time_sleep` and the `hashicorp/time` requirement in the harness template, the `terraform-harness-role-propagation` migration and its spec, and the generator assertion covering the wait.

### Ordering relative to #1270

If we split, I'd merge this before #1270 rather than after. #1270 doesn't fix the race, and its own `main` run will execute `terraform-deploy`, so merging it first risks another red `main` and another 45-minute retry. Landing this first also lets #1270 be regenerated by the `update-versions` workflow against the new `main`, instead of someone resolving dependency churn inside this PR — the two currently conflict in `package.json`, both lockfiles, `versions.ts` and a snapshot.

### Branch state

I've already merged current `main` into this branch (`45aad77d`):

- `applyWithRetries` was replaced by `applyInfra` from #1272, which retries **only** the CloudWatch Logs delivery-source `ConflictException` rather than any failure, and logs a marker so we can see when it fires. I folded your 30s inter-attempt delay into it. Retrying every failure would have masked the two genuine defects we hit since: the harness `environment_variables` bug (#1274) and this plugin-cache race.
- Conflicts with #1274 in `agentcore-harness/generator.spec.ts` and the `terraform-harness-vpc-role-and-memory-grant` convergence test were resolved additively. That convergence test now runs both later migrations, matching what `nx migrate` does.

Local verification after the merge: 878 tests pass across `src/migrations`, `src/agentcore-harness` and `src/terraform/project`; `compile`, `generate-migrations` and `lint` pass.

### Minor

The description calls the new migration `terraform-install-providers-target`, but the code has `terraform-init-target` and a `terraform-init` target. Worth aligning the description so the changelog and the migration name agree.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/app/agentcore-harness/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -10,6 +10,10 @@ terraform {
       source  = "hashicorp/random"
       version = "<%- randomProviderVersion %>"
     }
+    time = {
+      source  = "hashicorp/time"
+      version = "<%- timeProviderVersion %>"
+    }
```
shsrams, line 16: Remove the `time` provider requirement. Nothing else in this module uses it once the `time_sleep` below is gone.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/app/agentcore-harness/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -366,6 +370,23 @@ resource "aws_iam_role_policy" "execution_role" {
...
+  depends_on = [aws_iam_role.execution_role, aws_iam_role_policy.execution_role]
+
+  create_duration = "30s"
+
+  triggers = {
+    assume_role_policy = aws_iam_role.execution_role[0].assume_role_policy
+    policy             = aws_iam_role_policy.execution_role[0].policy
+  }
+}
+
```
shsrams, line 389: Remove this comment and the `time_sleep.execution_role_propagation` resource. Provider v6.64.0 (vended by #1270) retries `Role validation failed` on Create, so the module no longer needs its own wait.

packages/nx-plugin/src/utils/agent-core-constructs/files/terraform/app/agentcore-harness/__nameKebabCase__/__nameKebabCase__.tf.template
```diff
@@ -455,7 +476,7 @@ resource "aws_bedrockagentcore_harness" "this" {
 
   tags = var.tags
 
-  depends_on = [aws_iam_role_policy.execution_role]
+  depends_on = [time_sleep.execution_role_propagation]
```
shsrams, line 479: Restore this to what `main` has:

```hcl
  depends_on = [aws_iam_role_policy.execution_role]
```

packages/nx-plugin/src/migrations/latest/terraform-harness-role-propagation/migration.ts
```diff
@@ -0,0 +1,162 @@
+/**
```
shsrams, line 1: Delete this migration directory (`metadata.json`, `migration.ts`, `migration.spec.ts`). With the `time_sleep` removed from the template there is nothing for existing workspaces to converge on. `migrations.json` is generated, so removing the directory is enough; `generate-migrations` in the verification run will confirm the manifest still builds.

packages/nx-plugin/src/agentcore-harness/generator.spec.ts
```diff
@@ -545,6 +545,24 @@ describe('agentcore-harness generator', () => {
...
+      );
+      expect(tf).toContain(
+        'depends_on = [aws_iam_role.execution_role, aws_iam_role_policy.execution_role]',
+      );
+      expect(tf).toContain(
+        'depends_on = [time_sleep.execution_role_propagation]',
+      );
+      // The wait's provider is pinned like every other the module uses.
+      expect(tf).toContain('source  = "hashicorp/time"');
+    });
```
shsrams, line 564: Remove this test along with the `time_sleep`.

packages/nx-plugin/src/migrations/v1.0.0-rc.84/0005-terraform-harness-vpc-role-and-memory-grant/migration.spec.ts
```diff
@@ -6,6 +6,7 @@ import type { Tree } from '@nx/devkit';
 import { agentcoreHarnessGenerator } from '../../../agentcore-harness/generator.js';
 import { createTreeUsingTsSolutionSetup } from '../../../utils/test.js';
 import environmentVariablesMigration from '../../latest/terraform-harness-environment-variables/migration.js';
+import rolePropagationMigration from '../../latest/terraform-harness-role-propagation/migration.js';
```
shsrams, line 9: Remove this import and the `await rolePropagationMigration(tree);` call at line 553. Keep the `environmentVariablesMigration` step and the updated comment; the chain still needs to run every later migration that touches this module.

shsrams (dismissed):
Please drop the harness role-propagation half of this PR so the plugin-cache fix can land on its own. The reasoning is in https://github.com/awslabs/nx-plugin-for-aws/pull/1263#issuecomment-5724678612 — in short, provider v6.64.0 retries the `Role validation failed` error on Create, and #1270 vends that version, so the plugin would be carrying a template change plus a migration for something the provider now handles.

Changes requested, each anchored inline below:

1. Harness template: remove the `time` provider requirement, the `time_sleep.execution_role_propagation` resource, and restore `depends_on = [aws_iam_role_policy.execution_role]` on `aws_bedrockagentcore_harness.this`.
2. Delete `packages/nx-plugin/src/migrations/latest/terraform-harness-role-propagation/` (`metadata.json`, `migration.ts`, `migration.spec.ts`). `migrations.json` is generated at build time, so nothing else to update there.
3. Remove the `waits for the execution role to propagate before creating the Harness` test from `agentcore-harness/generator.spec.ts`.
4. Remove the `rolePropagationMigration` import and call from the `terraform-harness-vpc-role-and-memory-grant` convergence spec.
5. Update the PR description so it matches the code: the target is `terraform-init` (not `install-providers`) and the migration is `terraform-init-target` (not `terraform-install-providers-target`); the smoke-test change is now `applyInfra` from #1272, which retries only the CloudWatch Logs delivery-source `ConflictException`, not any failure; and remove the AgentCore Harness role-propagation section once the code is gone.

Keep everything under `terraform/project`, the `terraform-init-target` migration, and the `terraform-apply.ts` change as they are. Run `pnpm nx run-many -t compile generate-migrations lint -p @aws/nx-plugin` and the specs under `src/migrations`, `src/agentcore-harness` and `src/terraform/project` after the change.

shsrams:
Split done in 91c9793c: reverted 4fe59971, so the `time_sleep`, the `hashicorp/time` requirement, the `terraform-harness-role-propagation` migration and its spec, and the generator assertion are gone. The harness template, `agentcore-harness/generator.spec.ts` and the rc.84 convergence spec are now byte-identical to `main`. Description and title updated to match (`terraform-init` / `terraform-init-target`). 868 tests pass across `src/migrations`, `src/agentcore-harness`, `src/terraform/project`.

Note: this branch still has the `../../latest/terraform-harness-environment-variables` import in the rc.84 spec that #1276 fixes on `main`. Left as-is here so the two do not conflict; whichever merges second picks it up via `main`.

